PR #1, opened 2021-05-12
Update CI

GuillaumeGomez:


bilelmoussaoui:
Can we re-trigger a new build of the Windows jobs? it looks fine to me otherwise

GuillaumeGomez:
Restarted it.

.github/workflows/CI.yml
```diff
@@ -59,12 +55,6 @@ jobs:
       - run: cargo run --release -- --no-manual-traits ../glib-macros
         working-directory: checker
         if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
-      - run: cargo run --release -- --no-manual-traits ../gtk3-macros
-        working-directory: checker
-        if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
-      - run: ./check_init_asserts ../${{matrix.conf.name}}
-        working-directory: checker
```
bilelmoussaoui, line 66: I suppose those are useful because glib is still on this workspace

GuillaumeGomez:
cc @pbor

.github/workflows/CI.yml
```diff
@@ -59,16 +55,13 @@ jobs:
...
-        if: matrix.rust == 'nightly' && (matrix.conf.name == 'atk' || matrix.conf.name == 'gdk' || matrix.conf.name == 'gdkx11' || matrix.conf.name == 'gtk')
       # tests run
       - name: tests nightly
         run: xvfb-run --auto-servernum cargo test --manifest-path ${{ matrix.conf.name }}/Cargo.toml ${{ matrix.conf.nightly }}
```
sdroege, line 60: None of the tests here require X11

sdroege: I.e. you can remove all the `xvfb-run` stuff

pbor:
> cc @pbor

I would simply try to replace 

```
gvsbuildref: 69cdac34c36d74eeae37006bb0fcedc64ee518dd
```

with

```
gvsbuildref: 4330ae8d249f58ed69e67baa9e77de46ee0676b6
```

Which is the latest version of https://github.com/wingtk/gvsbuild

Explanation:
The Windows CI checks out that script, builds all the C libraries and then caches them, so that they are not rebuilt every time, so in the old repo they were built once when the CI was added and then not rebuilt anymore. However since now this is a new repo, on the first run this is rebuilding all C libraries. This fails because the old version of gvsbuild was pointing at graphene git, which in the mean time changed. This was fixed in https://github.com/wingtk/gvsbuild/commit/3a2459153e75820bef1583dbd7153b9977181855 so it should not be an issue anymore.

GuillaumeGomez:
`STATUS_ACCESS_VIOLATION` => that's new. Yeay windows. Merging.

pbor:
> `STATUS_ACCESS_VIOLATION` => that's new. Yeay windows. Merging.

This looks like an actual failure or the gio tests with the latest gio...

(with that said, I agree on going ahead and merge)

PR #56, opened 2021-05-14
Fix typo in function name

GuillaumeGomez:


GuillaumeGomez:
Windows is still broken as usual. I love dark magic. :)

sdroege (approved):
It didn't cause problems because apparently nobody ever tried calling this function :) Otherwise it would've failed linking.

PR #59, opened 2021-05-15
glib: Mark various Variant getter functions correctly as (transfer fu…

sdroege:
…ll) in their return value

Fixes a couple of memory leaks.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -198,7 +198,7 @@ impl Variant {
     #[inline]
     #[doc(alias = "get_variant")]
     pub fn as_variant(&self) -> Option<Variant> {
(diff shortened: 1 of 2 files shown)


sdroege:
@GuillaumeGomez Before you complain, the existing tests already cover that but apparently nobody runs them through valgrind except for me every now and then :)

GuillaumeGomez (approved):

GuillaumeGomez:
Could be interesting to run them through valgrind in the CI then. :p (Also, I never complain on fixes!)

bilelmoussaoui:
> Could be interesting to run them through valgrind in the CI then. :p (Also, I never complain on fixes!)

Can you open an issue so we can figure out whether it's possible to do so?

GuillaumeGomez:
I do it for another project of mine. Opening the issue.

PR #60, opened 2021-05-15
glib: Fix leaks in FromGlibContainer impls for GString

sdroege:


sdroege:
@GuillaumeGomez Before you complain, the existing tests already cover that but apparently nobody runs them through valgrind except for me every now and then :)

GuillaumeGomez (approved):

GuillaumeGomez:
Same answer as #59. :p

PR #67, opened 2021-05-16
Add proc macro for binding methods to action maps

andy128k:
Continuation of https://github.com/gtk-rs/gtk3-rs/pull/266
Closes https://github.com/gtk-rs/gtk3-rs/issues/240


GuillaumeGomez:
Before going any further: is it really worth it to create a whole new crate for this? Shouldn't we put it into glib-macros instead?

Also: please add error message tests for your macro (like there are in `./glib/clone_tests/clone_compilation_errors.py`).

bilelmoussaoui:
> Before going any further: is it really worth it to create a whole new crate for this? Shouldn't we put it into glib-macros instead?

fyi, the actions are part of gio not glib

GuillaumeGomez:
We can rename "glib-macros" into "gtk-rs-macros" or something along the line if you prefer? I'd simply prefer to avoid creating too many repositories.

sdroege:
> fyi, the actions are part of gio not glib

Don't we have a macro for `GResource` things somewhere? That would also be part of some kind of gio-macros crate then

bilelmoussaoui:
> > fyi, the actions are part of gio not glib
> 
> Don't we have a macro for `GResource` things somewhere? That would also be part of some kind of gio-macros crate then

Indeed we do

sdroege:
Ah but that's not a proc macro so irrelevant here.

We could put this into the `glib-macros` crate here, doesn't seem like much of a problem. `gio` already indirectly depends on it anyway, and `glib-macros` wouldn't have to depend on `gio`. And technically `gio` is a part of `glib` anyway.

andy128k:
Moving this to `glib-macros` will make a circular dependency. Doc-tests depend on `gio`.

sdroege:
On Sun, 2021-06-06 at 03:30 -0700, Andrey Kutejko wrote:
> Moving this to glib-macros will make a circular dependency. Doc-tests
> depend on gio.

Well that decides it then :)

GuillaumeGomez (approved):

Hofer-Julian:
Is something blocking this PR at the moment?

sdroege:
> Is something blocking this PR at the moment?

Someone who knows well about actions to review if the way how the macro exposes them makes sense, and someone to review if adding this macro causes us composibility problems in the future.

andy128k:
@Hofer-Julian You may try this macro with [my crate](https://github.com/andy128k/awesome-gtk/).

jf2048:
@andy128k I'd like to see this updated with the latest bits from awesome-gtk (async stuff, lint fixes) and then I can review it

gio-macros/src/actions.rs
```diff
@@ -0,0 +1,556 @@
...
+        this.#method(#state_arg #parameter_arg) #maybe_await
+    };
+    if let Some(state_type) = handler.state_type()? {
+        invoke = change_state(handler.sig.output.span(), &invoke, state_type);
+    }
+
+    if is_async {
+        invoke = quote! {
+            let action = action.clone();
+            let parameter: Option<glib::variant::Variant> = parameter.cloned();
```
jf2048, line 332:
```suggestion
            let action = ::std::clone::Clone::clone(&action);
            let parameter: ::std::option::Option<glib::variant::Variant> = parameter.cloned();
```

gio-macros/src/actions.rs
```diff
@@ -0,0 +1,556 @@
...
+    attrs: ActionAttributes,
+    sig: Signature,
+}
+
+impl ActionHandlerInfo {
+    fn action_name(&self) -> String {
+        if let Some(ref name) = self.attrs.name {
+            name.clone()
+        } else {
+            self.sig.ident.to_string()
```
jf2048, line 40:
```suggestion
            self.sig.ident.to_string().replace("_", "-")
```

The convention in GTK is to use dashes in action names, so I think this makes sense as a default? If that sounds good, a note can be added in the docs about this and you can mention it can still be changed manually to underscores with `#[action(name = "...")]`

gio-macros/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+/// ```
+///
+/// If you need an action with both `activate` and `change-state` handlers, just
+/// create two methods and annotate them with the same action name.
+///
+/// ## Defaults
+///
+/// A `stateful` annotation may be omited if `initial_state` or `change_state` is
+/// specified.
+///
```
jf2048, line 94: Maybe put a note in here about the caveats with async actions? Especially as it relates to returning state changes, it seems like most of the time a stateful action shouldn't be async. You can put a note similar to this one I put in the template callbacks macro: https://github.com/gtk-rs/gtk4-rs/blob/86c1548cc2a6fa05c4285802f0607b374f3bec0f/gtk4-macros/src/lib.rs#L122-L131

gio-macros/src/actions.rs
```diff
@@ -0,0 +1,556 @@
...
+            )*
+        }
+    };
+    parse(register_method.into()).unwrap()
+}
+
+pub fn actions(
+    attrs: ActionImplAttributes,
+    mut input: ItemImpl,
+) -> Result<TokenStream, TokenStream> {
```
jf2048, line 506: Instead of returning a `Result` and using `to_compile_error`, you can catch the error, pass it to `proc_macro_error::emit_error` and then generate an empty register method. I made the gtk4 macros do something like this and it works really well to reduce the error message if you have a nightly compiler.

gio-macros/src/lib.rs
```diff
@@ -0,0 +1,201 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+extern crate proc_macro;
+
```
jf2048, line 4:
```suggestion
```

gio-macros/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "gio-macros"
+documentation = "https://gtk-rs.org/docs/gio-macros/"
+homepage = "https://gtk-rs.org/"
+authors = ["The Gtk-rs Project Developers"]
+description = "Rust bindings for the Gio library, proc macros crate"
+version = "0.13.0"
```
jf2048, line 7:
```suggestion
version = "0.16.0"
```

jf2048 (changes requested):

jf2048:
A couple more things:

- I didn't bother tagging all of them because there's so many, but every usage of gio and glib inside a quote block should be fully qualified using proc-macro-crate. `crate_ident_new` can be copied from glib-macros. Then it can be `#crate_ident` for gio and `#crate_ident::glib` for glib.
- Does the `#[allow(clippy)]` bits actually do anything here? I thought those didn't have an effect inside macros but I could be wrong.

jf2048:
Do we want to require an `#[action]` attribute on the method in order to make it an action? Without that attribute the macro would skip over that function. That's what I did with `gtk::template_callbacks` but not sure if it makes sense here?

andy128k:
There are multiple examples of actions with and without an attribute in tests.

jf2048:
Let me modify the example to show what I mean. Would it help to be like this:

```rust
#[gio::actions]
impl MyApplication {
    #[action]
    fn action1(&self) { /* this method gets added with add_action */ }
    fn my_method(&self) { /* not an action, just an ordinary method */ }
}
```

Then `action1` becomes an action but `my_method` does not. Does that make sense? It would allow for putting multiple macros on the same block, like for a widget:

```rust
#[gio::actions]
#[gtk::template_callbacks]
impl MyWidget {
    #[action]
    fn action1(&self) { /* ... */ }
    #[template_callback]
    fn callback1(&self) { /* ... */ }
}
```

What do you think? Or does it make sense to enforce a separate `impl` block for each macro? I'm not sure.

andy128k:
My opinion is that it is better to keep such blocks separated. Such macros may modify signatures of functions, generate new ones and step on each others feet any other way.

jf2048:
Sounds good to me

gio-macros/src/lib.rs
andy128k: Why? Such actions work as expected. Returning a value is just a fancy way to invoke `set_state`.

jf2048: Right, but it is possible with async actions to have two or more invocations of the action get spawned. If a later future resolves quicker than an earlier one, that would be like calling `set_state` with a stale value.

sdroege:
@bilelmoussaoui As you worked on actions API, what do you think of this considering the existing helpers you added for making this easier?

andy128k:
Current API is still very verbose and repetitive. I gave it a try and that's how it looks like just for two actions.

```rust
            window
                .add_action_entries([
                    gio::ActionEntry::builder("find")
                        .activate(move |this: &Self::Type, _, _| {
                            let this = this.clone();
                            glib::MainContext::default().spawn_local(async move {
                                this.find().await;
                            });
                        })
                        .build(),
                    gio::ActionEntry::builder("copy")
                        .activate(move |this: &Self::Type, _, _| this.copy())
                        .build(),
                ])
                .unwrap();
```

PR #73, opened 2021-05-17
Fix graphene-sys tests

andy128k:


graphene/sys/tests/constant.c
```diff
@@ -21,6 +21,7 @@
                     unsigned long: "%lu", \
                     long long: "%lld", \
                     unsigned long long: "%llu", \
+                    float: "%f", \
```
GuillaumeGomez, line 24: Why adding this line? Just below there the same printf option.

andy128k: GCC (v 10.2.1) emits an error:

```
error: ‘_Generic’ selector of type ‘float’ is not compatible with any association
```

andy128k: Corresponding change in `gir` is [here](https://github.com/gtk-rs/gir/pull/1142).

graphene/sys/src/lib.rs
```diff
@@ -62,9 +62,6 @@ pub const GRAPHENE_RAY_INTERSECTION_KIND_ENTER: graphene_ray_intersection_kind_t
 pub const GRAPHENE_RAY_INTERSECTION_KIND_LEAVE: graphene_ray_intersection_kind_t = 2;
 
 // Constants
-pub const GRAPHENE_HAS_GCC: c_int = 1;
-pub const GRAPHENE_HAS_SCALAR: c_int = 1;
-pub const GRAPHENE_HAS_SSE: c_int = 1;
 pub const GRAPHENE_PI: c_double = 3.141593;
 pub const GRAPHENE_PI_2: c_double = 1.570796;
```
bilelmoussaoui, line 66: I would probably just ignore those & make them an alias of the std types

sdroege: Yeah, like I wrote in the gir PR :)

graphene/sys/Gir.toml
```diff
@@ -15,6 +15,11 @@ external_libraries = [
 ignore = [
     "Graphene.true",
     "Graphene.false",
+    "Graphene.HAS_GCC",
+    "Graphene.HAS_SCALAR",
+    "Graphene.HAS_SSE",
+    "Graphene.PI",
+    "Graphene.PI_2",
```
bilelmoussaoui, line 22: you will have to mark them as manual on the non-sys part & added them to lib.rs

andy128k: Why? These constants are [not documented](https://developer.gnome.org/graphene/stable/api-index.html).

bilelmoussaoui:
Can you re-enable graphene's sys tests on the CI? so we can be sure the PR actually fixes them. Thanks!

sdroege:
If you rebase on latest master then the docs CI should also pass.

graphene/sys/src/lib.rs
```diff
@@ -62,11 +62,6 @@ pub const GRAPHENE_RAY_INTERSECTION_KIND_ENTER: graphene_ray_intersection_kind_t
 pub const GRAPHENE_RAY_INTERSECTION_KIND_LEAVE: graphene_ray_intersection_kind_t = 2;
 
 // Constants
-pub const GRAPHENE_HAS_GCC: c_int = 1;
-pub const GRAPHENE_HAS_SCALAR: c_int = 1;
-pub const GRAPHENE_HAS_SSE: c_int = 1;
-pub const GRAPHENE_PI: c_double = 3.141593;
-pub const GRAPHENE_PI_2: c_double = 1.570796;
 pub const GRAPHENE_SIMD_S: *const c_char = b"sse\0" as *const u8 as *const c_char;
```
sdroege, line 65: This one should also be ignored while we're at it. It's completely dependent on where the graphene version was built that was used to generate the .gir file.

sdroege (commented):
Seems good to me otherwise :)

GuillaumeGomez (approved):

.github/workflows/CI.yml
```diff
@@ -77,6 +77,10 @@ jobs:
       - name: build
         run: cargo build --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "${{ matrix.conf.features }}"
         if: matrix.rust != 'nightly'
+      - name: Test ${{ matrix.conf.name }}/sys
+        run: cargo test
+        working-directory: ${{ matrix.conf.name }}/sys
+        if: matrix.conf.name == 'graphene'
```
bilelmoussaoui, line 83: You just need to change the value in https://github.com/gtk-rs/gtk-rs-core/pull/73/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R26

andy128k: Parameter `test_sys` does nothing. It is not used anywhere in the `CI` action.

graphene/sys/src/lib.rs
andy128k: Done

.github/workflows/CI.yml
sdroege: @bilelmoussaoui What should we do about this?

bilelmoussaoui: wait what. It definitely used to do something, any ideas @GuillaumeGomez ?

bilelmoussaoui: Please copy it from https://github.com/gtk-rs/gtk4-rs/blob/master/.github/workflows/CI.yml#L37-L39 for now :-)

bilelmoussaoui (approved):
Thanks!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #79, opened 2021-05-04
Add missing C doc aliases


GuillaumeGomez:
We realized with https://github.com/gtk-rs/gtk-rs/pull/504 that a lot of doc aliases were missing. We should check on all the manual code of the crates.

GuillaumeGomez:
Done.

PR #83, opened 2021-05-18
Add missing doc aliases

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/79.

I improved the previous python script:

<details>

```python
import os
import sys


IGNORE_C_FNS = ["_get_reference_count", "_status", "_free", "_destroy", "_unref"]
IGNORE_FNS = [
    "deref", "deref_mut", "drop", "fmt", "clone", "to_value", "g_value_get_flags",
    "g_value_set_flags", "get_type",
]
IGNORE_FNS_START = ["to_glib", "from_glib", "from_raw", "to_raw"]


def is_valid_name(s):
    for c in s:
        if not c.isalnum() and c != '_':
            return False
    for x in IGNORE_C_FNS:
        if s.endswith(x):
            return False
    return True


def get_fn_name(s):
    parts = s.split(" ")
    x = 0
    while x < len(parts) and parts[x] != "fn":
        x += 1
    x += 1
    if x >= len(parts):
        return None
    fn_name = parts[x].split("<")[0].split("(")[0]
    for x in IGNORE_FNS_START:
        if fn_name.startswith(x):
            return None
    if fn_name in IGNORE_FNS:
        return None
    return fn_name


# This to prevent to add the same doc alias more than once on a same function.
def need_doc_alias(content, pos, alias):
    pos -= 1
    while pos >= 0:
        stripped = content[pos].strip()
        if not stripped.startswith("#[") and not stripped.startswith("///"):
            return True
        elif stripped == alias:
            return False
        pos -= 1
    return True


def find_method_in_trait(content, start_line, fn_name):
    if start_line is None:
        return None
    start_line += 1
    while start_line < len(content):
        clean = content[start_line].strip()
        if clean.startswith("fn "):
            name = clean[3:].split("<")[0].split("(")[0]
            if name == fn_name:
                return start_line
        elif content[start_line] == "}":
            # We reached the end of the trait declaration and found nothing...
            break
        start_line += 1
    return None


def add_parts(path):
    print("=> Updating '{}'".format(path))
    with open(path, 'r') as f:
        content = f.read().split('\n')
    x = 0
    # The key is the trait name, the value is its line number.
    traits = {}
    is_in_trait = None
    while x < len(content):
        clean = content[x].lstrip()
        if not clean.startswith("pub fn") and not clean.startswith("fn"):
            if content[x] == "}":
                is_in_trait = None
            elif (clean.startswith("impl ") or clean.startswith("impl<")) and " for " in clean:
                name = clean.split(" for ")[0].split(" ")[-1].strip()
                if name.endswith("Ext") or name.endswith("ExtManual"):
                    is_in_trait = name
            # This is needed because we want to put Ext traits doc aliases on the trait methods
            # directly and not on their implementation.
            if clean.startswith("pub trait") or clean.startswith("trait"):
                name = clean.split("trait")[1].split("<")[0].split(":")[0].split("{")[0].strip()
                if name.endswith("Ext") or name.endswith("ExtManual"):
                    traits[name] = x
                    # Completely skip the trait declaration.
                    while x < len(content) and content[x] != "}":
                        x += 1
                    continue
            x += 1
            continue

        if clean.endswith(';'): # very likely a trait method declaration.
            x += 1
            continue
        fn_name = get_fn_name(clean)
        if fn_name is None:
            x += 1
            continue
        spaces = ''
        for _ in range(0, len(content[x]) - len(clean)):
            spaces += ' '
        spaces_and_braces = spaces + '}'
        start = x
        need_traits_update = False
        if is_in_trait is not None:
            trait_method_pos = find_method_in_trait(content, traits.get(is_in_trait), fn_name)
            if trait_method_pos is None:
                print("Cannot find `{}` in trait `{}`, putting doc aliases on implementation".format(fn_name, is_in_trait))
            else:
                start = trait_method_pos
                need_traits_update = True
        x += 1
        added = 0
        while x < len(content):
            if content[x] == spaces_and_braces:
                break
            if "ffi::" in content[x]:
                sys_name = content[x].split("ffi::")[1].split("(")[0]
                if is_valid_name(sys_name):
                    alias = '#[doc(alias = "{}")]'.format(sys_name)
                    if need_doc_alias(content, start, alias) and fn_name in sys_name:
                        content.insert(start, spaces + alias)
                        added += 1
                        start += 1
                        x += 1
            x += 1
        if need_traits_update and added > 0:
            # move their start pos
            for trait in traits:
                if traits[trait] > start:
                    traits[trait] += added
        x += 1
    with open(path, 'w') as f:
        f.write('\n'.join(content))


def run_dirs(path):
    for entry in os.listdir(path):
        full = os.path.join(path, entry)
        if os.path.isdir(full):
            # We don't want to go in auto code parts. They already have everything they need.
            if not entry in ["auto", "subclass"]:
                run_dirs(full)
        elif entry.endswith(".rs"):
            add_parts(full)


def main():
    if len(sys.argv) < 2:
        print("No folder given as argument, updating current `src` (if any)")
        run_dirs("src")
    else:
        for x in sys.argv[1:]:
            print("> Going into folder `{}`...".format(x))
            run_dirs(x)
            print("< Done!")


if __name__ == "__main__":
    main()
```

</details>

### DIFF
--- a/glib/src/date.rs
+++ b/glib/src/date.rs
@@ -27,6 +27,7 @@ unsafe impl Sync for Date {}
 
 impl Date {
     #[doc(alias = "g_date_valid_dmy")]
+    #[doc(alias = "g_date_new_dmy")]
     pub fn new_dmy(day: DateDay, month: DateMonth, year: DateYear) -> Result<Date, BoolError> {
         let month = month.into_glib();
         unsafe {
--- a/glib/src/date.rs
+++ b/glib/src/date.rs
@@ -26,7 +26,6 @@ unsafe impl Send for Date {}
 unsafe impl Sync for Date {}
 
 impl Date {
-    #[doc(alias = "g_date_valid_dmy")]
     #[doc(alias = "g_date_new_dmy")]
     pub fn new_dmy(day: DateDay, month: DateMonth, year: DateYear) -> Result<Date, BoolError> {
         let month = month.into_glib();
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1208,6 +1208,7 @@ pub trait ObjectExt: ObjectType {
     #[doc(alias = "get_interface")]
     fn interface<T: IsInterface>(&self) -> Option<InterfaceRef<T>>;
 
+    #[doc(alias = "g_object_set_property")]
     fn set_property<'a, N: Into<&'a str>, V: ToValue>(
         &self,
         property_name: N,
@@ -1222,6 +1223,7 @@ pub trait ObjectExt: ObjectType {
     fn set_properties_from_value(&self, property_values: &[(&str, Value)])
         -> Result<(), BoolError>;
     #[doc(alias = "get_property")]
+    #[doc(alias = "g_object_get_property")]
     fn property<'a, N: Into<&'a str>>(&self, property_name: N) -> Result<Value, BoolError>;
     fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> bool;
     #[doc(alias = "get_property_type")]
@@ -1334,6 +1336,7 @@ pub trait ObjectExt: ObjectType {
     where
         F: Fn(&[Value]) -> Option<Value>;
     /// Emit signal by signal id.
+    #[doc(alias = "g_signal_emitv")]
     fn emit(&self, signal_id: SignalId, args: &[&dyn ToValue]) -> Result<Option<Value>, BoolError>;
     /// Same as `emit` but takes `Value` for the arguments.
     fn emit_with_values(
@@ -1367,6 +1370,7 @@ pub trait ObjectExt: ObjectType {
         details: Quark,
         args: &[Value],
     ) -> Result<Option<Value>, BoolError>;
+    #[doc(alias = "g_signal_handler_disconnect")]
     fn disconnect(&self, handler_id: SignalHandlerId);
 
     fn connect_notify<F: Fn(&Self, &crate::ParamSpec) + Send + Sync + 'static>(
@@ -1384,7 +1388,9 @@ pub trait ObjectExt: ObjectType {
         name: Option<&str>,
         f: F,
     ) -> SignalHandlerId;
+    #[doc(alias = "g_object_notify")]
     fn notify<'a, N: Into<&'a str>>(&self, property_name: N);
+    #[doc(alias = "g_object_notify_by_pspec")]
     fn notify_by_pspec(&self, pspec: &crate::ParamSpec);
 
     fn downgrade(&self) -> WeakRef<Self>;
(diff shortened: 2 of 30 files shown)


pangocairo/src/font_map.rs
```diff
@@ -11,6 +11,7 @@ pub trait FontMapExtManual {
 }
 
 impl<O: IsA<FontMap>> FontMapExtManual for O {
+    #[doc(alias = "pango_cairo_font_map_get_font_type")]
```
bilelmoussaoui, line 14: This one should be in the trait, not the trait implementation

glib/src/object.rs
```diff
@@ -1508,6 +1508,7 @@ impl<T: ObjectType> ObjectExt for T {
         Ok(())
     }
 
+    #[doc(alias = "g_object_set_property")]
```
bilelmoussaoui, line 1511: All of those should be on the trait, not the trait implementation

glib/src/date.rs
```diff
@@ -27,6 +27,7 @@ unsafe impl Sync for Date {}
 
 impl Date {
     #[doc(alias = "g_date_valid_dmy")]
```
bilelmoussaoui, line 29: looks wrong as a doc alias no?

bilelmoussaoui:
Looks good to me other than the script adding the doc aliases to the wrong place for traits

GuillaumeGomez:
@bilelmoussaoui Actually I made it this way voluntarily.

bilelmoussaoui:
> @bilelmoussaoui Actually I made it this way voluntarily.

Does it make sense though? if so, gir should be updated to do it similarly

GuillaumeGomez:
Huuuuuuum... I honestly don't know...

glib/src/date.rs
GuillaumeGomez: Well, the old script was less efficient, I'll remove it. ;)

GuillaumeGomez:
Updated!

glib/src/object.rs
```diff
@@ -2400,6 +2407,7 @@ impl ObjectClass {
         }
     }
 
+    #[doc(alias = "g_object_class_list_properties")]
```
bilelmoussaoui, line 2410: same for this & the ones after

glib/src/object.rs
```diff
@@ -2385,6 +2391,7 @@ impl ObjectClass {
             .map(|pspec| pspec.value_type())
     }
 
+    #[doc(alias = "g_object_class_find_property")]
```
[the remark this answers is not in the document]
GuillaumeGomez: It's not a trait implementation here, so I'm a bit lost. What do you mean?

bilelmoussaoui (approved):

PR #85, opened 2021-05-19
Create Boxed(Value) newtype for Boxed GValue

mehmooda:
Part of #4

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -454,7 +454,7 @@ impl Uninitialized for Value {
     }
 }
 
-impl StaticType for Value {
+impl StaticType for BoxedValue {
     fn static_type() -> Type {
         unsafe { from_glib(gobject_ffi::g_value_get_type()) }
     }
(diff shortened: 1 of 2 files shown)


glib/src/value.rs
```diff
@@ -1005,45 +1005,47 @@ numeric!(
     gobject_ffi::g_value_set_double
 );
 
-impl ValueType for Value {
-    type Type = Value;
+pub struct Boxed(pub Value);
```
sdroege, line 1008:
```suggestion
pub struct BoxedValue(pub Value);
```

And then also re-export it from `src/lib.rs`

sdroege (commented):

glib/src/value.rs
```diff
@@ -1005,45 +1005,47 @@ numeric!(
     gobject_ffi::g_value_set_double
 );
 
-impl ValueType for Value {
-    type Type = Value;
+pub struct BoxedValue(pub Value);
```
sdroege, line 1008: Needs some docs and having an `impl Deref for BoxedValue` would seem useful

sdroege:
@mehmooda If you want it in this release then it has to be updated in the next few days.

glib/src/value.rs
sdroege: Added both

glib/src/value.rs
```diff
@@ -1010,45 +1010,56 @@ numeric!(
     gobject_ffi::g_value_set_double
 );
 
-impl ValueType for Value {
-    type Type = Value;
+/// A [`Value`] containing another [`Value`].
+pub struct BoxedValue(pub Value);
```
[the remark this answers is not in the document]
sdroege: Yes

bilelmoussaoui:
I suppose this change doesn't require any change from gir's side? otherwise looks good to me

sdroege:
> I suppose this change doesn't require any change from gir's side? otherwise looks good to me

They don't :)

PR #86, opened 2021-05-20
Add doc alias check

GuillaumeGomez:
Same as https://github.com/gtk-rs/gtk3-rs/pull/540.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -55,6 +55,12 @@ jobs:
       - run: cargo run --release -- --no-manual-traits ../glib-macros
         working-directory: checker
         if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
+      - name: Check doc aliases
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -55,6 +55,12 @@ jobs:
       - run: cargo run --release -- --no-manual-traits ../glib-macros
         working-directory: checker
         if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
+      - name: Check doc aliases
+        run: |
+          python3 doc_aliases.py ../${{matrix.conf.name}}
+          cd .. && git diff -R --exit-code
```
sdroege, line 61: Just like what I wrote in the gstreamer-rs MR, it would make more sense to not use `-R` here :)

GuillaumeGomez: Ok!

sdroege: `cargo fmt check` does the same btw

PR #89, opened 2021-05-21
Use `&[impl AsRef<str>]` instead of `&[&str]`

### LINKED ISSUE #77: Use `T: AsRef<str>, &[T]` instead of `&[&str]` in various APIs
Much nicer to use if your strings are actual `String`.

Same thing applies for other `&[&T]` we might have in API somewhere as with `AsRef` both `&[T]` and `&[&T]` would be possible to pass in (and other `AsRef` impls of course).

rafaelcaricio:
I've skip all code under `auto` folders as those are auto-generated. Maybe I could update the automatic code generation as well? 🤔


bilelmoussaoui:
> I've skip all code under `auto` folders as those are auto-generated. Maybe I could update the automatic code generation as well? 🤔 

Yes that would require some changes in gir (the utility used for generating the auto code).

Do you want to have a look at updating that one as well?

rafaelcaricio:
@bilelmoussaoui Yes, I just started browsing the gir codebase 🔍

bilelmoussaoui:
I still see various `: &str` in glib/src/object.rs from a quick search https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/object.rs#L1259-L1277 there are probably others btw :)

sdroege:
Let's focus only on `[&str]` in this PR and keep plain `&str` for a later PR with separate discussion. The advantage seems less big there to me.

Also this needs some changes in https://github.com/gtk-rs/gir :)

Thanks for looking into this, @rafaelcaricio!

rafaelcaricio:
I've tried to check the gir code to find where we would need to change to generate the `AsRef<str>`. Unfortunately, it is not clear to me where the changes need to be done. I've created a test case for the part of the code that I think needs to be changed: https://github.com/gtk-rs/gir/commit/857e5303d792debfbf00350d8b74ddcdd49681ea 

Could anyone shine some light if that is the correct place?

rafaelcaricio:
I've opened a PR in the gir repo to complement the changes: https://github.com/gtk-rs/gir/pull/1149

bilelmoussaoui:
> I've opened a PR in the gir repo to complement the changes: [gtk-rs/gir#1149](https://github.com/gtk-rs/gir/pull/1149)

Do you mind using that branch of gir and regenerate gtk-rs-core so we can see how those changes looks like?

rafaelcaricio:
@bilelmoussaoui Would you like me to commit the generated changes here?

bilelmoussaoui:
> @bilelmoussaoui Would you like me to commit the generated changes here?

yes please

gio/src/auto/app_info.rs
```diff
@@ -268,7 +268,7 @@ pub trait AppInfoExt: 'static {
     #[doc(alias = "g_app_info_launch_uris")]
     fn launch_uris<P: IsA<AppLaunchContext>>(
         &self,
-        uris: &[&str],
+        uris: &&str,
```
rafaelcaricio, line 271: Not completely successful after all.. 🤕

gio/src/auto/file_info.rs
```diff
@@ -2,6 +2,7 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
+use crate::str;
```
rafaelcaricio, line 5: This one is also strange. 🤔

MarijnS95:
Cross-posting https://github.com/gtk-rs/gir/pull/1149#issuecomment-846469739, perhaps `&[impl AsRef<str>]` might work to save on the extra trait bounds that otherwise have to be added.

rafaelcaricio:
@MarijnS95 Right, that would simplify things indeed. Just doesn't look natural!? Happy to make the change anyways.. just let me know what you all prefer.

MarijnS95:
> @MarijnS95 Right, that would simplify things indeed. Just doesn't look natural!? Happy to make the change anyways.. just let me know what you all prefer.

I personally prefer it since it's more concise, but this is not my project and I'm not a maintainer so :)

gio/src/auto/file_info.rs
bilelmoussaoui: I think gir automatically tries to "include" everything used for `AsRef/IsA` and would need special casing for `str`

sdroege:
I'd prefer `impl trait` if we don't ever need to use the type variable inside the code.

MarijnS95:
I wonder if we could reduce trait bound complexity even further by using `impl Trait` in more places. But that's definitely a thing for a separate commit/PR!

sdroege:
I'm all for it, please go ahead if you have some time :)

MarijnS95:
> I'm all for it, please go ahead if you have some time :)

Easy enough I'd say, #1153.

Still needs some more testing and scrutineering, but it's time for breakfast first :P

gio/src/auto/file_info.rs
MarijnS95: https://github.com/gtk-rs/gir/blob/395e9ab1395d369c288167684ea42709a9ec592b/src/analysis/bounds.rs#L280 For the record.

rafaelcaricio:
If I understand the discussion here correctly, the PR https://github.com/gtk-rs/gir/pull/1153 supersedes my PR https://github.com/gtk-rs/gir/pull/1149 ? Now I should probably revert the commit I have in this PR to remove the generated code from my PR to gir?

What would be the next steps? I'm a bit confused now.

MarijnS95:
> If I understand the discussion here correctly, the PR [gtk-rs/gir#1153](https://github.com/gtk-rs/gir/pull/1153) supersedes my PR [gtk-rs/gir#1149](https://github.com/gtk-rs/gir/pull/1149) ? Now I should probably revert the commit I have in this PR to remove the generated code from my PR to gir?
> 
> What would be the next steps? I'm a bit confused now.

Hey!

Apologies for the confusion!

Your PR is not superseded at all :). Ideally we get https://github.com/gtk-rs/gir/pull/1153 in first, as that only switches to `impl`, and then get https://github.com/gtk-rs/gir/pull/1149 plus this PR on top to use `AsRef<str>` strings too!

I still have some finishing up to do, GStreamer-rs is pretty lenient but gtk-rs uses all the special types (async and futures with special trait bounds). After that I can help rebase your gir PR on top unless you wan to take care of that yourself.

Let me know what approach you prefer! We can always get your gir PR merged first, and that I rebase my work on top (though that PR is already rather loaded and substantial, it's much easier to rebase smaller, trivial changes like yours on top instead of the other way around).

rafaelcaricio:
@MarijnS95 I'm fine with waiting for your PR to get in first. This is my first contribution to those projects so I'm still figuring out my way around the code.

rafaelcaricio:
On the case of using `impl AsRef<str>`, the current version of gtk-rs/gir#1149 is generic enough to not break on callbacks that need to use the bounds. But I guess the gtk-rs/gir#1153 will help with differentiating.

rafaelcaricio:
I've updated the PR with my latest changes, now I used `&[impl AsRef<str>]`. I still need to update the manual code, but that is easy. :)

MarijnS95:
> I've updated the PR with my latest changes, now we use `&[impl AsRef<str>]`. I still need to update the manual code, but that is easy. :)

It doesn't look like all occurences were replaced, some are still using an explicit type parameter. Was that intended?

rafaelcaricio:
@MarijnS95 Could you point to one example where it was not replaced?

gio/src/app_info.rs
```diff
@@ -25,19 +25,20 @@ pub trait AppInfoExtManual: 'static {
         P: IsA<AppLaunchContext>,
         Q: IsA<Cancellable>,
         R: FnOnce(Result<(), glib::Error>) + Send + 'static,
+        S: AsRef<str>,
     >(
         &self,
-        uris: &[&str],
+        uris: &[S],
```
MarijnS95, line 31: Here.

gio/src/app_info.rs
```diff
@@ -25,19 +25,20 @@ pub trait AppInfoExtManual: 'static {
...
         callback: R,
     );
 
     #[cfg(any(feature = "v2_60", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
-    fn launch_uris_async_future<P: IsA<AppLaunchContext> + Clone + 'static>(
+    fn launch_uris_async_future<P: IsA<AppLaunchContext> + Clone + 'static, S: AsRef<str>>(
         &self,
-        uris: &[&str],
+        uris: &[S],
```
MarijnS95, line 41: Here.

gio/src/app_info.rs
```diff
@@ -49,15 +50,16 @@ impl<O: IsA<AppInfo>> AppInfoExtManual for O {
         P: IsA<AppLaunchContext>,
         Q: IsA<Cancellable>,
         R: FnOnce(Result<(), glib::Error>) + Send + 'static,
+        S: AsRef<str>,
     >(
         &self,
-        uris: &[&str],
+        uris: &[S],
```
MarijnS95, line 56: Here.

gio/src/app_info.rs
```diff
@@ -91,12 +93,18 @@ impl<O: IsA<AppInfo>> AppInfoExtManual for O {
 
     #[cfg(any(feature = "v2_60", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
-    fn launch_uris_async_future<P: IsA<AppLaunchContext> + Clone + 'static>(
+    fn launch_uris_async_future<P: IsA<AppLaunchContext> + Clone + 'static, S: AsRef<str>>(
         &self,
-        uris: &[&str],
+        uris: &[S],
```
MarijnS95, line 98: Here.

gio/src/desktop_app_info.rs
```diff
@@ -54,9 +54,10 @@ pub trait DesktopAppInfoExtManual {
         T: AsRawFd,
         U: AsRawFd,
         V: AsRawFd,
+        S: AsRef<str>,
     >(
         &self,
-        uris: &[&str],
+        uris: &[S],
```
MarijnS95, line 60: Here.

gio/src/desktop_app_info.rs
```diff
@@ -75,9 +76,10 @@ impl<O: IsA<DesktopAppInfo>> DesktopAppInfoExtManual for O {
...
```
MarijnS95, line 82: Here.

glib/src/functions.rs
```diff
@@ -27,10 +27,17 @@ use std::ptr;
...
+    U: AsRawFd,
+    V: AsRawFd,
+    A: AsRef<str>,
+    E: AsRef<str>,
+>(
     working_directory: P,
-    argv: &[&str],
-    envp: &[&str],
+    argv: &[A],
+    envp: &[E],
```
MarijnS95, line 40: Here.

MarijnS95 (changes requested):
Quite a few places :wink:

glib/src/functions.rs
rafaelcaricio: Those are all manually added, that's why I said that I need to change the manuals... ;) I will update the PR later

rafaelcaricio:
@MarijnS95 those are all manually added. That's why I wrote "I still need to update the manual code, but that is easy." :) Those occurrences are not generated by gir.

gio/src/app_info.rs
```diff
@@ -80,7 +80,7 @@ impl<O: IsA<AppInfo>> AppInfoExtManual for O {
         unsafe {
             ffi::g_app_info_launch_uris_async(
                 self.as_ref().to_glib_none().0,
-                uris.to_glib_none().0,
+                uris.as_ref().to_glib_none().0,
```
MarijnS95, line 83: This has a normal `to_glib_none()` implementation now right? This change wasn't necessary in auto code, and returns the slice like an identity operator:

https://doc.rust-lang.org/std/primitive.slice.html#impl-AsRef%3C%5BT%5D%3E

gio/src/app_info.rs
```diff
@@ -93,10 +93,16 @@ impl<O: IsA<AppInfo>> AppInfoExtManual for O {
...
         context: Option<&P>,
     ) -> Pin<Box_<dyn std::future::Future<Output = Result<(), glib::Error>> + 'static>> {
-        let uris = uris.iter().copied().map(String::from).collect::<Vec<_>>();
+        let uris = uris
+            .as_ref()
+            .iter()
+            .map(|v| v.as_ref())
+            .copied()
+            .map(String::from)
+            .collect::<Vec<_>>();
```
MarijnS95, line 105: `launch_uris_async` takes a `uris: &[impl AsRef<str>],` as well, it should be possible to pass this as-is or by simply copying the slice? Otherwise, this should perhaps take a `Cow` instead. Note that the closure below re-borrows the items.

Unfortunately I cannot compile-test and double-check this to give you better suggestions, the rest of the code fails to compile with a missing trait bound for `to_glib_none()` on `&[impl AsRef<str>]`.

gio/src/desktop_app_info.rs
```diff
@@ -129,7 +129,7 @@ impl<O: IsA<DesktopAppInfo>> DesktopAppInfoExtManual for O {
             let mut error = ptr::null_mut();
             let _ = ffi::g_desktop_app_info_launch_uris_as_manager_with_fds(
                 self.as_ref().to_glib_none().0,
-                uris.to_glib_none().0,
+                uris.as_ref().to_glib_none().0,
```
MarijnS95, line 132: Might also not be necessary given the other review comment.

MarijnS95:
> @MarijnS95 those are all manually added. That's why I wrote "I still need to update the manual code, but that is easy." :) Those occurrences are not generated by gir.

Gotcha, I think I misread your original comment, but you definitely said that :)

I cannot mark the comments as resolved though, not enough rights here :/

rafaelcaricio:
> the rest of the code fails to compile with a missing trait bound for `to_glib_none()` on `&[impl AsRef<str>]`.

This is where I'm stuck. I cannot find a way to implement that trait bound without conflicting with existing trait bounds. 😢

gio/src/app_info.rs
```diff
@@ -93,19 +93,14 @@ impl<O: IsA<AppInfo>> AppInfoExtManual for O {
...
         let context = context.map(ToOwned::to_owned);
         Box_::pin(crate::GioFuture::new(self, move |obj, send| {
             let cancellable = Cancellable::new();
-            let uris = uris
-                .iter()
-                .map(::std::borrow::Borrow::borrow)
-                .collect::<Vec<_>>();
             obj.launch_uris_async(
-                uris.as_ref(),
+                uris,
```
rafaelcaricio, line 103: Not sure this will actually compile, maybe after the trait bounds validation the compiler might complain about this. But I can work out a solution for that later, that is fine..

MarijnS95:
> This is where I'm stuck. I cannot find a way to implement that trait bound without conflicting with existing trait bounds.

It's a really tricky bugger indeed. I tried the obvious in implementing (with `todo!()`, no time to waste on the actual code just yet):

```rust
impl<'a, S: AsRef<str> + ToGlibPtr<'a, *const c_char>>
    ToGlibContainerFromSlice<'a, *const *const c_char> for S
```
(And later the various `mut` variants)

And that conflicts with the existing implementations. Neat I'd say, those should all be able to go through `AsRef` as a single shared implementation... Comment them out but noupe, now other traits are missing :/

sdroege:
> It's a really tricky bugger indeed.

The alternative would be to go with `let x = x.iter().map(|i| i.as_ref()).collect::<Vec<_>>()` as a first step. All the strings have to be reallocated and copied to a new vec anyway for the NUL-termination, doing another shallow copy of the slice is not going to make things noticeably worse :)

MarijnS95:
> The alternative would be to go with `let x = x.iter().map(|i| i.as_ref()).collect::<Vec<_>>()` as a first step. All the strings have to be reallocated and copied to a new vec anyway for the NUL-termination, doing another shallow copy of the slice is not going to make things noticeably worse :)

We'd have to insert this for autogenerated code too, but I think the `Transformation` system already lets you specify that with ease?

Alternatively `Bound` has its own `extra_to_glib` (or something like that). `&x.iter().map(|i| i.as_ref()).collect::<Vec<_>>()` should work though as the temporary is only dropped after the function is called.

rafaelcaricio:
@MarijnS95 Yes, I tried multiple ways... nothing worked, it always conflicted with something else.

The suggestion from @sdroege was what I also think would be the best option. Any pointers to where to add that in gir are appreciated. 😄

sdroege:
This would rather use `impl IntoStrV` now, and that's tracked by another issue already

Issue #2, opened 2021-05-13
No bindings for the g_win32 methods


nacho:
There seems to be missing all those methods for windows. Probably because they are not part of the gir on linux

- [ ] `g_win32_check_windows_version ()`
- [ ] `g_win32_get_command_line ()`
- [ ] `g_win32_error_message ()`
- [ ] `g_win32_getlocale ()`
- [ ] `g_win32_get_package_installation_directory ()`
- [x] `g_win32_get_package_installation_directory_of_module ()`
- [ ] `g_win32_get_package_installation_subdirectory ()`
- [ ] `g_win32_get_windows_version ()`
- [ ] `g_win32_locale_filename_from_utf8 ()`

Details at https://developer.gnome.org/glib/stable/glib-Windows-Compatibility-Functions.html

sdroege:
Some are there but yes, many are missing :) You'd have to add them manually like the ones that already exist. Any specific ones you're missing?

nacho:
The one that I currently need is `g_win32_get_package_installation_directory_of_module`. I made a quick git grep in the repo and I could not find any g_win32 method.

sdroege:
gio/src/win32_input_stream.rs for example.

sdroege:
I guess there's more to fix here.

bilelmoussaoui:
I updated the issue and added a list of the functions based on GLib docs which I linked in the original issue as well

bilelmoussaoui:
Can be closed now

sdroege:
That were all?

PR #90, opened 2021-05-22
Manually bind g_win32_get_package_installation_directory_of_module

pbor:
Partially fixes https://github.com/gtk-rs/gtk-rs-core/issues/2

### DIFF
--- a/glib/src/win32.rs
+++ b/glib/src/win32.rs
@@ -1,0 +1,15 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::translate::*;
+use std::path::PathBuf;
+
+#[doc(alias = "g_win32_get_package_installation_directory_of_module")]
+pub fn win32_get_package_installation_directory_of_module(
+    hmodule: ffi::gpointer,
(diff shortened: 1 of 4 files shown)


GuillaumeGomez (approved):

glib/src/win32.rs
```diff
@@ -0,0 +1,18 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::translate::*;
+use std::ffi::OsString;
+
+#[doc(alias = "g_win32_get_package_installation_directory_of_module")]
+pub fn win32_get_package_installation_directory_of_module(
+    hmodule: ffi::gpointer,
+) -> Option<OsString> {
```
pbor, line 9: Shall I return `PathBuf` instead of `OsString`?

sdroege: `PathBuf` if it's actually a path :)

sdroege (commented):
Looks good to me otherwise, thanks! :)

sdroege (approved):

PR #104, opened 2021-05-29
Expose the type of pango::Attribute

jsparber:


bilelmoussaoui:
I don't think we should actually be exposing AttrClass at all but rather uses it directly in Attribute?

jsparber:
> I don't think we should actually be exposing AttrClass at all but rather uses it directly in Attribute?

I think it makes sense to move it to `Attribute::type_()`.

sdroege:
Yeah, can you add it there? Having it also on the class is OK though

jsparber:
> Yeah, can you add it there? Having it also on the class is OK though

done. :)

sdroege (approved):

PR #108, opened 2021-06-01
Generate missing doc aliases for newtypes

GuillaumeGomez:
Based on https://github.com/gtk-rs/checker/pull/27


glib/src/enums.rs
```diff
@@ -561,6 +564,7 @@ impl Clone for FlagsClass {
 
 /// Representation of a single flags value of a `FlagsClass`.
 #[derive(Debug, Clone)]
+#[doc(alias = "GFlagsValue, FlagsClass")]
```
GuillaumeGomez, line 567: Oh that's a funny bug!

sdroege: lol

sdroege (approved):

PR #113, opened 2021-06-02
Generate missing doc aliases

GuillaumeGomez:


glib/src/gobject/flags.rs
```diff
@@ -3,6 +3,7 @@
 use crate::translate::*;
 
 bitflags::bitflags! {
+    #[doc(alias = "GParamFlags")]
     pub struct ParamFlags: u32 {
         const READABLE = 1;
```
bilelmoussaoui, line 8: Would be nice if those are using the ffi values. Also they seem to lack a doc alias

GuillaumeGomez: Ah indeed! One stone, two pigeons. :D (I need the first to get the second)

glib/src/gobject/flags.rs
```diff
@@ -3,24 +3,30 @@
...
-        const USER_2 = 1024;
-        const USER_3 = 2048;
-        const USER_4 = 4096;
-        const USER_5 = 8192;
-        const USER_6 = 16384;
-        const USER_7 = 32768;
-        const USER_8 = 65536;
-        const EXPLICIT_NOTIFY = 1073741824;
-        const DEPRECATED = 2147483648;
+        #[doc(alias = "G_PARAM_READABLE as u32")]
```
bilelmoussaoui, line 8: Hmmm

bilelmoussaoui (changes requested):

glib/src/gobject/flags.rs
GuillaumeGomez: HAHA

bilelmoussaoui: `alias.split()[0]` :p

GuillaumeGomez: Something like that. ;)

bilelmoussaoui (approved):

glib/src/gobject/flags.rs
```diff
@@ -3,24 +3,30 @@
...
-        const LAX_VALIDATION = 16;
-        const USER_0 = 128;
-        const USER_1 = 256;
-        const USER_2 = 1024;
-        const USER_3 = 2048;
-        const USER_4 = 4096;
-        const USER_5 = 8192;
-        const USER_6 = 16384;
-        const USER_7 = 32768;
-        const USER_8 = 65536;
```
sdroege, line 21: Please add these back, thanks :)

glib/src/gobject/flags.rs
```diff
@@ -3,24 +3,30 @@
...
+        #[doc(alias = "G_PARAM_READWRITE")]
+        const READWRITE = gobject_ffi::G_PARAM_READWRITE as u32;
+        #[doc(alias = "G_PARAM_CONSTRUCT")]
+        const CONSTRUCT = gobject_ffi::G_PARAM_CONSTRUCT as u32;
+        #[doc(alias = "G_PARAM_CONSTRUCT_ONLY")]
+        const CONSTRUCT_ONLY = gobject_ffi::G_PARAM_CONSTRUCT_ONLY as u32;
+        #[doc(alias = "G_PARAM_LAX_VALIDATION")]
+        const LAX_VALIDATION = gobject_ffi::G_PARAM_LAX_VALIDATION as u32;
+        #[doc(alias = "G_PARAM_STATIC_NAME")]
+        const STATIC_NAME = gobject_ffi::G_PARAM_STATIC_NAME as u32;
```
sdroege, line 21: This makes no sense from Rust, same for the other `G_PARAM_STATIC_...` flags

GuillaumeGomez: Ok§

glib/src/gobject/flags.rs
```diff
@@ -3,13 +3,20 @@
...
+        const WRITABLE = gobject_ffi::G_PARAM_WRITABLE as u32;
+        #[doc(alias = "G_PARAM_READWRITE")]
+        const READWRITE = gobject_ffi::G_PARAM_READWRITE as u32;
+        #[doc(alias = "G_PARAM_CONSTRUCT")]
+        const CONSTRUCT = gobject_ffi::G_PARAM_CONSTRUCT as u32;
+        #[doc(alias = "G_PARAM_CONSTRUCT_ONLY")]
+        const CONSTRUCT_ONLY = gobject_ffi::G_PARAM_CONSTRUCT_ONLY as u32;
+        #[doc(alias = "G_PARAM_LAX_VALIDATION")]
+        const LAX_VALIDATION = gobject_ffi::G_PARAM_LAX_VALIDATION as u32;
         const USER_0 = 128;
```
sdroege, line 20: Actually `USER_0` makes no sense. It starts at `1<<8`, which is `USER_1 = 256`

GuillaumeGomez: Removed.

sdroege (approved):

PR #116, opened 2021-06-03
Fix glib re-export detection for macros

fengalin:
When a crate includes 2 `glib` dependent known crates, the first known is
selected to prepend to `glib` in macros. If this first crate is optional
and not selected, it is not reachable and causes compilation error.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -104,6 +104,10 @@ jobs:
         run: cargo build --manifest-path glib-macros/Cargo.toml
       - name: "glib-macros: test"
         run: cargo test --manifest-path glib-macros/Cargo.toml
+      # glib-macros in 2 levels dependent
+      - name: Check two levels glib dependent
+        run: cargo check
+        working-directory: tests/two-levels-glib-dependent
       # examples
       - name: "examples"
         run: cargo build --manifest-path examples/Cargo.toml --bins --examples --all-features
--- a/glib-macros/src/utils.rs
+++ b/glib-macros/src/utils.rs
@@ -168,7 +168,7 @@ const KNOWN_GLIB_EXPORTS: [&str; 10] = [
 pub fn crate_ident_new() -> TokenStream {
     use proc_macro_crate::FoundCrate;
 
-    let crate_path = match crate_name("glib") {
+    match crate_name("glib") {
         Ok(FoundCrate::Name(name)) => Some(name),
         Ok(FoundCrate::Itself) => Some("glib".to_string()),
         Err(_) => None,
@@ -176,26 +176,19 @@ pub fn crate_ident_new() -> TokenStream {
     .map(|s| {
         let glib = Ident::new(&s, Span::call_site());
         quote!(#glib)
-    });
-
-    let crate_path = crate_path.or_else(|| {
-        KNOWN_GLIB_EXPORTS.iter().find_map(|c| {
-            if let Ok(f) = crate_name(c) {
-                let crate_name = match f {
-                    FoundCrate::Name(name) => name,
-                    FoundCrate::Itself => c.to_string(),
-                };
-                let crate_root = Ident::new(&crate_name, Span::call_site());
+    })
+    .or_else(|| {
+        KNOWN_GLIB_EXPORTS.iter().find_map(|c| match crate_name(c) {
+            Ok(FoundCrate::Itself) => {
+                let crate_root = Ident::new(&c, Span::call_site());
                 Some(quote::quote! {
                     #crate_root::glib
                 })
-            } else {
-                None
             }
+            _ => None,
         })
-    });
-
-    crate_path.unwrap_or_else(|| {
+    })
+    .unwrap_or_else(|| {
         proc_macro_error::emit_call_site_warning!(
             "Can't find glib crate. Please ensure you have a glib in scope"
         );
--- a/tests/two-levels-glib-dependent/Cargo.toml
+++ b/tests/two-levels-glib-dependent/Cargo.toml
@@ -1,0 +1,7 @@
+[workspace]
+
+members = [
+    "gstreamer",
+    "gtk",
+    "glib-dependent-dependent"
+]
--- a/tests/two-levels-glib-dependent/README.md
+++ b/tests/two-levels-glib-dependent/README.md
@@ -1,0 +1,10 @@
+# glib-rs Two Levels Dependent Test
+
+This repository is intended at checking `glib` macro
+re-export detection in two levels dependencies.
+
+The detection mechanism used to stop at the first identified crate known
+to re-export `glib` and which was found in `Cargo.toml`. When used in a crate
+that depends both on `gstreamer` and `gtk` and `gtk` is optional, the detection
+mechanism stopped at `gtk` and preprended `gtk` to `glib`, leading to errors
+compiling `glib` proc-macros.
--- a/tests/two-levels-glib-dependent/glib-dependent-dependent/Cargo.toml
+++ b/tests/two-levels-glib-dependent/glib-dependent-dependent/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "glib-dependent-dependent"
+version = "0.1.0"
+authors = ["François Laignel <fengalin@free.fr>"]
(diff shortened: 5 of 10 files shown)


fengalin:
Need a couple iteration to fix CI new check.

.github/workflows/CI.yml
```diff
@@ -55,6 +55,11 @@ jobs:
       - run: cargo run --release -- --no-manual-traits ../glib-macros
         working-directory: checker
         if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
+      - name: Check two levels glib dependent
+        run: |
+          git clone https://github.com/fengalin/glib-rs-two-levels-dependent
+          cd glib-rs-two-levels-dependent && cargo c
```
sdroege, line 61:
```suggestion
          cd glib-rs-two-levels-dependent && cargo check
```

I guess?

fengalin: `c` is short for `check` :)

sdroege: TIL

.github/workflows/CI.yml
```diff
@@ -55,6 +55,11 @@ jobs:
       - run: cargo run --release -- --no-manual-traits ../glib-macros
         working-directory: checker
         if: matrix.rust == 'nightly' && matrix.conf.name == 'glib'
+      - name: Check two levels glib dependent
```
bilelmoussaoui, line 58: it should probably use the checkout action (see few lines above)

fengalin: Indeed. Thanks @bilelmoussaoui !

.github/workflows/CI.yml
```diff
@@ -61,6 +61,21 @@ jobs:
           cd .. && git diff --exit-code
         working-directory: checker
         if: matrix.rust == 'nightly'
+      - uses: actions/checkout@v2
+        with:
+          repository: fengalin/glib-rs-two-levels-dependent
```
GuillaumeGomez, line 66: I guess we'll have to move this repository into gtk-rs at some point (and rename it too)?

fengalin: OC. Let me try to fix CI, and we'll handle this.

GuillaumeGomez: Sure, it was mostly a reminder for future me. :)

.github/workflows/CI.yml
```diff
@@ -104,6 +104,27 @@ jobs:
...
+          echo "BRANCH_NAME=$(echo ${{ github.head_ref }})" >> $GITHUB_ENV
+      - name: Check two levels glib dependent
+        run: |
+          export SPECS="{ git = \"https://github.com/${{ env.ACTOR }}/gtk-rs-core\", branch = \"${{ env.BRANCH_NAME }}\" }"
+          printf "\n\n[patch.'https://github.com/gtk-rs/gtk-rs-core']\n" >> Cargo.toml
+          echo "glib-sys = ${SPECS}" >> Cargo.toml
+          echo "gobject-sys = ${SPECS}" >> Cargo.toml
+          echo "glib = ${SPECS}" >> Cargo.toml
+          cat Cargo.toml
+          cargo check
```
fengalin, line 126: I couldn't find a better way to patch `Cargo.toml` with the PR author's branch. This could be a problem if they didn't use `gtk-rs-core` for their fork.

.github/workflows/CI.yml
fengalin: How would you like to proceed? You create a fork of [my repository](https://github.com/fengalin/glib-rs-two-levels-dependent) on behalf of `gtk-rs` and I update the CI script?

GuillaumeGomez: Not sure actually... Will this repository be used somewhere else? If not, we could simply create another folder in this repository and then you run it from there. Otherwise you can give me admin rights on the repository so I can migrate it here.

fengalin: > Will this repository be used somewhere else?

I don't think so.

> we could simply create another folder in this repository and then you run it from there

I didn't want to do this initially because we need the first glib dependent to use a name in `KNOWN_GLIB_EXPORTS` which is not the first of the list. That's why I used `gstreamer`. But I guess that would be fine if it's under a `test` folder.

GuillaumeGomez: Then let's go for an extra folder. Maybe we should create a "glib-tests" folder and put them there instead of having multiple places...

.github/workflows/CI.yml
bilelmoussaoui: I think we can just have  a new tests directory with subprojects that do some different tests that can't fit in a classic test situation. This way we don't need to do much other than cargo build/cargo check the tests in that directory.

fengalin: Yes, that's the solution we came up with :) https://github.com/gtk-rs/gtk-rs-core/pull/116#pullrequestreview-675242780

bilelmoussaoui: Oh my bad

fengalin: np :)

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

bilelmoussaoui:
This is caused a regression in glib::clone!
```
error[E0433]: failed to resolve: use of undeclared crate or module `glib`
  --> examples/grid_packing/main.rs:49:33
   |
49 |     quit_button.connect_clicked(clone!(@weak window => move |_| window.destroy()));
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `glib`
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: use of undeclared crate or module `glib`
```

glib-macros/src/utils.rs
```diff
@@ -168,34 +168,27 @@ const KNOWN_GLIB_EXPORTS: [&str; 10] = [
...
     .map(|s| {
         let glib = Ident::new(&s, Span::call_site());
         quote!(#glib)
-    });
-
-    let crate_path = crate_path.or_else(|| {
-        KNOWN_GLIB_EXPORTS.iter().find_map(|c| {
-            if let Ok(f) = crate_name(c) {
-                let crate_name = match f {
-                    FoundCrate::Name(name) => name,
```
bilelmoussaoui, line 185: what's the reason for dropping this?

fengalin: See the PR description: it stopped at first known crate found in Cargo.toml, even if optional.

fengalin:
@bilelmoussaoui try `use gtk::glib::{self, clone} `

glib-macros/src/utils.rs
bilelmoussaoui: That would be an issue to report to  proc-macro-crate though, no?

fengalin: Might be yes. I'm away though, so can't take care of it this weekend.

fengalin: I'm not sure why it fails in this particular case though.

bilelmoussaoui: It fails because gtk4 re-exports glib and we expect people to use the re-exported glib without having to re-add glib as a dependency. So the applications/examples don't directly on glib at all but indirectly through gtk4/gtk3. I have opened https://github.com/gtk-rs/gtk-rs-core/issues/127 so it's not get forgotten

fengalin: Yes, that's the use case I addressed too (in gst-plugins-rs/utils).

bilelmoussaoui: Ah I see, I think the issue might come from the Downgrade/Upgrade traits.

bilelmoussaoui:
> When a crate includes 2 `glib` dependent known crates, the first known is
> selected to prepend to `glib` in macros. If this first crate is optional
> and not selected, it is not reachable and causes compilation error.

unrelated to the changes, but which crates have such use case?

fengalin:
gst-plugins-rs/utils

PR #119, opened 2021-06-04
CI docs: define releases env variable for all steps

bilelmoussaoui:
so that the pre-docs step generates a link to the latest release correctly

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,6 +13,9 @@ jobs:
     name: build
     container:
       image: ghcr.io/gtk-rs/gtk-rs/gtk:latest
+    env:
+      RELEASES: |
+        0.10=0.10
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


sdroege (approved):

GuillaumeGomez:
Just wondering: does it mean we have to add a new version every time we make a release?

bilelmoussaoui:
> Just wondering: does it mean we have to add a new version every time we make a release?

post release yes

GuillaumeGomez:
Urg... More manual stuff...

bilelmoussaoui:
btw, it's used here https://gtk-rs.org/gtk-rs-core/ and that page is generated everytime the CI runs (for now both on release and git, for testing purposes, but in the future it would be only when we push to master)

bilelmoussaoui:
> Urg... More manual stuff...

I don't think there's a way to make any of this more automatic, because then you will have to maintain more obscure scripts that no one understands :)

PR #126, opened 2021-06-05
Add minimal crate info

sophie-h:
Also adds minimum required version of the library to every crate doc.

### DIFF
--- a/cairo/src/lib.rs
+++ b/cairo/src/lib.rs
@@ -5,33 +5,35 @@
 //! This library contains safe Rust bindings for [Cairo](https://www.cairographics.org/).
 //! It is a part of [Gtk-rs](https://gtk-rs.org/).
 //!
-//! ## Crate features
+//! Cairo 1.14 is the lowest supported version for the underlying library.
 //!
-//! ### Default-on features
+//! # Crate features
+//!
+//! ## Default-on features
 //!
 //! * **use_glib** - Use with [glib](https://gtk-rs.org/docs/glib/)
 //!
-//! ### Fileformat features
+//! ## Fileformat features
 //!
 //! * **png** - Reading and writing PNG images
 //! * **pdf** - Rendering PDF documents
 //! * **svg** - Rendering SVG documents
 //! * **ps** - Rendering PostScript documents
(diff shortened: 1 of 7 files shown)


sophie-h:
Will do for gtk3/4 after review

GuillaumeGomez:
That sounds quite complicated to maintain... Could we use `#[doc = include_str!("version.md")]` instead for the line containing the version?

sdroege (approved):

sdroege:
> That sounds quite complicated to maintain... Could we use `#[doc = include_str!("version.md")]` instead for the line containing the version?

And where would this `version.md` come from? :)

sophie-h:
> And where would this `version.md` come from? :)

That would have to be gir's task I guess?

bilelmoussaoui:
> That sounds quite complicated to maintain... Could we use `#[doc = include_str!("version.md")]` instead for the line containing the version?

We had to maintain it as part of a README/ (website but not really because it was outdated)

sophie-h:
Still it's something that gir knows. Can gir just dump a `library_version.md` in `src`?

GuillaumeGomez:
> Still it's something that gir knows. Can gir just dump a `library_version.md` in `src`?

Good idea! Just need to check if `#[doc = include_str!()]` is stable or not first.

GuillaumeGomez:
Unfortunately it's not stable yet: https://github.com/rust-lang/rust/pull/83366 (it'll be part of 1.54).

sophie-h:
`#![cfg_attr(feature = "dox", doc = include_str!("version.md"))]` works for me

sophie-h:
Ah, only on nightly :facepalm:

sophie-h:
Merge as is for now?

GuillaumeGomez:
You were about to make a revolution in how I saw versions in rust, too bad haha. That would have been super cool if it worked though.

GuillaumeGomez (approved):

GuillaumeGomez:
Yes, thanks for doing this! Can you open an issue so we can update how we handle this once 1.54 is out please?

Issue #129, opened 2021-06-06
Update clone macro documentation to mention it needs glib in scope


GuillaumeGomez:

fengalin:
Did you think about the examples in https://github.com/gtk-rs/gtk-rs-core/blob/master/glib-macros/src/lib.rs or is there any other places that I should check?

GuillaumeGomez:
More like to mention that `glib` needs to be in scope.

fengalin:
Where? Why? I mean, I added this to the examples.

GuillaumeGomez:
In the `clone!` macro doc. Just in the code example is far from enough. It's definitely not obvious just by reading the code example that `glib` needs to be in scope. It's just there, but we don't really know why.

fengalin:
> It's just there, but we don't really know why.

Do we need to explain why? The compiler will complain explicitly anyway.

GuillaumeGomez:
Yes we need to, otherwise it just says "we need glib because the dark gods from the ancient times required a sacrifice but it was to refused to them and they put the blame on glib, etc", or whatever other reason you can think of. If it's not obvious, it needs to be explained.

fengalin:
I was not involved in the design nor the implementation of these macros. I just spent at least 2 hours reverse-engineering the code which doesn't document anything about the cyclic dependencies, nor anything required to understand how it is designed, and I still can't tell why it wasn't possible to achieve what `serde` was able to achieve. So I suggest that someone who knows better adds a commit to explain that.

sdroege:
Maybe @gdesmott remembers something?

gdesmott:
> Maybe @gdesmott remembers something?

Hum no, I wasn't involved with this macro.

sdroege:
Hm I thought the concern here was generally about the proc-macro-crate which is also used by the other macros. But that was apparently not added in your versions yet and was added much later by @GuillaumeGomez so he should be the best person to explain things ;) Sorry

GuillaumeGomez:
I basically asked to add this in the documentation:

> The `clone!` macro is expecting to have `glib` in scope in order to call traits from it, so please be sure it's available.

Doesn't seem too much for me...

sdroege:
And yes that would be good to document, but also it's unfortunate that this is needed at all :)

GuillaumeGomez:
glib is very rarely a direct dependance so we have the choice between trying to guess which crate to use for it or simply ask nicely our users to import it in the scope. :)

[...]

Issue #127, opened 2021-06-05
regression from #116


bilelmoussaoui:
see https://github.com/gtk-rs/gtk-rs-core/pull/116#issuecomment-855253011

logs from gtk4-rs examples:
``` 

error[E0433]: failed to resolve: use of undeclared crate or module `glib`
  --> examples/grid_packing/main.rs:49:33
   |
49 |     quit_button.connect_clicked(clone!(@weak window => move |_| window.destroy()));
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `glib`
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: use of undeclared crate or module `glib`
  --> examples/grid_packing/main.rs:49:33
   |
49 |     quit_button.connect_clicked(clone!(@weak window => move |_| window.destroy()));
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `glib::clone`
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this trait
   |
1  | use crate::clone::Downgrade;
   |

error[E0433]: failed to resolve: use of undeclared crate or module `glib`
  --> examples/grid_packing/main.rs:49:33
   |
49 |     quit_button.connect_clicked(clone!(@weak window => move |_| window.destroy()));
   |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `glib::clone`
   |
   = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this trait
   |
1  | use crate::clone::Upgrade;
   |

error: aborting due to 3 previous errors

For more information about this error, try `rustc --explain E0433`.
error: could not compile `gtk4-rs-examples`
```

fengalin:
After a quick check, this is how I see it:

[...]

PR #132, opened 2021-06-07
glib-macros: remove KNOWN_GLIB_EXPORTS

fengalin:
See https://github.com/gtk-rs/gtk-rs-core/issues/127 for details.

Crates checked locally:

- `gtk-rs-core`.
- `gtk3-rs`.
- `gtk4-rs`.
- `gstreamer-rs`.
- `gst-plugins-rs`.

Except for `clone!`, I didn't spot any explicit use of `glib::` in the other macros.

I think full CI should be launched on these crates after this is merged though.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/127
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/129

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -51,7 +51,8 @@ use syn::{parse_macro_input, DeriveInput, LitStr};
 /// ### Passing a strong reference
 ///
 /// ```
-/// # use glib_macros::clone;
+/// use glib;
+/// use glib_macros::clone;
 /// use std::rc::Rc;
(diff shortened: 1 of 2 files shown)


sdroege:
Does this require users to explicitly depend on glib now instead of indirectly via e.g. gtk?

fengalin:
> Does this require users to explicitly depend on glib now instead of indirectly via e.g. gtk?

sdroege (approved):

PR #135, opened 2021-06-10
glib: Duplicate some ObjectClass methods to Interface

jsparber:


glib/src/object.rs
```diff
@@ -2941,6 +2954,56 @@ impl<T: IsInterface> Interface<T> {
...
+    }
+
+    #[doc(alias = "g_object_interface_find_property")]
+    pub fn find_property<'a, N: Into<&'a str>>(
+        &self,
+        property_name: N,
+    ) -> Option<crate::ParamSpec> {
+        let property_name = property_name.into();
+        unsafe {
+            let klass = self as *const _ as *const gobject_ffi::GObjectClass;
```
sdroege, line 2986:
```suggestion
            let klass = self as *const _ as *const gobject_ffi::GTypeInterface;
```

glib/src/object.rs
```diff
@@ -2941,6 +2954,56 @@ impl<T: IsInterface> Interface<T> {
             }
         }
     }
+
+    pub fn has_property<'a, N: Into<&'a str>>(
```
sdroege, line 2958: Should be on an `impl<T: IsA<Object>> Interface<T> {}` block

glib/src/object.rs
```diff
@@ -2941,6 +2954,56 @@ impl<T: IsInterface> Interface<T> {
...
+                klass as *mut _,
+                property_name.to_glib_none().0,
+            ))
+        }
+    }
+
+    #[doc(alias = "g_object_interface_list_properties")]
+    pub fn list_properties(&self) -> Vec<crate::ParamSpec> {
+        unsafe {
+            let klass = self as *const _ as *const gobject_ffi::GObjectClass;
```
sdroege, line 2998:
```suggestion
            let klass = self as *const _ as *const gobject_ffi::GTypeInterface;
```

glib/src/object.rs
```diff
@@ -2908,6 +2908,19 @@ impl<T: IsInterface> Interface<T> {
         }
     }
 
+    /// Gets the interface struct for `Self` from `type_`.
+    pub fn from_type(type_: Type) -> InterfaceRef<'static, T> {
```
sdroege, line 2912: Needs to check for `NULL` and return an `Option`

sdroege (approved):

PR #136, opened 2021-06-10
Let cancellable be passed into GioFuture instead of creating it by the closure

sdroege:
Preparation for some new API.

Depends on https://github.com/gtk-rs/gir/pull/1182


sdroege:
Will update gtk3-rs / gtk4-rs once this is merged.

GuillaumeGomez (approved):

PR #139, opened 2021-06-12
CI: use apt instead of dnf

bilelmoussaoui:
because it's an ubuntu based image instead of a fedora one

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -41,7 +41,6 @@ jobs:
       - name: Grab gtk-rs LOGO
         if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
         run: |
-          apt-get install wget -y
           wget https://raw.githubusercontent.com/gtk-rs/gtk-rs.github.io/master/logo/gtk-rs-16.png -o ./target/doc/favicon-16x16.png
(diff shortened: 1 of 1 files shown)


sdroege:
How does anything work at all?

GuillaumeGomez:
I guess it doesn't.

bilelmoussaoui:
> How does anything work at all?

which things? the docs CI? it doesn't after my PR. The thing is this job step is only triggered on master/on release that's why

sdroege:
Oh I see

.github/workflows/docs.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Grab gtk-rs LOGO
         if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
         run: |
-          sudo dnf install wget -y
+          apt-get install wget -y
```
sdroege, line 44: Nowadays you'd usually just use `apt` instead of `apt-get` FWIW

sdroege (approved):

.github/workflows/docs.yml
bilelmoussaoui: Actually, let's move the wget install to the docker image, there's no reason to re-install it everytime here

bilelmoussaoui:
don't merge until #140 is merged and the image got updated

bilelmoussaoui:
@GuillaumeGomez this can be merged now as well :)

GuillaumeGomez (approved):

PR #145, opened 2021-06-14
Fix return value for File::for_commandline_arg

GuillaumeGomez:
Fixes similar case of https://github.com/gtk-rs/gtk3-rs/issues/351.

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -546,6 +546,8 @@ manual_traits = ["FileExtManual"]
         [[object.function.parameter]]
         name = "arg"
         string_type = "os_string"
+        [object.function.return]
+        nullable = true
     [[object.function]]
     name = "new_for_commandline_arg_and_cwd"
     # Not annotated as constructor in Gir => force it to apply naming convention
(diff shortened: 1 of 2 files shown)


gio/Gir.toml
```diff
@@ -546,6 +546,8 @@ manual_traits = ["FileExtManual"]
         [[object.function.parameter]]
         name = "arg"
         string_type = "os_string"
+        [object.function.return]
+        nullable = true
```
bilelmoussaoui, line 550: This can never be nullable per the docs "This operation never fails, but the returned object might not support any I/O operation if arg points to a malformed path."

GuillaumeGomez: The issue seems to say otherwise though...

GuillaumeGomez: And the source code seems to do it too: https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gfile.c#L6728

The doc is lying.

GuillaumeGomez:
The problem isn't the returned value. Closing then.

PR #146, opened 2021-06-18
glib: Make `SignalHandlerId` implement `Clone` and `Copy`

ids1024:
It's just a wrapper around a non-zero integer, so I don't see any reason it can't be `Copy`.

### DIFF
--- a/glib/src/signal.rs
+++ b/glib/src/signal.rs
@@ -41,7 +41,7 @@ use std::num::NonZeroU64;
 ///     }
 /// }
 /// ```
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct SignalHandlerId(NonZeroU64);
 
 impl SignalHandlerId {


GuillaumeGomez:
I recall that there was one though. If so, we should definitely add a comment explaining why it shouldn't implement `Copy` so other developers don't wonder about it as well.

cc @sdroege

ids1024:
I have trouble imagining what could be a problem with clone/copy here that couldn't also be a problem just moving the value, but perhaps someone can point something out.

GuillaumeGomez:
Let's wait for @sdroege. Maybe it's just me being confused with something else...

sdroege:
Yes there is a reason and we should document it together with some example code for the common reasons why people want it.

There were already a couple PRs with exactly the same change. Look there for the reason but I'll write it here later again too.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sdroege:
The reason for this is that if they were `Copy` / `Clone` then you could disconnect a signal handler multiple times. The first time this works correctly, the other times you get a critical warning at best or disconnect some other signal handler if you're unlucky enough.

While this is still not perfect (you can disconnect a signal handler id of one object from another object, for example), this forces you to think about handling the signal handler ids correctly.

sdroege:
Re-opening so we can document this properly somewhere. Want to update this PR accordingly?

ids1024:
> (you can disconnect a signal handler id of one object from another object, for example)

Too bad it isn't possible/practical to enforce this through the type system.

For my use case, I wanted a copy of `SignalHandlerId` in a couple places to block/unblock the signal. I guess using `Rc` is the correct solution, and could be mentioned in the documentation.

GuillaumeGomez:
Well, I mean, it's literally how Rust works... If you have one instance of an object you can't copy but still want to use it in multiple places, either you use a static or a `Rc`. :)

sdroege:
Yes but it's not obvious to people, we should document that (and `Option` / `Option::take()`) :)

GuillaumeGomez:
Seems a bit too much in my opinion but someone adds it, I'm good with it (never too much doc I guess :) ).

sdroege:
If we don't add that we'll get the seventeenth PR for this in a bit :P

GuillaumeGomez:
:fearful:

A6GibKm:
How to work around the inconvenient type is documented in https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/signal.rs#L13.

sdroege:
Hm so it's even documented already. Thanks!

PR #148, opened 2021-06-20
glib: implement Clone on SourceId

gdesmott:
I need to store the SourceId in a struct so it's automatically removed
on Drop. But I cannot do so as the SourceId does not implement Copy or
Clone and source_remove() consumes the SourceId.

### DIFF
--- a/glib/src/signal.rs
+++ b/glib/src/signal.rs
@@ -12,6 +12,9 @@ use std::num::NonZeroU64;
 
 /// The id of a signal that is returned by `connect`.
 ///
+/// This type does not implement `Clone` to prevent disconnecting
+/// the same signal handler multiple times.
+///
 /// ```ignore
 /// use glib::SignalHandlerId;
 /// use gtk::prelude::*;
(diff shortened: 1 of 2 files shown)


sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/146#issuecomment-864456305 , basically the same reason here.

You can either store an `Option<SourceId>` (and then use e.g. `Option::take()` once), or you can store a `glib::Source` directly and use the API on it instead.

sdroege:
Re-opening so we can document this properly somewhere. Want to update this PR accordingly, and maybe also for `SignalHandlerId`?

gdesmott:
Something like this?

GuillaumeGomez (approved):

GuillaumeGomez:
Yup, something like this, thanks!

sdroege:
I would've added a comment about `Option` and `Option::take` here, and in case of the `SourceId`s also the `glib::Source` API and its refcounting.

PR #149, opened 2021-06-21
Fix clippy warnings

GuillaumeGomez:
Following https://github.com/gtk-rs/gir/pull/1193.


.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.51.0"
```
bilelmoussaoui, line 20: Is it required for gir itself or for building the crates as well? if that's the case, we should bump the version in the readme as well

GuillaumeGomez: No it was simply a test, I updated the version for gir regen check, but it's not needed for the other crates. I removed this commit and made a more correct fix.

sdroege (approved):

PR #165, opened 2021-06-26
glib: Fix invalid types let from_variant panic

sophie-h:


glib/src/variant.rs
```diff
@@ -642,7 +646,22 @@ where
     V: FromVariant,
     H: BuildHasher + Default,
 {
+    /// ```
```
GuillaumeGomez, line 649: Maybe turn it into a code comment? Inner functions are not documented after all. :)

sdroege (approved):

glib/src/variant.rs
sophie-h: Will it still be picked up as a test?

sdroege: No, but you could also add a non-doctest here

```rust
#[test]
fn test_that_thing() {
    [...]
}
```

sdroege:
@GuillaumeGomez Should include that in the next bugfix release too

GuillaumeGomez:
Indeed, just waiting for the test to be moved.

sdroege (approved):

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/171

PR #167, opened 2021-06-27
uri: re-export URI parsing utilities

### LINKED ISSUE #164: `g_uri_parse_scheme()` no longer exposed
The `g_uri_parse_scheme()` is no longer available for builds below `v2_66` while that API has been available since GLib 2.16. https://developer.gnome.org/glib/stable/glib-URI-Functions.html#g-uri-parse-scheme

Can we have it back? Ubuntu 20.04 LTS doesn't have glib 2.66.

philn:
These are bindings for the GLib API available since GLib 2.16. They are also
exported in the `Uri` module available only when the v2_66 feature is enabled
though.

Fixes #164

### DIFF
--- a/glib/src/utils.rs
+++ b/glib/src/utils.rs
@@ -211,6 +211,54 @@ pub fn is_canonical_pspec_name(name: &str) -> bool {
     })
 }
 
+#[doc(alias = "g_uri_escape_string")]
+pub fn uri_escape_string(
+    unescaped: &str,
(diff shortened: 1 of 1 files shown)


sdroege (approved):

sdroege:
Should this also be backported to the 0.14 release?

philn:
Yes please, if it's not too much trouble!

PR #182, opened 2021-06-30
Run clippy on beta

GuillaumeGomez:


sdroege:
Do we only want to run on beta, or also on stable?

GuillaumeGomez:
Hum... This is a good question. Generally clippy adds more lints, rarely remove them or change their default level, no?

sdroege:
But it would probably be nice to also not have clippy warnings on current stable

GuillaumeGomez:
I'll update the commits to check on both.

sdroege (approved):

PR #188, opened 2021-07-01
Make stronger guarantees on `ToValue`

### LINKED ISSUE #4: impl ValueType for Value should be a no-op.
# Problem
`Value::get()` looks like it should act like `std::convert::TryInto` but it doesn't.

```rust
let x: Value = 0i32.to_value();
let y: Value = x.get().unwrap();
```

The above panics with `WrongValueType(ValueTypeMismatchError { actual: gint, requested: GValue })`

The above code expects the Value to be a Boxed Value.

What this means is that if a function returns a T: ValueType. It is impossible for that T to be a generic (unknown) Value).

an example from gtk-rs/gtk4-rs#332

`TreemodelExtManual::get() -> Value` does not have the same semantics as `TreeModelExtManual::get<T: ValueType>() -> T` as it is not possible to get an untouched Value from it.

# Possible Solution

Make `impl ValueType for Value` act as a no-op. Similar to how `impl<T> From<T> for T` acts.

The existing behaviour of a boxed value could be created by having a `impl ValueType for Box<Value>` or even a `struct Boxed(pub Value)`

piegamesde:
This more or less reverts https://github.com/gtk-rs/glib/commit/390ca3a2 ("Don't implement ToValue for Value").

Closes #4.

To be honest, I'm not 100% sure if this is the best way forward. The traits and trait bounds are really convoluted and hard to keep track of. For example, I don't know how well this works for `SendValue`. Let me know what you think.

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -246,6 +246,19 @@ where
 }
 
 /// Trait to convert a value to a `Value`.
+///
+/// Similar to other common conversion traits, the following invariants are guaranteed:
+///
+/// - **Invertibility**: `x.to_value().get().unwrap() == x`. In words, [`FromValue`] is the inverse of `ToValue`.
+/// - **Idempotence**: `x.to_value() == x.to_value().to_value()`.
+///   In words, applying `ToValue` multiple times yields the same result as applying it once.
+///   Idempotence also applies the other way around: `value.get::<Value>()` is a no-op.
+///
+/// There is also the possibility to wrap values within values, see [`BoxedValue`]. All (un-)boxing needs to be done
+/// manually, and will be preserved under the conversion methods.
+///
+/// The conversion methods may cause values to be cloned, which may result in reference counter changes or heap allocations depending
+/// on the source and target type.
 pub trait ToValue {
     /// Convert a value to a `Value`.
     fn to_value(&self) -> Value;
@@ -451,6 +464,71 @@ impl From<SendValue> for Value {
 impl Uninitialized for Value {
     unsafe fn uninitialized() -> Value {
         mem::zeroed()
+    }
+}
+
+impl ToValue for Value {
+    fn to_value(&self) -> Value {
+        self.clone()
+    }
+
+    fn value_type(&self) -> Type {
+        self.type_()
+    }
+}
+
+impl<'a> ToValue for &'a Value {
+    fn to_value(&self) -> Value {
+        (*self).clone()
+    }
+
+    fn value_type(&self) -> Type {
+        self.type_()
+    }
+}
+
+pub struct NopChecker;
+
+unsafe impl ValueTypeChecker for NopChecker {
+    type Error = std::convert::Infallible;
+    fn check(_value: &Value) -> Result<(), Self::Error> {
+        Ok(())
+    }
+}
+
+unsafe impl<'a> FromValue<'a> for Value {
+    type Checker = NopChecker;
+
+    unsafe fn from_value(value: &'a Value) -> Self {
+        value.clone()
+    }
+}
+
+unsafe impl<'a> FromValue<'a> for &'a Value {
+    type Checker = NopChecker;
+
+    unsafe fn from_value(value: &'a Value) -> Self {
+        value
+    }
+}
+
+impl ToValue for SendValue {
+    fn to_value(&self) -> Value {
+        self.0.clone()
+    }
+
+    fn value_type(&self) -> Type {
+        self.type_()
+    }
+}
+
+impl<'a> ToValue for &'a SendValue {
+    fn to_value(&self) -> Value {
+        self.0.clone()
+    }
+
+    fn value_type(&self) -> Type {
+        self.type_()
     }
 }
 
--- a/glib/tests/value.rs
+++ b/glib/tests/value.rs
@@ -1,0 +1,66 @@
+use glib::*;
+use std::ops::Deref;
+
+// FIXME all .get::<i32>() should be replaced with .get(); the compiler is totally able to infer the type itself.
+// But somehow without some tests are failing on Windows because the type inference doesn't work or something.
+
+/// Test that `ToValue` (and conversely, `FromValue`) uphold the promised invariants
+#[test]
+pub fn to_value_invariants() {
+    // Inverse
+    assert_eq!(0i32, 0i32.to_value().get::<i32>().unwrap());
+    assert_eq!(0i32, (&0i32.to_value()).get::<i32>().unwrap());
+
+    // Idempotence
+    assert_eq!(
+        &0i32.to_value().type_(),
+        &0i32.to_value().to_value().type_()
+    );
+    assert_eq!(0i32, 0i32.to_value().to_value().get::<i32>().unwrap());
+    assert_eq!(
+        0i32,
+        0i32.to_value()
+            .get::<Value>()
+            .unwrap()
+            .get::<i32>()
+            .unwrap()
+    );
+    assert_eq!(
+        0i32,
+        (&0i32.to_value())
+            .get::<Value>()
+            .unwrap()
+            .get::<i32>()
+            .unwrap()
+    );
+}
+
+/// Test that `ToValue` and `FromValue` handle nexted boxed values correctly (as per the documentation)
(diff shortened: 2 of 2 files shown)


glib/src/value.rs
```diff
@@ -454,6 +462,42 @@ impl Uninitialized for Value {
...
+    fn to_value(&self) -> Value {
+        self.clone()
+    }
+
+    fn value_type(&self) -> Type {
+        self.type_()
+    }
+}
+
+pub struct NopChecker;
```
sdroege, line 475: What is this good for and why is it public?

glib/src/value.rs
```diff
@@ -454,6 +462,42 @@ impl Uninitialized for Value {
     }
 }
 
+impl ToValue for Value {
```
sdroege, line 465:
```suggestion
impl<'a> ToValue for &'a Value {
```

Would (also) be useful to have. Can you add that?

glib/tests/value.rs
```diff
@@ -0,0 +1,13 @@
+use glib::*;
+
+/// Test that `ToValue` (and conversely, `FromValue`) uphold the promised invariants
+#[test]
+pub fn to_value_invariants() {
+    // Inverse
+    assert_eq!(0i32, 0i32.to_value().get().unwrap());
```
sdroege, line 7:
```suggestion
    assert_eq!(0i32, 0i32.to_value().get::<i32>().unwrap());
```

glib/src/value.rs
```diff
@@ -246,6 +246,14 @@ where
 }
 
 /// Trait to convert a value to a `Value`.
+///
+/// Similar to other common conversion traits, the following invariants are guaranteed:
+/// 
+/// - **Invertibility**: `x.to_value().get().unwrap() == x`. In words, `FromValue` is the inverse of `ToValue`.
+/// - **Idempotence**: `x.to_value() == x.to_value().to_value()`.
+///   In words, applying `ToValue` multiple times yields the same result as applying it once.
+///   Idempotence also applies the other way around: `value.get::<Value>()` is a no-op.
```
sdroege, line 255: That last part is not actually implemented here, is it?

piegamesde: Yes it is. The tests do `.to_value().get::<Value>().unwrap().get().unwrap()`.

glib/src/value.rs
piegamesde: It an unfortunate detail of the `FromValue` trait, see the lines that follow. It requires a `ValueTypeChecker` and it must be public. The existing `ValueTypeChecker` implementations didn't fit the current situation, so I had to write another one.

glib/src/value.rs
sdroege: Hm, where does the `FromValue` impl for `Value` come from?

sdroege: Oh!

glib/src/value.rs
sdroege: Makes sense

glib/tests/value.rs
piegamesde: I don't understand why this is required. The test compiles for me without it. But I've also seen the CI fail on this, and that's confusing.

piegamesde:
Maybe @mehmooda have a look on this too?

glib/src/value.rs
```diff
@@ -246,6 +246,14 @@ where
...
+
```
sdroege, line 256: Maybe add a comment here about `BoxedValue` for the rare case to actually store a `Value` inside a `Value`

glib/src/value.rs
```diff
@@ -454,6 +462,62 @@ impl Uninitialized for Value {
...
+}
+
+pub struct NopChecker;
+
+unsafe impl ValueTypeChecker for NopChecker {
+    type Error = std::convert::Infallible;
+    fn check(_value: &Value) -> Result<(), Self::Error> {Ok(())}
+}
+
+unsafe impl <'a> FromValue<'a> for Value {
```
sdroege, line 492:
```suggestion
unsafe impl <'a> FromValue<'a> for &'a Value {
```

Too maybe

glib/tests/value.rs
sdroege: Yeah also it happens only on that one Windows job for some reason... not sure, it shouldn't be required at all.

sdroege:
Needs a run of `cargo fmt` (in addition to the comments below) but otherwise this seems acceptable to me. @bilelmoussaoui @GuillaumeGomez ?

glib/src/value.rs
GuillaumeGomez: Please remove this empty line.

glib/src/value.rs
piegamesde: What for? The method already takes `&self`, there is no owned version. What you suggest would take an `&'a &'a self`, which is weird.

sdroege: No, it would return an `&'a Value` instead of a `Value`

piegamesde: Oh, I see.

sdroege (dismissed):
Good for me. How about others?

GuillaumeGomez (approved):

GuillaumeGomez:
Windows CI is still failing though. How weird...

piegamesde:
What do you suggest doing about the windows failure? I'm honestly out of debugging ideas. I could use `#cfg` flags to change the code when built on Windows or to disable that snippet entirely (it's only a test anyways).

sdroege:
Go with `get::<u32>()` and add a FIXME comment above

sdroege (dismissed):

sdroege:
@bilelmoussaoui @GuillaumeGomez So what do you think?

GuillaumeGomez:
I don't have opinion on this topic. I rarely use this type so I'm definitely not sure wether it's better or not...

glib/src/value.rs
```diff
@@ -246,6 +246,16 @@ where
 }
 
 /// Trait to convert a value to a `Value`.
+///
+/// Similar to other common conversion traits, the following invariants are guaranteed:
+///
+/// - **Invertibility**: `x.to_value().get().unwrap() == x`. In words, `FromValue` is the inverse of `ToValue`.
```
bilelmoussaoui, line 252:
```suggestion
/// - **Invertibility**: `x.to_value().get().unwrap() == x`. In words, [`FromValue`] is the inverse of [`ToValue`].
```
maybe?

bilelmoussaoui:
Otherwise looks fine to me, maybe we should document that the Value is cloned for the `impl ToValue for Value`? _

sdroege:
> maybe we should document that the Value is cloned for the `impl ToValue for Value`?

But where?

GuillaumeGomez:
On the impl?

sdroege:
The way how rustdoc presents docs on impls, almost nobody is reading them. Most people even miss the existence of impls in rustdocs regularly...

GuillaumeGomez:
Then add a note on the trait directly?

sdroege:
Or to that top part where it mentions the properties/relationships between the traits?

GuillaumeGomez:
Or all of the mentioned.

sdroege:
Ok!

piegamesde:
> maybe we should document that the Value is cloned for the impl ToValue for Value?

What for? What would be the alternative the people might expect?

sdroege:
It to be idempotent also in side effects, i.e. not causing copies and the like.

piegamesde:
I added a line about this to the `ToValue` trait documentation. This isn't specific to converting `Value` -> `Value`, thus I don't think it is appropriate on an `impl` block.

sdroege:
Sounds good to me. I'll merge once the CI passes (needs a `cargo fmt` run).

piegamesde:
Argh, I keep forgetting this. There really should be a commit hook or something.

GuillaumeGomez:
> Argh, I keep forgetting this. There really should be a commit hook or something.

Nope. CI is good.

sdroege:
> There really should be a commit hook or something.

Commit hooks are local only, you'd have to somehow install it manually after cloning the repository.

PR #190, opened 2021-07-08
0.14 backports

sdroege:


glib/tests/value.rs
```diff
@@ -0,0 +1,66 @@
+use glib::*;
+use std::ops::Deref;
+
+// FIXME all .get::<i32>() should be replaced with .get(); the compiler is totally able to infer the type itself.
+// But somehow without some tests are failing on Windows because the type inference doesn't work or something.
+
+/// Test that `ToValue` (and conversely, `FromValue`) uphold the promised invariants
```
GuillaumeGomez, line 7: Arf, didn't see the doc comments... It's useless on test functions. :-/

Oh well...

GuillaumeGomez (approved):

glib/tests/value.rs
piegamesde: I wouldn't say they're useless, but I agree that using `//` here instead wouldn't make a difference right now.

GuillaumeGomez: Arf, badly wrote it, I meant the doc comment format, not the content. Just like you said, using `//` would have been better.

PR #191, opened 2021-07-08
Update versions to 0.14.1

sdroege:
Versions are updated for all crates so that they're in sync (and so I can create a 0.14.1 tag now, and the next bugfix release will be 0.14.2 and things are not confusing).

I'm going to upload cairo, glib-macros and glib as those had actual bugfixes. The other crates had documentation fixes and clippy warning fixes, which seem not very relevant for crates.io.


GuillaumeGomez:
I was about to ask why updating all crates' version but you answered it so all good for me, thanks!

GuillaumeGomez (approved):

bilelmoussaoui (approved):

Issue #193, opened 2021-07-09
[BUG] MainContextAcquireGuard not exported?


lunaryorn:
**Bug description**

`MainContextAcquireGuard` doesn't seem to be exported in glib 0.14.1 (`use glib::MainContextAcquireGuard` fails to compile), which makes it impossible to pass the guard around to transfer the guard.

Rustdoc also fails to find the definition for this type at <https://docs.rs/glib/0.14.1/glib/struct.MainContext.html#method.acquire>:

![grafik](https://user-images.githubusercontent.com/224922/125044895-b7512300-e09c-11eb-9bf7-cae177a4dd64.png)

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/194 for the fix.

> Rustdoc also fails to find the definition for this type at https://docs.rs/glib/0.14.1/glib/struct.MainContext.html#method.acquire:

You probably want to use the docs at https://gtk-rs.org/gtk-rs-core/ . They're more complete than docs.rs.

sdroege:
Will be in glib 0.14.2 in an hour or so

lunaryorn:
Thanks :pray:

PR #194, opened 2021-07-09
glib: Re-export main context acquire guard

sdroege:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/193


glib/src/lib.rs
```diff
@@ -156,13 +156,14 @@ mod utils;
 pub use self::utils::*;
 mod main_context;
 mod main_context_channel;
+pub use self::main_context::MainContextAcquireGuard;
+pub use self::main_context_channel::{Receiver, Sender, SyncSender};
+mod date;
```
sdroege, line 161: Before you ask, for some arbitrary reason `cargo fmt` wants to move `mod date` up here now :man_shrugging:

GuillaumeGomez: Ok I won't ask then. :)

GuillaumeGomez (approved):

PR #200, opened 2021-07-14
glib: Implement FromGlibPtrBorrow for GValue

ZanderBrown:
Based on the GBoxed version

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -655,6 +655,20 @@ macro_rules! from_glib {
             }
         }
 
+        impl FromGlibPtrBorrow<*mut gobject_ffi::GValue> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *mut gobject_ffi::GValue) -> Borrowed<$name> {
+                $crate::translate::from_glib_borrow::<_, $name>(ptr as *mut gobject_ffi::GValue)
+            }
+        }
+
+        impl FromGlibPtrBorrow<*const gobject_ffi::GValue> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const gobject_ffi::GValue) -> Borrowed<$name> {
+                $crate::translate::from_glib_borrow::<_, $name>(ptr as *const gobject_ffi::GValue)
+            }
+        }
+
         impl FromGlibContainerAsVec<*mut gobject_ffi::GValue, *mut *mut gobject_ffi::GValue>
             for $name
         {


ZanderBrown:
Hmm, okay GitHub be weird

sdroege:
Hrm I see. Can you rebase manually with git on master and `git push --force` that here, then switch the base branch to master? Changes first go there and are then possibly backported (this one should be backported).

The change itself looks fine, but OOC where is this going to be used?

sdroege (approved):
Thanks!

ZanderBrown:
> The change itself looks fine, but OOC where is this going to be used?

I have a callback with a GValue param, `gir` assumes it can `from_glib_borrow` in the trampoline except, well, it can't (yet) :-)

sdroege (commented):
Actually does this work? Isn't this just calling the same impl recursively forever?

You can implement the `mut` one via the `const` one, but one of them at least needs to be implemented properly.

glib/src/value.rs
```diff
@@ -655,6 +655,20 @@ macro_rules! from_glib {
             }
         }
 
+        impl FromGlibPtrBorrow<*mut gobject_ffi::GValue> for $name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *mut gobject_ffi::GValue) -> Borrowed<$name> {
+                $crate::translate::from_glib_borrow::<_, $name>(ptr as *mut gobject_ffi::GValue)
```
sdroege, line 661:
```suggestion
                Borrowed::new(ptr::read(ptr as *const $name))
```

Might work, but it's all a bit ugly.

Ideally in your callback what you'd do instead (manually, not via gir... but I guess gir could get support for that) would be:

```
unsafe extern "C" your_trampoline(..., value: *const GValue, ...) {
    your_rust_function(..., &*(value as *const glib::Value), ...)
}
```

bilelmoussaoui: >  (manually, not via gir... but I guess gir could get support for that)

we have a bunch of functions that were manually implemented for this reason.  I would suggest that we rather fix this in gir directly

ZanderBrown: I've been playing with GValue-related stuff and come to the conclusion `gir` just isn't producing the right code

sdroege: Yes it doesn't do that correctly in many cases with `GValue`s. Your best bet for now would be to bind the function manually. If you can give me a link to your attempt I'd be happy to review that.

sdroege:
Let's close this then as it doesn't seem like this would work, and from what I understand you also don't need this anymore?

ZanderBrown:
Yes, Sorry I thought I'd closed this already

sdroege:
Actually I was wrong. This is now implemented in https://github.com/gtk-rs/gtk-rs-core/pull/275 . As long as the values are immutable this will work correctly and compiles down to basically the same code. For mutable values in callbacks we'll always need manual code, as before.

PR #205, opened 2021-07-22
Ignore find_with_equal_func for ListStore until a suitable API is provided by the C library

### LINKED ISSUE #202: `find_with_equal_func` is not implemented for ListStore
The bindings were not generated for this function.

keltrycroft:
Fixes #202.

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -704,6 +704,9 @@ manual_traits = ["ListStoreExtManual"]
     name = "sort"
     manual = true
     doc_trait_name = "ListStoreExtManual"
+    [[object.function]]
+    name = "find_with_equal_func"
+    ignore = true # See https://gitlab.gnome.org/GNOME/glib/-/issues/2447
 
 [[object]]
 name = "Gio.MemoryInputStream"
--- a/gio/src/auto/list_store.rs
+++ b/gio/src/auto/list_store.rs
@@ -63,13 +63,6 @@ impl ListStore {
         }
     }
 
-    //#[cfg(any(feature = "v2_64", feature = "dox"))]
-    //#[cfg_attr(feature = "dox", doc(cfg(feature = "v2_64")))]
-    //#[doc(alias = "g_list_store_find_with_equal_func")]
-    //pub fn find_with_equal_func<P: IsA<glib::Object>>(&self, item: &P, equal_func: /*Unimplemented*/FnMut(/*Unimplemented*/Option<Fundamental: Pointer>, /*Unimplemented*/Option<Fundamental: Pointer>) -> bool) -> Option<u32> {
-    //    unsafe { TODO: call ffi:g_list_store_find_with_equal_func() }
-    //}
-
     #[doc(alias = "g_list_store_insert")]
     pub fn insert<P: IsA<glib::Object>>(&self, position: u32, item: &P) {
         unsafe {


gio/src/list_store.rs
```diff
@@ -4,6 +4,9 @@ use crate::ListStore;
 use glib::translate::*;
 use glib::{IsA, Object};
 use std::cmp::Ordering;
+#[cfg(any(feature = "v2_64", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v2_64")))]
```
GuillaumeGomez, line 8: Please remove this line.

gio/src/list_store.rs
```diff
@@ -15,6 +18,15 @@ pub trait ListStoreExtManual {
 
     #[doc(alias = "g_list_store_sort")]
     fn sort<F: FnMut(&Object, &Object) -> Ordering>(&self, compare_func: F);
+
+    #[cfg(any(feature = "v2_64", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_64")))]
+    #[doc(alias = "g_list_store_find_with_equal_func")]
+    fn find_with_equal_func<P: IsA<glib::Object>, F: Fn(&Object, &Object) -> glib::ffi::gboolean>(
```
GuillaumeGomez, line 25: Shouldn't this be:

```suggestion
    fn find_with_equal_func<P: IsA<glib::Object>, F: Fn(&IsA<Object>, &IsA<Object>) -> glib::ffi::gboolean>(
```


?

gio/src/list_store.rs
```diff
@@ -52,6 +64,28 @@ impl<O: IsA<ListStore>> ListStoreExtManual for O {
...
+            let mut position = mem::MaybeUninit::uninit();
+            let ret = from_glib(ffi::g_list_store_find_with_equal_func(
+                self.as_ref().to_glib_none().0,
+                item.as_ref().to_glib_none().0,
+                Some(equal_func_trampoline),
+                position.as_mut_ptr(),
+            ));
+            let position = position.assume_init();
+            if ret {
+                Some(position)
```
GuillaumeGomez, line 83:
```suggestion
            if ret {
                let position = position.assume_init();
                Some(position)
```

gio/src/list_store.rs
```diff
@@ -70,3 +104,17 @@ unsafe extern "C" fn compare_func_trampoline(
         Ordering::Greater => 1,
     }
 }
+
+unsafe extern "C" fn equal_func_trampoline(
```
GuillaumeGomez, line 108: @sdroege Should we move this function into `find_with_equal_func` as well? I thought all callbacks were directly declared inside the block calling them but maybe I'm wrong there...

gio/src/list_store.rs
bilelmoussaoui: Just `Some(position.assume_init())`  should be fine :-)

gio/src/list_store.rs
bilelmoussaoui: They should unless they are re-used elsewhere

gio/src/list_store.rs
bilelmoussaoui: It also should be returning a normal bool and not a glib:ffi::gboolean

gio/src/list_store.rs
```diff
@@ -52,6 +64,28 @@ impl<O: IsA<ListStore>> ListStoreExtManual for O {
             )
         }
     }
+
+    fn find_with_equal_func<P: IsA<glib::Object>, F: Fn(&Object, &Object) -> glib::ffi::gboolean>(
```
bilelmoussaoui, line 68:
```suggestion
    #[cfg(any(feature = "v2_64", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_64")))]
    fn find_with_equal_func<P: IsA<glib::Object>, F: Fn(&Object, &Object) -> glib::ffi::gboolean>(
```

gio/src/list_store.rs
```diff
@@ -52,6 +64,28 @@ impl<O: IsA<ListStore>> ListStoreExtManual for O {
...
+        &self,
+        item: &P,
+        equal_func: F,
+    ) -> Option<u32> {
+        unsafe {
+            let mut position = mem::MaybeUninit::uninit();
```
bilelmoussaoui, line 74:
```suggestion
            let mut position = std::mem::MaybeUninit::uninit();
```
and drop the mem import you have at the top of the file

gio/src/list_store.rs
keltrycroft: Moved callback inside block as it is not re-used elsewhere

gio/src/list_store.rs
```diff
@@ -5,6 +5,7 @@ use glib::translate::*;
 use glib::{IsA, Object};
 use std::cmp::Ordering;
 
+
```
GuillaumeGomez, line 8: Why this new line?

gio/src/list_store.rs
```diff
@@ -52,6 +61,41 @@ impl<O: IsA<ListStore>> ListStoreExtManual for O {
...
+            b: glib::ffi::gconstpointer,
+        ) -> glib::ffi::gboolean {
+            let a: Borrowed<glib::Object> = from_glib_borrow(a as *mut glib::gobject_ffi::GObject);
+            let b: Borrowed<glib::Object> = from_glib_borrow(b as *mut glib::gobject_ffi::GObject);
+        
+            if *a == *b {
+                true.into_glib()
+            } else {
+                false.into_glib()
+            }
```
GuillaumeGomez, line 82:
```suggestion
            (*a == *b).into_glib()
```

gio/src/list_store.rs
```diff
@@ -70,3 +114,5 @@ unsafe extern "C" fn compare_func_trampoline(
         Ordering::Greater => 1,
     }
 }
+
```
GuillaumeGomez, line 117: Why adding those lines?

gio/src/list_store.rs
keltrycroft: My mistake, no reason

gio/src/list_store.rs
keltrycroft: Much cleaner, thanks

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks good to me, thanks! Just one question to @bilelmoussaoui and @sdroege: should we rename the function? Not sure if `find_with_equal_func` follows the naming conventions...

gio/src/list_store.rs
```diff
@@ -52,6 +60,37 @@ impl<O: IsA<ListStore>> ListStoreExtManual for O {
...
+        item: &P,
+    ) -> Option<u32> {
+        unsafe extern "C" fn equal_func_trampoline(
+            a: glib::ffi::gconstpointer,
+            b: glib::ffi::gconstpointer,
+        ) -> glib::ffi::gboolean {
+            let a: Borrowed<glib::Object> = from_glib_borrow(a as *mut glib::gobject_ffi::GObject);
+            let b: Borrowed<glib::Object> = from_glib_borrow(b as *mut glib::gobject_ffi::GObject);
+        
+            (*a == *b).into_glib()
```
bilelmoussaoui, line 77: Just noticed that this doesn't actually make use of a compare_func. The `find_with_equal_func` is supposed to take a function used to find the item you're looking for. There's already a `find` function that works without having to pass a compare callback. 

The issue we have here and the reason why this wasn't automatically generated I guess is the original C function doesn't provide a way for passing the compare callback as user_data.

bilelmoussaoui: See https://gitlab.gnome.org/GNOME/glib/-/issues/2447

sdroege: You could in theory work around this by storing the closure in TLS but...

gio/Gir.toml
```diff
@@ -704,6 +704,10 @@ manual_traits = ["ListStoreExtManual"]
     name = "sort"
     manual = true
     doc_trait_name = "ListStoreExtManual"
+    [[object.function]]
+    name = "find_with_equal_func"
+    manual = true
+    doc_trait_name = "ListStoreExtManual"
```
bilelmoussaoui, line 710:
```suggestion
    ignore = true # See https://gitlab.gnome.org/GNOME/glib/-/issues/2447
```

sdroege (commented):
Makes sense to me but please squash all commits into one :) There's no point in having 5 commits for this change

sdroege (approved):

PR #207, opened 2021-07-22
Add an Iterator for ListModel

V02460:
This PR implements an Iterator for the `ListModel` interface.

The Iterator can be created by the interface’s `iter` method. Example usage:
```rust
for member in members.iter::<User>() {
    …
}
```
As you can see, the `iter` method takes a type parameter `T` that is the type of the yielded items. At creation time of the Iterator a type check is performed to make sure it is a subtype of the model’s item type. Having a type-wise more convenient Iterator is also the reason I decided against implementing `IntoIterator` as it forces you to yield `Object`-type items. (This restriction could be lifted, if `ListModel` itself had the correct type parameter.)

The Iterator gives always sensible results, even if the model underneath it changes. That means no duplicates or unnecessary omissions. It accomplishes that cheaply by listening to the `items-changed` signal of the underlying model and adjusting its internal index accordingly.

Implementing `DoubleEndedIterator` and `ExactSizeIterator` should be possible, too, I might do that after a round of feedback.

### DIFF
--- a/gio/src/list_model.rs
+++ b/gio/src/list_model.rs
@@ -1,0 +1,88 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::cell::Cell;
+use std::marker::PhantomData;
+use std::rc::Rc;
+
+use glib::{clone, IsA, Object};
+
+use crate::prelude::*;
+use crate::ListModel;
+
+pub trait ListModelExtManual: ListModelExt {
+    /// Returns an iterator over the ListModel elements.
+    fn iter<T: IsA<Object>>(&self) -> ListModelIter<Self, T>
+    where
+        Self: Sized + Clone;
+}
+
+impl<O: IsA<ListModel>> ListModelExtManual for O {
+    fn iter<T: IsA<Object>>(&self) -> ListModelIter<Self, T>
+    where
+        Self: Sized + Clone,
+    {
+        ListModelIter::new(self.clone())
+    }
+}
+
+pub struct ListModelIter<M: ListModelExt, T: ObjectExt> {
+    data_type: PhantomData<T>,
+    model: M,
+    i: Rc<Cell<u32>>,
+}
+
+impl<M: ListModelExt + Clone, T: IsA<Object>> ListModelIter<M, T> {
+    /// Creates a new ListModel iterator.
+    /// 
+    /// ### Panics
+    /// Panics if `T::static_type()` is not of the model’s item type.
+    fn new(model: M) -> Self {
+        if !T::static_type().is_a(model.item_type()) {
+            panic!(
+                "Item type {} is not a subtype of model type {}.",
+                T::static_type(),
+                model.item_type()
+            );
+        }
+
+        let iter = ListModelIter::<M, T> {
+            data_type: PhantomData,
+            model: model.clone(),
+            i: Rc::new(Cell::new(0)),
+        };
+
+        // Adjust index when the underlying model changes.
+        model.connect_items_changed(
+            clone!(@weak iter.i as i_rc => move |_model, pos, n_removed, n_added| {
+                i_rc.set(adjust_index(i_rc.get(), pos, n_removed, n_added));
+            }),
+        );
+
+        iter
+    }
+}
+
+impl<M: ListModelExt, T: IsA<Object>> Iterator for ListModelIter<M, T> {
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let item = self.model.item(self.i.get());
+        self.i.set(self.i.get() + 1);
+        item.map(|x| x.downcast::<T>().unwrap())
+    }
+}
+
+/// Calculates the new index from the given change.
+fn adjust_index(index: u32, pos: u32, n_removed: u32, n_added: u32) -> u32 {
+    if index <= pos {
+        // Before the changed area; do nothing.
+        index
+    } else if index < pos + n_removed {
+        // In the changed area; skip the new elements.
+        pos + n_added
+    } else {
+        // index >= pos + n_removed
+        // After the changed area; move by the difference.
+        index + n_added - n_removed
+    }
+}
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
What happens when an item gets removed while you're iterating through the list model?

GuillaumeGomez:
> What happens when an item gets removed while you're iterating through the list model?

Funny, I asked the same question. I also asked for an alternative: what happens if you remove the current item and all the following ones?

V02460:
First of all, the iterator never panics.

> What happens when an item gets removed while you're iterating through the list model?

The iterator does the natural thing and continues relative to the items it was working on before the change of model. More specifically, I differentiate three cases: before, after and inside the change. The before and after cases continue relative to where they were before. For the inside case, I used the hopefully least unexpected thing and skip to after the change. The exact function of adjusting the underlying index can be seen at [`adjust_index`](https://github.com/gtk-rs/gtk-rs-core/pull/207/files#diff-44dd61d7102d5082b87c3c510cd4ceee05ca4bd2c23be922486f4f477bc90dc0R76-R88).

> what happens if you remove the current item and all the following ones?

The `next` method of the iterator returns `None`, so the iteration ends when it is called the next time.

bilelmoussaoui:
I think a test case that ensures that it works as expected is needed here

gio/src/list_model.rs
```diff
@@ -0,0 +1,88 @@
...
+    i: Rc<Cell<u32>>,
+}
+
+impl<M: ListModelExt + Clone, T: IsA<Object>> ListModelIter<M, T> {
+    /// Creates a new ListModel iterator.
+    /// 
+    /// ### Panics
+    /// Panics if `T::static_type()` is not of the model’s item type.
+    fn new(model: M) -> Self {
+        if !T::static_type().is_a(model.item_type()) {
```
sdroege, line 40: This is probably inversed. If you have a model of widgets, you don't want to be able to iterate over it as buttons (that might fail), but if you have a model of buttons you want to be able to iterate over it as widgets (that can't fail).

gio/src/list_model.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+/// Calculates the new index from the given change.
+fn adjust_index(index: u32, pos: u32, n_removed: u32, n_added: u32) -> u32 {
+    if index <= pos {
+        // Before the changed area; do nothing.
+        index
+    } else if index < pos + n_removed {
+        // In the changed area; skip the new elements.
```
sdroege, line 81: Why is this the correct behaviour? Seems as good / bad as not skipping the elements to me

sdroege:
This definitely needs many tests and documentation for the behaviour, and some justification why this is the right behaviour. @bilelmoussaoui as someone actually doing things with list models, would this behaviour be the one you would expect?

gio/src/list_model.rs
V02460: That’s a bad one. You’re right, it should be contravariant. Thank you for taking the time *really* looking into this.

gio/src/list_model.rs
V02460: My main reason is atomicity of a change: If a block of items changes under you, it will not be handled only partially. The whole affair is a race-condition anyway, so there is no advantage in including the items.

sdroege: You're skipping over newly added items, but items that got removed would not appear in the iteration (of course, they're gone). So there's at least some asymmetry here but that might be fine.

FWIW, IMHO it would be equally fine to immediately finish the iterator on the first modification during iteration.

I guess whatever behaviour we select here will have its advantages and disadvantages, but the main problem for me is that from a user point of view it's not obvious during the iteration that something happened and there was a modification. You probably want to restart your iteration in many cases if this happens.

We could make the iterator return a `Result<T, ModificationDuringIterationError>` but that's kind of annoying to use.

V02460: Simply finishing the iterator on any model change would be bad, because you can’t differentiate between a failure and coming to an orderly end. Returning `Result` is nice, because it holds the iterator to a transactional standard, but it might be overkill for all the applications that don’t expect/don’t care about model changes.

I thought a bit more about the index adjustment. If we assume adding and removing can be a replacement of an item, then jumping to anywhere but the end of the change would cause duplicates; yielding the old and then the new version of the item. Together with the atomicity mentioned above, these two criteria give us a definitive algorithm for recalculating the index.

I think there are two ways of doing this iterator:

**The transactional way** Yields the whole list consistently, but might fail. The user has high demands on consistency and is willing to implement retry logic and/or error handling by themself.
**The messy way** Easy to use and infallible. This yields elements orderly, but the first might be older than the last. The user doesn’t expect/care about model changes. This is the current implementation.

sdroege: The messy way would also potentially cause some items to not have been touched after the end of the iteration, which might lead to hard to debug bugs.

From a correctness point of view I would go with a `Result`, which is also how the `Iterator` adapter on top of `gst::Iterator` is implemented: https://gstreamer.pages.freedesktop.org/gstreamer-rs/stable/0.17/docs/gstreamer/struct.Iterator.html#impl-IntoIterator

The question is mostly who/how are people going to use this? Why would anybody ever iterate over the whole model? I can't think of any reason where it would be OK to get an inconsistent result ("messy way") and either I would assert that no modifications ever happen (because I know my code won't modify during the iteration), or I would have to restart the iteration on modification and hope that this time it finishes without yet another modification.

sdroege:
@GuillaumeGomez @bilelmoussaoui What do you think, which behaviour would make most sense here? Personally I think the `Result`/transactional way would make sense.

ranfdev:
I'm new to this thread, so sorry if I've missed something.

Since a list model contains a list of GObjects, it's just a list of pointers. They're cheap to clone.
Can't we just get the entire list of GObjects in a sweep, put them in a `Vec<glib::Object>` and then return `Vec::into_iter`?
It's not the best solution, but hey, it works, it's transactional, and it's better than nothing. It can be improved later

sdroege:
That would be possible, yes.

V02460:
That… yes. :sweat_smile:

sdroege:
That one is merged now but I'd still like to get something around this here in

PR #208, opened 2021-07-23
Add ListModel::find_sorted and ListStore::find_or_insert_sorted

V02460:
This PR adds two convenience methods: `find_sorted` to ListModel and `find_or_insert_sorted` to ListStore.

Both methods provide search functionality in O(log n). The naming schema is borrowed from the native ListStore methods.

### DIFF
--- a/gio/src/lib.rs
+++ b/gio/src/lib.rs
@@ -47,6 +47,7 @@ mod io_stream;
 pub use crate::io_stream::IOStreamAsyncReadWrite;
 mod input_stream;
 pub use crate::input_stream::{InputStreamAsyncBufRead, InputStreamRead};
+mod list_model;
 mod list_store;
 #[cfg(test)]
 mod memory_input_stream;
--- a/gio/src/list_model.rs
+++ b/gio/src/list_model.rs
@@ -1,0 +1,55 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::cmp::Ordering;
+
+use glib::{IsA, Object};
+
+use crate::prelude::*;
+use crate::ListModel;
+
+pub trait ListModelExtManual {
+    /// Searches the sorted list with `compare_func` in log time.
+    ///
+    /// On success returns the index of the found item.
+    /// On failure returns the index where a matching element can be inserted
+    /// to mainatain order.
+    ///
+    /// ### Panics
+    /// Panics if `T::static_type()` is not of the model’s item type.
+    fn find_sorted<T, F>(&self, compare_func: F) -> Result<usize, usize>
+    where
+        T: IsA<Object>,
+        F: FnMut(&T) -> Ordering;
+}
+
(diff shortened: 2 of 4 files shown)


GuillaumeGomez:
It looks good to me, however I'm definitely not an expert on this matter so I'll let @sdroege and @bilelmoussaoui "judge" on this matter. :)

bilelmoussaoui:
I'm not sure if it makes sense to add new API in the bindings directly. Are those actually that useful in general?

V02460:
@bilelmoussaoui I obviously find these functions useful (there is log n search now where none was before), but they can be implemented on top of the existing API. I’m not sure what the aims of the C API are, maybe they try to be minimalistic? I would like to see them in the rust bindings to have it as useful and convenient as it gets, and if these functions were to later find their way into gio, one can swap them out.

GuillaumeGomez:
I think it should be the other way around: having them in the C libraries before we bind them in the Rust bindings. We're not opposed to adding API on gtk-rs directly if they make sense though. It just seems like a very specific use case...

V02460:
@GuillaumeGomez I think it’s genuinely useful, so I would try to get it into the C API. Adding it here first would just be a shorter path to have it actually usable in Rust. I don’t see getting it in here first doing any harm or does that clash with some policies you have regarding API stability?

`ListModel::find_sorted` is the more general and maybe more useful function of the two. It gives a fast search and fits with the existing API with special-casing a sorted model. It is roughly equivalent to [`Vec::binary_search_by`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.binary_search_by).

`ListStore::find_or_insert_sorted` is a bit more specific in its usecase: Making sure, there is exactly one item of the given kind in the model; prioritizing an existing one. It is a relatively thin wrapper around `ListModel::find_sorted`, but it takes some lines to write out the logic. I lean towards saying it is useful, while being biased by having a problem at hand that it applies to ;)

GuillaumeGomez:
> I don’t see getting it in here first doing any harm or does that clash with some policies you have regarding API stability?

One word: maintenance.

sdroege:
I would also prefer to have this in either the C library or in a utility crate on top of the bindings. The bindings should really only bind to the C API and in addition provide integration API into the Rust ecosystem (and as part of that also potentially small API additions that make the API more Rust'y).

jsparber:
I'm not fan of adding a method to `gio::ListModel` this way, since the implementation should happen in the object implementing the interface and not be part of the `ListModel`. Also you can't really assume that the ListModel is sorted. 

`ListStore::find_or_insert_sorted` on the other hand looks like something that can be easily implemented, and doesn't seam to be really useful in general except for specific cases.

Anyways i think this API change really needs to happen in gio and not in the bindings.

sdroege:
I guess let's close this here then. Thanks for your changes nonetheless! Maybe they can end up in some kind of gio utility crate, or maybe you can submit them to GLib in C so they can end up in all bindings.

PR #216, opened 2021-07-28
Variant,VariantDict: More type lookup methods

cgwalters:
In ostree/rpm-ostree we use GVariant extensively and particularly
`a{sv}`s.  I have this helper trait
https://github.com/ostreedev/ostree-rs-ext/blob/6fcdde1697a851f9c29d9f8b15d77e114a7a5ae7/lib/src/variant_utils.rs#L32
but we can do better by offering an API equivalent to
`lookup()` + `Variant::get()` that returns the native type.

Plus for `Variant` it's often convenient to have a version that
returns a `Result<T>` with a useful error message about
the actual+expected types.

---

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -167,6 +167,32 @@ impl crate::value::ToValueOptional for Variant {
     }
 }
 
+/// An error returned from the [`try_get`](struct.Variant.html#method.try_get) function
+/// on a [`Variant`](struct.Variant.html) when the expected type does not match the actual type.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct VariantTypeMismatchError {
+    pub actual: VariantType,
+    pub expected: VariantType,
+}
+
+impl VariantTypeMismatchError {
+    pub fn new(actual: VariantType, expected: VariantType) -> Self {
+        Self { actual, expected }
+    }
+}
+
+impl fmt::Display for VariantTypeMismatchError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Type mismatch: Expected '{}' got '{}'",
+            self.expected, self.actual
+        )
+    }
+}
+
+impl std::error::Error for VariantTypeMismatchError {}
+
 impl Variant {
     /// Returns the type of the value.
     pub fn type_(&self) -> &VariantTy {
@@ -185,6 +211,16 @@ impl Variant {
     #[inline]
     pub fn get<T: FromVariant>(&self) -> Option<T> {
         T::from_variant(self)
+    }
+
+    /// Tries to extract a value of type `T`.
+    pub fn try_get<T: FromVariant>(&self) -> Result<T, VariantTypeMismatchError> {
+        self.get().ok_or_else(|| {
+            VariantTypeMismatchError::new(
+                self.type_().to_owned(),
+                (&*T::static_variant_type()).to_owned(),
+            )
+        })
     }
 
     /// Boxes value.
@@ -975,5 +1011,16 @@ mod tests {
             <Vec<(&str, u8, u32)>>::static_variant_type().to_str(),
             "a(syu)"
         );
+    }
+
+    #[test]
+    fn test_get() -> Result<(), Box<dyn std::error::Error>> {
+        let u = 42u32.to_variant();
+        assert!(u.get::<i32>().is_none());
+        assert_eq!(u.get::<u32>().unwrap(), 42);
+        assert!(u.try_get::<i32>().is_err());
+        // Test ? conversion
+        assert_eq!(u.try_get::<u32>()?, 42);
+        Ok(())
     }
 }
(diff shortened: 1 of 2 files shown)


glib/src/variant_dict.rs
```diff
@@ -59,6 +59,20 @@ impl VariantDict {
         }
     }
 
+    /// Look up a typed value from this `VariantDict`.
+    ///
+    /// The given `key` is looked up in `self`.
+    ///
+    /// This will return `None` if the `key` is not present in the dictionary
+    /// or if it is present but the type of the value does not match the
+    /// expected type.
```
sdroege, line 68: Should non-presence maybe be distinguished in the return type from presence-but-wrong-type? Using a `Result<Option<T>, WrongTypeError>` probably will make it harder to use?

GuillaumeGomez: I think returning a `Result<Option<T>>` makes more sense though, otherwise we can't know what failed.

sdroege: `headers::HeaderMapExt` has `typed_get()` and `typed_try_get()`, the latter returning a `Result` and the former flattening into a single `Option`.

GuillaumeGomez: Should we provide both then?

sdroege: https://docs.rs/headers/0.3.4/headers/trait.HeaderMapExt.html#tymethod.typed_get for reference

cgwalters: I made this method mirroring the existing `Variant::get()`.  Should we also add `try_get()` then?

sdroege: `Variant::get()` can only fail if the type doesn't match, but this here would fail if either that happens or the given key is not stored in the dict without the caller being able to distinguish the two cases.

Arguably, `Variant::get()` should also return a `Result` as that's an actual error case and not just a non-existing value.

cgwalters: Right.   I was thinking to add `Variant::try_get()` because in my code I usually end up converting `Option` :arrow_right: `Result` for that case too; see e.g. https://github.com/coreos/rpm-ostree/pull/3034/files#diff-e618d2cef0ae2cb833c18e5603d050edbf916eeac5b66392487572aab2d8d32cR114

sdroege: If it was a `Result<T, glib::BoolError>` or something else generic that wouldn't probably help you much, you'd still need a `map_err()` in your code I assume?

sdroege: (But yes that should really return a `Result` IMHO)

cgwalters: OK, updated!

glib/src/variant.rs
```diff
@@ -187,6 +211,16 @@ impl Variant {
         T::from_variant(self)
     }
 
+    /// Tries to extract a value of type `T`.
+    pub fn try_get<T: FromVariant>(&self) -> Result<T, VariantTypeMismatchError> {
```
sdroege, line 215:
```suggestion
    pub fn get<T: FromVariant>(&self) -> Result<T, VariantTypeMismatchError> {
```

IMHO (and remove the old `get()`). If someone needs the old behaviour then it's only a `.ok()` away.

cgwalters: Ah hmm you're thinking to queue this for a new semver incompatible bump?  Is that already planned?  The 0.14 bump was *mostly* straightforward for us but it also reverberates up the stack - we had to do a chain of semver bumps for our code too.

I pushed a patch that is semver compatible for now that addresses the other comments.

cgwalters: Ah I see https://github.com/gtk-rs/gtk-rs-core/commit/7c3814fc088f5ae876c7ebabd251f2d1071bf29c so that seems to answer that.

cgwalters: Now that I try to change `get()` like you suggest it then raises the question of whether the `FromVariant` trait should also return a `Result<T, VariantTypeError>`.

sdroege: Keep it in a separate commit in the same PR and then we can backport the semver compatible changes to 0.14 and keep the breaking ones in master.

For `FromVariant`, I guess so. Ideally these traits should become closer to the ones for `Value`. But that's out of scope for this PR IMHO.

glib/src/variant_dict.rs
cgwalters: > If it was a Result<T, glib::BoolError> or something else generic that wouldn't probably help you much, you'd still need a map_err() in your code I assume?

That's a good point.  And I also notice we were missing the `impl Error`, fixed that.

sdroege:
How do we move forward here?

cgwalters:
> How do we move forward here?

OK I updated this with a patch that cherry picks cleanly to 0.14 and is semver compatible, etc.  Since I think the change of `s/try_get/get/` is going to be pretty large, can we do it as a separate PR?

I added more unit tests, including `?` etc.

sdroege (approved):

GuillaumeGomez (approved):

PR #217, opened 2021-07-29
cairo: silence self_named_constructor

elmarco:
Silence a clippy warning.

Or do we want to break API here and rename it "new"?

### DIFF
--- a/cairo/src/utils.rs
+++ b/cairo/src/utils.rs
@@ -37,8 +37,7 @@ pub struct Version {
 impl Version {
     #[doc(alias = "cairo_version")]
     #[doc(alias = "get_version")]
-    #[allow(clippy::self_named_constructor)]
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
An API break is not a problem.

cairo/src/utils.rs
```diff
@@ -47,6 +47,12 @@ impl Version {
     }
 }
 
+impl std::default::Default for Version {
```
bilelmoussaoui, line 50:
```suggestion
impl Default for Version {
```
I think the Default trait is in-scope by default?

sdroege (commented):
Looks good to me otherwise

bilelmoussaoui (approved):

Issue #22, opened 2020-11-10
Add bindings for GInitiable / GAsyncInitiable


sdroege:

sdroege:
Bindings are there but calling/checking on construction is still awkward because of the library boundary. CC @jf2048

jf2048:
I think the best we could do is call `g_type_interfaces` and do some string comparisons on the type name before calling `g_object_new`, but that's not really appealing

sdroege:
Yeah and `dlsym()` the `g_initable_get_type()` from the current process. That's the best I came up with so far too :)

jf2048:
A little bit better would be:

- Use `Type::qname` instead to do quark comparisons
- Error from `Object::new`, `Object::with_type` immediately if the interface is detected
- Require use of other constructors `Initable::new`, `Initable::with_type`, `AsyncInitable::new` etc for those classes

PR #218, opened 2021-07-29
Added basic gio::Initable implementation

xanathar:
This adds a basic implementation of `gio::Initable` which can be used to allow objects to implement `GInitable`.

Limitations:
- It does not attempt to solve the current issue with graceful failure of object creation on the Rust side (https://github.com/gtk-rs/gtk-rs-core/issues/22)
- It does not wrap the helper functions (`g_initable_new`, `g_initable_new_valist` and `g_initable_newv`); they are left commented as generated by gir, in the same vein of other types already in gtk-rs
- Unit test is implemented only using the ffi functions for now


gio/src/auto/initable.rs
```diff
@@ -0,0 +1,67 @@
...
+impl Initable {
+    //#[doc(alias = "g_initable_new")]
+    //pub fn new<P: IsA<Cancellable>>(object_type: glib::types::Type, cancellable: Option<&P>, error: &mut glib::Error, first_property_name: Option<&str>, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> glib::Object {
+    //    unsafe { TODO: call ffi:g_initable_new() }
+    //}
+
+    //#[doc(alias = "g_initable_new_valist")]
+    //pub fn new_valist<P: IsA<Cancellable>>(object_type: glib::types::Type, first_property_name: &str, var_args: /*Unknown conversion*//*Unimplemented*/Unsupported, cancellable: Option<&P>) -> Result<glib::Object, glib::Error> {
+    //    unsafe { TODO: call ffi:g_initable_new_valist() }
+    //}
```
bilelmoussaoui, line 29: You can disable both new_valist/new as they are not bindings friendly. Probably rename newv to new as well

gio/src/auto/initable.rs
```diff
@@ -0,0 +1,67 @@
...
+
+pub const NONE_INITABLE: Option<&Initable> = None;
+
+pub trait InitableExt: 'static {
+    #[doc(alias = "g_initable_init")]
+    fn init<P: IsA<Cancellable>>(&self, cancellable: Option<&P>) -> Result<(), glib::Error>;
+}
+
+impl<O: IsA<Initable>> InitableExt for O {
+    fn init<P: IsA<Cancellable>>(&self, cancellable: Option<&P>) -> Result<(), glib::Error> {
```
sdroege, line 46: Calling this multiple times is not necessarily safe AFAIU from the docs

gio/src/auto/initable.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub struct Initable(Interface<ffi::GInitable, ffi::GInitableIface>);
+
+    match fn {
+        type_ => || ffi::g_initable_get_type(),
+    }
+}
+
+impl Initable {
+    //#[doc(alias = "g_initable_new")]
+    //pub fn new<P: IsA<Cancellable>>(object_type: glib::types::Type, cancellable: Option<&P>, error: &mut glib::Error, first_property_name: Option<&str>, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> glib::Object {
```
sdroege, line 22: You could implement this manually around `glib::Object::new()` and `Initable::init()` if you want. That should be like 5 lines of code :)

gio/src/subclass/initable.rs
```diff
@@ -0,0 +1,177 @@
...
+use glib::Error;
+
+use glib::subclass::prelude::*;
+
+use std::ptr;
+
+use crate::Cancellable;
+use crate::Initable;
+
+pub trait InitableImpl: ObjectImpl + Send {
```
sdroege, line 14:
```suggestion
pub trait InitableImpl: ObjectImpl {
```

`Initable` does not require the object to be send'able (and GObjects are never only `Send`, if anything they must be `Send + Sync` due to refcounting).

gio/src/subclass/initable.rs
```diff
@@ -0,0 +1,177 @@
...
+            let test = InitableTestType::new();
+            let test: *mut ffi::InitableTestType = test.as_ptr();
+            let mut error: *mut glib::ffi::GError = std::ptr::null_mut();
+
+            assert_ne!(0x123456789abcdef, ffi::initable_test_type_get_value(test));
+
+            let result = crate::ffi::g_initable_init(
+                test as *mut crate::ffi::GInitable,
+                std::ptr::null_mut(),
+                &mut error,
```
sdroege, line 170: Maybe also assert that the error is null afterwards

gio/src/subclass/initable.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    impl InitableTestType {
+        pub fn new() -> Self {
+            glib::Object::new(&[]).expect("Failed creation of InitableTestType object")
+        }
+    }
+
+    #[test]
+    fn test_initable_through_ffi() {
```
sdroege, line 159: You could also add another test that goes via the safe Rust API only

gio/src/auto/initable.rs
sdroege: So maybe this needs to be `unsafe` (and that's why having a safe `new()` would be useful). Not entirely sure.

sdroege:

sdroege:
The CI failure will be gone if you rebase on latest master

gio/src/auto/initable.rs
xanathar: Yes, it's not safe to be called more than once (or anything other than exactly once).
It's not clear to me however how to enforce that: in pure Rust I would have used a type-state / builder-object pattern kind of thing where the init takes ownership of self and returns a type which does not support Initable anymore, but in this interop scenario with the gobject model I'm not sure of the approach to take.

sdroege: You mark it `unsafe` and add a `new()` that internally calls it :)

xanathar: I wrote my message in between your first two and didn't refresh :) it was actually pretty clear what to do

gio/src/auto/initable.rs
xanathar: Done in new version

gio/src/auto/initable.rs
xanathar: Done. I called it `initable_new` to avoid name conflicts with `new` which is in widespread use, but I can easily change that.

gio/src/initable.rs
```diff
@@ -0,0 +1,34 @@
...
+use crate::Cancellable;
+use crate::Initable;
+use glib::object::IsA;
+use glib::object::IsClass;
+use glib::value::ToValue;
+use glib::Object;
+
+pub trait InitableExtManual: Sized + 'static {
+    /// Helper function to safely create and initialize an object implementing [Initable].
+    fn initable_new<P: IsA<Cancellable>>(
```
sdroege, line 11: You'd put that into an `impl Initiable` block, and then

```suggestion
    fn initable_new<T: IsA<Initable>, P: IsA<Cancellable>>(...) -> Result<T, Error>
```

gio/src/subclass/initable.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    glib::wrapper! {
+        pub struct InitableTestType(ObjectSubclass<imp::InitableTestType>)
+            @implements Initable;
+    }
+
+    impl InitableTestType {
+        #[allow(clippy::new_without_default)]
+        pub fn new() -> Self {
+            glib::Object::new(&[]).expect("Failed creation of InitableTestType object")
```
sdroege, line 157: This should go via the initable interface

xanathar:
I pushed a new version which addresses the comments.
I accidentally pushed an intermediate one before rebasing, sorry, so please give the CI time to run twice 🤷🏻‍♂️

gio/src/initable.rs
sdroege: Or maybe not... mh!

gio/src/subclass/initable.rs
xanathar: I use the type in a couple of tests (`test_initable_with_init` and `test_initable_through_ffi`) which need the object in a non-initialized state. I can remove the tests, but I think the FFI one is good to check that a declaration of a Rust initable type would work called from C

xanathar: `test_initable_with_initable_new` does the test using the initable interface

gio/src/initable.rs
xanathar: The reason I did it in an external trait (`InitableExtManual`) was to preserve the generation of the base interface from gir. I can move the whole Initable to manual if preferred, though.

xanathar:
BTW - the CI failures are not bogus - there are a few types for which gir now adds "Initable" as an implemented interface. Tomorrow I'll re-run gir and push an update.

gio/src/initable.rs
sdroege: You can keep the autogenerated code and have a separate `impl Initiable` in `src/initable.rs`. That's not a problem :)

gio/src/subclass/initable.rs
sdroege: Ok then, but please add a comment about this :)

sdroege (approved):

bilelmoussaoui (dismissed):

xanathar:
Exercising CI for minor findings.. as soon as it's green I will convert to "ready".

sdroege (approved):

PR #220, opened 2021-08-01
Add function to consume raw data of ImageSurface

### LINKED ISSUE #29: How to get ownership of data back from ImageSurface
## Summary

You can quickly create an `ImageSurface` from some byte array using `ImageSurface::create_for_data`. However, I don't see a way to get ownership of that byte array back from the `ImageSurface` without performing a copy. I would like to figure out a way to avoid copying my image data.

## More context

I am working on an application that uses a background thread to render an image and uses double buffering to quickly update the main display. My approach is inspired by [this tutorial](https://www.cairographics.org/threaded_animation_with_cairo/), but updated for compatibility with GTK+ 3 and obviously using the Rust APIs (and ideally just safe Rust). I am using a `Vec<u8>` as a pixel map which I can pass between threads, and from it construct an `ImageSurface` using `ImageSurface::create_for_data`. The problem is, after drawing on the `ImageSurface`, I'd like to be able to take back ownership of my pixel map without cloning the data into a new `Vec` from the `ImageSurfaceData` returned by `ImageSurface::get_data`. Right now, it seems my options are to either to clone the data or drop into `unsafe`.

## Possible approaches

I see from [the code](https://github.com/gtk-rs/cairo/blob/master/src/image_surface.rs#L62) that the byte array is stored as user data on the created surface. However, this user data key is not accessible publicly. If it were, one could presumably get the `Rc` handle from the user data, drop the `ImageSurface` and use `Rc::try_unwrap` and `Box::downcast` to recover the byte array. That's kind of gross though. An idea that's only slightly better at best would be to return the `Rc<Box<dyn AsMut<u8>>>` from `create_for_data` along with the `ImageSurface`, which at least doesn't expose the user data key.

Probably the cleanest API I can think of would be a new generic struct like `ImageSurfaceWithData<D: AsMut<u8>>` which is `Deref<Target = ImageSurface>`. But that starts to deviate even more from the C API. That's basically the API that I will likely implement myself with `unsafe` if there is no upstream solution.

Another idea would be to add a method `ImageSurface::finish_output_data(&self) -> Result<Box<dyn AsMut<u8>>, Error>` which mirrors the functionality of `Surface::finish_output_stream`. That seems like the most consistent option, if not the safest.

Any thoughts on if there already exists an approach that I missed or whether this is worth addressing?

lehmanju:
This introduces a new function for `ImageSurface` which consumes the object, checks for exclusivity on the reference counter and returns a mutable slice of the raw data. Also added a minimal test case to verify.

Fixes #29

### DIFF
--- a/cairo/src/image_surface.rs
+++ b/cairo/src/image_surface.rs
@@ -107,6 +107,25 @@ impl ImageSurface {
         }
     }
 
+    pub fn take_data(self) -> Result<ImageSurfaceDataOwned, BorrowError> {
+        unsafe {
+            if ffi::cairo_surface_get_reference_count(self.to_raw_none()) > 1 {
+                return Err(BorrowError::NonExclusive);
+            }
+
+            self.flush();
+            let status = ffi::cairo_surface_status(self.to_raw_none());
+            if let Some(err) = status_to_result(status).err() {
+                return Err(BorrowError::from(err));
+            }
+            if ffi::cairo_image_surface_get_data(self.to_raw_none()).is_null() || is_finished(&self)
+            {
+                return Err(BorrowError::from(Error::SurfaceFinished));
+            }
+            Ok(ImageSurfaceDataOwned { surface: self })
+        }
+    }
+
     pub fn with_data<F: FnOnce(&[u8])>(&self, f: F) -> Result<(), BorrowError> {
         self.flush();
         unsafe {
@@ -146,6 +165,46 @@ impl ImageSurface {
     #[doc(alias = "get_width")]
     pub fn width(&self) -> i32 {
         unsafe { ffi::cairo_image_surface_get_width(self.to_raw_none()) }
+    }
+}
+
+pub struct ImageSurfaceDataOwned {
+    surface: ImageSurface,
+}
+
+impl AsRef<[u8]> for ImageSurfaceDataOwned {
+    fn as_ref(&self) -> &[u8] {
+        let len = (self.surface.stride() as usize) * (self.surface.height() as usize);
+        unsafe {
+            let ptr = ffi::cairo_image_surface_get_data(self.surface.to_raw_none());
+            debug_assert!(!ptr.is_null());
+            slice::from_raw_parts(ptr, len)
+        }
+    }
+}
+
+impl AsMut<[u8]> for ImageSurfaceDataOwned {
+    fn as_mut(&mut self) -> &mut [u8] {
+        let len = (self.surface.stride() as usize) * (self.surface.height() as usize);
+        unsafe {
+            let ptr = ffi::cairo_image_surface_get_data(self.surface.to_raw_none());
+            debug_assert!(!ptr.is_null());
+            slice::from_raw_parts_mut(ptr, len)
+        }
+    }
+}
+
+impl Deref for ImageSurfaceDataOwned {
+    type Target = [u8];
+
+    fn deref(&self) -> &Self::Target {
+        self.as_ref()
+    }
+}
+
+impl DerefMut for ImageSurfaceDataOwned {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.as_mut()
     }
 }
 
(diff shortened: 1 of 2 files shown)


cairo/src/image_surface.rs
```diff
@@ -107,6 +107,26 @@ impl ImageSurface {
         }
     }
 
+    pub fn consume_data<'a>(self) -> Result<&'a mut [u8], BorrowError> {
```
sdroege, line 110: This is unsound. The data is owned by the surface, but the surface is destroyed after this function returned so you'll end up with a reference to already freed data.

I'm not sure how this can be implemented as the data of the surface could be of any type and we don't even know how to free it.

You could maybe create a new type that `Deref`s / `DerefMut`s into `[u8]`, keeps the reference to the image surface alive and then on `Drop` finally gets rid of the surface?

lehmanju: I guess the test case works because the freed memory is not overwritten. I'll redo this, thanks for the comment.

lehmanju:
Taking ownership of the raw `ImageSurface` data is not possible if the buffer is allocated by cairo itself. This only allows borrowing via a wrapper struct which is quite tedious to keep around if you only want the data and not use `ImageSurface` ever again. Therefore, `consume_data` is not possible to implement in this context.

However, if we allocate memory manually and assume ownership of the raw data at all times, an implementation of `consume_data` is possible. At the end of the object's lifetime, a custom `Drop` implementation would free the buffer though.

Thoughts?

sdroege:
I don't understand :) You'd only need some like this returned from the function you created here.

```rust
struct ImageSurfaceDataOwned {
    surface: *mut ffi::CairoImageSurface,
}

impl Deref {
    type Target = [u8];
    fn deref(&self) -> &[u8] { cairo_image_surface_get_data() and slice::from_raw_parts }
}

impl DerefMut {
    fn deref_mut(&mut self) -> &mut [u8] { cairo_image_surface_get_data() and slice::from_raw_parts_mut }
}

impl Drop for ImageSurfaceOwnedData {
    fn drop(&mut self) {
        ffi::cairo_surface_destroy(self.surface);
    }
}
```

lehmanju:
But that would still only be a reference to the underlying data buffer, no?

In my case, I want to create a surface, draw something, get ownership of the buffer and send it to a MemoryTexture without cloning it since I don't need the cairo surface anymore.

sdroege:
As what type do you need it for the memory texture?

lehmanju:
Basically, I want to have a queue of rendered images that provides a view of the data. But because displaying the rendered images happens somewhere else, I want to transfer ownership of the data since I won't need it anymore at the source location.

From the documentation, `gdk4::MemoryTexture` takes a `Bytes` struct which can either be created from an owned slice or from static data, this is without cloning. It is possible to use a reference but that would unnecessarily clone the buffer.

sdroege:
> From the documentation, `gdk4::MemoryTexture` takes a `Bytes` struct which can either be created from an owned slice or from static data, this is without cloning. It is possible to use a reference but that would unnecessarily clone the buffer.

So you would do what I wrote above and also implement `AsRef<[u8]>` on that struct. Then you can create a `Bytes` from it via `from_owned()`.

lehmanju:
Oh dear, I get it now. I somehow thought that one would need to keep a reference to the original object in order for the buffer to not be destroyed. But `AsRef<[u8]> + 'static` effectively takes ownership of that wrapper object which wasn't obvious to me.

lehmanju:
Updated to use a wrapper struct

sdroege (dismissed):

cairo/src/image_surface.rs
```diff
@@ -149,6 +168,46 @@ impl ImageSurface {
     }
 }
 
+pub struct ImageSurfaceDataOwned {
```
sdroege, line 171: You need to export that from `lib.rs`: https://github.com/gtk-rs/gtk-rs-core/blob/master/cairo/src/lib.rs#L110

lehmanju: Done

sdroege (approved):

GuillaumeGomez (approved):

PR #223, opened 2021-08-02
glib/variant: Expose `g_variant_get_normal_form()`

cgwalters:
In ostree we compute `sha256(variant)` and expect this to be
in "normal form".  While this is a relatively obscure feature
it is public GLib API and trivial to expose.

The main benefit is this will help us drop a dependency on
`glib-sys` and move towards the re-export model from the 0.14
series.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -362,6 +362,12 @@ impl Variant {
         unsafe { from_glib_full(ffi::g_variant_get_data_as_bytes(self.to_glib_none().0)) }
     }
 
+    /// Returns a copy of the variant in normal form.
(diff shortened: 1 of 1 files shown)


sdroege (approved):

sdroege:
@cgwalters Would you need a 0.14 release with this and the other new variant API?

cgwalters:
> @cgwalters Would you need a 0.14 release with this and the other new variant API?

Hi yep, though I may have a few more coming.  Don't hold a 0.14 release but it can also wait a bit too!  I can actually work on a "rollup" PR against 0.14 and test my code with it.

PR #224, opened 2021-08-03
glib/variant: add fallible child-value getter

lucab:
This adds an additional method to retrieve child value from a
container in a fallible way. It returns an optional type in order
to avoid panics if the input does not meet the requirements.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -247,10 +247,24 @@ impl Variant {
     #[doc(alias = "get_child_value")]
     #[doc(alias = "g_variant_get_child_value")]
     pub fn child_value(&self, index: usize) -> Variant {
-        assert!(index < self.n_children());
         assert!(self.is_container());
+        assert!(index < self.n_children());
 
         unsafe { from_glib_full(ffi::g_variant_get_child_value(self.to_glib_none().0, index)) }
+    }
+
+    /// Try to read a child item out of a container `Variant` instance.
+    ///
+    /// It returns `None` if `self` is not a container type or if the given
+    /// `index` is larger than number of children.
+    pub fn try_child_value(&self, index: usize) -> Option<Variant> {
+        if !(self.is_container() && index < self.n_children()) {
+            return None;
+        }
+
+        let v =
+            unsafe { from_glib_full(ffi::g_variant_get_child_value(self.to_glib_none().0, index)) };
+        Some(v)
     }
 
     /// Tries to extract a `&str`.
@@ -1030,5 +1044,14 @@ mod tests {
         // Test ? conversion
         assert_eq!(u.try_get::<u32>()?, 42);
         Ok(())
+    }
+
+    #[test]
+    fn test_try_child_value() {
+        let a = ["foo"].to_variant();
+        assert!(a.try_child_value(0).is_some());
+        assert!(a.try_child_value(1).is_none());
+        let u = 42u32.to_variant();
+        assert!(u.try_child_value(0).is_none());
     }
 }


glib/src/variant.rs
```diff
@@ -244,15 +244,27 @@ impl Variant {
     ///
     /// * if `self` is not a container type.
     /// * if given `index` is larger than number of children.
-    #[doc(alias = "get_child_value")]
-    #[doc(alias = "g_variant_get_child_value")]
     pub fn child_value(&self, index: usize) -> Variant {
-        assert!(index < self.n_children());
         assert!(self.is_container());
+        assert!(index < self.n_children());
```
lucab, line 249: Self-note: this re-ordering is happening in order to align with the other conditionals, and also because `n_children()` does panic on non-container input.

glib/src/variant.rs
```diff
@@ -244,15 +244,27 @@ impl Variant {
...
 
         unsafe { from_glib_full(ffi::g_variant_get_child_value(self.to_glib_none().0, index)) }
     }
 
+    /// Try to read a child item out of a container `Variant` instance.
+    ///
+    /// It returns `None` if `self` is not a container type or if the given
+    /// `index` is larger than number of children.
+    #[doc(alias = "g_variant_get_child_value")]
+    pub fn get_child_value(&self, index: usize) -> Option<Variant> {
```
bilelmoussaoui, line 259: Would call the method try_child_value maybe ?

lucab: Hmm, I'd probably say no, naming pattern is already similar to other index-based getters in the stdlib like https://doc.rust-lang.org/std/primitive.slice.html#method.get.

glib/src/variant.rs
```diff
@@ -244,15 +244,27 @@ impl Variant {
     ///
     /// * if `self` is not a container type.
     /// * if given `index` is larger than number of children.
-    #[doc(alias = "get_child_value")]
-    #[doc(alias = "g_variant_get_child_value")]
```
sdroege, line 248: Why?

glib/src/variant.rs
sdroege: I think the naming is a bit strange now. `child_value()` and `get_child_value()` sound like they would do the same, but they actually don't.

As such I would also call this `try_child_value()`

lucab: Ack, renamed that way.

sdroege (approved):

glib/src/variant.rs
lucab: It was moved below.
But the CI complains about this too, so I've restored it back and left the new helper without aliases.

GuillaumeGomez (approved):

glib/src/variant.rs
```diff
@@ -247,12 +247,26 @@ impl Variant {
...
+        assert!(index < self.n_children());
 
         unsafe { from_glib_full(ffi::g_variant_get_child_value(self.to_glib_none().0, index)) }
     }
 
+    /// Try to read a child item out of a container `Variant` instance.
+    ///
+    /// It returns `None` if `self` is not a container type or if the given
+    /// `index` is larger than number of children.
+    pub fn try_child_value(&self, index: usize) -> Option<Variant> {
```
cgwalters, line 260: Do you see any use cases for this where the caller wouldn't immediately want to do `.get()` on this?  Searching the source...actually right now most of the uses of `child_value()` immediately call `.get()`.

lucab: I think you are wondering about directly having a fallible generic getter. Yes, I think it would make sense, I also observed that pattern in basically all consumers. (Not sure if we are biased though as we look more or less at the same codebases, and I didn't experience many other glib-based projects other than those.)

cgwalters: https://github.com/gtk-rs/gtk-rs-core/pull/226

PR #227, opened 2021-08-04
variant: Add a `bytes()` method.

cgwalters:
Similarly to how we have a `str()` method which avoids copying
string data just to read it, add a method to access a byte array
directly.  Right now, dealing with GVariant type `ay` is extremely
expensive because we'll end up heap allocating a variant for *each
byte* whether one does `let data: Vec<u8> = v.get()` or iterates.

Note that following previous design discussion, this API returns
`Result` and not `Option`.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -34,6 +34,16 @@
 //! assert_eq!(hello.str(), Some("Hello!"));
 //! assert_eq!(num.str(), None);
 //!
+//! // `bytes` tries to borrow a byte array (GVariant type `ay`),
+//! // rather than creating a deep copy which would be expensive for
+//! // nontrivially sized byte arrays.
+//! // The test data here is the zstd compression header, which
+//! // stands in for arbitrary binary data (e.g. not UTF-8).
+//! let bufdata = b"\xFD\x2F\xB5\x28";
+//! let bufv = bufdata.to_variant();
+//! assert_eq!(bufv.bytes().unwrap(), bufdata);
+//! assert!(num.bytes().is_err());
+//!
 //! // Variant carrying a Variant
 //! let variant = Variant::from_variant(&hello);
 //! let variant = variant.as_variant().unwrap();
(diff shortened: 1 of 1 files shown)


glib/src/variant.rs
```diff
@@ -290,6 +290,28 @@ impl Variant {
         }
     }
 
+    /// Tries to extract a `&[u8]` from a variant of type `ay`.
```
GuillaumeGomez, line 293: What does `ay` mean?

GuillaumeGomez (approved):

glib/src/variant.rs
cgwalters: https://gitlab.gnome.org/GNOME/glib/-/blob/be4b8cbe96ecd8d7d738fb9c80729285123ecdf3/glib/gvarianttype.c#L125

GuillaumeGomez: Maybe write it in full text in the documentation then. Not everyone is a gnome developer (myself included).

cgwalters: Done.

glib/src/variant.rs
```diff
@@ -34,6 +34,14 @@
 //! assert_eq!(hello.str(), Some("Hello!"));
 //! assert_eq!(num.str(), None);
 //!
+//! // `bytes` tries to borrow a byte array (GVariant type `ay`).
```
GuillaumeGomez, line 37: Please add the explanation here as well.

GuillaumeGomez (approved):

glib/src/variant.rs
```diff
@@ -290,6 +300,28 @@ impl Variant {
         }
     }
 
+    /// Tries to extract a `&[u8]` from a variant of type `ay` (array of bytes).
+    ///
+    /// Returns an error if the type is not `ay`.
+    pub fn bytes(&self) -> Result<&[u8], VariantTypeMismatchError> {
```
sdroege, line 306: Should really update all the other `Option`s to `Result`s :)

sdroege (approved):

glib/src/variant.rs
cgwalters: Yeah, I'd just also like to use this stuff on 0.14 so will try to do a larger pass for "0.15 API changes" at some point later.

PR #229, opened 2021-08-05
variant: Add an API to iterate over `as`

cgwalters:
In ostree and rpm-ostree in particular we have quite a lot of
variants of type `as` (array of string).  Right now, accessing
those requires lots and lots of allocations.  Honestly, it's
mostly not truly performance sensitive, but I'd like to avoid
doing obviously silly things so that it doesn't become
a problem down the line.

We've had `.str()` for quite a while to access a variant of type `s`
without copying out the string data.  This extends that model to
variants of type `as` by allowing iteration over the values.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -88,9 +88,9 @@ use crate::gstring::GString;
 use crate::translate::*;
 use crate::StaticType;
 use crate::Type;
-use crate::VariantIter;
 use crate::VariantTy;
 use crate::VariantType;
+use crate::{VariantIter, VariantStrIter};
 use std::borrow::Cow;
 use std::cmp::{Eq, Ordering, PartialEq, PartialOrd};
 use std::collections::HashMap;
@@ -391,10 +391,44 @@ impl Variant {
     }
 
     /// Create an iterator over items in the variant.
+    ///
+    /// Note that this heap allocates a variant for each element,
+    /// which can be particularly expensive for large arrays.
     pub fn iter(&self) -> VariantIter {
         assert!(self.is_container());
 
         VariantIter::new(self.clone())
+    }
+
+    /// Create an iterator over borrowed strings from a GVariant of type `as` (array of string).
+    ///
+    /// This will fail if the variant is not an array of with
+    /// the expected child type.
+    ///
+    /// A benefit of this API over [`Self::iter()`] is that it
+    /// minimizes allocation, and provides strongly typed access.
+    ///
+    /// ```
+    /// # use glib::prelude::*;
+    /// let strs = &["foo", "bar"];
+    /// let strs_variant: glib::Variant = strs.to_variant();
+    /// for s in strs_variant.array_iter_str()? {
+    ///     println!("{}", s);
+    /// }
+    /// # Ok::<(), Box<dyn std::error::Error>>(())
+    /// ```
+    pub fn array_iter_str(&self) -> Result<VariantStrIter, VariantTypeMismatchError> {
+        let child_ty = String::static_variant_type();
+        let actual_ty = self.type_();
+        let expected_ty = child_ty.with_array();
+        if actual_ty != expected_ty {
+            return Err(VariantTypeMismatchError {
+                actual: actual_ty.to_owned(),
+                expected: expected_ty,
+            });
+        }
+
+        Ok(VariantStrIter::new(self))
     }
 
     /// Variant has a container type.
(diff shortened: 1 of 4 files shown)


glib/src/variant.rs
```diff
@@ -391,12 +391,46 @@ impl Variant {
...
+    /// ```
+    /// # use glib::prelude::*;
+    /// let strs = &["foo", "bar"];
+    /// let strs_variant: glib::Variant = strs.to_variant();
+    /// for s in strs_variant.array_iter_str()? {
+    ///     println!("{}", s);
+    /// }
+    /// # Ok::<(), Box<dyn std::error::Error>>(())
+    /// ```
+    pub fn array_iter_str<'a>(&'a self) -> Result<VariantStrIter<'a>, VariantTypeMismatchError> {
```
sdroege, line 420: We would probably want the same around `Variant::bytes()`?

glib/src/variant_iter.rs
```diff
@@ -59,6 +60,73 @@ impl DoubleEndedIterator for VariantIter {
 
 impl ExactSizeIterator for VariantIter {}
 
+/// Iterator over items in a variant of type `as`.
+#[derive(Debug)]
+pub struct VariantStrIter<'a> {
```
sdroege, line 65: And in that case this should probably become a macro as the code will be the same, just the getter for each child value, the returned type and the name would be different for each iterator.

Could also be done generically but as long as it's only for 2-3 iterators...

sdroege (dismissed):
Looks good to me otherwise. Let me know if you want to add the bytes container iterator here

glib/src/variant.rs
```diff
@@ -391,12 +391,46 @@ impl Variant {
...
+        let child_ty = String::static_variant_type();
+        let actual_ty = self.type_();
+        let expected_ty = child_ty.with_array();
+        if actual_ty != &expected_ty {
```
sdroege, line 424:
```
error: taken reference of right operand
   --> glib/src/variant.rs:424:12
    |
424 |         if actual_ty != &expected_ty {
    |            ^^^^^^^^^^^^^------------
    |                         |
    |                         help: use the right value directly: `expected_ty`
    |
    = note: `-D clippy::op-ref` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#op_ref
```

glib/src/variant.rs
sdroege:
```
error: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
   --> glib/src/variant.rs:420:5
    |
420 |     pub fn array_iter_str<'a>(&'a self) -> Result<VariantStrIter<'a>, VariantTypeMismatchError> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::needless-lifetimes` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
```

cgwalters: > We would probably want the same around Variant::bytes()?

For `aay` you mean?  I've used this type signature occasionally but it seems pretty rare.  It's also not very expensive to do `v.iter().map(|v| v.bytes())` right?

sdroege: That won't work because iteration creates a new variant, and the lifetime of the byte slice is constrained by the lifetime of that new variant.

But yes, I meant `aay`

cgwalters: Right, you'd need to do instead something like:

```
let aay = &[b""].to_variant();
for buf in aay.iter() {
  let buf: &[u8] = buf.bytes();
  ...
}
```
which I think would be fine right?

glib/src/variant_iter.rs
cgwalters: Yeah, I actually have another mostly-done PR to add `iter_get()` which would heavily overlap with this, but I realized I mostly needed it for `as` and then realized I wanted this special case to avoid cloning to `String`.

In the end...I'm not sure because what we may really want is the equivalent of `g_variant_get_fixed_array()` for e.g. `au` (a generalization of `.bytes()`) - but those are also pretty obscure IME.

My general rule of thumb is that duplicating (similar but not identical) code twice can be OK, but if it becomes 3 versions (and definitely more) then that's usually the line where it makes sense to factor out common bits.  Which in this case would have to be a macro because we can't do [generic borrowing iterators yet](https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html).

cgwalters:
I have been looking through the CI pages here and I cannot find the actual error.  I just see "This check failed" with no additional information in e.g. https://github.com/gtk-rs/gtk-rs-core/pull/229/checks?check_run_id=3253268213

sdroege:
Strange. Let's re-run.

sdroege:
I think github actions have a problem.

sdroege (approved):

PR #230, opened 2021-08-05
0.14 backports

sdroege:


GuillaumeGomez (approved):

sdroege:
@GuillaumeGomez What should we do about the CI? Needs newer Rust version because of cfg-expr but we don't really want to increase the MSRV for the stable release. Also will probably cause new clippy warnings because you wanted to run clippy with beta ;)

GuillaumeGomez:
I guess we don't have the choice: we need to remove commits changing the MSRV...

sdroege:
> I guess we don't have the choice: we need to remove commits changing the MSRV...

That's outside our control. We could add a `Cargo.lock` though.

GuillaumeGomez:
Not a big fan of `.lock` files. What about using `crate = "=0.x.y"` for `cfg-expr`?

sdroege:
> Not a big fan of `.lock` files. What about using `crate = "=0.x.y"` for `cfg-expr`?

That will break things for people using newer Rust and having newer cfg-expr in their dependency tree. How is that better than Cargo.lock? Seems worse in every aspect :) Cargo.toml also affects users, Cargo.lock only affects people building the gtk-rs-core repository directly (and not as a dependency).

cgwalters:
I hit the cfg-expr bit and dug into it a little and AFAICS the only thing was the new `Some(a | b)` bits in https://github.com/EmbarkStudios/cfg-expr/commit/7f856d7dec36943fbc0cc1f8310750d59b434137 which seems trivial to undo and I can't imagine would be a burden to carry for them.  Let's just put together the argument to do so in a PR?

cgwalters:
https://github.com/EmbarkStudios/cfg-expr/pull/31

GuillaumeGomez:
Should be fixed now.

sdroege:
@GuillaumeGomez Can you do the same fix for the other unconnected test crate?

GuillaumeGomez:
Sure!

GuillaumeGomez:
Fixed as well, merging.

PR #235, opened 2021-08-13
glib: add manual bindings for collation utils

pbor:


glib/src/unicollate.rs
```diff
@@ -0,0 +1,54 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::translate::*;
+
+#[derive(Clone, Debug, PartialEq, PartialOrd)]
+pub struct CollateKey(String);
```
sdroege, line 6: How would you actually use this? This should probably either return a `String` directly as a standalone function, or `Deref`/etc into the contained `String`.

A standalone function seems to make most sense.

pbor: Why? you just need something that can be used for sorting, you do not need the string itself, even the C doc says that the only thing you can do with the returned string is to call `strcmp` on it.

So for example, if you have a `ListModel` where you want to show a list of files, you would implement your item object with

```
struct Item {
      label: String,
      collate_key: CollateKey,
}
```

and then use the `.collate_key` to sort your list

glib/src/unicollate.rs
```diff
@@ -0,0 +1,54 @@
...
+
+impl CollateKey {
+    #[doc(alias = "g_utf8_collate_key")]
+    pub fn new<T: AsRef<str>>(s: T) -> Self {
```
sdroege, line 10: Or should that be a `From` impl?

glib/src/unicollate.rs
sdroege: Hmm I guess that makes sense then, yes. @GuillaumeGomez @bilelmoussaoui ?

GuillaumeGomez: I'm definitely not convinced. :-/

glib/src/unicollate.rs
pbor: sure

glib/src/unicollate.rs
pbor: I would need a bit more explanation on what does not convince you to try to explain :)

It does not seem so controversial to me: these C function return a "key" whose only property is to be sortable, you are not supposed to make any assumption on how the string is formatted, you are not supposed to print them, you are not supposed to edit them etc. Since C is not very expressive, the functions return a `char *` and just document these assumptions.

In rust we can easily return a newtype that implements `Ord` and prevent a class of errors like using the key in a user visible string just because in latin languages the key at the moment happens to be the same as the string itself. It looks to me like a fairly common use of newtypes in rust...

GuillaumeGomez: Well, first your object only takes owned strings, you could use `Cow` instead. Also, this is a very specific API that people have to know about otherwise it'll never be used. That's this second part which is problematic for me. It seems to be providing something C-specific whereas we can achieve the same in rust (or at least I think).

sdroege: > Well, first your object only takes owned strings, you could use Cow instead.

The strings come from C, there's no way to use `Cow` here.

> It seems to be providing something C-specific whereas we can achieve the same in rust (or at least I think).

It's not C specific. It's unicode API and for some reason `std` does not provide corresponding API. There are various crates for this though, like ones depending on the ICU C library.

GuillaumeGomez: > The strings come from C, there's no way to use `Cow` here.

And what if I want to create a `CollateKey` on my own to compare if another coming from C?

> It's not C specific. It's unicode API and for some reason `std` does not provide corresponding API. There are various crates for this though, like ones depending on the ICU C library.

Then I'm missing information. Unless I'm missin something, it's simply a case insensitive comparison, no?

sdroege: > And what if I want to create a CollateKey on my own to compare if another coming from C?

You better don't because unless you implemented unicode collation correctly you'll probably end up doing things wrong :)

> it's simply a case insensitive comparison, no?

It's language/locale dependent, and even without that there's no concept of case insensitivity in all scripts and complex rules (is ä an a with diacritics or an a-umlaut, i.e. same letter or different letter? Is I the capital form of i or ı? in Japenese カアー«< カイー and キイー«< キイー are considered equivalent, ...). If you're bored, https://unicode-org.github.io/icu/userguide/collation/concepts.html is a good explanation.

GuillaumeGomez: > You better don't because unless you implemented unicode collation correctly you'll probably end up doing things wrong :)

I meant:

```rust
let x  = CollateKey("my string");
```

pbor: You can't do that, can you? the inner string is not marked `pub`

GuillaumeGomez: Ah damn. Ok, now I get it, just ignore my comment.

GuillaumeGomez:
Seems like a good idea. Please add documentation and it's good for me.

glib/src/unicollate.rs
sdroege: Also you really *shouldn't* do that. Your random string is probably not a valid collate key.

GuillaumeGomez: That's what I understood afterwards. Sometimes I'm a bit slow...

pbor:
Documenting made me realize we can do even better in rust: by using separate newtypes we can also prevent mixing collation keys for strings and for filenames by mistake

glib/src/lib.rs
```diff
@@ -89,6 +89,8 @@ pub mod translate;
 mod gstring;
 pub use self::gstring::GString;
 pub mod types;
+mod unicollate;
+pub use self::unicollate::CollationKey;
```
sdroege, line 93:
```suggestion
pub use self::unicollate::{CollationKey, FilenameCollationKey};
```

sdroege (approved):

PR #239, opened 2021-08-18
gio: Fix behavior of `SettingsBindFlags::INVERT_BOOLEAN`

ids1024:
`g_settings_bind_with_mapping` doesn't support `G_SETTINGS_BIND_INVERT_BOOLEAN` and asserts it isn't in the flags.

To fix that, use `g_settings_bind` when both mapping functions are `None`.

(In gio `g_settings_bind` just calls `g_settings_bind_with_mapping`, and when `G_SETTINGS_BIND_INVERT_BOOLEAN` is passed it removes the flag and passes static mapping function to invert the value.)

### DIFF
--- a/gio/src/settings.rs
+++ b/gio/src/settings.rs
@@ -76,31 +76,44 @@ impl<'a> BindingBuilder<'a> {
         unsafe extern "C" fn destroy_closure(ptr: *mut libc::c_void) {
             Box::<Mappings>::from_raw(ptr as *mut _);
         }
-        let get_trampoline: Option<unsafe extern "C" fn(_, _, _) -> _> =
-            if self.get_mapping.is_none() {
-                None
-            } else {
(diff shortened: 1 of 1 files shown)


sdroege:
That still allows running into that assertion or am I missing something? And it allows setting that flag while providing mapping functions, which is probably also not very useful?

ids1024:
Yes, if you pass one or both mapping functions the same error shows up. Not sure there's a good way to statically prevent that, other than having separate functions/builders for `g_settings_bind` and `g_settings_bind_with_mapping` with separate flag types.

(Also I guess it's not technically an "assertion"; it calls `g_return_if_fail`.)

sdroege (approved):

sdroege:
Yeah, let's get this in then for now. It's an improvement at least :)

PR #241, opened 2021-08-20
glib: Provide default implementations for `IsSubclassable` methods

ids1024:
Also provides `parent_class_init()` and `parent_instance_init()` methods to make manual implementation a bit neater.

The implementation here is a bit of a mess, since specialization doesn't exist and `Self::default_class_init()` would error due to not being able to infer `T`, without using fully qualified syntax.

The Resulting API seems like an improvement though.

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -2,7 +2,7 @@
 
 //! Module that contains the basic infrastructure for subclassing `GObject`.
 
-use crate::object::{Cast, ObjectSubclassIs, ObjectType};
+use crate::object::{Cast, IsClass, ObjectSubclassIs, ObjectType, ParentClassIs};
 use crate::translate::*;
 use crate::{Closure, Object, StaticType, Type, Value};
 use std::marker;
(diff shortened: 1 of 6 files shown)


sdroege (commented):
Nice improvement but why do we need both traits? They look basically the same and the Ext one seems sufficient?

ids1024:
It's definitely annoying to have two traits, but as far as I can tell it's necessary for the API here.

`IsSubclassableDefault<T>` has the generic `T`. Calling `Self::default_class_init(class)` in a manual implementation of `class_init` errors because it can't infer `T`. It can be called as `<Self as IsSubclassableDefault<T>>::default_class_init(class)`, but that isn't a huge improvement (though it is at least the same for every implementation).

`IsSubclassableExt` moves the trait bound to the method, so it can be called as just `Self::parent_class_init::<T>(class)`. But it is only implemented for subclasses, not for `Object`, since the methods need a `where Self::Parent: IsSubclassable<T>` bound.

glib/src/subclass/types.rs
```diff
@@ -115,19 +115,72 @@ pub unsafe trait ClassStruct: Sized + 'static {
...
     ///
     /// This is automatically called during instance initialization and must call `instance_init()`
     /// of the parent class.
-    fn instance_init(instance: &mut InitializingObject<T>);
+    fn instance_init(instance: &mut InitializingObject<T>) {
+        Self::default_instance_init(instance);
+    }
+}
+
+pub trait IsSubclassableDefault<T: ObjectSubclass>: IsClass {
```
sdroege, line 137: In that case, if this is only meant to be called from the default impl, we don't really need to make this `pub`, right?

ids1024: Ideally yes, but Rust wouldn't allow the type bound in the definition of `IsSubclassable` if the trait isn't public.

ids1024: I guess with specialization, `IsSubclassableDefault` could be implemented with empty function bodies for everything that implements `IsSubclassable`, with a specialization to call the parent method when `ParentClassIs` is implemented. Then the `IsSubclassable<T>: IsSubclassableDefault<T>`  bound would be unneeded and the trait could be private.

Too bad specialization doesn't seem like it will be stable particularly soon.

sdroege: Ok thanks for the explanation :) Can you put some FIXME comment about that in the code to remind us to possibly look for a better solution once Rust is ready for that?

sdroege:
@bilelmoussaoui @GuillaumeGomez Once this is merged, you can simplify things in `gtk/src/subclass` btw.

glib/src/subclass/types.rs
ids1024: Added `#[doc(hidden)]` and `FIXME` comment.

sdroege (approved):

sdroege:
See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/840 for the gstreamer-rs update.

sdroege:
This broke gstreamer-rs, needed some more fixes.

sdroege:
This also breaks gtk3-rs:

```
error[E0277]: the trait bound `T: BinImpl` is not satisfied
   --> /root/.cargo/git/checkouts/gtk3-rs-a818089b55180ee1/eed0e28/gtk/src/subclass/button.rs:47:28
    |
47  | unsafe impl<T: ButtonImpl> IsSubclassable<T> for Button {
    |                            ^^^^^^^^^^^^^^^^^ the trait `BinImpl` is not implemented for `T`
    | 
   ::: /root/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/de54393/glib/src/subclass/types.rs:118:53
    |
118 | pub unsafe trait IsSubclassable<T: ObjectSubclass>: IsSubclassableDefault<T> {
    |                                                     ------------------------ required by this bound in `glib::subclass::types::IsSubclassable`
    |
```

I'm going to revert this again until there's a PR for gtk3-rs and gtk4-rs.

PR #243, opened 2021-08-23
Remove the `links` annotations

nagisa:
The `links` annotation in `Cargo.toml` is intended to ensure that in the
crate graph there's at most one crate that's an implementation of
some sort concept.

This can make sense in some scenarios, most prominent of which is when
the crate defines `#[no_mangle]` symbols (e.g. by compiling a vendored C
library.) In that situation linking a binary that depends on two
versions of the library cannot work because of colliding symbol names.

There does not appear to be a similar reason to impose such a
restriction on the users of `glib-sys` and similar, however. All of
these crates link to a system library, they do not define any
`#[no_mangle]` symbols nor they don't vendor or build any C libraries as
part of their process. All they do is linking to a system library, and
most likely all the different versions of the bindings will link to the
exact same library too.

I haven't seen any global resources that these bindings use to ensure
soundness of the library, either.


nagisa:
From the looks of it, what I need to do instead is to adjust the `gir` crate to stop generating this field instead...

sdroege:
> I haven't seen any global resources that these bindings use to ensure
> soundness of the library, either.

What do you mean with that?

> From the looks of it, what I need to do instead is to adjust the `gir` crate to stop generating this field instead...

Yes, but fortunately that shouldn't be a too difficult change :)

> There does not appear to be a similar reason to impose such a
> restriction on the users of `glib-sys` and similar, however.
> [...]

That explanation makes sense to me, thanks.

nagisa:
> What do you mean with that?

So one example would be having something like:

```
static GLOBAL_MUTEX_RESOURCE: Mutex<()> = StaticMutex::new();

fn do_something_serial() {
    let lock = GLOBAL_MUTEX_RESOURCE.lock();
    gently_prod_the_resource(); // this must not be called concurrently in the executable, or the rocket explodes.
    lock.unlock();
}
```

Here, having a duplicate implementation in the dependency tree would allow you to call `crate1::do_something_serial()` and `crate2::do_something_serial()` concurrently nevertheless, because the mutex also gets duplicated.

A resource in these situations tends to be something that manifests some phenomena in our physical world, but sometimes also comes up as a solution to some library that does not acknowledge concurrent execution exists.

sdroege:
Right. There's no such thing anywhere in the -sys crates. In the non-sys crates there is something similar, e.g. https://github.com/gtk-rs/gtk-rs-core/blob/75c847827a349d233f872f22540d6feebc54f240/gio/src/read_input_stream.rs#L26

If you have two `gio` crates in your dependency tree, only of the two will succeed to register that type and if both are trying then the second one would panic.

As the non-sys crates don't even have `links = ` in the `Cargo.toml` and this doesn't cause soundness issues (it just panics), it doesn't seem too much of a problem though.

nagisa:
Updated `gir` and regenerated everything, too. At the very least this acts as a confirmation that regenerating `gir` won't add `links` back in again.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

sdroege:
@nagisa Is this important for you to have in a 0.14.4 bugfix release or doesn't matter?

nagisa:
I'm good, no need for a release.

PR #245, opened 2021-08-23
glib: adapt to signed vs. unsigned libc::c_char to fix non-x86 builds

### LINKED ISSUE #244: [BUG] glib 0.14.3 fails to typecheck / compile on non-x86 architectures
Version 0.14.3 of the `glib` crate does not compile on any non-x86 architecture (aarch64, armv7hfp, powerpc64le, s390x).

This commit seems to have introduced code that does not typecheck correctly on non-x86 architectures:
https://github.com/gtk-rs/gtk-rs-core/commit/9b9d47876a5898f5d9c756f8ea0f09bf5c54a204

This is the compiler error message:

```
error[E0308]: mismatched types
  --> src/variant_iter.rs:88:17
   |
88 |                 s as *const u8 as *const i8,
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
   |
   = note: expected raw pointer `*const u8`
              found raw pointer `*const i8`
error: aborting due to previous error
For more information about this error, try `rustc --explain E0308`.
error: could not compile `glib`
```

decathorpe:
Fixes #244

Changing to the simpler `s as *const _` does not work, as it creates the opposite effect (glib fails to compile on x86 targets), so the double cast seems to be necessary.

I have confirmed that the glib crate with this patch compiles successfully on x86_64, i686, ppc64le, aarch64, armv7hl, and s390x.

### DIFF
--- a/glib/src/variant_iter.rs
+++ b/glib/src/variant_iter.rs
@@ -85,7 +85,7 @@ impl<'a> VariantStrIter<'a> {
             ffi::g_variant_get_child(
                 self.variant.to_glib_none().0,
                 i,
-                s as *const u8 as *const i8,
+                s as *const u8 as *const _,
                 &p,
                 std::ptr::null::<i8>(),
             );


sdroege (approved):

sdroege:
@decathorpe I'll do another bugfix release for that some time later. Thanks!

decathorpe:
No hurry, I've included this patch in our builds on Fedora. But thanks :)

sdroege:
Also CC @cgwalters

glib/src/variant_iter.rs
```diff
@@ -85,7 +85,7 @@ impl<'a> VariantStrIter<'a> {
             ffi::g_variant_get_child(
                 self.variant.to_glib_none().0,
                 i,
-                s as *const u8 as *const i8,
+                s as *const u8 as *const _,
                 &p,
                 std::ptr::null::<i8>(),
```
sdroege, line 90: @decathorpe This one is not a problem?

cgwalters: Would it work to just do `s as *const _`?

cgwalters: Nevermind, I see this was covered in the commit message.

decathorpe: @sdroege apparently not ... which data types and which pointers can be converted to which others implicitly is a bit beyond me, but probably NULL is always NULL.

cgwalters: Regarding `std::ptr::null::<i8>()` I suspect it doesn't matter because it's part of the varargs parameters, which inherently aren't type checked.

sdroege: Ah, of course. Thanks :)

PR #254, opened 2021-08-31
[Regen] Update system-deps version

GuillaumeGomez:
Part of gtk-rs/gir#1215.


sdroege (changes requested):
You forgot cairo

GuillaumeGomez:
@sdroege Keep forgetting about it... Great catch, thanks! Updated.

sdroege (approved):

PR #255, opened 2021-08-31
gio: Bindings for `DBusObjectManager*`, `Gio.DBusObjectSkeleton`

ids1024:
At the moment, Gir doesn't seem to be able to generate the async constructors for `DBusObjectManagerClient`, but the blocking versions should work anyway...


sdroege:
CI needs fixing but otherwise looks OK to me. Can this actually be used like this, without a way to implement subclasses of those types?

ids1024:
These types all have constructors, so presumably they should be usable without subclassing. Though `DBusInterfaceSkeleton` (which already has bindings) is an abstact base class, so I think subclassing support for that at least is probably necessary to properly provide a dbus service.

sdroege:
> so I think subclassing support for that at least is probably necessary to properly provide a dbus service.

Yeah that was my point. Can all this be used without that? :)

sdroege:
@ids1024 ping?

bilelmoussaoui:
Let us close this one, feel free to re-open if you want to push this forward

PR #256, opened 2021-08-31
glib: Provide default implementations for `IsSubclassable` methods

ids1024:
Reopening https://github.com/gtk-rs/gtk-rs-core/pull/241.

https://github.com/gtk-rs/gtk3-rs/pull/629 seems sufficient to fix the issues using this with gtk3-rs. Of course, once this is merged further improvement can be made there.


sdroege:
@ids1024 @bilelmoussaoui GTK4 is also passing and doesn't need any changes?

ids1024:
It didn't seem to require any changes to gtk4 when I tried it. Not sure exactly what is needed in gstreamer-rs.

sdroege:
That's already taken care of in my MR, but it also means an addition of many useless empty `impl GstObjectImpl for MyType {}` everywhere, which makes me a bit unhappy :)

sdroege:
Sorry, I completely forgot about this one. I'll try to get it in during the weekend :)

PR #261, opened 2021-09-04
bind gtask run in thread

pbor:


gio/src/task.rs
```diff
@@ -60,6 +60,51 @@ impl Task {
...
+    pub fn run_in_thread<Q>(&self, task_func: Q)
+    where
+        Q: FnOnce(&Self, Option<&glib::Object>, Option<&Cancellable>),
+        Q: Send + 'static,
+    {
+        let task_func_data = Box_::new(task_func);
+
+        // FIXME: we store the func pointer into the task data
+        // for now this is okaish, since we do not expose a way to
+        // set the task data in the bindings
```
sdroege, line 73: Where else would you store it?

gio/src/task.rs
```diff
@@ -60,6 +60,51 @@ impl Task {
...
+        unsafe {
+            ffi::g_task_set_task_data(
```
sdroege, line 75: If `run_in_thread()` is called multiple times, this will probably cause interesting problems? Or is that actually OK?

gio/src/task.rs
pbor: Yeah, I think this is the only place where we can store it: I left the fixme just to double check with you.

Shall I explicitly mark set_data as ignore in the gir.toml()

gio/src/task.rs
pbor: I do not think you can call `run_in_thread` multiple times, no?

sdroege: What's preventing it from being called twice currently? :)

gio/src/task.rs
sdroege: Would make sense, thanks :)

gio/src/task.rs
pbor: mmm... it never even occurred to me to call it multiple times even from the C api, but you are right.

Any suggestion on how to enforce this?

sdroege: Maybe check if there is task data, and if there is panic

sdroege (approved):

Issue #265, opened 2021-09-15
[BUG] Surface map_to_image assertion failed


danielronnkvist:
**Bug description**

Trying to use the `map_to_image` method on a ImageSurface but I'm getting an error:
>Assertion failed: (CAIRO_REFERENCE_COUNT_HAS_REFERENCE (&surface->ref_count)), function cairo_surface_reference, file cairo-surface.c, line 930.

Even with a minimal example like this it fails:
```rust
fn main() {
    let surface = cairo::ImageSurface::create(cairo::Format::ARgb32, 1080, 1080).unwrap();
    let mapped = surface.map_to_image(Some(cairo::RectangleInt {
        x: 100,
        y: 100,
        width: 500,
        height: 500
    })).unwrap();
}
```

sdroege:
Thanks for reporting. This should be fixed by https://github.com/gtk-rs/gtk-rs-core/pull/266 but let me know if there's more broken around there.

danielronnkvist:
Thank you @sdroege!

sdroege:
I'll create a bugfix release for crates.io once you confirmed that there's not more broken, some time in the next days :)

danielronnkvist:
@sdroege It works as intended for my use case, thanks!

sdroege:
Thanks for confirming!

sdroege:
Released as 0.14.7 on crates.io now

PR #266, opened 2021-09-15
cairo: Fix up reference counting in Surface::map_to_image()

sdroege:
The C function is technically (transfer none) as one must not call
destroy() on it to unreference it but instead
cairo_surface_unmap_image().

This was tried to be avoided in the Drop impl of MappedImageSurface but
not correctly, leading to assertions.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/265


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #275, opened 2021-09-25
Implement FromGlibPtrBorrow for glib::Value and glib::SendValue

sdroege:


GuillaumeGomez:
What does it allow? Can we add test to ensure there is no regression too please?

sdroege:
> What does it allow?

In what sense? It implements that trait so that a bit more autogenerated code can actually compile :) I need this for some new GStreamer API.

> Can we add test to ensure there is no regression too please?

What kind of test? We don't have any such tests for any of the other `FromGlibPtrXXX` trait impls either.

GuillaumeGomez:
Well, basically, what couldn't be done before?

sdroege:
> Well, basically, what couldn't be done before?

```rust
let value: glib::translate::Borrow<glib::Value> = from_glib_borrow(value_ptr);
```

See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/856/diffs?commit_id=f3369ea0369918b48764922b7e4ab0937949f654#9c3895a1520865a3ba4e8cd33f18b489ebfb6917_0_228

GuillaumeGomez (approved):

PR #276, opened 2021-09-27
macros: emit an Instance associated type for interfaces

pbor:
This is useful in the ffi, since this is the type of the pointers
to exchange with C. This is consistent with the associated type
already emitted for GObjects.


pbor:
mmm... I guess I have to properly think about it instead of blind cut&paste

PR #279, opened 2021-10-01
docs: Fix the examples for variants

Hofer-Julian:
Before it tested the wrong variable


sdroege (approved):

sdroege:
This does not actually change the behaviour though, you just renamed the variable. But like this is clearer.

Hofer-Julian:
It doesn't but only by accident if you look at line 61.

PR #280, opened 2021-10-02
glib: Merge variants of `wrapper!` as much as possible

ids1024:
Handling object wrappers with or without `@extends` and with or without `@implements` in one variant is problematic, since a comma between the two is necessary if and only if both are used.

But assuming no issues turn up, this should be a good improvement making this a fair bit simpler.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1017,7 +1017,7 @@ macro_rules! glib_object_wrapper {
 
     // This case is only for glib::Object itself below. All other cases have glib::Object in its
     // parent class list
-    (@object [$($attr:meta)*] $name:ident, $ffi_name:ty, $ffi_class_name:ty, @type_ $get_type_expr:expr) => {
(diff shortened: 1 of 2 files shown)


sdroege:
Did you check if gtk3-rs, gtk4-rs and gstreamer-rs still build fine with this and end up with the same expanded code?

ids1024:
Both seem to build fine. Not sure how best to compare the expanded code.

sdroege:
> Not sure how best to compare the expanded code.

Good question. `cargo expand` and diffing might work?

GuillaumeGomez:
Just like @sdroege said: please give us before/after expanded code (because reviewing changes to this macro is a bit complicated...).

ids1024:
There seems to be no difference in the output of `cargo expand` for the `gtk` and `gtk4` crate. On `gtk-rs-examples`, `cargo expand` errors (https://github.com/dtolnay/cargo-expand/issues/105).

sdroege (approved):

sdroege:
Ok, the changes also look good to me so let's get it in then. We can take care of any fallout afterwards, but generally it should work fine :)

Issue #199, opened 2021-07-12
`graphene::Rect::contains_point(&self, p: &Point)` is wrong for Points above the Rect [BUG]


flxzt:
**Bug description**
`Rect::contains_point(&self, p: &Point)` returns `true` for points which are above the rectangle. It should return `false` when they are not inside it.
This is the code that can reproduce the issue:

```
use gtk4::graphene::{Point, Rect};

fn main() {
    let rect = Rect::new(100., 100., 100., 100.);

    let right = Point::new(250., 150.);
    let below = Point::new(150., 50.);
    let left = Point::new(50., 150.);
    let above = Point::new(150., 250.);

    assert!(!rect.contains_point(&right));
    assert!(!rect.contains_point(&below));
    assert!(!rect.contains_point(&left));
    assert!(!rect.contains_point(&above));
}
```

**Backtrace**

```
thread 'main' panicked at 'assertion failed: !rect.contains_point(&above)', src/main.rs:14:5
stack backtrace:
   0:     0x55897d544380 - std::backtrace_rs::backtrace::libunwind::trace::ha5edb8ba5c6b7a6c
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x55897d544380 - std::backtrace_rs::backtrace::trace_unsynchronized::h0de86d320a827db2
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55897d544380 - std::sys_common::backtrace::_print_fmt::h97b9ad6f0a1380ff
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x55897d544380 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h14be7eb08f97fe80
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x55897d55e2cf - core::fmt::write::h2ca8877d3e0e52de
                               at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/fmt/mod.rs:1094:17
[...]

PR #282, opened 2021-10-07
Use bool instead of gboolean in graphene for FFI

sdroege:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/199


GuillaumeGomez:
Thanks!

sdroege:
Should maybe make a bugfix release with this but that also needs a bugfix branch for gir then...

GuillaumeGomez:
Well, I don't expect people to regenerate with gir on the released branches...

GuillaumeGomez:
Well, I'm not sure it's strictly needed but doesn't hurt to have a "clean" code way in case we need it. Thanks for doing it!

Issue #211, opened 2021-07-26
Inline-allocate graphene structs instead of going via heap allocations


sdroege:
Currently each graphene struct is heap allocated by us, which is not really necessary and is also not done usually when using the library from C. All the heap allocations are expensive when doing lots of calculations.

gir doesn't allow generating such code but it would probably make sense to add support for it, and a `glib::wrapper!` variant that takes something like a `StackBoxed` or similar instead of just `Boxed` to inline-allocate the struct.

CC @bilelmoussaoui

PR #284, opened 2021-10-09
Add support for inline-allocated Boxed types and various graphene fixups

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/issues/211


sdroege:
> I didn't change it for quite a long time now...

Works again :shrug:

PR #292, opened 2021-10-16
Remove iOS target from quartz_surface in Cairo

ScattyNatalie:


sdroege:
I'll make a 0.14 bugfix release with this some time later this weekend.

PR #306, opened 2021-10-29
pango: regen with pre 1.50

bilelmoussaoui:
Missing since annotations were fixed in https://gitlab.gnome.org/GNOME/pango/-/merge_requests/488


pango/src/auto/font.rs
```diff
@@ -132,6 +138,16 @@ impl<O: IsA<Font>> FontExt for O {
     //    unsafe { TODO: call ffi:pango_font_get_hb_font() }
     //}
 
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    fn languages(&self) -> Option<Language> {
```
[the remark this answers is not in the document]
bilelmoussaoui: It should, it doesn't have an array & null-terminated annotation. Will send a patch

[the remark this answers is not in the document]
bilelmoussaoui: https://gitlab.gnome.org/GNOME/pango/-/merge_requests/489/

pango/src/auto/attribute.rs
```diff
@@ -30,6 +45,67 @@ glib::wrapper! {
 }
 
 impl Attribute {
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    #[doc(alias = "pango_attribute_as_color")]
+    pub fn as_color(&mut self) -> Option<AttrColor> {
```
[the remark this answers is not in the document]
bilelmoussaoui: I ignored them as we expect people to use the functions in the IsAttribute trait

sdroege (commented):
Let me know once the gir-files submodule is updated to the right version

pango/sys/Cargo.toml
```diff
@@ -54,3 +55,6 @@ version = "1.46"
 
 [package.metadata.system-deps.pango.v1_48]
 version = "1.48"
+
+[package.metadata.system-deps.pango.v1_50]
+version = "1.50"
```
sdroege, line 60: You probably want something like

```suggestion
version = "1.49.2"
```
here or not? There's no 1.50 yet

pango/src/auto/versions.txt
```diff
@@ -1,2 +1,2 @@
 Generated by gir (https://github.com/gtk-rs/gir @ 7b93dbf37ad7)
-from gir-files (https://github.com/gtk-rs/gir-files @ fa1fb41533be)
+from gir-files (https://github.com/gtk-rs/gir-files @ fa1fb41533be+)
```
sdroege, line 2: This one and then we're good :)

pango/sys/src/lib.rs
```diff
@@ -843,19 +866,18 @@ impl ::std::fmt::Debug for PangoGlyphGeometry {
 }
 
 #[repr(C)]
-#[derive(Copy, Clone)]
 pub struct PangoGlyphInfo {
     pub glyph: PangoGlyph,
     pub geometry: PangoGlyphGeometry,
-    pub attr: PangoGlyphVisAttr,
+    _truncated_record_marker: c_void,
+    // /*Ignored*/field attr has incomplete type
```
bilelmoussaoui, line 873: Meh this change is problematic, I wonder why...

pango/sys/src/lib.rs
```diff
@@ -865,13 +887,19 @@ impl ::std::fmt::Debug for PangoGlyphInfo {
 pub struct PangoGlyphItem {
     pub item: *mut PangoItem,
     pub glyphs: *mut PangoGlyphString,
+    pub y_offset: c_int,
+    pub start_x_offset: c_int,
+    pub end_x_offset: c_int,
```
bilelmoussaoui, line 892: Are those fine? don't we need different ones for < & > 1.50?

pango/sys/src/lib.rs
```diff
@@ -924,9 +952,10 @@ impl ::std::fmt::Debug for PangoGlyphString {
 }
 
 #[repr(C)]
-#[derive(Copy, Clone)]
 pub struct PangoGlyphVisAttr {
     pub is_cluster_start: c_uint,
+    _truncated_record_marker: c_void,
+    // field is_color has incomplete type
```
bilelmoussaoui, line 958: I'm not sure what it means to have an incomplete type, it seems to be correct here
```xml
    <record name="GlyphVisAttr" c:type="PangoGlyphVisAttr">
      <doc xml:space="preserve">A `PangoGlyphVisAttr` structure communicates information between
the shaping and rendering phases.

Currently, it contains cluster start and color information.
More attributes may be added in the future.</doc>
      <field name="is_cluster_start" writable="1" bits="1">
        <doc xml:space="preserve">set for the first logical glyph in each cluster.
  (Clusters are stored in visual order, within the cluster, glyphs
  are always ordered in logical order, since visual order is meaningless;
  that is, in Arabic text, accent glyphs follow the glyphs for the
  base character.)</doc>
        <type name="guint" c:type="guint"/>
      </field>
      <field name="is_color" writable="1" bits="1">
        <doc xml:space="preserve">set if the the font will render this glyph with color. Since 1.50</doc>
        <type name="guint" c:type="guint"/>
      </field>
    </record>
```

pango/sys/src/lib.rs
sdroege: Yes that's bad. You need to know the proper struct layout for being able to make it boxed inline.

What changed here in the headers for it to be completely broken now? :)

pango/sys/src/lib.rs
sdroege: That means that we were never supposed to allocate this ourselves and that it shouldn't be boxed-inline. No other changes needed apart from that.

pango/sys/src/lib.rs
sdroege: Because it's a bit field and bit fields are difficult (and behave different between compilers, etc). The support for them in our code generator is very minimal.

sdroege:
Your explanation for `GlyphInfo` is wrong (and that could very well be boxed-inline: check how it's used in C). The explanation is correct for `PangoGlyphItem`.

sdroege (approved):

PR #315, opened 2021-10-30
glib: add simpler variants of ParamSpec::new_override

### LINKED ISSUE #201: Simplify ParamSpec::new_override
Currently ParmSpec::new_override takes two parameters. The property name & the ParamSpec. Instead, it should take the type we want to override it's property and the property we want to override.

The code then should check whether it's an Interface or an Object and use `Interface::from_type` / `ObjectClass::from_type`

bilelmoussaoui:
Create two variants, one for a Class & the other for an interface.
We can't make them into one call as we will have to lose the IsInterface/IsClass checks

Fixes #201

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -2,10 +2,15 @@
 
 use crate::translate::*;
 use crate::utils::is_canonical_pspec_name;
+use crate::IsA;
 use crate::ParamFlags;
 use crate::StaticType;
 use crate::Type;
 use crate::Value;
+use crate::{
+    object::{Interface, InterfaceRef, IsClass, IsInterface, ObjectClass},
+    Object,
+};
 
 use std::char::CharTryFromError;
 use std::convert::TryFrom;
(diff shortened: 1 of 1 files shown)


glib/src/param_spec.rs
```diff
@@ -529,6 +534,57 @@ impl ParamSpec {
...
+    /// ```
+    pub fn new_override_interface<T: IsA<Object> + IsInterface>(name: &str) -> Self {
+        assert!(
+            is_canonical_pspec_name(name),
+            "{} is not a valid canonical parameter name",
+            name
+        );
+        // in case it's an interface
+        let interface_ref: InterfaceRef<T> = Interface::from_type(T::static_type()).unwrap();
+        let pspec = interface_ref.find_property(name).unwrap();
```
bilelmoussaoui, line 553: I'm not sure if we need to check if the property exists? :)

bilelmoussaoui: I suppose yes, and we can panic with a proper error message

sdroege (approved):
Make it an expect or so, otherwise seems good

sdroege (approved):

PR #316, opened 2021-10-30
gio: add manual bindings for GIOExtension

pbor:


gio/src/extension.rs
```diff
@@ -0,0 +1,200 @@
...
+    pub fn set_required_type(&mut self, type_: Type) {
+        unsafe { ffi::g_io_extension_point_set_required_type(self.0.as_ptr(), type_.into_glib()) }
+    }
+
+    #[doc(alias = "g_io_extension_point_implement")]
+    pub fn implement(
+        extension_point_name: &str,
+        type_: Type,
+        extension_name: &str,
+        priority: i32,
```
bilelmoussaoui, line 100: Could use glib::Priority instead?

gio/src/extension.rs
```diff
@@ -0,0 +1,200 @@
...
+            );
+            from_glib_none(e)
+        }
+    }
+}
+
+/// The implementation of an `IOExtensionPoint`.
+#[doc(alias = "GIOExtension")]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct IOExtension(ptr::NonNull<ffi::GIOExtension>);
```
bilelmoussaoui, line 117: Should be on a separate file for rustdoc-stripper. Otherwise it won't be documented

gio/src/extension.rs
sdroege: This is not a mainloop/source priority but just an integer

gio/src/extension.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    #[doc(alias = "g_io_extension_point_lookup")]
+    pub fn lookup(name: &str) -> Option<Self> {
+        unsafe {
+            let ep = ffi::g_io_extension_point_lookup(name.to_glib_none().0);
+            if ep.is_null() {
+                None
+            } else {
+                Some(from_glib_none(ep))
+            }
```
sdroege, line 52:
```suggestion
            from_glib_none(ep)
```

Should also work

gio/src/extension.rs
```diff
@@ -0,0 +1,200 @@
...
+        unsafe {
+            let e = ffi::g_io_extension_point_get_extension_by_name(
+                self.0.as_ptr(),
+                name.to_glib_none().0,
+            );
+            if e.is_null() {
+                None
+            } else {
+                Some(from_glib_none(e))
+            }
```
sdroege, line 81:
```suggestion
            from_glib_none(e)
```

gio/src/extension.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    #[doc(alias = "g_io_extension_point_get_required_type")]
+    pub fn required_type(&self) -> Type {
+        unsafe { from_glib(ffi::g_io_extension_point_get_required_type(self.0.as_ptr())) }
+    }
+
+    #[doc(alias = "g_io_extension_point_set_required_type")]
+    pub fn set_required_type(&mut self, type_: Type) {
```
sdroege, line 91: The `&mut` here makes no sense because the type is `Copy` but any change affects all copies anyway. Maybe this should be part of a builder together with `register()`?

gio/src/extension.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    #[doc(alias = "g_io_extension_point_implement")]
+    pub fn implement(
+        extension_point_name: &str,
+        type_: Type,
+        extension_name: &str,
+        priority: i32,
+    ) -> IOExtension {
+        unsafe {
+            let e = ffi::g_io_extension_point_implement(
```
sdroege, line 103: This should probably assert that `type_` is a / implements the required type and that the extension point actually exists

gio/src/extension.rs
pbor: I did not know! I will split... does it need to follow a specific naming rule? E.g. be called `io_extension.rs` instead of just `extension.rs`?

gio/src/extension.rs
pbor: This is done internally by the glib function, do we need to duplicate it? Or do you mean that we have a way to enforce it with the type system?

sdroege: Well, what will happen if you call the C function wrong here? `g_critical()` and then it panics because the returned pointer is `NULL`?

pbor: `g_warning` and then panics because of `NULL`... shall I make it fallible and return `Option`?

pbor: Updated to do that

sdroege: No, it should panic in one way or another. It's a programmer error, not something the programmer is supposed to handle at runtime.

pbor: mmm, but the glib side are warnings... I am wondering if that is intentional so that a broken extension does not crash the entire app

sdroege: Just like `g_return_val_if_fail()`. If that ever happens then your application is not going to behave as it should.

pbor: Note also that returning an `Option` does not make the code harder for the caller, since normally you do not really need to check that return value

sdroege: Yes and then the error is completely silently ignored :)

pbor: `g_return_val_if_fail` is a critical you can turn into an assert, these ones are more legitimate runtime warnings IMHO.

If eg my app has an extension point and at some point I bump the ABI and require extensions to implement a new GType, I want my app to not crash even if there are stray .so in the folder and I want my app to log that the extension is ignored.

Sure, on the caller side the error will be ignored, but then again the C api returns a pointer just for the far fetched case you want to call it multiple times.

sdroege: Seems like a case where you clearly want your application to stop functioning because the whole environment is not set up the way it's expected to, and it won't work as expected (i.e. the extension implementor wouldn't be used).

pbor: I do not agree: if eg a third party nautilus plugin that adds some simple convenience as an extra item in the popup menu is outdated with a new version of nautilus, I do not want the entire nautilus to crash: I want the plugin ignored and if I really care about that plugin, I can see in the logs that it has been ignored.

Also, the policy here is set in gio, which logs a warning and ignores the extension, I do not think the Rust bindings should enforce a different policy in this regard.

With that said, I promise this is my last comment on the matter :-)
If you still think we should panic, I will add back the assert and push an updated patch.

sdroege: Ok, let's keep it like this for now then.

gio/src/io_extension_point.rs
```diff
@@ -0,0 +1,186 @@
...
+use glib::translate::*;
+use glib::Type;
+use std::fmt;
+use std::ptr;
+
+use crate::io_extension::IOExtension;
+
+/// Builder for extension points.
+#[derive(Debug)]
+pub struct IOExtensionPointBuilder<'a> {
```
sdroege, line 12:
```suggestion
#[must_use("Builder doesn't do anything unless built")]
pub struct IOExtensionPointBuilder<'a> {
```

Or something like that

gio/src/io_extension.rs
```diff
@@ -0,0 +1,51 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+use glib::Type;
+use std::fmt;
+use std::ptr;
+
+/// The implementation of an `IOExtensionPoint`.
+#[doc(alias = "GIOExtension")]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
```
sdroege, line 10: You might want to do a manual `Debug` impl that prints the name of the extension at least. That would seem useful for debugging purposes.

pbor: Should I leave `Display` as is, or should I change that to show the name as well?

bilelmoussaoui: The display seems fine to me

gio/src/io_extension.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::ptr;
+
+/// The implementation of an `IOExtensionPoint`.
+#[doc(alias = "GIOExtension")]
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct IOExtension(ptr::NonNull<ffi::GIOExtension>);
+
+impl fmt::Debug for IOExtension {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "IOExtension {}", self.name())
```
bilelmoussaoui, line 15:
```suggestion
        f.debug_struct("IOExtension")
           .field("name", &self.name())
           .field("priority", &self.priority())
           .finish()
```
Would do something like this

pbor: Cool, I learned something new :)

gio/src/io_extension.rs
sdroege: Whatever you prefer, personally I don't think any of these types should implement `Display`...

sdroege (dismissed):

sdroege:
> #### Conflicting files
> `gio/Gir.toml`

Needs merge conflicts to be solved

PR #319, opened 2021-10-31
gio: Add various `gio::File` missing functions

sdroege:


gio/src/auto/file_attribute_info_list.rs
```diff
@@ -0,0 +1,48 @@
...
+    match fn {
+        ref => |ptr| ffi::g_file_attribute_info_list_ref(ptr),
+        unref => |ptr| ffi::g_file_attribute_info_list_unref(ptr),
+        type_ => || ffi::g_file_attribute_info_list_get_type(),
+    }
+}
+
+impl FileAttributeInfoList {
+    #[doc(alias = "g_file_attribute_info_list_new")]
+    pub fn new() -> FileAttributeInfoList {
```
bilelmoussaoui, line 22:
```suggestion
    pub fn new() -> Self {
```

sdroege: That's autogenerated code :P

bilelmoussaoui: Oups, i shouldn't have reviewed that from my phone :P

gio/src/file_attribute_info_list.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use glib::translate::*;
+
+use std::fmt;
+use std::mem;
+use std::ptr;
+
+glib::wrapper! {
+    #[doc(alias = "GFileAttributeInfo")]
+    pub struct FileAttributeInfo(BoxedInline<ffi::GFileAttributeInfo>);
```
bilelmoussaoui, line 13: should be on a `file_attribute_info.rs` file for rustdoc-stripper to pick it up. Otherwise it won't be documented

sdroege: Didn't know that, thanks :) Changed accordingly.

bilelmoussaoui: I tend to run embed-docs and then strip them to see if a manual code is going to be documented. Pretty handy

sdroege:
So it seems like the GLib tests consistently deadlock on Windows since a while. That probably need some debugging.

sdroege:
Ah, interestingly I can reproduce that one.

sdroege:
The update to the 2021 edition is breaking that test. I'll investigate :)

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/321.

Otherwise this one here should be ready now @GuillaumeGomez @bilelmoussaoui

bilelmoussaoui (approved):

PR #320, opened 2021-10-31
glib: Add documentation to all undocumented `Object` / `ObjectExt` an…

sdroege:
…d related functions

Also change the `PropertyBindingBuilder` to return a `Result` when
building instead of an `Option`, change `stop_signal_emission` to
work with a signal id and add a new `stop_signal_emission_by_name` that
takes a string.

Also add a few more missing doc aliases.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/6

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1106,6 +1106,10 @@ glib_object_wrapper!(@object
 pub type ObjectClass = Class<Object>;
 
 impl Object {
+    /// Create a new instance of an object with the given properties.
+    ///
+    /// This fails if the object is not instantiable, doesn't have all the given properties or
+    /// property values of the wrong type are provided.
     #[allow(clippy::new_ret_no_self)]
     pub fn new<T: IsA<Object> + IsClass>(
         properties: &[(&str, &dyn ToValue)],
@@ -1115,6 +1119,10 @@ impl Object {
             .unwrap())
     }
 
+    /// Create a new instance of an object of the given type with the given properties.
+    ///
+    /// This fails if the object is not instantiable, doesn't have all the given properties or
+    /// property values of the wrong type are provided.
     pub fn with_type(
         type_: Type,
         properties: &[(&str, &dyn ToValue)],
@@ -1141,6 +1149,10 @@ impl Object {
         unsafe { Object::new_internal(type_, &params) }
     }
 
+    /// Create a new instance of an object of the given type with the given properties.
+    ///
+    /// This fails if the object is not instantiable, doesn't have all the given properties or
+    /// property values of the wrong type are provided.
     pub fn with_values(type_: Type, properties: &[(&str, Value)]) -> Result<Object, BoolError> {
         use std::ffi::CString;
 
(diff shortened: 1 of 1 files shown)


sdroege:
I hate writing docs, but apparently so does everybody else or these would've been documented already :)

glib/src/object.rs
```diff
@@ -1227,50 +1239,108 @@ pub trait ObjectExt: ObjectType {
...
+    /// If no type is provided then only the existence of the property is checked.
     fn has_property<'a, N: Into<&'a str>>(&self, property_name: N, type_: Option<Type>) -> bool;
+
+    /// Get the type of the property `property_name` of this object.
+    ///
+    /// This returns `None` if the property does not exist.
     #[doc(alias = "get_property_type")]
     fn property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type>;
+
+    /// Get the [`ParamSpec`](crate::ParamSpec) of the property `property_name` of this object.
```
GuillaumeGomez, line 1323: I don't think you need to pass the URL. Did you try:

```suggestion
    /// Get the [`ParamSpec`] of the property `property_name` of this object.
```

?

glib/src/object.rs
```diff
@@ -1227,50 +1239,108 @@ pub trait ObjectExt: ObjectType {
...
     fn find_property<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<crate::ParamSpec>;
+
+    /// Return all [`ParamSpec`](crate::ParamSpec) of the properties of this object.
```
GuillaumeGomez, line 1326: Same.

glib/src/object.rs
sdroege: Then it prints `crate::ParamSpec` in the docs, that's ugly.

glib/src/object.rs
```diff
@@ -2416,12 +2644,16 @@ impl ObjectClass {
...
+    /// Get the type of the property `property_name` of this object class.
+    ///
+    /// This returns `None` if the property does not exist.
     #[doc(alias = "get_property_type")]
     pub fn property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type> {
         self.find_property(property_name)
             .map(|pspec| pspec.value_type())
     }
 
+    /// Get the [`ParamSpec`](crate::ParamSpec) of the property `property_name` of this object class.
```
GuillaumeGomez, line 2656: Same.

glib/src/object.rs
```diff
@@ -2438,6 +2670,7 @@ impl ObjectClass {
         }
     }
 
+    /// Return all [`ParamSpec`](crate::ParamSpec) of the properties of this object class.
```
GuillaumeGomez, line 2673: Same.

glib/src/object.rs
```diff
@@ -3014,12 +3267,16 @@ impl<T: IsA<Object> + IsInterface> Interface<T> {
...
+    /// Get the type of the property `property_name` of this interface.
+    ///
+    /// This returns `None` if the property does not exist.
     #[doc(alias = "get_property_type")]
     pub fn property_type<'a, N: Into<&'a str>>(&self, property_name: N) -> Option<Type> {
         self.find_property(property_name)
             .map(|pspec| pspec.value_type())
     }
 
+    /// Get the [`ParamSpec`](crate::ParamSpec) of the property `property_name` of this interface.
```
GuillaumeGomez, line 3279: Same.

glib/src/object.rs
```diff
@@ -3036,6 +3293,7 @@ impl<T: IsA<Object> + IsInterface> Interface<T> {
         }
     }
 
+    /// Return all [`ParamSpec`](crate::ParamSpec) of the properties of this interface.
```
GuillaumeGomez, line 3296: Same.

glib/src/object.rs
GuillaumeGomez: Ah no ok, now I know why: you need to specify `crate::` to make the link work, which is why you created a link to improve the rendering. Makes sense!

GuillaumeGomez (approved):

glib/src/object.rs
```diff
@@ -1106,6 +1106,10 @@ glib_object_wrapper!(@object
 pub type ObjectClass = Class<Object>;
 
 impl Object {
+    /// Create a new instance of an object with the given properties.
+    ///
+    /// This fails if the object is not instantiable, doesn't have all the given properties or
+    /// property values of the wrong type are provided.
```
bilelmoussaoui, line 1112:
```suggestion
    /// Create a new instance of an object with the given properties.
    ///
    /// # Panics
    ///
    /// This fails if the object is not instantiable, doesn't have all the given properties or
    /// property values of the wrong type are provided.
```

I think something like this would look nicer & make it clear that it's the cause of a potential panic

bilelmoussaoui:
Otherwise looks good to me

bilelmoussaoui:
Maybe cc @Hofer-Julian

glib/src/object.rs
sdroege: Indeed, I missed that this one doesn't return a `Result` but just panics :)

sdroege: Hm no, this returns a `Result`. It won't panic by itself.

glib/src/object.rs
```diff
@@ -1115,6 +1119,10 @@ impl Object {
             .unwrap())
     }
 
+    /// Create a new instance of an object of the given type with the given properties.
+    ///
+    /// This fails if the object is not instantiable, doesn't have all the given properties or
+    /// property values of the wrong type are provided.
     pub fn with_type(
```
Hofer-Julian, line 1126: I wonder if there are situations where you want to use `with_type`  instead of `new` or if `new_type` should rather be private.

glib/src/object.rs
```diff
@@ -1141,6 +1149,10 @@ impl Object {
         unsafe { Object::new_internal(type_, &params) }
     }
 
+    /// Create a new instance of an object of the given type with the given properties.
+    ///
+    /// This fails if the object is not instantiable, doesn't have all the given properties or
+    /// property values of the wrong type are provided.
     pub fn with_values(type_: Type, properties: &[(&str, Value)]) -> Result<Object, BoolError> {
```
Hofer-Julian, line 1156: Same question for this function

glib/src/object.rs
```diff
@@ -1227,50 +1239,108 @@ pub trait ObjectExt: ObjectType {
...
+    /// Returns the interface `T` of the object.
+    ///
+    /// `None` is returned if the object does not implement the interface `T`.
     #[doc(alias = "get_interface")]
     fn interface<T: IsInterface>(&self) -> Option<InterfaceRef<T>>;
 
+    /// Sets the property `property_name` of the object to value `value`.
+    ///
+    /// This fails if the property does not exist, or if the type of the property is different than
+    /// the provided value, or if the property is not writable.
```
Hofer-Julian, line 1273:
```suggestion
    /// This fails if the property does not exist, the type of the property is different than
    /// the provided value, or if the property is not writable.
```

glib/src/object.rs
```diff
@@ -1227,50 +1239,108 @@ pub trait ObjectExt: ObjectType {
...
     fn set_property<'a, N: Into<&'a str>, V: ToValue>(
         &self,
         property_name: N,
         value: V,
     ) -> Result<(), BoolError>;
+
+    /// Sets the property `property_name` of the object to value `value`.
+    ///
+    /// This fails if the property does not exist, or if the type of the property is different than
+    /// the provided value, or if the property is not writable.
```
Hofer-Julian, line 1284:
```suggestion
    /// This fails if the property does not exist, the type of the property is different than
    /// the provided value, or if the property is not writable.
```

glib/src/object.rs
```diff
@@ -1227,50 +1239,108 @@ pub trait ObjectExt: ObjectType {
...
     fn set_property_from_value<'a, N: Into<&'a str>>(
         &self,
         property_name: N,
         value: &Value,
     ) -> Result<(), BoolError>;
+
+    /// Sets multiple properties of the object at once.
+    ///
+    /// This does not set any properties if one or more properties don't exist, or values of the
+    /// wrong type are provided, or if any of the properties is not writable.
```
Hofer-Julian, line 1294:
```suggestion
    /// This does not set any properties if one or more properties don't exist, values of the
    /// wrong type are provided, or if any of the properties is not writable.
```

glib/src/object.rs
```diff
@@ -1227,50 +1239,108 @@ pub trait ObjectExt: ObjectType {
...
+    #[doc(alias = "g_object_set")]
     fn set_properties(&self, property_values: &[(&str, &dyn ToValue)]) -> Result<(), BoolError>;
+
+    /// Sets multiple properties of the object at once.
+    ///
+    /// This does not set any properties if one or more properties don't exist, or values of the
+    /// wrong type are provided, or if any of the properties is not writable.
```
Hofer-Julian, line 1301:
```suggestion
    /// This does not set any properties if one or more properties don't exist, values of the
    /// wrong type are provided, or if any of the properties is not writable.
```

glib/src/object.rs
```diff
@@ -1376,64 +1528,127 @@ pub trait ObjectExt: ObjectType {
...
         f: F,
     ) -> SignalHandlerId;
+
+    /// Connect to the `notify` signal of the object.
+    ///
+    /// This is emitted whenever a property is changed. If `name` is provided then the signal
+    /// handler is only called for this specific property.
+    ///
+    /// This is like `connect_notify` but doesn't require a `Send+Sync` or `'static` closure. No
+    /// runtime checks for wrongly calling the closure are performed.
```
Hofer-Julian, line 1616: Maybe there should be a `#Safety` section in the end

Hofer-Julian (commented):
This looks really good apart from a few minor things. :)
Thanks a lot @sdroege!

glib/src/object.rs
sdroege: Yes, if all you have is the `GType`. I need this in some places :)

sdroege: Think of factories, for example.

sdroege:
@Hofer-Julian Thanks for the review, I'll update it accordingly tomorrow :)

PR #321, opened 2021-11-01
glib: Fix bug in main context channel test that caused it to deadlock

sdroege:
With edition 2021 the closures are only borrowing what they have to, so
here in this case it didn't actually borrow the helper as it was not
actually used inside the closure. With that, it was not dropped when the
main context channel was dropped and the main loop was never quit.


bilelmoussaoui (approved):

sdroege:
Reported this also here: https://github.com/rust-lang/rust/issues/90465 . Just in case this was not an intentional change.

PR #323, opened 2021-11-01
ObjectExt helpers

### LINKED ISSUE #307: Add property getter and setter that panic when an error occurs
At the moment, when you want to get a custom property you have to accept a lot of boilerplate:

```rust
let current_state = switch
        .property("state")
        .expect("The property needs to exist and be readable.")
        .get::<bool>()
        .expect("The property needs to be of type `bool`.");
```

Typically, these properties are created within your program without any decision tree involved.
This means that you always want to panic if either `property` or `get` returns `Err`.

To accommodate this use case, I suggest to change 
`fn property<'a, N: Into<&'a str>>(&self, property_name: N) -> Result<Value, BoolError>`
to
`fn property<'a, N: Into<&'a str>, T: FromValue>(&self, property_name: N) -> T`

This would simplify the example mentioned before to this:

```rust
let current_state = switch.property::<bool>("state");
```

The old behavior could then be exposed with a method called `property_value`.

@sdroege

bilelmoussaoui:
Fixes #307


glib/src/object.rs
```diff
@@ -1838,7 +1870,23 @@ impl<T: ObjectType> ObjectExt for T {
         Ok(())
     }
 
-    fn property<'a, N: Into<&'a str>>(&self, property_name: N) -> Result<Value, BoolError> {
+    fn try_property<'a, N: Into<&'a str>, V: FromValue<'static>>(
+        &self,
+        property_name: N,
+    ) -> Result<V, BoolError> {
+        let prop = self.try_property_value(property_name)?;
+        prop.get::<V>()
```
bilelmoussaoui, line 1878: This doesn't work for lifetime reasons 
```

error[E0597]: `prop` does not live long enough
    --> glib/src/object.rs:1878:9
     |
1878 |         prop.get::<V>()
     |         ^^^^-----------
     |         |
     |         borrowed value does not live long enough
     |         argument requires that `prop` is borrowed for `'static`
1879 |             .map_err(|e| crate::bool_error!("Failed to get cast value to a different type {}", e))
```
tried everything i know but failed everytime

sdroege:
> glib: add panicing helpers to ObjectExt

It's written `panicking` with `k` btw (English :tada: )

glib/src/object.rs
sdroege: As discussed on matrix, we should add a `FromValueOwned` trait that is used here and everywhere else where currently a `FromValue<'static>`  is used. Something like

```rust
pub trait FromValueOwned: for<'a> FromValue<'a> {
    unsafe fn from_value_owned(v: &Value) -> Self;
}
```

sdroege: And a `Value::get_owned()` that makes use of this trait

bilelmoussaoui: > As discussed on matrix, we should add a `FromValueOwned` trait that is used here and everywhere else where currently a `FromValue<'static>`  is used. Something like
> 
> ```rust
> pub trait FromValueOwned: for<'a> FromValue<'a> {
>     unsafe fn from_value_owned(v: &Value) -> Self;
> }
> ```


Should I go ahead and add it as part of the same PR or do it in a different one ?

sdroege: A different one would be best, and then this one can be rebased on top of that once merged.

If you can write a `Value::get_owned()` function then this PR will also be implementable.

glib/src/object.rs
```diff
@@ -1272,42 +1272,77 @@ pub trait ObjectExt: ObjectType {
...
+    /// Gets the property `property_name` of the object and cast it to the type V.
+    ///
+    /// Fails if the property doesn't exist or is not readable or of a different type than V.
+    fn try_property<'a, N: Into<&'a str>, V: FromValue<'static> + 'static>(
+        &self,
+        property_name: N,
+    ) -> Result<V, BoolError>;
+
+    /// Similar to `try_property` but panics instead of failing.
+    fn property<'a, N: Into<&'a str>, V: FromValue<'static> + 'static>(
```
sdroege, line 1329: This and the others can be updated to the `for<'b> FromValue<'b>` stuff now, and make use of `Value::get_owned()`

glib/src/object.rs
```diff
@@ -1524,72 +1604,124 @@ pub trait ObjectExt: ObjectType {
...
     /// If the signal has a return value then this is returned here.
-    fn emit_with_details(
+    fn try_emit_with_details(
         &self,
         signal_id: SignalId,
         details: Quark,
         args: &[&dyn ToValue],
     ) -> Result<Option<Value>, BoolError>;
 
+    /// Similar to `try_emit_with_details` but panics instead of failing.
```
sdroege, line 1696: Maybe it would be good to have the long docs on the function we expect people to use

sdroege: And make the other ones properly link to the documented function. Needs to be surrounded by `[]` I guess

bilelmoussaoui:
So as expected, it does fail on glib tests. No idea why

glib/src/object.rs
bilelmoussaoui: yeah i will look into improving the docs slightly a bit once we get green CI

bilelmoussaoui: Should be good enough now

bilelmoussaoui:
I will look into updating gtk4-rs today

sdroege:
I'll try updating gstreamer-rs/gst-plugins-rs over the weekend too, but there a few hundred calls to the affected functions that have to be changed. It will take a while :)

bilelmoussaoui:
Note that we need to update gir to fix the `emit_by_name` use case. While we are at it; we could maybe make it use `property`/`set_property`/`emit_by_name` from ObjectExt directly instead of going through FFI.

Code such as 
```rust

    fn n_axes(&self) -> u32 {
        unsafe {
            let mut value = glib::Value::from_type(<u32 as StaticType>::static_type());
            glib::gobject_ffi::g_object_get_property(
                self.to_glib_none().0 as *mut glib::gobject_ffi::GObject,
                b"n-axes\0".as_ptr() as *const _,
                value.to_glib_none_mut().0,
            );
            value
                .get()
                .expect("Return Value for property `n-axes` getter")
        }
    }

    fn set_seat<P: IsA<Seat>>(&self, seat: Option<&P>) {
        unsafe {
            glib::gobject_ffi::g_object_set_property(
                self.to_glib_none().0 as *mut glib::gobject_ffi::GObject,
                b"seat\0".as_ptr() as *const _,
                seat.to_value().to_glib_none().0,
            );
        }
    }
```

```rust
pub fn emit_change_current_page(&self, object: i32) -> bool {
        let stash: Stash<*mut ffi::GtkShortcutsSection, _> = self.to_glib_none();
        let gobject =
            unsafe { glib::Object::from_glib_borrow(stash.0 as *mut glib::gobject_ffi::GObject) };
        let res = gobject
            .emit_by_name("change-current-page", &[&object])
            .unwrap();
        res.unwrap()
            .get()
            .expect("Return Value for `emit_change_current_page`")
    }
```
could become


```rust
    fn n_axes(&self) -> u32 {
        self.property("n-axes")
    }

    fn set_seat<P: IsA<Seat>>(&self, seat: Option<&P>) {
        self.set_property("seat", &seat);
    }
```

```rust
pub fn emit_change_current_page(&self, object: i32) -> bool {
        let value = self
            .emit_by_name("change-current-page", &[&object])
            .unwrap();
        value.get()
            .expect("Return Value for `emit_change_current_page`")
    }
```

sdroege:
If you want to update gir accordingly, go ahead :)

bilelmoussaoui:
Sure, I will give it a try later this weekend

PR #325, opened 2021-11-03
glib: add a Value::get_owned

bilelmoussaoui:
The current FromValue trait can't handle the FromValue<'static>
case as that means having to borrow self for 'static.

Needed for having a generic ObjectExt::property that would return T instead of Value. See #323

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -622,10 +622,6 @@ impl<T: Send + ToValue + ?Sized> ToSendValue for T {
     }
 }
 
-impl ValueType for &'static str {
-    type Type = Self;
-}
-
 unsafe impl<'a> FromValue<'a> for &'a str {
     type Checker = GenericValueTypeOrNoneChecker<Self>;
 
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -52,7 +52,7 @@ use crate::translate::*;
 use crate::types::{StaticType, Type};
 
 /// A type that can be stored in `Value`s.
-pub trait ValueType: ToValue + FromValue<'static> + 'static {
+pub trait ValueType: ToValue + for<'a> FromValue<'a> + 'static {
     /// Type to get the `Type` from.
     ///
     /// This exists only for handling optional types.
@@ -71,7 +71,7 @@ pub trait ValueTypeOptional:
 
 impl<T, C> ValueType for Option<T>
 where
-    T: FromValue<'static, Checker = C> + ValueTypeOptional + StaticType + 'static,
+    T: for<'a> FromValue<'a, Checker = C> + ValueTypeOptional + StaticType + 'static,
     C: ValueTypeChecker<Error = ValueTypeMismatchOrNoneError>,
 {
     type Type = T::Type;
@@ -392,6 +392,17 @@ impl Value {
         unsafe {
             T::Checker::check(self)?;
             Ok(T::from_value(self))
+        }
+    }
+
+    /// Tries to get a value of an owned type `T`.
+    pub fn get_owned<T>(&self) -> Result<T, <<T as FromValue>::Checker as ValueTypeChecker>::Error>
+    where
+        T: for<'b> FromValue<'b> + 'static,
+    {
+        unsafe {
+            T::Checker::check(self)?;
+            Ok(FromValue::from_value(self))
         }
     }
 
@@ -964,6 +975,11 @@ mod tests {
 
         let some_v = Some("test").to_value();
         assert_eq!(some_v.get::<&str>(), Ok("test"));
+        assert_eq!(some_v.get_owned::<String>(), Ok("test".to_string()));
+        assert_eq!(
+            some_v.get_owned::<Option<String>>(),
+            Ok(Some("test".to_string()))
+        );
         assert_eq!(some_v.get::<Option<&str>>(), Ok(Some("test")));
         assert_eq!(
             some_v.get::<i32>(),


glib/src/value.rs
```diff
@@ -205,6 +205,18 @@ pub unsafe trait FromValue<'a>: Sized {
     unsafe fn from_value(value: &'a Value) -> Self;
 }
 
+/// Similar to `FromValue` but works for owned `Value.
```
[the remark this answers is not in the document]
bilelmoussaoui: it's hard to do because the `ToValue` trait is not implemented for `&'static str`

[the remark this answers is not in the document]
sdroege: Why not? It really should, that's how `obj.set_property("foo", "bar")` works :)

glib/src/value.rs
```diff
@@ -205,6 +205,18 @@ pub unsafe trait FromValue<'a>: Sized {
...
+pub unsafe trait FromOwnedValue: Sized {
```
sdroege, line 209:
```suggestion
pub unsafe trait FromOwnedValue: for<'a> FromValue<'a> {
```

Then you can remove the `Checker` probably

glib/src/value.rs
```diff
@@ -205,6 +205,18 @@ pub unsafe trait FromValue<'a>: Sized {
...
+    /// Value type checker.
+    type Checker: ValueTypeChecker;
+
+    /// Get the contained value from a `Value`
+    ///
+    /// # Safety
+    /// `Self::Checker::check()` must be called first and must not fail.
+    unsafe fn from_value(value: &Value) -> Self;
```
sdroege, line 217:
```suggestion
    unsafe fn from_value_owned(value: &Value) -> Self;
```

glib/src/value.rs
```diff
@@ -205,6 +205,18 @@ pub unsafe trait FromValue<'a>: Sized {
     unsafe fn from_value(value: &'a Value) -> Self;
```
sdroege, line 205: Search for `FromValue<'static>` in this file. These should all become `FromValueOwned` AFAICS

glib/src/value.rs
```diff
@@ -205,6 +205,18 @@ pub unsafe trait FromValue<'a>: Sized {
...
+    /// Value type checker.
+    type Checker: ValueTypeChecker;
+
+    /// Get the contained value from a `Value`
+    ///
+    /// # Safety
+    /// `Self::Checker::check()` must be called first and must not fail.
+    unsafe fn from_value(value: &Value) -> Self;
+}
+
```
sdroege, line 219: You'll actually have to implement the new trait for `String`, `i32`, etc. (and also put it into the `glib::wrapper!` macros for boxed/shared/boxed-inline/object)

glib/src/value.rs
bilelmoussaoui: All of them? the one in ValueType trait definition can't be made as FromValueOwned, it breaks a ton of code

sdroege: All two of them, and especially that one was the one I was thinking of. How does it break?

glib/src/value.rs
```diff
@@ -205,6 +205,21 @@ pub unsafe trait FromValue<'a>: Sized {
...
+/// let x = "gtk-rs".to_value();
+///
+/// let y = x.get_owned::<&str>();
+/// ```
+pub unsafe trait FromOwnedValue: for<'a> FromValue<'a> {
+    /// Get the contained value from a `Value`
+    ///
+    /// # Safety
+    /// `Self::Checker::check()` must be called first and must not fail.
+    unsafe fn from_value_owned(value: &Value) -> Self;
```
sdroege, line 220:
```suggestion
    unsafe fn from_value_owned(value: &Value) -> Self {
        <Self as FromValue>::from_value(v)
    }
```

Does that work? Then you don't need to implement any functions anywhere

glib/src/value.rs
```diff
@@ -964,6 +1038,7 @@ mod tests {
 
         let some_v = Some("test").to_value();
         assert_eq!(some_v.get::<&str>(), Ok("test"));
+        assert_eq!(some_v.get_owned::<String>(), Ok("test".to_string()));
```
sdroege, line 1041: Can you also add a test for `Option<String>`?

glib/src/value.rs
```diff
@@ -783,6 +851,12 @@ macro_rules! numeric {
             type Checker = GenericValueTypeChecker<Self>;
 
             unsafe fn from_value(value: &'a Value) -> Self {
+                FromOwnedValue::from_value_owned(value)
+            }
+        }
+
+        unsafe impl FromOwnedValue for $name {
```
sdroege, line 858: Please also shortly grep in the sources to see if there are any manual `FromValue` impls left that should get this one too

glib/src/value.rs
```diff
@@ -205,6 +205,21 @@ pub unsafe trait FromValue<'a>: Sized {
...
 }
 
+/// Similar to `FromValue` but works for owned `Value.
+///
+/// ```compile_fail
+/// let x = "gtk-rs".to_value();
+///
+/// let y = x.get_owned::<&str>();
+/// ```
+pub unsafe trait FromOwnedValue: for<'a> FromValue<'a> {
```
sdroege, line 215:
```suggestion
pub unsafe trait FromOwnedValue: for<'a> FromValue<'a> + 'static {
```

glib/src/value.rs
```diff
@@ -228,6 +243,15 @@ mod private {
     }
 }
 
+pub trait FromOwnedValueOptional: for<'a> FromValueOptional<'a> {}
```
sdroege, line 246:
```suggestion
pub trait FromOwnedValueOptional: for<'a> FromValueOptional<'a> + 'static {}
```

glib/src/value.rs
bilelmoussaoui: Indeed, thanks!

glib/src/value.rs
bilelmoussaoui: the only one left are Value & BoxedValue. Should they also get a FromOwnedValue impl?

glib/src/value.rs
bilelmoussaoui: There are different cases:
1 - We need to generate the new trait as part of gir & do a regen (for flags/enums)
2 - we can no longer impl ValueType for `&'static str`
3 - BoxedValue 

Otherwise I fixed the remaining ones

bilelmoussaoui: For 1- https://github.com/gtk-rs/gir/pull/1254

glib/src/value.rs
sdroege: Yes

glib/src/value.rs
sdroege: 2 - seems expected and correct
3 - what's the problem there?

bilelmoussaoui: 3 - None, i did implement FromOwnedValue for it as well
2 - so, we drop the ValueType impl of it?

bilelmoussaoui:
This one is blocked by https://github.com/gtk-rs/gir/pull/1254 now

glib/src/value.rs
sdroege: 2 - Yes, it makes no sense. You can't get a `&'static str` from a `Value`

glib/src/value.rs
```diff
@@ -205,6 +205,23 @@ pub unsafe trait FromValue<'a>: Sized {
     unsafe fn from_value(value: &'a Value) -> Self;
 }
 
+/// Similar to `FromValue` but works for owned `Value.
+///
+/// ```compile_fail
+/// let x = "gtk-rs".to_value();
```
GuillaumeGomez, line 211: Maybe add some comments like "This example cannot compile because `&str` is a temporary value" or something along the line. And please also add a working code example (with `String` for example). So both are always checked.

GuillaumeGomez (approved):

glib/src/value.rs
```diff
@@ -205,6 +205,29 @@ pub unsafe trait FromValue<'a>: Sized {
...
+///
+/// let y = x.get_owned::<&str>();
+/// ```
+/// 
+/// ```
+/// let owned_str = "hello".to_string().to_value();
+/// assert_eq!(owned_str.owned_get::<String>(), "hello".to_string());
+/// ```
+pub unsafe trait FromOwnedValue: for<'a> FromValue<'a> + 'static {
```
sdroege, line 221: I wonder if we can make a blanket implementation for this instead of implementing it for every type. Does something like

```rust
unsafe impl<'a, T: FromValue<'a> + 'static> FromOwnedValue for T {}
```

work?

bilelmoussaoui: That can't work because we don't have a FromValue impl for String. If we manually implement FromOwnedValue for String only, it complains that it's already implemented as part of the blanket implementation :/

glib/src/object.rs
```diff
@@ -45,7 +45,7 @@ pub unsafe trait ObjectType:
     + crate::value::ValueType
     + crate::value::ToValue
     + crate::value::ToValueOptional
-    + crate::value::FromValueOptional<'static>
+    + crate::value::FromOwnedValueOptional
```
sdroege, line 48:
```suggestion
    + crate::value::FromOwnedValueOptional
    + for<'a> crate::value::FromValueOptional<'a>
```

too maybe?

glib/src/value.rs
```diff
@@ -395,6 +432,17 @@ impl Value {
         }
     }
 
+    /// Tries to get a value of an owned type `T`.
+    pub fn get_owned<T>(&self) -> Result<T, <<T as FromValue>::Checker as ValueTypeChecker>::Error>
+    where
+        T: FromOwnedValue,
```
sdroege, line 438:
```suggestion
        T: for<'b> FromValue<'b> + 'static,
```

Does this maybe also work?

glib/src/param_spec.rs
```diff
@@ -43,7 +43,7 @@ impl crate::value::ValueType for ParamSpec {
 unsafe impl<'a> crate::value::FromValue<'a> for ParamSpec {
     type Checker = crate::value::GenericValueTypeOrNoneChecker<Self>;
 
-    unsafe fn from_value(value: &'a crate::Value) -> Self {
+    unsafe fn from_value(value: &crate::Value) -> Self {
```
sdroege, line 46: Why this change, and why does this even compile? :)

bilelmoussaoui: Heh, not an intended change.

glib/src/value.rs
sdroege: If that works we won't need a new trait, if it doesn't then that's just how it is. With the blanket implementation it should be good enough.

glib/src/value.rs
```diff
@@ -52,7 +52,7 @@ use crate::translate::*;
 use crate::types::{StaticType, Type};
 
 /// A type that can be stored in `Value`s.
-pub trait ValueType: ToValue + FromValue<'static> + 'static {
+pub trait ValueType: ToValue + FromValue<'static> {
```
sdroege, line 55:
```suggestion
pub trait ValueType: ToValue + for<'a> FromValue<'a> + 'static {
```

sdroege: Also the other `FromValue<'static>` cases in this file

sdroege (approved):

PR #327, opened 2021-11-03
None constants as associated constants

fengalin:
See https://github.com/gtk-rs/gir/issues/1252.

- [x] Depends on https://github.com/gtk-rs/gir/pull/1253
- [x] Remove `.gitmodules` changes
- [x] Update gir.
- [x] Regen.


fengalin:
CI failure seems to be caused be a network error. I don't think I can restart without force updating.

sdroege (dismissed):

PR #330, opened 2021-11-04
Add missing Matrix methods

### LINKED ISSUE #329: [FEATURE REQUEST] need a way to construct a pango::Matrix
I want to construct my own `pango::Matrix`, but there's no constructor.  I'm using this:

```rust
        let pango_matrix = pango::ffi::PangoMatrix {
            // my own values here
            xx,
            xy,
            yx,
            yy,
            x0,
            y0,
        };

        let pango_matrix_ptr: *const pango::ffi::PangoMatrix = &pango_matrix;

        let matrix = unsafe { pango::Matrix::from_glib_none(pango_matrix_ptr) };

        my_pango_context.set_matrix(Some(&matrix));
```

@sdroege suggested that `pango::Matrix` should be a `BoxedInline` instead of a `Boxed`, too.

GuillaumeGomez:
Fixes #329.

### DIFF
--- a/pango/src/matrix.rs
+++ b/pango/src/matrix.rs
@@ -1,0 +1,72 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{Matrix, Rectangle};
+use glib::translate::*;
+use std::fmt;
+
+impl Matrix {
+    pub fn new(xx: f64, xy: f64, yx: f64, yy: f64, x0: f64, y0: f64) -> Self {
+        Self(ffi::PangoMatrix {
+            xx,
+            xy,
+            yx,
+            yy,
+            x0,
+            y0,
+        })
+    }
+
+    #[doc(alias = "pango_matrix_transform_pixel_rectangle")]
+    pub fn transform_pixel_rectangle(&self, rect: &mut Rectangle) {
+        unsafe {
+            ffi::pango_matrix_transform_pixel_rectangle(
+                self.to_glib_none().0,
+                rect.to_glib_none_mut().0,
+            )
+        }
+    }
+
+    #[doc(alias = "pango_matrix_transform_rectangle")]
+    pub fn transform_rectangle(&self, rect: &mut Rectangle) {
+        unsafe {
+            ffi::pango_matrix_transform_rectangle(self.to_glib_none().0, rect.to_glib_none_mut().0)
+        }
+    }
+
+    pub fn xx(&self) -> f64 {
+        self.0.xx
+    }
+
+    pub fn xy(&self) -> f64 {
+        self.0.xy
+    }
+
+    pub fn yx(&self) -> f64 {
+        self.0.yx
+    }
+
+    pub fn yy(&self) -> f64 {
+        self.0.yy
+    }
+
+    pub fn x0(&self) -> f64 {
+        self.0.x0
+    }
+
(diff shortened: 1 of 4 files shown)


pango/src/matrix.rs
```diff
@@ -0,0 +1,34 @@
...
+            )
+        }
+    }
+
+    #[doc(alias = "pango_matrix_transform_rectangle")]
+    pub fn transform_rectangle(&self, mut rect: Option<&mut Rectangle>) {
+        unsafe {
+            ffi::pango_matrix_transform_rectangle(self.to_glib_none().0, rect.to_glib_none_mut().0)
+        }
+    }
```
bilelmoussaoui, line 33: Could implement Debug

pango/src/matrix.rs
```diff
@@ -0,0 +1,34 @@
...
+            xy,
+            yx,
+            yy,
+            x0,
+            y0,
+        })
+    }
+
+    #[doc(alias = "pango_matrix_transform_pixel_rectangle")]
+    pub fn transform_pixel_rectangle(&self, mut rect: Option<&mut Rectangle>) {
```
bilelmoussaoui, line 19: Why the rect needs to be mut?

pango/src/matrix.rs
```diff
@@ -0,0 +1,34 @@
...
+        unsafe {
+            ffi::pango_matrix_transform_pixel_rectangle(
+                self.to_glib_none().0,
+                rect.to_glib_none_mut().0,
+            )
+        }
+    }
+
+    #[doc(alias = "pango_matrix_transform_rectangle")]
+    pub fn transform_rectangle(&self, mut rect: Option<&mut Rectangle>) {
```
sdroege, line 29: Put this and the other new function into `manual = true` in the `Gir.toml` so that the commented out autogenerated code disappears and docs are generated.

pango/src/matrix.rs
sdroege: Because it's modified by the function

sdroege: Also why `Option`? While technically the function allows that, it has absolutely no effect if no rectangle is passed in :) That's both input and output of the function. Same for the other one.

GuillaumeGomez: I thought that the function was doing something even with a NULL rectangle. Didn't check the source code though...

pango/src/matrix.rs
GuillaumeGomez: I suppose the `Matrix` type already implements it. Or are you talking about something else?

GuillaumeGomez:
Updated.

pango/src/matrix.rs
sdroege: It doesn't and you have to implement it :)

GuillaumeGomez: Ah. Damn.

GuillaumeGomez:
Added the `Debug` impl and methods to access the fields.

sdroege (dismissed):

sdroege:
@GuillaumeGomez You lost the changes to the autogenerated code somehow, now the regen check fails.

GuillaumeGomez:
Remember kids: `git add` works well with folders, not just files! :D

sdroege (approved):

Issue #301, opened 2021-10-25
Rename `glib::GString` to `glib::CString` to avoid confusion with the C `GString` (aka `glib::String`)


sdroege:
And potentially also rename `glib::String` to `glib::StringBuilder` (which Vala did, for example), but OTOH the Rust "string builder" type is also just called `String`.

@bilelmoussaoui @GuillaumeGomez Opinions?

pbor:
For what is worth I found `GString` super confusing at the beginning. I like the plan proposed in the last comments!

sdroege:
> For what is worth I found `GString` super confusing at the beginning. I like the plan proposed in the last comments!

You mean removing `glib::String` (the C `GString` string builder) and keeping `glib::GString` (the string with the GLib allocator)?

pbor:
Yes, and renaming that to `glib::String`

sdroege:
First part in https://github.com/gtk-rs/gtk-rs-core/pull/332 . I'll do the renaming once this is in, it requires gir changes and manual changes all over the place.

sdroege:
@GuillaumeGomez for renaming, you were very much against CString, but String would be OK for you?

GuillaumeGomez:
I'd still prefer `GString` but don't want to drag the debate forever.

sdroege:
IMHO `GString` is better than `String` (our thing is a `Box<str>` but `String` is a string builder!), but `CString` would be best (our thing is an owned, NUL-terminated C/FFI string just like `CString` but in addition also guaranteed UTF-8).

I'm a bit hesitant to rename it to `String` because of that :)

sdroege:
Let's keep it like this then. No more `glib::String` and keeping `glib::GString`

PR #332, opened 2021-11-05
glib: Remove `glib::String` bindings

sdroege:
It's not supposed to be used by any C APIs available to bindings, and
Rust's `std::String` fulfills the same purpose more tightly integrated
into the language.

See https://github.com/gtk-rs/gtk-rs-core/issues/301


gio/src/auto/dbus_interface_info.rs
```diff
@@ -19,16 +19,10 @@ glib::wrapper! {
...
-    pub fn generate_xml(&self, indent: u32, string_builder: &mut glib::String) {
-        unsafe {
-            ffi::g_dbus_interface_info_generate_xml(
-                self.to_glib_none().0,
-                indent,
-                string_builder.to_glib_none_mut().0,
-            );
-        }
-    }
+    //#[doc(alias = "g_dbus_interface_info_generate_xml")]
```
bilelmoussaoui, line 22: would be nice to generate those manually and make them return a String :)

sdroege: If someone needs them we can add them back. I don't understand why anybody is using the `GDbus` bindings as they are (without some high-level API that makes it convenient to use) anyway, but this is even more fringe API than the other parts.

sdroege: Or said otherwise: I'm not going to spend more time on `GDbus` and I still think it was a mistake to add at all without something more useful around it. It's enough of a burden to keep it compile with all the other changes going on :)

sdroege: FWIW, outside GLib these two functions are used by no code inside Debian. The only other references are from autogenerated bindings for various languages, where there's no code in Debian making use of that bindings API either.

PR #334, opened 2021-11-05
variant: Add support for `BTreeMap`

cgwalters:
In ostree, we heavily use `a{sv}` and further we compute
cryptographic checksums over them.  It is hence important for
reproducibility that we insert values in a defined (sorted)
order.

Expressing this by serializing a `BTreeMap` in the same way we
support `HashMap` fixes this.


cgwalters:
xref https://github.com/ostreedev/ostree-rs-ext/pull/149#discussion_r743877438

sdroege (approved):

PR #336, opened 2021-11-06
Fix #335 - repr(transparent) for cairo_bool_t; correct opaque types

### LINKED ISSUE #335: [BUG] cairo-rs-sys: Several repr(C) types need to be repr(transparent)
This simple test fails on s/390x; the assertion triggers:

```rust
#[test]
fn initial_cairo_context_has_no_current_point() {
    let surf = cairo::ImageSurface::create(cairo::Format::ARgb32, 100, 100).unwrap();
    let cr = cairo::Context::new(&surf).unwrap();
    assert!(!cr.has_current_point().unwrap());
}
```

The root cause is this, from cairo-sys-rs:

```rust
#[repr(C)]
#[derive(Debug, Clone, Copy)]
pub struct cairo_bool_t {
    value: c_int,
}
extern "C" {
    pub fn cairo_has_current_point(cr: *mut cairo_t) -> cairo_bool_t;
}
```

In the C code, there is `typedef int cairo_bool_t;`.

Changing the test to this - a duplicated version of `cairo_bool_t` from Rust and calling `cairo_has_current_point` by hand - shows the problem clearly.  I'm using my own `my_cairo_bool_t` because the `value` field is private in cairo-sys-rs, and I want to access it.

```rust
 #[repr(C)]
 #[derive(Debug, Clone, Copy)]
 struct my_cairo_bool_t {
     value: libc::c_int,
 }
#[test]
fn cairo_return_types() {
    let output = cairo::ImageSurface::create(cairo::Format::ARgb32, 100, 100).unwrap();
    let cr = cairo::Context::new(&output).unwrap();

    let cr_raw = cr.to_raw_none();
    let res = unsafe { cairo::ffi::cairo_has_current_point(cr_raw) };
    let res_ptr = &res;
    let converted: &my_cairo_bool_t = unsafe { std::mem::transmute(res_ptr) };

    assert_eq!(converted.value, 0);
}
```

Failure:
```
[  924s] ---- api::cairo_return_types stdout ----
[  924s] thread 'main' panicked at 'assertion failed: `(left == right)`
[  924s]   left: `-1667192928`,
[  924s]  right: `0`', tests/src/api.rs:241:5
[  924s] stack backtrace:
[  924s]   11:      0x2aa077c7aec - core::panicking::assert_failed::h9cd50c53c9f2529d
[  924s]                                at /home/abuild/rpmbuild/BUILD/rustc-1.54.0-src/library/core/src/panicking.rs:117:5
[  924s]   12:      0x2aa0713c7dc - src::api::cairo_return_types::h686e39ee8bfebfac
[  924s]                                at /home/abuild/rpmbuild/BUILD/librsvg-2.52.3/tests/src/api.rs:241:5
```

Summary: the `cairo_bool_t` struct should be `repr(transparent)`, not `repr(C)`.  I think this was not caught before because has_current_point is probably the first bool-returning function that librsvg calls from Cairo.

I'll submit a PR for this and other similar types.  For example, I think `#[repr(C)] pub struct cairo_t;` works by chance; weren't opaque structs supposed to be represented with something like

```rust
#[repr(C)]
pub struct cairo_t {
    _private: [u8; 0]
}
```

I haven't checked the Unsafe Code Guidelines in a while :)

(I'm filing this for posterity instead of just submitting a PR.)

federicomenaquintero:
Fixes #123.

The important fix is that `cairo_bool_t` needs to be `repr(transparent)` - it doesn't work with `repr(C)` on s/390x otherwise.

Then, declare opaque types like `cairo_t` with the Rustonomicon's trick.

### DIFF
--- a/cairo/sys/src/lib.rs
+++ b/cairo/sys/src/lib.rs
@@ -76,13 +76,16 @@ pub type cairo_ps_level_t = c_int;
 pub type cairo_mesh_corner_t = c_uint;
 
 macro_rules! opaque {
-    ($name:ident) => {
+    ($(#[$attr:meta])*
+     $name:ident) => {
         // https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs
+        $(#[$attr])*
         #[repr(C)]
         pub struct $name {
             _data: [u8; 0],
             _marker: core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
         }
+        $(#[$attr])*
         impl ::std::fmt::Debug for $name {
             fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
                 write!(f, "{} @ {:?}", stringify!($name), self as *const _)
@@ -96,11 +99,10 @@ opaque!(cairo_surface_t);
 opaque!(cairo_device_t);
 opaque!(cairo_pattern_t);
 
-#[cfg(any(feature = "xcb", feature = "dox"))]
-#[repr(C)]
-pub struct xcb_connection_t(c_void);
-#[cfg(any(feature = "xcb", feature = "dox"))]
-debug_impl!(xcb_connection_t);
+opaque!(
+    #[cfg(any(feature = "xcb", feature = "dox"))]
+    xcb_connection_t
+);
 
 #[cfg(any(feature = "xcb", feature = "dox"))]
 pub type xcb_drawable_t = u32;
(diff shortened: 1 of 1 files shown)


sdroege (commented):
Looks good to me but please put a `Fixes #123` into the commit message too, and run `cargo fmt` once to make the CI happy :)

GuillaumeGomez:
I added the `Fixes #123` part in his message.

@federicomenaquintero Please run `cargo fmt`.

sdroege:
> I added the `Fixes #123` part in his message.

Why didn't you just run `cargo fmt` while at that? :) I did that now.

GuillaumeGomez:
I simply edited the first message and did nothing locally on my computer?

GuillaumeGomez (approved):

sdroege:
Oh that's possible from the web UI? TIL :)

federicomenaquintero:
Oops, thanks for fixing it up for me :)

cairo/sys/src/lib.rs
```diff
@@ -85,50 +94,36 @@ macro_rules! debug_impl {
...
-pub struct cairo_pattern_t(c_void);
-
-#[cfg(any(feature = "xcb", feature = "dox"))]
-#[repr(C)]
-pub struct xcb_connection_t(c_void);
-#[cfg(any(feature = "xcb", feature = "dox"))]
-debug_impl!(xcb_connection_t);
+opaque!(
+    #[cfg(any(feature = "xcb", feature = "dox"))]
+    xcb_connection_t
```
GuillaumeGomez, line 104: Ah, we forgot to add `#[cfg_attr(feature = "dox", doc(cfg(feature = "xcb")))]`...

cairo/sys/src/lib.rs
```diff
@@ -85,50 +94,36 @@ macro_rules! debug_impl {
...
 #[cfg(any(feature = "xcb", feature = "dox"))]
 pub type xcb_pixmap_t = u32;
 
-#[cfg(any(feature = "xcb", feature = "dox"))]
-#[repr(C)]
-pub struct xcb_visualtype_t(c_void);
-#[cfg(any(feature = "xcb", feature = "dox"))]
-debug_impl!(xcb_visualtype_t);
+opaque!(
+    #[cfg(any(feature = "xcb", feature = "dox"))]
```
GuillaumeGomez, line 114: Same here and below.

Issue #23, opened 2020-11-09
Add bindings for g_signal_connect_object()


haecker-felix:
More details: https://valadoc.org/gobject-2.0/GLib.Signal.connect_object.html

jf2048:
If #338 is good enough then possibly this could be closed. It should be usable for the same purpose. `g_signal_connect_object` can't really be bound as-is because there is nowhere to pass in the trampoline.

sdroege:
Yeah I think this can be closed.

Issue #337, opened 2021-11-08
New variant of ObjectExt::emit_with_name


bilelmoussaoui:
Currently, you can emit a signal and it might returns a Option<Value>. Similar to the new property getter; we would like to have something that automatically returns `Option<T> where T:  for <'b> FromValue<'b> + 'static`

sdroege:
Or even better something that returns `()` for `None` (and panics otherwise), and `T` for other cases

GuillaumeGomez:
I'm not sure to understand why we would want to return `T` directly. Why would we want to remove the `Option` wrapping?

PR #338, opened 2021-11-09
Add gclosure macro

### LINKED ISSUE #19: Provide proc macro attribute to generate generic signal handler functions
This would generate a function that works on `GValue`s and automatically converts to the given types in the Rust function signature.

This needs to wait for https://github.com/gtk-rs/gtk-rs/issues/48 and would be useful for solving https://github.com/gtk-rs/gtk4-rs/issues/24 and https://github.com/gtk-rs/gtk-rs/issues/128

jf2048:
Fixes #19, probably also good enough for #23

I tried to copy the syntax of `clone!` as much as I could. This should be a convenient improvement on `connect_object` since this can take weak references on multiple objects in one call.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -1,6 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 mod clone;
+mod closure;
 mod downgrade_derive;
 mod gboxed_derive;
 mod gboxed_shared_derive;
@@ -264,6 +265,165 @@ use syn::{parse_macro_input, DeriveInput, LitStr};
 #[proc_macro_error]
 pub fn clone(item: TokenStream) -> TokenStream {
     clone::clone_inner(item)
+}
+
+/// Macro for creating a [`Closure`] object. This is a wrapper around [`Closure::new`] that
+/// automatically type checks its arguments at run-time.
+///
+/// A `Closure` takes [`Value`] objects as inputs and output. This macro will automatically convert
+/// the inputs to Rust types when invoking its callback, and then will convert the output back to a
+/// `Value`. All inputs must implement the [`FromValue`] trait, and outputs must either implement
+/// the [`ToValue`] trait or be the unit type `()`. Type-checking of inputs is done at run-time; if
+/// incorrect types are passed via [`Closure::invoke`] then the closure will panic. Note that when
+/// passing input types derived from [`Object`] or [`Interface`], you must take care to upcast to
+/// the exact object or interface type that is being received.
+///
+/// Similarly to [`clone!`](crate::clone!), this macro can be useful in combination with signal
+/// handlers to reduce boilerplate when passing references. Unique to `Closure` objects is the
+/// ability to watch an object using a the `@watch` directive. Only an [`Object`] value can be
+/// passed to `@watch`, and only one object can be watched per closure. When an object is watched,
+/// a weak reference to the object is held in the closure. When the object is destroyed, the
+/// closure will become invalidated: all signal handlers connected to the closure will become
+/// disconnected, and any calls to [`Closure::invoke`] on the closure will be silently ignored.
+/// Internally, this is accomplished using [`Object::watch_closure`] on the watched object.
+///
+/// The `@weak-allow-none` and `@strong` captures are also supported and behave the same as in
+/// [`clone!`](crate::clone!), as is aliasing captures with the `as` keyword. Notably, these
+/// captures are able to reference `Rc` and `Arc` values in addition to `Object` values.
+///
+/// [`Closure`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/closure/struct.Closure.html
+/// [`Closure::new`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/closure/struct.Closure.html#method.new
+/// [`Closure::new_local`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/closure/struct.Closure.html#method.new_local
+/// [`Closure::invoke`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/closure/struct.Closure.html#method.invoke
+/// [`Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
+/// [`FromValue`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/trait.FromValue.html
+/// [`ToValue`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/trait.ToValue.html
+/// [`Interface`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/object/struct.Interface.html
+/// [`Object`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/object/struct.Object.html
+/// [`Object::watch_closure`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/object/trait.ObjectExt.html#tymethod.watch_closure
+/// **⚠️ IMPORTANT ⚠️**
+///
+/// `glib` needs to be in scope, so unless it's one of the direct crate dependencies, you need to
+/// import it because `closure!` is using it. For example:
+///
+/// ```rust,ignore
+/// use gtk::glib;
+/// ```
+///
+/// ### Using as a closure object
+///
+/// ```
+/// use glib_macros::closure;
+///
+/// let concat_str = closure!(|s: &str| s.to_owned() + " World");
+/// let result = concat_str.invoke(&[&"Hello"]).unwrap().get::<String>().unwrap();
+/// assert_eq!(result, "Hello World");
+/// ```
+///
+/// ### Connecting to a signal
+///
+/// For wrapping closures that can't be sent across threads, the
+/// [`closure_local!`](crate::closure_local!) macro can be used. It has the same syntax as
+/// `closure!`, but instead uses [`Closure::new_local`] internally.
+///
+/// ```
+/// use glib;
+/// use glib::prelude::*;
+/// use glib_macros::closure_local;
+///
+/// let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
+/// obj.connect_closure(
+///     "notify", false,
+///     closure_local!(|_obj: glib::Object, pspec: glib::ParamSpec| {
+///         println!("property notify: {}", pspec.name());
+///     }));
+/// ```
+///
+/// ### Object Watching
+///
+/// ```
+/// use glib;
+/// use glib::prelude::*;
+/// use glib_macros::closure_local;
+///
+/// let closure = {
+///     let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
+///     let closure = closure_local!(@watch obj => move || {
+///         obj.type_().name()
+///     });
+///     assert_eq!(
+///         closure.invoke(&[]).unwrap().get::<String>().unwrap(),
+///         "GObject"
+///     );
+///     closure
+/// };
+/// // `obj` is dropped, closure invalidated so it always does nothing and returns None
+/// assert!(closure.invoke(&[]).is_none());
+/// ```
+///
+/// `@watch` has special behavior when connected to a signal:
+///
+/// ```
+/// use glib;
+/// use glib::prelude::*;
+/// use glib_macros::closure_local;
+///
+/// let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
+/// {
+///     let other = glib::Object::new::<glib::Object>(&[]).unwrap();
+///     obj.connect_closure(
+///         "notify", false,
+///         closure_local!(@watch other as b => move |a: glib::Object, pspec: glib::ParamSpec| {
+///             let value = a.property_value(pspec.name());
+///             b.set_property(pspec.name(), &value);
+///         }));
+///     // The signal handler will disconnect automatically at the end of this
+///     // block when `other` is dropped.
+/// }
+/// ```
+///
+/// ### Weak and Strong References
+///
+/// ```
+/// use glib;
+/// use glib::prelude::*;
+/// use glib_macros::closure;
+/// use std::sync::Arc;
+///
+/// let closure = {
+///     let a = Arc::new(String::from("Hello"));
+///     let b = Arc::new(String::from("World"));
+///     let closure = closure!(@strong a, @weak-allow-none b => move || {
+///         // `a` is Arc<String>, `b` is Option<Arc<String>>
+///         format!("{} {}", a, b.as_ref().map(|b| b.as_str()).unwrap_or_else(|| "Moon"))
+///     });
+///     assert_eq!(
+///         closure.invoke(&[]).unwrap().get::<String>().unwrap(),
+///         "Hello World"
+///     );
+///     closure
+/// };
+/// // `a` still kept alive, `b` is dropped
+/// assert_eq!(
+///     closure.invoke(&[]).unwrap().get::<String>().unwrap(),
+///     "Hello Moon"
+/// );
+/// ```
+#[proc_macro]
+#[proc_macro_error]
+pub fn closure(item: TokenStream) -> TokenStream {
+    closure::closure_inner(item, "new")
+}
+
+/// The same as [`closure!`](crate::closure!) but uses [`Closure::new_local`] as a constructor.
+/// This is useful for closures which can't be sent across threads. See the documentation of
+/// [`closure!`](crate::closure!) for details.
+///
+/// [`Closure::new_local`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/closure/struct.Closure.html#method.new_local
+#[proc_macro]
+#[proc_macro_error]
+pub fn closure_local(item: TokenStream) -> TokenStream {
+    closure::closure_inner(item, "new_local")
 }
 
 #[proc_macro_derive(GEnum, attributes(genum))]
--- a/glib-macros/tests/test.rs
+++ b/glib-macros/tests/test.rs
@@ -339,3 +339,127 @@ fn derive_variant() {
     assert_eq!(var.type_().as_str(), "()");
     assert_eq!(var.get::<Variant5>(), Some(v));
 }
+
+#[test]
+fn closure() {
+    let empty = glib::closure!(|| {});
+    assert!(empty.invoke(&[]).is_none());
+
+    let no_arg = glib::closure!(|| 2i32);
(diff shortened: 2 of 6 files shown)


sdroege:
Thanks, this looks really promising. I'll review it in detail later.

> probably also good enough for #23

If you pass the object into the closure as a weak reference then for normal usage yes. However `GtkBuilder` signal handlers can also use this flag, and then connection to the signal must happen slightly differently (I forgot the details).

glib-macros/src/gclosure.rs
```diff
@@ -0,0 +1,220 @@
...
+                }
+            });
+        let arg_names = &self.args;
+        let closure = &self.closure;
+
+        tokens.extend(quote! {
+            {
+                let #closure_ident = {
+                    #(#outer_before)*
+                    #crate_ident::closure::Closure::new_local(move |#values_ident| {
```
sdroege, line 202: This should probably get a `_local` and "normal" variant for the macro so both can be created

glib-macros/src/gclosure.rs
```diff
@@ -0,0 +1,220 @@
...
+        let outer_before = self.captures.iter().map(|c| c.outer_before_tokens(&crate_ident));
+        let inner_before = self.captures.iter().map(|c| c.inner_before_tokens(&crate_ident));
+        let outer_after = self.captures.iter().map(|c| c.outer_after_tokens(&crate_ident, &closure_ident));
+
+        let arg_values = self.args
+            .iter()
+            .enumerate()
+            .map(|(index, arg)| {
+                let err_msg = format!(
+                    "Wrong type for argument {} in gclosure: {{:?}}",
```
sdroege, line 185:
```suggestion
                    "Wrong type for argument {} in GClosure: {{:?}}",
```

glib/src/closure.rs
```diff
@@ -118,6 +118,22 @@ impl Closure {
     }
 }
 
+pub trait ToOptionValue {
```
sdroege, line 121: Maybe should be named `ToClosureReturn` or `ToClosureReturnValue`?

Something like this is basically what I was thinking of in https://github.com/gtk-rs/gtk-rs-core/issues/337 btw @bilelmoussaoui , just the other way around

glib/src/object.rs
```diff
@@ -1636,6 +1636,51 @@ pub trait ObjectExt: ObjectType {
...
+    /// Same as `connect` but takes a `Closure` instead of a `Fn`.
+    #[doc(alias = "g_signal_connect_closure")]
+    fn connect_closure<'a, N>(
+        &self,
+        signal_name: N,
+        after: bool,
+        closure: Closure,
+    ) -> Result<SignalHandlerId, BoolError>
+    where
+        N: Into<&'a str>;
```
sdroege, line 1655: Use plain `&str` for the arguments

glib-macros/tests/test.rs
```diff
@@ -280,3 +280,85 @@ fn subclassable() {
...
+
+        weak_none_test
+    };
+    assert_eq!(weak_none_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 0);
+
+    {
+        let obj1 = glib::Object::new::<glib::Object>(&[]).unwrap();
+        let obj2 = glib::Object::new::<glib::Object>(&[]).unwrap();
+
+        let obj_arg_test = glib::gclosure!(|a: glib::Object, b: glib::Object| {
```
sdroege, line 340: This wants a run through `cargo fmt`

glib/src/object.rs
```diff
@@ -1636,6 +1636,51 @@ pub trait ObjectExt: ObjectType {
...
+    /// Connect a closure to the signal `signal_name` on this object.
+    ///
+    /// If `after` is set to `true` then the callback will be called after the default class
+    /// handler of the signal is emitted, otherwise before.
+    ///
+    /// This fails if the signal does not exist.
+    ///
+    /// Same as `connect` but takes a `Closure` instead of a `Fn`.
+    #[doc(alias = "g_signal_connect_closure")]
+    fn connect_closure<'a, N>(
```
bilelmoussaoui, line 1648: Similar to the changes that happened in ObjectExt lately, this needs to be renamed to `try_connect_closure` & you need to provide a panicking variant named `connect_closure`

glib/src/object.rs
```diff
@@ -1636,6 +1636,51 @@ pub trait ObjectExt: ObjectType {
...
         F: Fn(&[Value]) -> Option<Value>;
 
+    /// Connect a closure to the signal `signal_name` on this object.
+    ///
+    /// If `after` is set to `true` then the callback will be called after the default class
+    /// handler of the signal is emitted, otherwise before.
+    ///
+    /// This fails if the signal does not exist.
+    ///
+    /// Same as `connect` but takes a `Closure` instead of a `Fn`.
```
bilelmoussaoui, line 1646:
```suggestion
    /// Same as [`Self::connect`] but takes a [`crate::Closure`](`Closure`) instead of a `Fn`.
```

sdroege:
> > probably also good enough for #23
> 
> If you pass the object into the closure as a weak reference then for normal usage yes. However `GtkBuilder` signal handlers can also use this flag, and then connection to the signal must happen slightly differently (I forgot the details).

Which you implement differently in the `GtkBuilder` case in the other PR, so that's indeed covered then :)

glib/src/object.rs
```diff
@@ -1636,6 +1636,51 @@ pub trait ObjectExt: ObjectType {
...
+    /// Connect a closure to the signal `signal_id` on this object.
+    ///
+    /// If `after` is set to `true` then the callback will be called after the default class
+    /// handler of the signal is emitted, otherwise before.
+    ///
+    /// This fails if the signal does not exist.
+    ///
+    /// Same as `connect_closure` but takes a `SignalId` instead of a signal name.
+    #[doc(alias = "g_signal_connect_closure_by_id")]
+    fn connect_closure_id(
```
bilelmoussaoui, line 1666: Similar to connect_closure

glib-macros/src/gclosure.rs
```diff
@@ -0,0 +1,220 @@
...
+                let err_msg = format!("failed to upgrade `{}`", alias.to_string());
+                quote! {
+                    let #alias = match #crate_ident::clone::Upgrade::upgrade(&#alias) {
+                        Some(val) => val,
+                        None => panic!(#err_msg)
+                    };
+                }
+            },
+            CaptureKind::WeakAllowNone => quote! {
+                let #alias = #crate_ident::clone::Upgrade::upgrade(&#alias);
```
sdroege, line 65: Should there maybe be a variant that automatically invalidates the closure (i.e. doesn't actually call the body of the closure) if upgrading any of the weak references fails? That seems generally useful and then you don't need to implement that part in the gtk4-rs PR.

sdroege: Calling `g_closure_invalidate()` from here seems a bit tricky as you somehow need to channel the `GClosure*` into the closure itself (needs some `Arc<Mutex<_>>` construct in the general case), but simply not calling the closure body in this case seems a good start already.

glib-macros/src/gclosure.rs
jf2048: Added a second macro for this.

glib-macros/src/gclosure.rs
jf2048: I just removed that part, it was supposed to be the macro name, but it's unnecessary.

glib-macros/src/gclosure.rs
jf2048: I think this should already be handled by calling `watch_closure`. That was my major motivation for doing this. :) See the code in `outer_after_tokens`. Maybe the documentation is not as clear as it could be?

glib-macros/tests/test.rs
jf2048: I ran it, this line is 99 characters long, one short of getting reformatted...

jf2048:
> Which you implement differently in the `GtkBuilder` case in the other PR, so that's indeed covered then :)

Indeed, I thought I would use this macro there initially but it turned out to not be very practical to do it that way, since `BuilderRustScope` needs to take callbacks and only creates the closure objects on demand.

glib-macros/src/gclosure.rs
sdroege: I think I just missed that part in the documentation :) Sorry

glib-macros/src/lib.rs
```diff
@@ -265,6 +266,101 @@ pub fn clone(item: TokenStream) -> TokenStream {
     clone::clone_inner(item)
 }
 
+/// The same as [`gclosure_local!`](crate::gclosure_local) but uses [`Closure::new`] as a
+/// constructor.
```
sdroege, line 270: Can you move the docs to this one, which is more generic? And in the other one mention that it's the same but will panic if the closure is called or dropped from a different thread than it was called on?

glib-macros/src/lib.rs
```diff
@@ -265,6 +266,101 @@ pub fn clone(item: TokenStream) -> TokenStream {
...
+/// [`FromValue`] trait, and outputs must either implement the
+/// [`ToValue`] trait or be the unit type `()`.
+///
+/// Similarly to `clone`, this macro can be useful in combination with signal handlers to reduce
+/// boilerplate when passing references. The `@weak`, `@weak-allow-none` and `@strong` captures are
+/// supported. There is one major difference compared to `clone`: references captured with `@weak`
+/// will always cause the closure to become invalidated when the object is destroyed. When a
+/// closure connected to a signal becomes invalidated, the signal handler will automatically be
+/// disconnected. Internally, this is accomplished using
+/// [`Object::watch_closure`] on each of the weak references.
```
sdroege, line 294: This also brings a big different to the `clone!` macro: That one can also support weak references for `Arc` and `Rc`, for example. But this only works for `Object` now. Which is fine, but should probably be documented :)

glib-macros/src/lib.rs
```diff
@@ -265,6 +266,101 @@ pub fn clone(item: TokenStream) -> TokenStream {
...
+}
+
+/// Macro for creating a [`Closure`] object. This is a wrapper around
+/// [`Closure::new_local`] that type checks its arguments at runtime.
+///
+/// A `Closure` takes [`Value`] objects as inputs and output. This macro will
+/// automatically convert the inputs to Rust types when invoking its callback, and then will
+/// convert the output back to a `Value`. All inputs must implement the
+/// [`FromValue`] trait, and outputs must either implement the
+/// [`ToValue`] trait or be the unit type `()`.
```
sdroege, line 286: Might be worth mentioning here that this will panic if the closure is called with different types than what is used inside the macro

glib/src/object.rs
```diff
@@ -1636,6 +1636,59 @@ pub trait ObjectExt: ObjectType {
...
+        after: bool,
+        closure: Closure,
+    ) -> Result<SignalHandlerId, BoolError>;
+
+    /// Connect a closure to the signal `signal_name` on this object.
+    ///
+    /// If `after` is set to `true` then the callback will be called after the default class
+    /// handler of the signal is emitted, otherwise before.
+    ///
+    /// This fails if the signal does not exist.
```
sdroege, line 1652:
```suggestion
    /// This panics if the signal does not exist.
```

Also below

glib-macros/src/gclosure.rs
sdroege: This is of course not threadsafe: the closure might just be called, then is invalidated, then you try upgrading the weak reference but the object is just gone.

So it would probably be good to also handle this case here instead of panicking

glib/src/object.rs
```diff
@@ -2388,22 +2432,86 @@ impl<T: ObjectType> ObjectExt for T {
...
+        self.try_connect_closure_id(signal_id, details, after, closure)
+    }
+
+    fn try_connect_closure(
+        &self,
+        signal_name: &str,
+        after: bool,
+        closure: Closure,
+    ) -> Result<SignalHandlerId, BoolError> {
+        let signal_name: &str = signal_name.into();
```
sdroege, line 2445:
```suggestion
```

glib-macros/src/lib.rs
jf2048: I was going to but couldn't figure out a good way to get the examples/doctests to work without the `local` variant. It seems there would have to be a longer example for `gclosure` with an object subclass that implements `Send` + `Sync`. Would that work?

glib-macros/src/lib.rs
jf2048: Is there a case to be made for supporting the same `@weak` behavior as `clone!` has? It would be possible to do that and then move the `watch_closure` behavior to a new directive like `@weak-obj` or something like that. But that would also mean supporting `@default-return` and `@default-panic` as well.

glib-macros/src/lib.rs
sdroege: Oh I see. Maybe the examples could stay with the local version (they work the same way anyway), but would be on the non-local version of the macro?

glib-macros/src/lib.rs
sdroege: I think it's fine for now, just needs to be documented. If someone wants this behaviour we can add it later, and for now the `weak-none` variant can be used together with an `unwrap()` or so.

A6GibKm:
Is there a reason it is called `glib::gclosure` rather than `glib::closure`? In my opinion it goes against how namespaces work in Rust.

glib-macros/src/gclosure.rs
jf2048: That should never happen, `watch_closure` will ensure the object has at least one strong reference while the closure is being called. See [here](https://gitlab.gnome.org/GNOME/glib/-/blob/b4b5b3d029c1b5f8c859657bc269c39f276b3c68/gobject/gobject.c#L4426). And you can see this in the unit test for `gclosure!`, the ref count of a `@weak` capture always has a ref count of at least 3.

jf2048: Actually, now that I try this, using `@weak` on multiple objects doesn't work, the `GClosure` is limited to watching only one object. So that will have to be removed.

jf2048:
It was `closure!` first, then I named it `gclosure!` because I thought maybe it made sense because the macro returns a `Closure`, not a rust native type.  It probably should be changed back to `closure!` though now that I think about it.

glib-macros/src/gclosure.rs
sdroege: How do you want to handle that? By just not calling `watch()` and simply not calling the closure when object can't be upgraded? That however won't invalidate the closure and disconnect the signal handler, so we somehow need to also invalidate it.

sdroege:
> It was `closure!` first, then I named it `gclosure!` because I thought maybe it made sense because the macro returns a `Closure`, not a rust native type. It probably should be changed back to `closure!` though now that you mention it.

See also https://github.com/gtk-rs/gtk-rs-core/pull/341#discussion_r747258004 . My reasoning would've been the same for calling this one `gclosure!`, but OTOH it's `glib::Closure` and not `glib:GClosure`. I have no strong opinion here, there are arguments for both.

@bilelmoussaoui @GuillaumeGomez ?

sdroege:
Let's keep this one as `gclosure!` and worry about it in general in https://github.com/gtk-rs/gtk-rs-core/issues/343

glib-macros/src/gclosure.rs
jf2048: I was thinking to have the macro abort if more than one `@weak` is used. To support more would probably need to be changed in gobject by adding an internal mutex to `GClosure`.

glib-macros/src/lib.rs
jf2048: Sorry, not sure I understand. Leave the `gclosure_local!` examples on the local version, but also copy the `gclosure_local!` ones to the non-local version?

glib-macros/src/gclosure.rs
sdroege: That also works, yes

glib-macros/src/lib.rs
sdroege: Move the docs you have now on `gclosure_local!` to `gclosure!` and keep the examples in there using `gclosure_local!`, and then put docs on `gclosure_local!` that refer to `gclosure!` for more details.

glib-macros/src/lib.rs
```diff
@@ -265,6 +266,101 @@ pub fn clone(item: TokenStream) -> TokenStream {
     clone::clone_inner(item)
 }
 
+/// The same as [`gclosure_local!`](crate::gclosure_local) but uses [`Closure::new`] as a
+/// constructor.
+///
+/// [`Closure::new`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/closure/struct.Closure.html#method.new
+#[proc_macro]
+#[proc_macro_error]
+pub fn gclosure(item: TokenStream) -> TokenStream {
```
sdroege, line 275: Consensus in #343 is to remove the `g` prefix, so let's get this in as `closure!` and `closure_local!` directly.

glib-macros/src/lib.rs
jf2048: On thinking about this, I feel it should be changed from `@weak` here, to `@weak-obj` or `@watch` or something like that. The semantics are totally different from the use of `@weak` in `clone!` and it seems that giving it the same name will cause more confusion.

GuillaumeGomez: This is a very good point @jf2048. The naming would be confusing between the two.

sdroege: Good idea!

glib-macros/src/gclosure.rs
jf2048: The macro now aborts when more than one object is watched

glib-macros/src/lib.rs
jf2048: Moved all docs into `closure!` and included some more examples there.

glib-macros/src/lib.rs
jf2048: I changed it to `@watch` and updated the docs to say that only `Object` can be passed there, also added a note specifically mentioning that `@weak-allow-none` and `@strong` still take `Arc` and `Rc`.

glib/src/lib.rs
```diff
@@ -15,7 +15,7 @@ pub use bitflags;
 pub use once_cell;
 
 pub use glib_macros::{
-    clone, gflags, object_interface, object_subclass, Downgrade, GBoxed, GEnum, GErrorDomain,
+    clone, closure, closure_local, gflags, object_interface, object_subclass, Downgrade, GBoxed, GEnum, GErrorDomain,
```
sdroege, line 18: This needs a `cargo fmt` run

sdroege (dismissed):

glib/src/lib.rs
sdroege: Done

sdroege (approved):

bilelmoussaoui (approved):

PR #340, opened 2021-11-10
glib: Lots of `Variant` / `VariantType` improvements

sdroege:
This should lower the number of allocations a bit.

### DIFF
--- a/glib/src/auto/enums.rs
+++ b/glib/src/auto/enums.rs
@@ -674,3 +674,133 @@ impl ErrorDomain for UriError {
         }
     }
 }
+
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
+#[non_exhaustive]
+#[doc(alias = "GVariantClass")]
+pub enum VariantClass {
+    #[doc(alias = "G_VARIANT_CLASS_BOOLEAN")]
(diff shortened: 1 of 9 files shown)


glib/src/auto/enums.rs
```diff
@@ -674,3 +674,133 @@ impl ErrorDomain for UriError {
...
+                Self::Double => "Double",
+                Self::String => "String",
+                Self::ObjectPath => "ObjectPath",
+                Self::Signature => "Signature",
+                Self::Variant => "Variant",
+                Self::Maybe => "Maybe",
+                Self::Array => "Array",
+                Self::Tuple => "Tuple",
+                Self::DictEntry => "DictEntry",
+                _ => "Unknown",
```
GuillaumeGomez, line 746: Can you instead bind `Self::_Unknown(_)` please? It's especially important in case we add new variants: we'll be sure to not forget to add them into the pattern matching thanks to it.

sdroege: This is autogenerated code

GuillaumeGomez: Oh. Ok then it's good, I'll open an issue on `gir` because it's not great.

GuillaumeGomez (approved):

PR #341, opened 2021-11-10
glib-macros: Add a `Variant` trait derive macro

sdroege:
This allows storing arbitrary structs in `Variant`s as long as all their fields can be stored in `Variant`s.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -7,6 +7,7 @@ mod gboxed_shared_derive;
 mod genum_derive;
 mod gerror_domain_derive;
 mod gflags_attribute;
+mod gvariant_derive;
 mod object_interface_attribute;
 mod object_subclass_attribute;
 mod utils;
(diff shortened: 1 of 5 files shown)


GuillaumeGomez (dismissed):

glib-macros/src/lib.rs
```diff
@@ -518,3 +519,28 @@ pub fn downgrade(input: TokenStream) -> TokenStream {
...
+///     some_int: i32,
+/// }
+///
+/// let v = Foo { some_string: String::from("bar"), some_int: 1 };
+/// let var = v.to_variant();
+/// assert_eq!(var.get::<Foo>(), Some(v));
+/// ```
+///
+/// [`glib::Variant`]: variant/struct.Variant.html
+#[proc_macro_derive(Variant)]
```
bilelmoussaoui, line 542: It's not that great that some macros are prefixed with G and others not :/

sdroege: I agree, but I don't know which option is nicer. Do you have a preference?

sdroege: Let's add the `G` here for consistency with all the other derive macros and create a new issue to discuss this? @bilelmoussaoui

bilelmoussaoui: Sounds good to me

bilelmoussaoui (approved):

PR #347, opened 2021-11-11
glib: Fix `Deref` impls for `FixedSizeVariantArray`

sdroege:
These should deref to the contained type, not the underlying slice type.

Also get rid of all the `PartialEq` impls: having these is not too
useful, and there would be a lot more needed to cover all the special
cases similar to the ones the current ones cover.


sdroege:
CI passed, let's get this in? :)

bilelmoussaoui (approved):

PR #348, opened 2021-11-12
glib: Mark wrapper structs as `#[repr(transparent)]`

sdroege:
This way they're completely equivalent to the underlying FFI pointer and
can be used interchangeably.

Also extend a test to assert this.

----

This does not help us much *yet* because our conversion traits for containers always go to Rust traits or have to add a `NULL` at the end, but this allows reducing the number of temporary allocations in the future and in manual code.


bilelmoussaoui (dismissed):

GuillaumeGomez (approved):

sdroege:
Please review again. This comes with some API that makes use of this now to reduce some allocations. In the future we can also teach `gir` about these but that's for another day.

glib/src/translate.rs
```diff
@@ -2274,6 +2274,440 @@ where
...
+enum ContainerTransfer {
+    Full,
+    Container,
+    None,
+}
+
+/// Slice of elements of type `T` allocated by the GLib allocator.
+///
+/// This can be used like a `&[T]`.
+pub struct GPtrSlice<T: GlibPtrDefault + FromGlibPtrFull<<T as GlibPtrDefault>::GlibType>> {
```
sdroege, line 2287: Maybe all this should be moved to some kind of `glib::collections` module and also get re-exported from the crate root?

@bilelmoussaoui @GuillaumeGomez ?

GuillaumeGomez (dismissed):

glib/src/translate.rs
GuillaumeGomez: What would be the advantage?

sdroege: 1. We usually don't have `glib::translate` types in the public API
2. We usually don't have `glib::some_module::Type` in the public API but just `glib::Type`
3. I'm going to add more things in the near future (something around `GSList` / `GList` and `GHashTable`, for example)

GuillaumeGomez: Sounds reasonable. Please move them into `glib::collections` then! :)

sdroege: Done!

GuillaumeGomez (approved):

PR #351, opened 2021-11-12
glib: Re-use paramspec names instead of creating a new NUL-terminated…

sdroege:
… property name when setting/getting properties


GuillaumeGomez (approved):

GuillaumeGomez:
The gtk-rs wrapper is getting really thin. :o

PR #352, opened 2021-11-12
glib: Simplify enum/flags class implementation

sdroege:
Less pointer wrapping and fewer allocations, and better `Debug` impl.

### DIFF
--- a/glib/src/enums.rs
+++ b/glib/src/enums.rs
@@ -2,9 +2,9 @@
 
 use crate::translate::*;
 use crate::value::Value;
-use crate::CStr;
 use crate::Type;
-use std::cmp;
+use std::ffi::CStr;
+use std::{cmp, fmt, ptr};
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 pub enum UserDirectory {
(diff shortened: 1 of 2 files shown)


glib/src/enums.rs
```diff
@@ -5,6 +5,8 @@ use crate::value::Value;
 use crate::CStr;
 use crate::Type;
 use std::cmp;
+use std::fmt;
+use std::ptr;
```
GuillaumeGomez, line 9: Why not merging all these imports into one?

glib/src/enums.rs
```diff
@@ -72,14 +83,15 @@ impl EnumClass {
...
+                ptr::NonNull::new(gobject_ffi::g_type_class_ref(type_.into_glib()) as *mut _)
+                    .unwrap(),
             ))
         }
     }
 
     /// `Type` of the enum.
     pub fn type_(&self) -> Type {
-        unsafe { from_glib((*self.0).g_type_class.g_type) }
+        unsafe { from_glib(self.0.as_ref().g_type_class.g_type) }
```
GuillaumeGomez, line 94: That's super cool.

glib/src/enums.rs
```diff
@@ -88,13 +100,13 @@ impl EnumClass {
...
-    pub fn value(&self, value: i32) -> Option<EnumValue> {
+    pub fn value(&self, value: i32) -> Option<&EnumValue> {
         unsafe {
-            let v = gobject_ffi::g_enum_get_value(self.0, value);
+            let v = gobject_ffi::g_enum_get_value(self.0.as_ptr(), value);
             if v.is_null() {
                 None
             } else {
-                Some(EnumValue(v, self.clone()))
+                Some(&*(v as *const EnumValue))
```
GuillaumeGomez, line 109: No more clone! \o/

GuillaumeGomez (approved):
This is really great! Especially the ptr to slice conversions!

bilelmoussaoui (approved):

glib/src/enums.rs
```diff
@@ -551,75 +575,79 @@ impl FlagsClass {
...
+                if v.value() & f != 0 {
+                    res.push(&*(v as *const FlagsValue));
                 }
             }
-            res
+            Some((flags_class, res))
         }
     }
-
-    /// Get `FlagsClass` to which the flags value belongs.
```
bilelmoussaoui, line 618: I suppose it's fine to lose access to the FlagClass/EnumClass

sdroege: Yes, you can't really have the value without the class.

PR #354, opened 2021-11-13
pango: take a generic attr in AttrList methods

bilelmoussaoui:
Noticed that when bumping librsvg to next gtk-rs stack https://gitlab.gnome.org/GNOME/librsvg/-/merge_requests/634


bilelmoussaoui:
i shouldn't use the web ui for such things :p will update in a bit

sdroege:
Looks good in principle though :)

sdroege (approved):

PR #357, opened 2021-11-13
pango cleanups

bilelmoussaoui:


pango/src/attr_list.rs
```diff
@@ -27,18 +27,18 @@ impl AttrList {
     }
 
     #[doc(alias = "pango_attr_list_insert")]
-    pub fn insert(&self, attr: impl IsAttribute) {
+    pub fn insert(&self, attr: impl AsRef<Attribute>) {
         unsafe {
-            let attr = attr.upcast();
+            let attr = attr.as_ref();
             ffi::pango_attr_list_insert(self.to_glib_none().0, attr.to_glib_none().0 as *mut _);
             mem::forget(attr); //As attr transferred fully
```
bilelmoussaoui, line 34: is the mem::forget still needed here?

bilelmoussaoui: or reworded: how should i handle the clippy warning because now i have a reference

.github/workflows/CI.yml
```diff
@@ -24,7 +24,7 @@ jobs:
           - { name: "gio", features: "v2_66", nightly: "--all-features", test_sys: false }
           - { name: "glib", features: "v2_66", nightly: "--all-features", test_sys: false }
           - { name: "graphene", features: "", nightly: "", test_sys: true }
-          - { name: "pango", features: "v1_46", nightly: "--all-features", test_sys: false }
+          - { name: "pango", features: "v1_46", nightly: "--all-features", test_sys: true }
```
GuillaumeGomez, line 27: :tada:

pango/src/attr_list.rs
sdroege: Well you would have to copy the attribute now :) It's better how the function was before because then it took actual ownership of the value and passed on ownership to the C function.

pango/src/color.rs
```diff
@@ -2,8 +2,39 @@
...
 use crate::Color;
 use glib::translate::*;
+use std::{fmt, ptr};
 
 impl Color {
+    #[doc(alias = "pango_color_parse")]
+    pub fn parse(spec: &str) -> Result<Self, glib::BoolError> {
+        unsafe {
+            let color = ptr::null_mut();
+            let is_success = ffi::pango_color_parse(color, spec.to_glib_none().0);
```
sdroege, line 12:
```suggestion
            let color = Color::uninitialized();
            let is_success = ffi::pango_color_parse(color.to_glib_none_mut().0, spec.to_glib_none().0);
```

etc

sdroege: (same for the alpha function)

pango/src/attr_list.rs
bilelmoussaoui: indeed, but then we have to manually cast when adding an attribute to the AttrList

sdroege: Why? `attr: impl IsAttribute` gives you ownership

bilelmoussaoui: librsvg has a use case where it creates multiples attributes; stores them on a `Vec` then iterate the attributes to set a start_index/end_index; only after that it adds them to the AttrList. Which gives you `Attribute` in the Vec & you can't make Attribute implement IsAttribute

sdroege: I don't know, you either need to copy or pass by value one way or another. There's no third way :)

sdroege: Copying it is then :)

sdroege (dismissed):

.github/workflows/CI.yml
bilelmoussaoui: heh i had to drop the commit; we need to figure out newer pango on CI first :/

pango/src/attr_list.rs
```diff
@@ -1,14 +1,14 @@
...
 
-use crate::{AttrList, IsAttribute};
+use crate::{AttrList, Attribute};
 use glib::translate::*;
 use std::mem;
 
 impl AttrList {
     #[doc(alias = "pango_attr_list_change")]
-    pub fn change(&self, attr: impl IsAttribute) {
+    pub fn change(&self, attr: impl AsRef<Attribute>) {
```
sdroege, line 9:
```suggestion
    pub fn change(&self, attr: impl Into<Attribute>) {
```

Maybe? Every `FooAttribute` can be `Into<Attribute>` via `upcast()`, and `T: Into<T>` blanket implementation exists

bilelmoussaoui: Seems to work fine :-)

sdroege: You actually have to implement `impl From<AttrColor> for Attribute { ... }` etc, otherwise this only works for the literal `Attribute`

bilelmoussaoui: Oups, I thought it was already implemented :)

pango/src/attribute.rs
```diff
@@ -154,6 +154,12 @@ macro_rules! define_attribute_struct {
             }
         }
 
+        impl Into<crate::Attribute> for $rust_type {
```
sdroege, line 157:
```suggestion
        impl From<$rust_type> for crate::Attribute {
```

Like I told you in the comment ;) Then clippy will also be happy.

bilelmoussaoui: Agh... fixed

sdroege (approved):

PR #362, opened 2021-11-13
gio: generate a bunch of auto bindings

bilelmoussaoui:


bilelmoussaoui:
There seems to be more simple things that can be generated, will look into it in a bit :)

gio/src/auto/functions.rs
```diff
@@ -8,6 +8,7 @@ use crate::Cancellable;
 use crate::DBusConnection;
 use crate::File;
 use crate::IOErrorEnum;
+use crate::IOModule;
```
sdroege, line 11:
```
 error[E0432]: unresolved import `crate::IOModule`
Error:   --> gio\src\auto\functions.rs:11:5
   |
11 | use crate::IOModule;
   |     ^^^^^^^^^^^^^^^ no `IOModule` in the root
```

gio/Gir.toml
```diff
@@ -715,6 +728,11 @@ manual_traits = ["InputStreamExtManual"]
     manual = true
     doc_trait_name = "InputStreamExtManual"
 
+[[object]]
+name = "Gio.IOModule"
+status = "generate"
+cfg_condition = "unix"
```
sdroege, line 734: Why wouldn't that exist on Windows? :) That's how you get a TLS module on Windows, for example.

sdroege: But without `GTypeModule` / `GModule` bindings this is not very useful

gio/src/auto/datagram_based.rs
```diff
@@ -0,0 +1,110 @@
...
+        unsafe {
+            from_glib_full(ffi::g_datagram_based_create_source(
+                self.as_ref().to_glib_none().0,
+                condition.into_glib(),
+                cancellable.map(|p| p.as_ref()).to_glib_none().0,
+            ))
+        }
+    }
+
+    //fn receive_messages(&self, messages: /*Ignored*/&[&InputMessage], flags: i32, timeout: i64, cancellable: Option<&impl IsA<Cancellable>>) -> Result<i32, glib::Error> {
```
sdroege, line 97: These two are the two basic functions of this interface. Without it is not really useable. That's why I didn't add it yet :)

gio/src/auto/dtls_client_connection.rs
```diff
@@ -0,0 +1,205 @@
...
+use glib::signal::SignalHandlerId;
+use glib::translate::*;
+use std::boxed::Box as Box_;
+use std::fmt;
+use std::mem::transmute;
+use std::ptr;
+
+glib::wrapper! {
+    #[doc(alias = "GDtlsClientConnection")]
+    pub struct DtlsClientConnection(Interface<ffi::GDtlsClientConnection, ffi::GDtlsClientConnectionInterface>) @requires DatagramBased;
```
sdroege, line 20: ... and without the interface, this one (and the server part) is also not really useful.

gio/src/auto/enums.rs
```diff
@@ -1784,6 +1871,88 @@ impl ToValue for IOErrorEnum {
     }
 }
 
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
+#[non_exhaustive]
+#[doc(alias = "GIOModuleScopeFlags")]
+pub enum IOModuleScopeFlags {
```
sdroege, line 1877: This should be a flags type, not an enum. Misses the corresponding annotation

gio/src/auto/functions.rs
```diff
@@ -440,13 +441,17 @@ pub fn io_error_from_errno(err_no: i32) -> IOErrorEnum {
...
+    unsafe {
+        FromGlibPtrContainer::from_glib_full(ffi::g_io_modules_load_all_in_directory(
+            dirname.as_ref().to_glib_none().0,
+        ))
+    }
+}
 
 //#[doc(alias = "g_io_modules_load_all_in_directory_with_scope")]
-//pub fn io_modules_load_all_in_directory_with_scope(dirname: impl AsRef<std::path::Path>, scope: /*Ignored*/&mut IOModuleScope) -> /*Ignored*/Vec<IOModule> {
+//pub fn io_modules_load_all_in_directory_with_scope(dirname: impl AsRef<std::path::Path>, scope: /*Ignored*/&mut IOModuleScope) -> Vec<IOModule> {
```
sdroege, line 454: The scope should be easily bindable: that's just a new/free function, plus a block function

gio/src/auto/native_volume_monitor.rs
```diff
@@ -0,0 +1,27 @@
+// This file was generated by gir (https://github.com/gtk-rs/gir)
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use crate::VolumeMonitor;
+use std::fmt;
+
+glib::wrapper! {
+    #[doc(alias = "GNativeVolumeMonitor")]
+    pub struct NativeVolumeMonitor(Object<ffi::GNativeVolumeMonitor, ffi::GNativeVolumeMonitorClass>) @extends VolumeMonitor;
```
sdroege, line 10: This probably shouldn't have been in the .gir file. AFAIU it's private and exposed only via the parent class

gio/src/auto/socket.rs
```diff
@@ -352,15 +350,6 @@ impl<O: IsA<Socket>> SocketExt for O {
         }
     }
 
-    fn condition_check(&self, condition: glib::IOCondition) -> glib::IOCondition {
```
sdroege, line 355: Why does this go away? It is still needed for non-datagram sockets, AFAIU. The two APIs are named the same but not doing exactly the same, right?

gio/src/auto/unix_connection.rs
```diff
@@ -0,0 +1,260 @@
...
+        cancellable: Option<&impl IsA<Cancellable>>,
+        callback: P,
+    );
+
+    fn receive_credentials_async_future(
+        &self,
+    ) -> Pin<Box_<dyn std::future::Future<Output = Result<Credentials, glib::Error>> + 'static>>;
+
+    #[doc(alias = "g_unix_connection_receive_fd")]
+    fn receive_fd(&self, cancellable: Option<&impl IsA<Cancellable>>) -> Result<i32, glib::Error>;
```
sdroege, line 49: That `i32` there should use one of the fd traits from std

sdroege: (probably)

gio/src/auto/unix_connection.rs
```diff
@@ -0,0 +1,260 @@
...
+    );
+
+    fn send_credentials_async_future(
+        &self,
+    ) -> Pin<Box_<dyn std::future::Future<Output = Result<(), glib::Error>> + 'static>>;
+
+    #[doc(alias = "g_unix_connection_send_fd")]
+    fn send_fd(
+        &self,
+        fd: i32,
```
sdroege, line 71: And this

gio/Gir.toml
bilelmoussaoui: because the windows CI failed at linking stage

gio/src/auto/functions.rs
bilelmoussaoui: heh a gir issue, if a parameter is behind a feature flag; the function & its imports are not behind that feature flag

GuillaumeGomez: Please open an issue if not already done. ;)

bilelmoussaoui:
I am going to go ahead and close this one, if some specific bindings are needed in the future we can add them separately.

PR #363, opened 2021-11-14
simplify IsImplementable implementations

bilelmoussaoui:
Similar to the simplification made to IsSubclassable.

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -2,7 +2,7 @@
 
 //! Module that contains the basic infrastructure for subclassing `GObject`.
 
-use crate::object::{Cast, IsClass, ObjectSubclassIs, ObjectType, ParentClassIs};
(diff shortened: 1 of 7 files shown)


glib/src/subclass/types.rs
```diff
@@ -186,21 +186,41 @@ impl<U: IsClass + ParentClassIs> IsSubclassableExt for U {
     }
 }
 
+// FIXME: same thing as IsSubclassable
+#[doc(hidden)]
+pub trait IsImplementableDefault<T: ObjectSubclass>: IsInterface {
```
bilelmoussaoui, line 191: The extra trait doesn't seem to be necessary; we can just make trait methods do nothing by default? :)

sdroege: Yrd

sdroege (approved):

Issue #343, opened 2021-11-11
Consistently name macros


sdroege:
Currently many macros have a `G`/`g` as prefix while others don't. We should do this consistently.

Any objections to drop the prefix on all macros?

CC @bilelmoussaoui @GuillaumeGomez

GuillaumeGomez:
Can you show examples please?

sdroege:
> Can you show examples please?

https://gtk-rs.org/gtk-rs-core/git/docs/glib_macros/index.html

GuillaumeGomez:
I think we should do the following: any attribute/derive macro should start with `g` and the others shouldn't. The only thing I don't have an opinion on is either we should keep the capital letters or not...

So from the link you provided, only `clone` shouldn't be prepended.

sdroege:
IMO none of them should have a `G` at the front

bilelmoussaoui:
- For derive macros we currently have: 1 without a prefix, the rest with one. I think we should drop the prefix from the rest
- For macros like clone/gclosure (in the PR), i think the g prefix should be dropped as well; most of the times you will just call `glib::clone` / `glib::closure` without having to have it in scope. 
- For attribute macros, we only have a gflags that has a g prefix; I don't think it makes sense there neither :-)

A6GibKm:
+1, those prefixes seem like a C artifact.

GuillaumeGomez:
So that mean the derive will become `Boxed` or `Enum`. It's hard to make the name less informative at this point... As for the attributes, `flags` means literally nothing as well...

A6GibKm:
> So that mean the derive will become Boxed or Enum. It's hard to make the name less informative at this point... As for the attributes, flags means literally nothing as well...

The derive will become `glib::Enum`, of course if you import `use glib::Enum` it it is not clear, but the same could be said about a bunch of structs if imported blindly.

sdroege:
> So that mean the derive will become `Boxed` or `Enum`. It's hard to make the name less informative at this point... As for the attributes, `flags` means literally nothing as well...

It's a `glib::Enum` then, which is very closely connected to `glib::EnumClass`. Same for flags, etc.
`glib::Variant` would derive the things needed for a... `glib::Variant`.

Currently then prefix on the macros is odd also because there's no such prefix on the corresponding types.

GuillaumeGomez:
But in `glib` it'll be `crate::Enum`.

sdroege:
[...]

PR #364, opened 2021-11-14
Update macro syntax for enum, flags, boxed, and variant

pbor:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/343

### DIFF
--- a/glib-macros/src/enum_derive.rs
+++ b/glib-macros/src/enum_derive.rs
@@ -7,7 +7,7 @@ use quote::{quote, quote_spanned};
 use syn::{punctuated::Punctuated, spanned::Spanned, token::Comma, Data, Ident, Variant};
 
 use crate::utils::{
-    crate_ident_new, gen_enum_from_glib, parse_item_attributes, parse_type_name, ItemAttribute,
+    crate_ident_new, gen_enum_from_glib, parse_item_attributes, parse_name, ItemAttribute,
 };
 
 // Generate glib::gobject_ffi::GEnumValue structs mapping the enum such as:
@@ -73,7 +73,7 @@ pub fn impl_enum(input: &syn::DeriveInput) -> TokenStream {
         _ => abort_call_site!("#[derive(glib::Enum)] only supports enums"),
     };
 
-    let gtype_name = match parse_type_name(input, "enum_type") {
+    let gtype_name = match parse_name(input, "enum_type") {
         Ok(name) => name,
         Err(e) => abort_call_site!(
             "{}: #[derive(glib::Enum)] requires #[enum_type(name = \"EnumTypeName\")]",
--- a/glib-macros/src/enum_derive.rs
+++ b/glib-macros/src/enum_derive.rs
@@ -16,7 +16,7 @@ use crate::utils::{
 //         value_name: "Goat\0" as *const _ as *const _,
 //         value_nick: "goat\0" as *const _ as *const _,
 //     },
-fn gen_genum_values(
+fn gen_enum_values(
     enum_name: &Ident,
     enum_variants: &Punctuated<Variant, Comma>,
 ) -> (TokenStream, usize) {
@@ -29,11 +29,11 @@ fn gen_genum_values(
         let mut value_name = name.to_string().to_camel_case();
         let mut value_nick = name.to_string().to_kebab_case();
 
-        let attrs = parse_item_attributes("genum", &v.attrs);
+        let attrs = parse_item_attributes("enum_value", &v.attrs);
         let attrs = match attrs {
             Ok(attrs) => attrs,
             Err(e) => abort_call_site!(
-                "{}: GEnum enum supports only the following optional attributes: #[genum(name = \"The Cat\", nick = \"chat\")]",
+                "{}: derive(glib::Enum) enum supports only the following optional attributes: #[enum_value(name = \"The Cat\", nick = \"chat\")]",
                 e
             ),
         };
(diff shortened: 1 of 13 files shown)


glib-macros/src/enum_derive.rs
```diff
@@ -29,11 +27,11 @@ fn gen_genum_values(
         let mut value_name = name.to_string().to_camel_case();
         let mut value_nick = name.to_string().to_kebab_case();
 
-        let attrs = parse_item_attributes("genum", &v.attrs);
+        let attrs = parse_item_attributes("enum_value", &v.attrs);
         let attrs = match attrs {
             Ok(attrs) => attrs,
             Err(e) => abort_call_site!(
-                "{}: GEnum enum supports only the following optional attributes: #[genum(name = \"The Cat\", nick = \"chat\")]",
```
bilelmoussaoui, line 36: that should probably be called member

pbor: I think `enum_value` is more consistent since it corresponds to creating a `GEnumValue`

sdroege:
> Start work on #343

Am I missing something or did you do all of them here? :) So not really a "start".

pbor:
> Am I missing something or did you do all of them here? :) So not really a "start".

Yeah, at the end I think I covered all of them :)

glib-macros/tests/test.rs
```diff
@@ -182,14 +183,14 @@ fn derive_gboxed_nullable() {
 }
 
 #[test]
-fn attr_gflags() {
-    #[gflags("MyFlags")]
+fn attr_flags() {
+    #[glib::flags("MyFlags")]
```
bilelmoussaoui, line 187: why it's not called flags_type?

pbor: `glib::flags` is not a derive macro, so the type name is passed directly to the attribute itself

glib-macros/src/enum_derive.rs
```diff
@@ -65,26 +63,31 @@ fn gen_genum_values(
...
+        _ => abort_call_site!("#[derive(glib::Enum)] requires #[enum_type(\"EnumTypeName\")]"),
     };
 
-    let gtype_name = match parse_type_name(input, "genum") {
-        Ok(v) => v,
+    let gtype_name = match enum_type.parse_args::<LitStr>() {
+        Ok(name) => name,
         Err(e) => abort_call_site!(
-            "{}: derive(GEnum) requires #[genum(type_name = \"EnumTypeName\")]",
+            "{}: #[derive(glib::Enum)] requires #[enum_type(\"EnumTypeName\")]",
```
sdroege, line 82: In theory we can also derive all these with a default from the type name btw, but we don't do that intentionally because these need C-style namespacing @pbor

glib-macros/tests/test.rs
sdroege: I think like this is fine

sdroege (dismissed):
Looks all good to me

glib-macros/src/enum_derive.rs
pbor: yeah, we could pass the namespace instead of the full name, but does not seem to be worth it

sdroege:
@pbor @bilelmoussaoui @GuillaumeGomez Should we get this in then? Needs to be rebased on latest master but otherwise seems fine to me.

glib-macros/tests/test.rs
```diff
@@ -90,15 +90,15 @@ fn derive_shared_arc_nullable() {
 }
 
 #[test]
-fn derive_genum() {
-    #[derive(Debug, Eq, PartialEq, Clone, Copy, GEnum)]
+fn derive_enum() {
+    #[derive(Debug, Eq, PartialEq, Clone, Copy, glib::Enum)]
     #[repr(u32)]
-    #[genum(type_name = "TestAnimalType")]
+    #[enum_type("TestAnimalType")]
```
GuillaumeGomez, line 96: Can we have something similar to `enum_value` here (using `name = `)?

glib-macros/tests/test.rs
```diff
@@ -133,9 +133,9 @@ fn derive_genum() {
 }
 
 #[test]
-fn derive_gboxed() {
-    #[derive(Clone, Debug, PartialEq, Eq, GBoxed)]
-    #[gboxed(type_name = "MyBoxed")]
+fn derive_boxed() {
+    #[derive(Clone, Debug, PartialEq, Eq, glib::Boxed)]
+    #[boxed_type("MyBoxed")]
```
GuillaumeGomez, line 138: Same.

glib-macros/tests/test.rs
```diff
@@ -149,9 +149,10 @@ fn derive_gboxed() {
 }
 
 #[test]
-fn derive_gboxed_nullable() {
-    #[derive(Clone, Debug, PartialEq, Eq, GBoxed)]
-    #[gboxed(type_name = "MyNullableBoxed", nullable)]
+fn derive_boxed_nullable() {
+    #[derive(Clone, Debug, PartialEq, Eq, glib::Boxed)]
+    #[boxed_type("MyNullableBoxed")]
+    #[boxed_nullable]
```
GuillaumeGomez, line 155: I really preferred when it was inlined alongside the name.

glib-macros/tests/test.rs
GuillaumeGomez: Can we have a `name = ` here too?

sdroege:
@GuillaumeGomez Is it good for you now? Then I'll go over it another time too

GuillaumeGomez (dismissed):

GuillaumeGomez:
Yes, it looks good to me now. Thanks @pbor !

glib-macros/src/lib.rs
```diff
@@ -443,20 +502,20 @@ pub fn genum_derive(input: TokenStream) -> TokenStream {
 /// use glib::prelude::*;
 /// use glib::subclass::prelude::*;
 ///
-/// #[derive(Debug, Copy, Clone, glib::GErrorDomain)]
-/// #[gerror_domain(name = "ExFoo")]
+/// #[derive(Debug, Copy, Clone, glib::ErrorDomain)]
+/// #[error_domain("ExFoo")]
```
sdroege, line 506:
```suggestion
/// #[error_domain(name = "ExFoo")]
```

That's what @GuillaumeGomez wanted everywhere else so I guess this is just an oversight?

sdroege (commented):
Seems good to me otherwise too, it's all more consistent now at least

glib-macros/src/lib.rs
GuillaumeGomez: Ah yes! Missed this one, good catch.

sdroege (approved):

PR #366, opened 2021-11-15
glib: Add new trait for allowing to unwrap the return value of `Closu…

sdroege:
…re::emit()` directly instead of working with a `glib::Value`

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/337

### DIFF
--- a/glib-macros/src/closure.rs
+++ b/glib-macros/src/closure.rs
@@ -232,6 +232,7 @@ impl ToTokens for Closure {
             }
         });
         let arg_names = &self.args;
+        let args_len = self.args.len();
         let closure = &self.closure;
         let constructor = Ident::new(self.constructor, Span::call_site());
 
@@ -239,7 +240,8 @@ impl ToTokens for Closure {
             {
                 let #closure_ident = {
                     #(#outer_before)*
-                    #crate_ident::closure::Closure::#constructor(move |#values_ident| {
+                    #crate_ident::closure::RustClosure::#constructor(move |#values_ident| {
+                        assert_eq!(#values_ident.len(), #args_len);
                         #(#inner_before)*
                         #(#arg_values)*
                         #crate_ident::closure::ToClosureReturnValue::to_closure_return_value(
--- a/glib-macros/tests/test.rs
+++ b/glib-macros/tests/test.rs
@@ -343,23 +343,16 @@ fn derive_variant() {
 #[test]
 fn closure() {
     let empty = glib::closure!(|| {});
-    assert!(empty.invoke(&[]).is_none());
+    empty.invoke::<()>(&[]);
 
     let no_arg = glib::closure!(|| 2i32);
-    assert_eq!(no_arg.invoke(&[]).unwrap().get::<i32>().unwrap(), 2);
+    assert_eq!(no_arg.invoke::<i32>(&[]), 2);
 
     let add_1 = glib::closure!(|x: i32| x + 1);
-    assert_eq!(add_1.invoke(&[&3i32]).unwrap().get::<i32>().unwrap(), 4);
+    assert_eq!(add_1.invoke::<i32>(&[&3i32]), 4);
 
     let concat_str = glib::closure!(|s: &str| s.to_owned() + " World");
-    assert_eq!(
-        concat_str
-            .invoke(&[&"Hello"])
-            .unwrap()
-            .get::<String>()
-            .unwrap(),
-        "Hello World"
-    );
+    assert_eq!(concat_str.invoke::<String>(&[&"Hello"]), "Hello World");
 
     let weak_test = {
         let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
@@ -367,12 +360,12 @@ fn closure() {
         assert_eq!(obj.ref_count(), 1);
         let weak_test = glib::closure_local!(@watch obj => move || obj.ref_count());
         assert_eq!(obj.ref_count(), 1);
-        assert_eq!(weak_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 3);
+        assert_eq!(weak_test.invoke::<u32>(&[]), 3);
         assert_eq!(obj.ref_count(), 1);
 
         weak_test
     };
-    assert!(weak_test.invoke(&[]).is_none());
+    weak_test.invoke::<()>(&[]);
 
     {
         trait TestExt {
@@ -382,7 +375,7 @@ fn closure() {
         impl TestExt for glib::Object {
             fn ref_count_in_closure(&self) -> u32 {
                 let closure = glib::closure_local!(@watch self as obj => move || obj.ref_count());
-                closure.invoke(&[]).unwrap().get::<u32>().unwrap()
+                closure.invoke::<u32>(&[])
             }
         }
 
@@ -399,7 +392,7 @@ fn closure() {
             fn ref_count_in_closure(&self) -> u32 {
                 let closure =
                     glib::closure_local!(@watch self.obj as obj => move || obj.ref_count());
-                closure.invoke(&[]).unwrap().get::<u32>().unwrap()
+                closure.invoke::<u32>(&[])
             }
         }
 
@@ -413,11 +406,11 @@ fn closure() {
         let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
 
         let strong_test = glib::closure_local!(@strong obj => move || obj.ref_count());
-        assert_eq!(strong_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 2);
+        assert_eq!(strong_test.invoke::<u32>(&[]), 2);
 
         strong_test
     };
-    assert_eq!(strong_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 1);
+    assert_eq!(strong_test.invoke::<u32>(&[]), 1);
 
     let weak_none_test = {
         let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
@@ -425,11 +418,11 @@ fn closure() {
         let weak_none_test = glib::closure_local!(@weak-allow-none obj => move || {
             obj.map(|o| o.ref_count()).unwrap_or_default()
         });
-        assert_eq!(weak_none_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 2);
+        assert_eq!(weak_none_test.invoke::<u32>(&[]), 2);
 
         weak_none_test
     };
-    assert_eq!(weak_none_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 0);
+    assert_eq!(weak_none_test.invoke::<u32>(&[]), 0);
 
     {
         let obj1 = glib::Object::new::<glib::Object>(&[]).unwrap();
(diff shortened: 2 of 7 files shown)


glib/src/subclass/object.rs
```diff
@@ -362,7 +362,7 @@ mod test {
                             .get()
                             .expect("type conformity checked by 'Object::set_property'");
                         self.name.replace(name);
-                        obj.emit_by_name("name-changed", &[&*self.name.borrow()]);
+                        obj.emit_by_name::<()>("name-changed", &[&*self.name.borrow()]);
```
sdroege, line 365: This is a bit unfortunate: if you just don't use the return value the compiler doesn't know what type to expect.

sdroege:
CC @jf2048

bilelmoussaoui:
Shouldn't this move the current methods to a different name in case someone still wants the original behaviour of returning a glib::Value, the same as the other ObjectExt changes ?

sdroege:
> Shouldn't this move the current methods to a different name in case someone still wants the original behaviour of returning a glib::Value, the same as the other ObjectExt changes ?

Those exist already with the `_values()` prefix

GuillaumeGomez:
I'm not sure to understand why we would want to return `T` directly. Why would we want to remove the `Option` wrapping?

sdroege:
> I'm not sure to understand why we would want to return `T` directly. Why would we want to remove the `Option` wrapping?

`None` is only for `()` as return value, and in all other cases you get a `T: FromValue`.

sdroege:
(see also the `glib::closure!` macro etc)

GuillaumeGomez:
> `None` is only for `()` as return value, and in all other cases you get a `T: FromValue`.

Hum yes, it's exactly why the `Option` type exists. Seems perfectly fine to me.

sdroege:
> Hum yes, it's exactly why the `Option` type exists. Seems perfectly fine to me.

Your statement makes absolutely no sense :) `Option` exists for optional/nullable values, which is not what we have here. `None` is returned if the return type of the closure is `()`, `Some(value)` is returned if it's any other kind of return type (including `Option<String>`, `String`, `i32`, etc. but not `()`)

GuillaumeGomez (dismissed):

sdroege:
This also needs a `gir` update for the autogenerated `emit` functions

glib-macros/tests/test.rs
```diff
@@ -343,36 +343,29 @@ fn derive_variant() {
...
         assert_eq!(obj.ref_count(), 1);
         let weak_test = glib::closure_local!(@watch obj => move || obj.ref_count());
         assert_eq!(obj.ref_count(), 1);
-        assert_eq!(weak_test.invoke(&[]).unwrap().get::<u32>().unwrap(), 3);
+        assert_eq!(weak_test.invoke::<u32>(&[]), 3);
         assert_eq!(obj.ref_count(), 1);
 
         weak_test
     };
-    assert!(weak_test.invoke(&[]).is_none());
```
sdroege, line 375: Now this one is interesting. Invalidated closures return nothing too, even if the closure before had an actual return type that was not `()`.

I think we'll need to have `invoke()` return a `Result` and check for invalidation afterwards to distinguish the two no-result cases.

sdroege: Actually worse: the return value has to be initialized to the correct type before invoking and we currently don't do that. And then it's impossible to distinguish `()` from any other type as passing another type in will simply keep it at that type.

This only works for closures we create right now because we also initialize the type of the value when returning.

Also for closures from C the arguments are apparently not type-checked at runtime either.

I think `invoke()` has to be an `unsafe` function, unfortunately.

----

For signal closures this is no problem because there the return type and the arguments are introspectable at runtime. However when connecting a `Closure` as signal handler there will only be criticals at runtime if the return type does not match.

FYI @jf2048

sdroege: The main problem  I see here now is that `connect_closure()` is actually unsafe. If your closure returns a `String` and the C code assumes an `Object`, then it would silently use the string pointer as an object pointer and things fall apart.

That's why we have the additional checks inside the `GClosure` closures inside the normal `connect()`.

sdroege: I.e. the whole `closure!` and `connect_closure()` machinery is currently not good

sdroege: I think we'd go with having a `struct RustClosure { closure: glib::Closure, return_type: glib::Type, parameters: Vec<glib::Type> }` from the macro (the constructor of that would have to be unsafe) and then make use of that in `connect_closure()` to check if things match up.

`RustClosure` would have a bit of safe API (including `invoke()` but not the constructor for obvious reasons), and all `glib::Closure` related API would be unsafe

sdroege: I have this half-implemented now, just need to put all the pieces together.

sdroege: Hrm, I have some problems coming up with a way that infers the closure argument/return types while also working for all the different special `@` captures we have. I have one version that compiles and causes a memory leak for `@strong`, and versions that fail compiling on one or another for the closure variants.

glib-macros/src/closure.rs
```diff
@@ -232,25 +267,33 @@ impl ToTokens for Closure {
...
                     #(#outer_before)*
-                    #crate_ident::closure::Closure::#constructor(move |#values_ident| {
+                    #crate_ident::closure::RustClosure::#constructor(#types_ident.0, #types_ident.1, move |#values_ident| {
                         #(#inner_before)*
                         #(#arg_values)*
+
+                        // TODO: assert values.len() == args.len()
+
+                        let #closure_ident = #closure;
+                        let _ = #get_types_ident(Some(&#closure_ident));
```
sdroege, line 284: This doesn't work because of

```
error[E0373]: closure may outlive the current function, but it borrows `obj`, which is owned by the current function
   --> glib-macros/tests/test.rs:394:73
    |
394 |                     glib::closure_local!(@watch self.obj as obj => move || obj.ref_count());
    |                     ----------------------------------------------------^^-----------------
    |                     |                                                   |  |
    |                     |                                                   |  `obj` is borrowed here
    |                     |                                                   may outlive borrowed value `obj`
    |                     lifetime `'1` appears in the type of `____get_types`
```

now. I'm not entirely sure what the problem is. The `____get_types` above seems to cause the closure to require `'static` or otherwise some kind of lifetime mismatch. The problem only happens when calling the `____get_types` function.

glib-macros/src/closure.rs
```diff
@@ -222,6 +224,39 @@ impl ToTokens for Closure {
...
+        let arg_and_return_types = {
+            let type_parameter_idents = self
+                .args
+                .iter()
+                .enumerate()
+                .map(|(index, _arg)| Ident::new(&format!("T{}", index), Span::call_site()))
+                .collect::<Vec<_>>();
+
+            quote! {
+                let #get_types_ident: Box<dyn Fn(Option<&_>) -> _ + Send + Sync> = Box::new(|f| {
```
sdroege, line 238: This is all a bit convoluted but I couldn't find another way. The `Box` is needed so that we can move this very closure into our other closure below in a meaningful way. But I think it should be possible without a `Box` somehow.

sdroege: Should probably also move this block into the other one at the bottom. No advantage of having it separate.

glib-macros/src/closure.rs
sdroege: If someone wants to try their luck, this is the expanded code as a standalone testcase against glib-rs git master:

```rust
use glib::prelude::*;

fn new<F: Fn(&[glib::Value]) -> Option<glib::Value>>(r: glib::Type, a: Vec<glib::Type>, f: F) -> glib::Closure { todo!() }

fn main() {
    let obj = glib::Object::new::<glib::Object>(&[]).unwrap();

    {
        let ____get_types: Box<dyn Fn(Option<&_>) -> _ + Send + Sync> = Box::new(|f| {
            fn ____get_types<R: glib::StaticType, F: FnOnce() -> R>(
                f: Option<&F>,
            ) -> Option<(glib::Type, Vec<glib::Type>)> {
                if f.is_some() {
                    return None;
                }
                Some((<R as glib::StaticType>::static_type(), vec![]))
            }
            ____get_types(f)
        });
        let ____types = ____get_types(None).unwrap();
        let ____closure = unsafe {
            let obj = glib::clone::Downgrade::downgrade(&obj);
            new(____types.0, ____types.1, move |____values| {
                let obj = match glib::clone::Upgrade::upgrade(&obj) {
                    Some(val) => val,
                    None => panic!("failed to upgrade `obj`"),
                };
                let ____closure = || obj.ref_count();
                let _ = ____get_types(Some(&____closure));
                glib::closure::ToClosureReturnValue::to_closure_return_value(&(____closure)())
            })
        };
        glib::object::Object::watch_closure(&obj, &____closure);
    }
}
```

glib-macros/src/closure.rs
sdroege: Got rid of the box and simplified this part of the code a bit.

glib-macros/src/closure.rs
sdroege: This is all actually unnecessary: the caller is supposed to initialize the return `&mut Value` with the correct type, so we can check if the return value of our closure is of the correct type and otherwise panic.

For the arguments we already have checks in place via the `Value` getters.

It only means that for non-Rust created closures the invocations are still unsafe, but otherwise this simplifies things a lot.

I'll update the PR accordingly later.

sdroege:
This is good for review now.

glib/src/closure.rs
```diff
@@ -27,17 +30,131 @@ wrapper! {
...
+    }
+}
+
+impl AsRef<Closure> for RustClosure {
+    fn as_ref(&self) -> &Closure {
+        &self.0
+    }
+}
+
+impl AsRef<Closure> for Closure {
```
bilelmoussaoui, line 116: This should behind a `#[doc(hidden)]` maybe?

glib/src/subclass/object.rs
```diff
@@ -612,7 +609,7 @@ mod test {
                     .to_value(),
             )
         });
-        let value = obj.emit_by_name("create-child-object", &[]).unwrap();
+        let value: glib::Object = obj.emit_by_name("create-child-object", &[]);
```
bilelmoussaoui, line 612:
```suggestion
        let value = obj.emit_by_name::<glib::Object>("create-child-object", &[]);
```
Any reason why this one doesn't use the turbo-fish annotation?

bilelmoussaoui (approved):

glib/src/subclass/object.rs
sdroege: In this one I wanted to check if type inference of the return type works, that's why :)

glib/src/closure.rs
sdroege: I don't like that we're hiding trait impls in the docs in general.

glib/src/closure.rs
```diff
@@ -27,17 +30,139 @@ wrapper! {
...
+impl RustClosure {
+    /// Creates a new closure around a Rust closure.
+    ///
+    /// See [`glib::closure!`] for a way to create a closure with concrete types.
+    ///
+    /// # Panics
+    ///
+    /// Invoking the closure with wrong argument types or returning the wrong return value type
+    /// will panic.
+    #[doc(alias = "g_closure_new")]
```
GuillaumeGomez, line 45: Would be nice to have a code example (run when using `cargo test`) for each of these methods.

glib/src/closure.rs
GuillaumeGomez: I don't think it matter much here...

glib/src/closure.rs
sdroege: We tests covering them already. Ideally nobody should use these functions unless they already know what they're doing. There's the macro for the other case.

GuillaumeGomez: Code examples don't hurt. Even if you know what you're doing.

sdroege: I forward all people using the API wrongly to you then :)

I'll copy&paste one of the tests into the docs later.

GuillaumeGomez: That's another debate.

sdroege: There is a code example now on the two constructors.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks for adding the code examples! :)

PR #377, opened 2021-11-18
glib: Fix missing indirection in previous commit and add tests

sdroege:


glib/src/error.rs
```diff
@@ -260,4 +261,18 @@ mod tests {
         let true_dynamic_res = result_from_gboolean!(ffi::GTRUE, "{} message", "Dynamic");
         assert!(true_dynamic_res.is_ok());
     }
+
+    #[test]
+    fn test_value() {
+        let e1 = Error::new(crate::FileError::Failed, "Failed");
+        // This creates a copy ...
```
sdroege, line 268: For this I have a solution planned for the release after this one FWIW so that we don't create copies unless needed.

glib/src/date_time.rs
```diff
@@ -0,0 +1,17 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
```
GuillaumeGomez, line 1: It seems weird to add a file containing only a test. Can you put it into the `tests` folder instead please?

GuillaumeGomez: It increases build time when running tests only. Or am I misunderstanding you?

sdroege: Yes, but it seems completely unnecessary :) We might add new manual datetime API there soonish anyway.

sdroege: Also conceptually this is an unit test, not an integration test

GuillaumeGomez: That triggers strong OCD in me but whatever. Let's go with that.

glib/src/error.rs
GuillaumeGomez: Don't forget to open an issue if not already done. :)

GuillaumeGomez (approved):

glib/src/date_time.rs
sdroege: Funny because the other way would do that for me :)

PR #379, opened 2021-11-18
use must_use for manual builder patterns

bilelmoussaoui:


glib/src/object.rs
```diff
@@ -3348,6 +3348,7 @@ impl<'a> BindingBuilder<'a> {
     /// Establish the property binding.
     ///
     /// This fails if the provided properties do not exist.
+    #[must_use = "Builder needs to be built to be used"]
     pub fn try_build(self) -> Result<crate::Binding, crate::BoolError> {
```
bilelmoussaoui, line 3352: I don't know how to handle this one as we have a try_build/build variants; how we are supposed to use must_use? or as build calls try_build it can know it's actually used?

cc @GuillaumeGomez

sdroege: You probably want to annotate the builder type with this annotation, not the function?

GuillaumeGomez: `must_use` is applied on the returned item, not the function itself.

We can go another way and `impl Drop` on the `Builder` and it wasn't built, panic.

bilelmoussaoui: What do you mean it's applied on the returned item? So a build function on a builder with a must_use, is useless?

GuillaumeGomez: Well, unless you want to force the usage of the returned type, yes.

@sdroege: It doesn't force to use `*build`, only to get the object into a variable.

bilelmoussaoui: heh wait, that doesn't make it useful at all. then, the change on gir side where we added it to the build function wasn't good neither i guess

bilelmoussaoui:
This one is no longed needed

PR #380, opened 2021-11-19
glib: Validate signal class handler return value and arguments when chaining up

sdroege:


glib/src/subclass/signal.rs
```diff
@@ -526,10 +538,29 @@ impl Signal {
...
+
         let class_handler = class_handler.map(|class_handler| {
             Closure::new(move |values| unsafe {
                 let instance = gobject_ffi::g_value_get_object(values[0].to_glib_none().0);
-                class_handler(&SignalClassHandlerToken(instance as *mut _), values)
+                let res = class_handler(&SignalClassHandlerToken(instance as *mut _, return_type.into(), values.as_ptr()), values);
+
+                if return_type == Type::UNIT {
+                    if let Some(ref v) = res {
+                        assert!(res.is_none(), "Signal has no return value but class handler returned a value of type {}", v.type_());
```
GuillaumeGomez, line 550: How can this `assert` be triggered since we checked just before `res` was `Some`?

glib/src/subclass/signal.rs
```diff
@@ -526,10 +538,29 @@ impl Signal {
...
+                    }
+                } else {
+                    match res {
+                        None => {
+                            assert!(res.is_some(), "Signal has a return value but class handler returned none");
```
GuillaumeGomez, line 555: Why not simply panic?

glib/src/subclass/types.rs
```diff
@@ -876,20 +876,58 @@ pub(crate) unsafe fn signal_override_class_handler<F>(
...
+                assert!(
+                    res.is_none(),
+                    "Signal has no return value but class handler returned a value of type {}",
+                    v.type_()
+                );
+            }
+        } else {
+            match res {
+                None => {
+                    assert!(
```
GuillaumeGomez, line 907: panic?

glib/src/subclass/types.rs
```diff
@@ -876,20 +876,58 @@ pub(crate) unsafe fn signal_override_class_handler<F>(
...
+                return_type.into(),
+                values.as_ptr(),
+            ),
+            values,
+        );
+
+        if return_type == Type::UNIT {
+            if let Some(ref v) = res {
+                assert!(
+                    res.is_none(),
```
GuillaumeGomez, line 899: Cannot be triggered.

glib/src/subclass/signal.rs
sdroege: It will always hit, that's how :)

I went with an assert here so it's a bit clearer from the error what went wrong. It could also just be a plain `panic!`.

glib/src/subclass/signal.rs
sdroege: See above

glib/src/subclass/types.rs
sdroege: See above

glib/src/subclass/types.rs
sdroege: Will always trigger ;) See above

glib/src/subclass/signal.rs
GuillaumeGomez: Well, my comment makes it very obvious that the code isn't clear. Also, just realized that you match on `res` just after. Might be better to match directly on it and `panic` on the `Some` branch. That'll remove one extra check.

sdroege: The check is optimized away. The difference is in the output: just a plain panic with that message, vs. a `assertion res.is_none() failed: that message`. It seems more helpful to the user, but if you think the code is too hard to read otherwise I can change that.

GuillaumeGomez: The code is very convoluted here. Please change it. :)

sdroege:
Done

glib/src/subclass/types.rs
```diff
@@ -876,20 +876,54 @@ pub(crate) unsafe fn signal_override_class_handler<F>(
...
+        let res = class_handler(
+            &super::SignalClassHandlerToken(
+                instance as *mut _,
+                return_type.into(),
+                values.as_ptr(),
+            ),
+            values,
+        );
+
+        if return_type == Type::UNIT {
```
GuillaumeGomez, line 896: Considering it's the exact same code as above, do you think it's worth it to write a function where we make this check instead of duplicating the code?

sdroege: As it's just in two places, and those are in different modules, having a function shared between both modules for that seems a bit overkill

GuillaumeGomez: Fine by me!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #381, opened 2021-11-19
glib: Use plain FFI functions instead of a GClosure for the property …

sdroege:
…binding transform functions

Less overhead and less complicated FFI conversion.


GuillaumeGomez:
Code looks good but do you have a regen PR so we can check the result?

sdroege:
What regenerate PR? This doesn't change any API and is not changing gir :)

GuillaumeGomez:
I am very tired... Let me look again correctly. XD

GuillaumeGomez (approved):

PR #384, opened 2021-11-20
Rename various constructors for consistency

sdroege:
See individual commits

### DIFF
--- a/glib/src/date.rs
+++ b/glib/src/date.rs
@@ -26,7 +26,7 @@ unsafe impl Sync for Date {}
 
 impl Date {
     #[doc(alias = "g_date_new_dmy")]
-    pub fn new_dmy(day: DateDay, month: DateMonth, year: DateYear) -> Result<Date, BoolError> {
+    pub fn from_dmy(day: DateDay, month: DateMonth, year: DateYear) -> Result<Date, BoolError> {
         let month = month.into_glib();
         unsafe {
             let check: bool = from_glib(ffi::g_date_valid_dmy(day, month, year));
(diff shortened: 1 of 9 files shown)


glib/src/date.rs
```diff
@@ -26,7 +26,7 @@ unsafe impl Sync for Date {}
 
 impl Date {
```
bilelmoussaoui, line 27: Should be marked as renamed on gir's side as well

glib/src/thread_pool.rs
```diff
@@ -15,7 +15,7 @@ unsafe impl Sync for ThreadPool {}
 
 impl ThreadPool {
     #[doc(alias = "g_thread_pool_new")]
-    pub fn new_shared(max_threads: Option<u32>) -> Result<Self, crate::Error> {
+    pub fn shared(max_threads: Option<u32>) -> Result<Self, crate::Error> {
```
bilelmoussaoui, line 18: Same here

glib/src/date.rs
sdroege: It's not listed at all in `Gir.toml` and if it was it wouldn't have much effect because we don't inject docs in `glib`. There are lots of types missing from `Gir.toml`.

sdroege: Added this one for now

sdroege:
@bilelmoussaoui @bilelmoussaoui PRs for gtkX-rs and gstreamer-rs and gst-plugins-rs are all ready, so this could go in any time.

bilelmoussaoui (approved):

bilelmoussaoui:
Oups, mis-clicked

PR #385, opened 2021-11-20
glib: Fix FromValue impls for &Boxed, &Shared and &Object

sdroege:
This time it's hopefully all correct, for real. Third time!

### DIFF
--- a/glib/src/boxed.rs
+++ b/glib/src/boxed.rs
@@ -254,10 +254,11 @@ macro_rules! glib_boxed_wrapper {
             type Checker = $crate::value::GenericValueTypeOrNoneChecker<Self>;
 
             unsafe fn from_value(value: &'a $crate::Value) -> Self {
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Can you add tests where the `assert` is triggered please?

sdroege:
Which assertion? These assertion in the changes here can only be hit if we get invalid data from C

sdroege:
FWIW the old tests covered the bugs fixed here already. You just had to run them in valgrind.

GuillaumeGomez (approved):

GuillaumeGomez:
Well, let's hope they'll never get triggered then.

GuillaumeGomez:
Thanks!

sdroege:
> Well, let's hope they'll never get triggered then.

If they get triggered we can report a bug to the C library, so that's useful :) It will be clearer with the assertion than if it blows up later in completely strange ways.

PR #386, opened 2021-11-20
glib: Add API for non-destructive iteration over a `List` / `SList`

sdroege:
The `Iterator` implemention on the type itself directly consumes the
list so add another iterator that does not consume the list, and provide
an `iter()` method for getting it.

### DIFF
--- a/glib/src/collections.rs
+++ b/glib/src/collections.rs
@@ -1135,3 +1135,123 @@ impl<
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn slice() {
+        let items = [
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Add tests please.

sdroege:
There's unfortunately no suitable API we could use in GLib. There are tests for this in GStreamer though.

sdroege:
I'll add a test that manually (via FFI) creates a list of datetimes and works with that

sdroege:
@GuillaumeGomez Tests added (also some others), these are all valgrind clean now. Looks ok for you like this?

sdroege:
@GuillaumeGomez CI also passed

GuillaumeGomez (approved):

GuillaumeGomez:
Yes, thanks!

PR #389, opened 2021-11-25
graphene: add getters/setters for Size/Point/Point3D

bilelmoussaoui:


GuillaumeGomez (approved):

bilelmoussaoui:
Could have used a proper debug implementation but can't be done without fixing https://github.com/gtk-rs/gir/issues/1202

sdroege:
> Could have used a proper debug implementation but can't be done without fixing [gtk-rs/gir#1202](https://github.com/gtk-rs/gir/issues/1202)

These are `BoxedInline`, they have no automatic `Debug` impl. So go for it!

bilelmoussaoui:
> > Could have used a proper debug implementation but can't be done without fixing [gtk-rs/gir#1202](https://github.com/gtk-rs/gir/issues/1202)
> 
> These are `BoxedInline`, they have no automatic `Debug` impl. So go for it!

that's sadly not the case https://github.com/gtk-rs/gtk-rs-core/blob/master/graphene/src/auto/size.rs#L7-L17

sdroege:
Oh you can override that, and that's currently explicitly enabled in `Gir.toml`:

```toml
[[object]]
name = "Graphene.Size"
status = "generate"
boxed_inline = true
    [[object.derive]]
    name = "Debug"
```

sdroege:
`Matrix`, `Plane`, `Quaternion` and `Quad` seem to have no getters currently

bilelmoussaoui:
> `Matrix`, `Plane`, `Quaternion` and `Quad` seem to have no getters currently



> `Matrix`, `Plane`, `Quaternion` and `Quad` seem to have no getters currently

Matrix seems to have two getters, `to_float` to get a `[f32; 16]` or `value(col, row) -> f32` .I added a `values` variant though that returns a `[[f32;4];4]`

graphene/src/matrix.rs
```diff
@@ -200,4 +201,41 @@ impl Matrix {
             out.assume_init()
         }
     }
+
+    pub fn values(&self) -> [[f32; 4]; 4] {
+        [
```
bilelmoussaoui, line 206: I don't remember that much about my math classes, i hope i got this right :p

graphene/src/quad.rs
```diff
@@ -59,4 +60,23 @@ impl Quad {
...
         unsafe { from_glib_none(ffi::graphene_quad_get_point(self.to_glib_none().0, index_)) }
     }
+
+    pub fn points(&self) -> [Point; 4] {
+        unsafe {
+            [
+                from_glib_none(&self.0.points[0] as *const _),
+                from_glib_none(&self.0.points[1] as *const _),
+                from_glib_none(&self.0.points[2] as *const _),
+                from_glib_none(&self.0.points[3] as *const _),
```
sdroege, line 70: Why not just `&*(&self.0.points as *const [ffi::graphene_point_t; 4] as *const [Point; 4])`

sdroege:
Looks good to me otherwise

graphene/src/quad.rs
bilelmoussaoui: I wasn't aware of this, cleaned up a few cases like this one as well

graphene/src/box_.rs
```diff
@@ -12,24 +13,15 @@ impl Box {
...
-                from_glib_none(&out[0] as *const _),
-                from_glib_none(&out[1] as *const _),
-                from_glib_none(&out[2] as *const _),
-                from_glib_none(&out[3] as *const _),
-                from_glib_none(&out[4] as *const _),
-                from_glib_none(&out[5] as *const _),
-                from_glib_none(&out[6] as *const _),
-                from_glib_none(&out[7] as *const _),
-            ]
+            *(&out as *const [ffi::graphene_vec3_t; 8] as *const [Vec3; 8])
```
sdroege, line 16: and return a `&[Vec3; 8]` etc. No point in copying the whole thing

graphene/src/box_.rs
```diff
@@ -41,35 +33,35 @@ impl Box {
...
+            let mut b = Self::uninitialized();
             ffi::graphene_box_init_from_box(b.to_glib_none_mut().0, src.to_glib_none().0);
             b
         }
     }
 
     #[doc(alias = "graphene_box_init_from_points")]
     #[doc(alias = "new_from_points")]
-    pub fn from_points(&mut self, points: &[Point3D]) -> Box {
+    pub fn from_points(&mut self, points: &[Point3D]) -> Self {
```
sdroege, line 47: Erm this one is wrong

```suggestion
    pub fn from_points(points: &[Point3D]) -> Self {
```

`self` was not even used.

graphene/src/matrix.rs
```diff
@@ -200,4 +201,41 @@ impl Matrix {
             out.assume_init()
         }
     }
+
+    pub fn values(&self) -> [[f32; 4]; 4] {
```
sdroege, line 205: I think technically you could also return a `&[[f32; 4]; 4]` here.

```rust
&*(&self.value as *const ffi::graphene_simd4x4f_t as *const [[f32; 4]; 4])
```

But that should come with a test please. I think that would be useful to do though as otherwise you'd always copy 16 floats here on every access, which is a bit :)

graphene/src/quad.rs
```diff
@@ -59,4 +60,16 @@ impl Quad {
         assert!(index_ < 4);
         unsafe { from_glib_none(ffi::graphene_quad_get_point(self.to_glib_none().0, index_)) }
     }
+
+    pub fn points(&self) -> [Point; 4] {
+        unsafe { *(&self.0.points as *const [ffi::graphene_point_t; 4] as *const [Point; 4]) }
```
sdroege, line 65:
```suggestion
    pub fn points(&self) -> &[Point; 4] {
        unsafe { &*(&self.0.points as *const [ffi::graphene_point_t; 4] as *const [Point; 4]) }
```

graphene/src/vec2.rs
```diff
@@ -2,35 +2,36 @@
...
+            let mut vec = Self::uninitialized();
             ffi::graphene_vec2_init_from_vec2(vec.to_glib_none_mut().0, src.to_glib_none().0);
             vec
         }
     }
 
     #[doc(alias = "graphene_vec2_init_from_float")]
     #[doc(alias = "new_from_float")]
-    pub fn from_float(src: &[f32; 2]) -> Vec2 {
+    pub fn from_float(src: &[f32; 2]) -> Self {
```
sdroege, line 31: Why is this taking a reference but e.g. in `Triangle` it takes the values directly? Should probably make that consistent. For these things directly the values is probably nicest to use.

graphene/src/vec3.rs
```diff
@@ -2,35 +2,36 @@
...
+            let mut vec = Self::uninitialized();
             ffi::graphene_vec3_init(vec.to_glib_none_mut().0, x, y, z);
             vec
         }
     }
 
     #[doc(alias = "graphene_vec3_init_from_vec3")]
     #[doc(alias = "new_from_vec3")]
-    pub fn from_vec3(src: &Vec3) -> Vec3 {
+    pub fn from_vec3(src: &Vec3) -> Self {
```
sdroege, line 20: This one is a bit strange. That's basically `clone()` :)

graphene/src/vec3.rs
```diff
@@ -2,35 +2,36 @@
...
+            let mut vec = Self::uninitialized();
             ffi::graphene_vec3_init_from_vec3(vec.to_glib_none_mut().0, src.to_glib_none().0);
             vec
         }
     }
 
     #[doc(alias = "graphene_vec3_init_from_float")]
     #[doc(alias = "new_from_float")]
-    pub fn from_float(src: &[f32; 3]) -> Vec3 {
+    pub fn from_float(src: &[f32; 3]) -> Self {
```
sdroege, line 31:
```suggestion
    pub fn from_float(src: [f32; 3]) -> Self {
```

Also here

graphene/src/vec2.rs
sdroege:
```suggestion
    pub fn from_float(src: [f32; 2]) -> Self {
```

graphene/src/vec4.rs
```diff
@@ -4,57 +4,58 @@ use crate::Vec2;
...
+            let mut vec = Self::uninitialized();
             ffi::graphene_vec4_init_from_vec4(vec.to_glib_none_mut().0, src.to_glib_none().0);
             vec
         }
     }
 
     #[doc(alias = "graphene_vec4_init_from_float")]
     #[doc(alias = "new_from_float")]
-    pub fn from_float(src: &[f32; 4]) -> Vec4 {
+    pub fn from_float(src: &[f32; 4]) -> Self {
```
sdroege, line 55:
```suggestion
    pub fn from_float(src: [f32; 4]) -> Self {
```

graphene/src/matrix.rs
```diff
@@ -5,47 +5,48 @@ use crate::Point3D;
...
+            let mut mat = Self::uninitialized();
             ffi::graphene_matrix_init_from_2d(mat.to_glib_none_mut().0, xx, yx, xy, yy, x_0, y_0);
             mat
         }
     }
 
     #[doc(alias = "graphene_matrix_init_from_float")]
     #[doc(alias = "new_from_float")]
-    pub fn from_float(v: &[f32; 16]) -> Matrix {
+    pub fn from_float(v: &[f32; 16]) -> Self {
```
sdroege, line 24:
```suggestion
    pub fn from_float(v: [f32; 16]) -> Self {
```

graphene/src/vec3.rs
sdroege: Other types also have such functions. Not sure how useful it is to keep these :shrug:

bilelmoussaoui: I dropped them, we can re-add them if someone complains

bilelmoussaoui:
Thanks for the review @sdroege, should be updated now :)

bilelmoussaoui:
A ton of methods are not properly marked as manual, so they are not documented :/ I will add that in this PR as well

sdroege (approved):

PR #392, opened 2021-11-28
pango: opt-in for gi-docgen

bilelmoussaoui:


sdroege (approved):
I assume pango is the only one in gtk-rs-core so far?

bilelmoussaoui:
> I assume pango is the only one in gtk-rs-core so far?

Yes, I will test later with pangocairo

PR #401, opened 2021-12-06
Regenerate with latest gir (assert safe gerror)

lucab:
See https://github.com/gtk-rs/gir/pull/1281.


sdroege:
You also need to run `generator.py` and then commit the changes (in a separate commit)

sdroege (approved):

PR #405, opened 2021-12-08
glib: add ObjectSubclassIsExt trait with impl_ method

ranfdev:
So, I haven't tested it, but it doesn't use `unsafe` and it compiles, so I'm _pretty_ sure this just works

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -86,12 +86,13 @@ pub unsafe trait InstanceStruct: Sized + 'static {
     }
 }
 
-/// Trait implemented by `ObjectSubclassIs` to return the private rust struct
+/// Trait implemented by any type implementing `ObjectSubclassIs` to return the implementation, private Rust struct.
 pub trait ObjectSubclassIsExt: ObjectSubclassIs {
     fn impl_(&self) -> &Self::Subclass;
 }
 
 impl<T: ObjectSubclassIs<Subclass = S>, S: ObjectSubclass<Type = Self>> ObjectSubclassIsExt for T {
+    /// Returns the implementation (the private Rust struct) of this class instance
     fn impl_(&self) -> &T::Subclass {
         T::Subclass::from_instance(self)
     }
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -88,11 +88,11 @@ pub unsafe trait InstanceStruct: Sized + 'static {
 
 /// Trait implemented by any type implementing `ObjectSubclassIs` to return the implementation, private Rust struct.
 pub trait ObjectSubclassIsExt: ObjectSubclassIs {
+    /// Returns the implementation (the private Rust struct) of this class instance
     fn impl_(&self) -> &Self::Subclass;
 }
 
 impl<T: ObjectSubclassIs<Subclass = S>, S: ObjectSubclass<Type = Self>> ObjectSubclassIsExt for T {
-    /// Returns the implementation (the private Rust struct) of this class instance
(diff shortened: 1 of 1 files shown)


glib/src/subclass/types.rs
```diff
@@ -86,6 +86,17 @@ pub unsafe trait InstanceStruct: Sized + 'static {
     }
 }
 
+/// Trait implemented by `ObjectSubclassIs` to return the private rust struct
+pub trait ObjectSubclassIsExt: ObjectSubclassIs {
+    fn impl_(&self) -> &Self::Subclass;
```
sdroege, line 91: Can you add some docs here too on the function?

glib/src/subclass/types.rs
```diff
@@ -86,6 +86,17 @@ pub unsafe trait InstanceStruct: Sized + 'static {
     }
 }
 
+/// Trait implemented by `ObjectSubclassIs` to return the private rust struct
```
sdroege, line 89:
```suggestion
/// Trait implemented by any type implementing `ObjectSubclassIs` to return the implementation, private Rust struct.
```

sdroege (commented):
Seems good to me otherwise

GuillaumeGomez (dismissed):

GuillaumeGomez:
Seems good to me, minus the comments from @sdroege. ;)

glib/src/subclass/types.rs
```diff
@@ -86,6 +86,18 @@ pub unsafe trait InstanceStruct: Sized + 'static {
     }
 }
 
+/// Trait implemented by any type implementing `ObjectSubclassIs` to return the implementation, private Rust struct.
+pub trait ObjectSubclassIsExt: ObjectSubclassIs {
+    fn impl_(&self) -> &Self::Subclass;
```
Hofer-Julian, line 91: How about `imp` instead of `impl_`?

ranfdev: Since this method will probably be used a lot when subclassing, I also think `imp` is better.
Now I've called it `impl_` because it does a similar thing to `InstanceStruct::impl_`.

sdroege: If you want to rename it, rename it everywhere :) But `imp` is also a word in English, so not the best choice :imp:

glib/src/subclass/types.rs
```diff
@@ -86,6 +86,18 @@ pub unsafe trait InstanceStruct: Sized + 'static {
...
+}
+
+impl<T: ObjectSubclassIs<Subclass = S>, S: ObjectSubclass<Type = Self>> ObjectSubclassIsExt for T {
+    /// Returns the implementation (the private Rust struct) of this class instance
```
sdroege, line 95: You have to move this to the trait definition, not the impl block

sdroege:
```suggestion
    /// Returns the implementation (the implementation, private Rust struct) of this class instance.
```

glib/src/subclass/types.rs
ranfdev: Nevermind, let's keep `impl_`

glib/src/subclass/types.rs
ranfdev: but why repeat "the implementation"?

ranfdev:
Anyway, that's the code. Please don't tell me again to change the comments. If you want to change them, just do it, I'm ok with it

glib/src/subclass/types.rs
sdroege: Because I failed reading, sorry :)

PR #409, opened 2021-12-10
glib: export Signal

A6GibKm:


sdroege:
Not sure this makes much sense. It's only relevant for subclasses and we intentionally don't re-export any subclass API from the crate root. Instead it's all re-exported into the `glib::subclass` module.

What's the problem with having it there?

A6GibKm:
It is the only struct that I have to manually import as `glib::subclass::Signal`. Can't it be on glib::subclass::prelude either?

bilelmoussaoui:
> It is the only struct that I have to manually import as `glib::subclass::Signal`. Can't it be on glib::subclass::prelude either?

Well you do that right now because you have to manually define signals, if there were a macro... you wouldn't need to do that :)

sdroege:
> Can't it be on glib::subclass::prelude either?

No, only traits in preludes.

A6GibKm:
Closing as there are good reasons not to merge this.

PR #410, opened 2021-12-10
glib: add AnyGObject

ranfdev:
# Motivation:
Rust has a lot of crates with already-defined types. Those types can't be reimplemented as `GObject`s without a huge amount of efforts.
I find the type `AnyGObject` to be extremely useful, especially when putting Rust types inside a `gio::ListModel`

# `SendAnyGObject`
I don't know how to create an `Any + Send + Sync` variant. 
I know how to make a Rust type `Send`, but I don't know how to make a glib type `Send`.


sdroege:
> I know how to make a Rust type `Send`, but I don't know how to make a glib type `Send`.

Let's first get this one complete and then worry about that, but what problems did you run into? Use `glib::SendValue` and `Mutex` instead of `RefCell` and then `unsafe impl Send/Sync for WrapperType { }`.

glib/src/lib.rs
```diff
@@ -59,6 +59,8 @@ pub mod error;
 #[macro_use]
 pub mod object;
 
+pub mod any_gobject;
```
sdroege, line 62:
```suggestion
mod any_gobject;
pub use any_gobject::AnyGObject;
```

glib/src/any_gobject.rs
```diff
@@ -0,0 +1,108 @@
...
+    impl ObjectSubclass for AnyGObject {
+        const NAME: &'static str = "AnyGObject";
+        type Type = super::AnyGObject;
+        type ParentType = Object;
+    }
+    impl ObjectImpl for AnyGObject {}
+}
+
+wrapper! {
+    pub struct AnyGObject(ObjectSubclass<imp::AnyGObject>);
```
sdroege, line 31:
```suggestion
    pub struct BoxedAnyObject(ObjectSubclass<imp::AnyGObject>);
```

Etc IMHO. @bilelmoussaoui ?

glib/src/any_gobject.rs
```diff
@@ -0,0 +1,108 @@
+use crate as glib;
+use crate::subclass::prelude::*;
+use crate::{object_subclass, wrapper, Object};
+use std::any::Any;
+use std::cell::{Ref, RefCell, RefMut};
+
+mod imp {
+    use super::*;
+
+    #[derive(Debug)]
```
sdroege, line 10:
```suggestion
    #[derive(Debug, Default)]
```

glib/src/any_gobject.rs
```diff
@@ -0,0 +1,108 @@
...
+        type ParentType = Object;
+    }
+    impl ObjectImpl for AnyGObject {}
+}
+
+wrapper! {
+    pub struct AnyGObject(ObjectSubclass<imp::AnyGObject>);
+}
+
+impl AnyGObject {
```
sdroege, line 34: Also `impl Default for AnyGObject { ... }`

glib/src/any_gobject.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub fn new(item: Box<dyn Any>) -> Self {
```
sdroege, line 35: Should maybe take an `Option`, also the other API?

sdroege: And maybe the constructor and replace should take a `T: Any + 'static` and all the boxing happens internally?

sdroege:
That's for now what I saw when shortly looking over the code

glib/src/any_gobject.rs
ranfdev: > take an Option


Makes sense



> also the other API?


I don't understand, what's the other API?
I didn't make the `Send` one because I didn't know what to do. But now I know, so I'll do it.

ranfdev: Ok, I guess the "other API" is replace. I've seen your second comment only after replying.

sdroege: Yes

bilelmoussaoui:
The code overall looks pretty fine, but it lacks a usage example. It  could be added as a test or as a documentation

glib/src/any_gobject.rs
```diff
@@ -0,0 +1,108 @@
+use crate as glib;
```
bilelmoussaoui, line 1: it needs the license header as any file

bilelmoussaoui: what's the point of this?
```suggestion
```

glib/src/any_gobject.rs
```diff
@@ -0,0 +1,108 @@
...
+            Self {
+                item: RefCell::new(None),
+            }
+        }
+    }
+    #[object_subclass]
+    impl ObjectSubclass for AnyGObject {
+        const NAME: &'static str = "AnyGObject";
+        type Type = super::AnyGObject;
+        type ParentType = Object;
```
bilelmoussaoui, line 25:
```suggestion
```
it's the default nowadays

glib/src/any_gobject.rs
bilelmoussaoui: Hmm, not sure :) but i'm pretty bad at naming stuff! Maybe just BoxedObject?

ranfdev:
So, I've set all methods to take a type `T` and not an `Option<T>`. Why? Because I think the internal `item` is always going to be set **unless** `BoxedAnyObject` was instantied with `Default::default` or the specific function `new_empty`.

I've tried setting all the methods to take `Option<T>` but if the user puts `Option<T>`, he also expects to get an `Option<T>` when getting the value out (borrowing). That implies I have to return `Option<T>` instead of `T` everywhere, and I've found that becomes a mess.

It's a bit hard to describe a good API for this, because this is neither a `RefCell` wrapper, nor an `Option` wrapper, nor a `Box<dyn Any>` wrapper nor a `GObject` wrapper.

It's a fricking monster wrapping 4 things.

Also, I've introduced the dependency on `thiserror` (which was already present on `gio` and `cairo`, but not `glib`) because the methods `try_borrow*` should return a custom error. `try_borrow*` can fail in too many ways and returning an `Option<T>` is not enough.

ranfdev:
The `clear` method currently doesn't work, but I will fix it (and implement the rest) in the next days.

sdroege:
> I've tried setting all the methods to take `Option<T>` but if the user puts `Option<T>`, he also expects to get an `Option<T>` when getting the value out (borrowing). That implies I have to return `Option<T>` instead of `T` everywhere, and I've found that becomes a mess.

But that also means that various methods would panic if created via `default()` or if `clear()` was called. That doesn't seem ideal.

ranfdev:
Yes, calling `borrow` and `borrow_mut` may panic (for multiple reasons): that's why there's `try_borrow`.
The other methods can't panic

sdroege:
What's the use-case of allowing to create an empty one? That's the only way how you could end up with `None` right now.

Maybe it would be better to let this be a problem of the caller? `Option<T>` can also be stored in here without having every user deal with the `None` case explicitly or implicitly.

ranfdev:
> Option<T> can also be stored in here without having every user deal with the None case explicitly or implicitly.
Well, the type could always be initialized with `Self::default`, so users should have a way to deal with the `None` case without panicking, with `try_borrow`.
If you want I can remove any function for setting the item to `None`: `clear`, `new_empty`.


Anyway, in the meantime I've tried again to make methods take and return `Option<T>`.
For example, I've tried to make `try_borrow_mut` return `Result<RefMut<'_, Option<T>>, BorrowError>` but that's hard, if not impossible.
Explanation:
Given a type `RefCell<Option<Box<T>>>` `RefMut` can give a reference to the type inside
`RefCell`, which is `Option<Box<T>>` or directly to `Box<T>` or `T`.
Returning a `RefMut<Option<T>>` is impossible, because the memory layout contains a `Box` in the middle.

That means even making `borrow` return `Ref<Option<T>>` is impossible, for the same reason: the memory layout contains a `Box` which can't be skipped.
A reference can't selectively skip some parts of the memory.
Well, there are workarounds, but I would need to write 3x the code and probably even use unsafe.

sdroege:
You'd store a `RefCell<Box<Option<T>>>` for that very reason. But what's the use-case for allowing to store a `None` here at all, and shouldn't we just not worry about that and make it the user's problem?

The user can always store an `Option` inside this if it should be an optional value, it's nothing that has to be prescribed on our side.

glib/src/any_gobject.rs
sdroege: That would be a bug in glib-macros then, I guess

sdroege:
> The user can always store an `Option` inside this if it should be an optional value, it's nothing that has to be prescribed on our side.

Just to be clear, my suggestion is:

```rust
impl BoxedAnyObject {
    pub fn new<T: 'static>(item: T) -> Self;

    /// Returns `Err(t)` if the object does not store an `R`.
    pub fn replace<T: 'static, R: 'static>(&self, t: T) -> Result<R, T>;

    /// Fails both if can't borrow or if doesn't store a `T`.
    pub fn try_borrow<'a, T: 'static>(&'a self) -> Result<Ref<'a, T>, BorrowError>;
    pub fn try_borrow_mut<'a, T: 'static>(&'a mut self) -> Result<RefMut<'a, T>, BorrowError>;

    /// Like the ones above plus `unwrap()`.
    pub fn borrow<'a, T: 'static>(&'a self) -> Ref<'a, T>;
    pub fn borrow_mut<'a, T: 'static>(&'a mut self) -> RefMut<'a, T>;
}
```

If the user wants to handle an empty one, they can work with e.g. `BoxedAnyObject::new(None::<String>)` and deal with it accordingly.

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+    // Unset the internal item. Returns true if the internal item was set
+    pub fn clear(&self) -> bool {
+        if self.impl_().item.borrow_mut().is_some() {
+            true
+        } else {
+            false
+        }
+    }
+    // Replaces the wrapped value with a new one, returning the old value, without deinitializing either one.
```
sdroege, line 61: Please use `///` rustdoc comments here and elsewhere :)

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,126 @@
...
+            .as_ref()
+            .ok_or(BorrowError::IsNone)?
+            .downcast_ref::<T>()
+            .ok_or(BorrowError::InvalidType)?;
+        Ok(self.borrow()) // Now this won't panic
+    }
+
+    // Mutably borrows the wrapped value, returning none if the value is currently immutably borrowed or
+    // the inner value has never been seet
+    pub fn try_borrow_mut<'a, T: 'static>(&'a mut self) -> Result<RefMut<'a, T>, BorrowError> {
```
sdroege, line 89: The `&mut` here is meaningless because `GObject`s are refcounted. It's also not useful, borrowing a `RefCell` content mutably doesn't require a `&mut` reference to the refcell.

ranfdev:
> But what's the use-case for allowing to store a None here at all

I thought I needed to put `None` to initialize the type in the `default` function, but you are right, I can just Initialize the inner item to a `Box<None>` (therefore removing the outer `Option`)!

sdroege:
> I thought I needed to put `None` to initialize the type in the `default` function, but you are right, I can just Initialize the inner item to a `Box<None>` (therefore removing the outer `Option`)!

Not sure why you need that. You'd store an `Option<RefCell<Box<dyn Any>>>>` inside the instance (or `OnceCell` instead of `Option`), and then assume that after construction it is always not `None`.

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+        obj
+    }
+
+    /// Replaces the wrapped value with a new one, returning the old value, without deinitializing either one.
+    pub fn replace<T: 'static>(&self, t: T) -> Box<dyn Any> {
+        self.impl_().item.replace(Box::new(t) as Box<dyn Any>)
+    }
+
+    /// Immutably borrows the wrapped value, returning none if the value is currently mutably borrowed or
+    /// the inner value has never been seet
```
sdroege, line 83:
```suggestion
    /// the inner value has never been set
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+        let borrowed = self.impl_().item.try_borrow()?;
+        borrowed
+            .as_ref()
+            .downcast_ref::<T>()
+            .ok_or(BorrowError::InvalidType)?;
+        Ok(self.borrow()) // Now this won't panic
+    }
+
+    /// Mutably borrows the wrapped value, returning none if the value is currently immutably borrowed or
+    /// the inner value has never been seet
```
sdroege, line 98: As above

sdroege: And also below

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// or if the conversion to `T` fails
+    pub fn borrow<T: 'static>(&self) -> Ref<'_, T> {
+        Ref::map(self.impl_().item.borrow(), |item| {
+            item.as_ref().downcast_ref::<T>().unwrap()
+        })
+    }
+    /// Mutably borrows the wrapped value. The value cannot be borrowed while this borrow is active.
+    ///
+    /// # Panics
+    /// Panics if the value is currently borrowed or if the conversion to `T` fails
```
sdroege, line 127:
```suggestion
    /// Panics if the value is currently borrowed or if the conversion to `T` fails.
```

Also elsewhere

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+        drop(borrowed_mut);
+        Ok(self.borrow_mut()) // Now this won't panic
+    }
+
+    /// Immutably borrows the wrapped value. Multiple immutable borrows can be taken out at
+    /// the same time.
+    ///
+    /// # Panics
+    /// Panics if the value is currently mutably borrowed or the inner value has never been set
+    /// or if the conversion to `T` fails
```
sdroege, line 118: There is no "conversion to `T`", either it is a `T` or it's the wrong type

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+            Self {
+                item: RefCell::new(Box::new(None::<usize>)),
+            }
+        }
+    }
+    impl ObjectImpl for BoxedAnyObject {}
+}
+
+wrapper! {
+    /// # Examples
```
sdroege, line 42: Above the example, add some explanation/description what this type is good for

sdroege:
Looks good to me otherwise, thanks! @GuillaumeGomez or @bilelmoussaoui might have some more opinions

glib/Cargo.toml
```diff
@@ -32,6 +32,7 @@ gobject_ffi = { package = "gobject-sys", path = "gobject-sys" }
 glib-macros = { path = "../glib-macros" }
 rs-log = { package = "log", version = "0.4", optional = true }
 smallvec = "1.0"
+thiserror = "1"
```
GuillaumeGomez, line 35: I strongly vote against this. The gtk-rs crates are already big enough to not add more dependencies.

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// ```rust
```
GuillaumeGomez, line 43: No need for `rust`, it's already the default.

glib/src/boxed_any_object.rs
GuillaumeGomez: And also please add a dot at the end of the sentence.

glib/src/boxed_any_object.rs
GuillaumeGomez: And the dot at the end too.

glib/Cargo.toml
bilelmoussaoui: Other crates of gtk-rs seems to use it already iirc

glib/src/boxed_any_object.rs
ranfdev: Ups, that line shouldn't even be there, since `Option` is not a problem anymore. There's always some value inside...
I have to remember Rust checks the code, but not the comments 😆

glib/Cargo.toml
sdroege: `cairo` and `gio` and `gstreamer`

ranfdev: Mmmm, that's already used by gio and cairo, so it's not really a new dependency

GuillaumeGomez: I'll remove it from these crates. Please remove it.

sdroege: Good luck getting the `Error` and `Display` impls right then :) I wonder how often we'll have this very discussion

GuillaumeGomez: We have very different positions on this unfortunately. I don't think we'll ever agree. As long as I provide the PRs, I guess it's fine? ;)

Did this one already: https://github.com/gtk-rs/gtk-rs-core/pull/415

ranfdev:
I hope this is the good one

glib/Cargo.toml
sdroege: You can keep it here FWIW

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// let list = gio::ListStore::new(BoxedAnyObject::static_type());
+    /// list.append(boxed.clone().upcast());
+    ///
+    /// // And can be retrieved with `borrow`
+    /// let author: Author = boxed.borrow();
+    /// ```
+    pub struct BoxedAnyObject(ObjectSubclass<imp::BoxedAnyObject>);
+}
+
+impl Default for BoxedAnyObject {
```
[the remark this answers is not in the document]
sdroege: If you insist on keeping this, please at least document what the object will contain after construction :) I think it will be mostly confusing though.

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// list.append(boxed.clone().upcast());
+    ///
+    /// // And can be retrieved with `borrow`
+    /// let author: Author = boxed.borrow();
+    /// ```
+    pub struct BoxedAnyObject(ObjectSubclass<imp::BoxedAnyObject>);
+}
+
+impl BoxedAnyObject {
+    pub fn new<T: 'static>(value: T) -> Self {
```
sdroege, line 76: Needs some docs

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+            Self {
+                value: RefCell::new(Box::new(None::<usize>)),
+            }
+        }
+    }
+    impl ObjectImpl for BoxedAnyObject {}
+}
+
+wrapper! {
+    /// This is a type subclassing `glib::object::Object` capable of storing any Rust type.
```
sdroege, line 47:
```suggestion
    /// This is a subclass of `glib::object::Object` capable of storing any Rust type.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// It let's you insert a Rust type anywhere a `glib::object::Object` is needed.
+    /// The inserted value can then be borrowed as a Rust type, by using the various
+    /// included methods.
```
sdroege, line 50:
```suggestion
    /// provided methods.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl BoxedAnyObject {
+    pub fn new<T: 'static>(value: T) -> Self {
+        let obj: Self = Object::new(&[]).expect("Failed to create BoxedAnyObject");
+        obj.replace(value);
+        obj
+    }
+
+    /// Replaces the wrapped value with a new one, returning the old value, without deinitializing either one.
+    /// The returned value is inside a Box and must be manually downcasted.
```
sdroege, line 83:
```suggestion
    /// The returned value is inside a `Box` and must be manually downcasted if needed.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn replace<T: 'static>(&self, t: T) -> Box<dyn Any> {
+        self.impl_().value.replace(Box::new(t) as Box<dyn Any>)
+    }
+
+    /// Immutably borrows the wrapped value, returning an error if the value is currently mutably
+    /// borrowed or if it's not of type T.
```
sdroege, line 89:
```suggestion
    /// borrowed or if it's not of type `T`.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+        let borrowed = self.impl_().value.try_borrow()?;
+        borrowed
+            .as_ref()
+            .downcast_ref::<T>()
+            .ok_or(BorrowError::InvalidType)?;
+        Ok(self.borrow()) // Now this won't panic
+    }
+
+    /// Mutably borrows the wrapped value, returning an error if the value is currently borrowed.
+    /// or if it's not of type T.
```
sdroege, line 109:
```suggestion
    /// or if it's not of type `T`.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+    ///
+    /// The borrow lasts until the returned `RefMut` or all `RefMut`s derived
+    /// from it exit scope. The value cannot be borrowed while this borrow is
```
sdroege, line 112:
```suggestion
    /// from it exit scope. The value cannot be borrowed another time while this borrow is
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Immutably borrows the wrapped value.
+    ///
+    /// The borrow lasts until the returned `Ref` exits scope. Multiple
+    /// immutable borrows can be taken out at the same time.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value is currently mutably borrowed or if it's not of type T.
```
sdroege, line 137:
```suggestion
    /// Panics if the value is currently mutably borrowed or if it's not of type `T`.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Mutably borrows the wrapped value.
+    ///
+    /// The borrow lasts until the returned `RefMut` or all `RefMut`s derived
+    /// from it exit scope. The value cannot be borrowed while this borrow is
+    /// active.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the value is currently borrowed or if it's not of type T.
```
sdroege, line 155:
```suggestion
    /// Panics if the value is currently borrowed or if it's not of type `T`.
```

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn borrow<T: 'static>(&self) -> Ref<'_, T> {
+        Ref::map(self.impl_().value.borrow(), |value| {
+            value.as_ref().downcast_ref::<T>().unwrap()
+        })
+    }
+
+    /// Mutably borrows the wrapped value.
+    ///
+    /// The borrow lasts until the returned `RefMut` or all `RefMut`s derived
+    /// from it exit scope. The value cannot be borrowed while this borrow is
```
sdroege, line 150:
```suggestion
    /// from it exit scope. The value cannot be borrowed another time while this borrow is
```

sdroege (approved):
Looks good to me apart from the docs issues. @bilelmoussaoui?

Do you also want to do the thread-safe version as part of this PR?

glib/src/boxed_any_object.rs
ranfdev: I'm not gonna change this phrase, since it's copied from `std::cell::RefMut` xD

sdroege: Ok :)

sdroege (dismissed):

ranfdev:
Ok, sorry if I pushed too much commits. I'm done. To me, the PR is ready to be merged

sdroege (dismissed):

sdroege:
> To me, the PR is ready to be merged

For me too :)

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,165 @@
...
+            Self {
+                value: RefCell::new(Box::new(None::<usize>)),
+            }
+        }
+    }
+    impl ObjectImpl for BoxedAnyObject {}
+}
+
+wrapper! {
+    /// This is a subclass of `glib::object::Object` capable of storing any Rust type.
```
bilelmoussaoui, line 47:
```suggestion
    // rustdoc-stripper-ignore-next
    /// This is a subclass of `glib::object::Object` capable of storing any Rust type.
```

All the documentations comments, should be prefixed with rustdoc-stripper one. Otherwise they get removed with `python3 generate.py --strip-docs`. Normally we don't test that as part of the CI yet, but we will be doing so soon enough

sdroege: Not a problem in glib btw, it doesn't use `rustdoc-stripper`

bilelmoussaoui: > Not a problem in glib btw, it doesn't use `rustdoc-stripper`

It definitely does :)

sdroege: Hrm, but we don't have the ignore-next before basically any of the docs :)

bilelmoussaoui: If you rebase your changes on top of the master branch, the CI would properly complain now

ranfdev:
I've rebased to master to get the updated CI

ranfdev:
I had rebased on an old master 🤦🏼 ...

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,172 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate as glib;
```
bilelmoussaoui, line 3: Unrelated to this merge request, but please open an issue regarding the limitation /bug of glib-macros

bilelmoussaoui:
otherwise looks good to me, thanks!

sdroege (dismissed):

sdroege:
Needs a `cargo fmt` run

sdroege:
@GuillaumeGomez @bilelmoussaoui Hm why do the doctests only fail on Windows? :)

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///     subscribers: usize
+    /// }
+    /// // BoxedAnyObject can contain any custom type
+    /// let boxed = BoxedAnyObject::new(Author {
+    ///     name: String::from("GLibAuthor"),
+    ///     subscribers: 1000
+    /// });
+    ///
+    /// // The boxed data can be stored as a `glib::object::Object`
+    /// let list = gio::ListStore::new(BoxedAnyObject::static_type());
```
sdroege, line 67: This can't compile of course as we're in `glib` and `gio` can't have been compiled so far. @GuillaumeGomez what to do here? Just `no_compile`?

bilelmoussaoui:
> @GuillaumeGomez @bilelmoussaoui Hm why do the doctests only fail on Windows? :)

Something weird happened here https://github.com/gtk-rs/gtk-rs-core/blob/master/.github/workflows/CI.yml#L77, the first shouldn't be a build :/

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,178 @@
...
+
+wrapper! {
+    // rustdoc-stripper-ignore-next
+    /// This is a subclass of `glib::object::Object` capable of storing any Rust type.
+    /// It let's you insert a Rust type anywhere a `glib::object::Object` is needed.
+    /// The inserted value can then be borrowed as a Rust type, by using the various
+    /// provided methods.
+    ///
+    /// # Examples
+    /// ```
```
GuillaumeGomez, line 54:
```suggestion
    /// ```no_run
```

sdroege: It won't compile, not just not run :)

bilelmoussaoui:
> > @GuillaumeGomez @bilelmoussaoui Hm why do the doctests only fail on Windows? :)
> 
> Something weird happened here https://github.com/gtk-rs/gtk-rs-core/blob/master/.github/workflows/CI.yml#L77, the first shouldn't be a build :/

can you create an issue for this? will handle it tomorrow

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// use glib::BoxedAnyObject;
```
sdroege, line 55:
```suggestion
    /// use glib::BoxedAnyObject;
    /// use glib::prelude::*;
```

Or `static_type()` won't be in scope

glib/src/boxed_any_object.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///     name: String::from("GLibAuthor"),
+    ///     subscribers: 1000
+    /// });
+    ///
+    /// // The boxed data can be stored as a `glib::object::Object`
+    /// let list = gio::ListStore::new(BoxedAnyObject::static_type());
+    /// list.append(boxed.clone().upcast());
+    ///
+    /// // And can be retrieved with `borrow`
+    /// let author: Author = boxed.borrow();
```
sdroege, line 71:
```suggestion
    /// let author: Ref<Author> = boxed.borrow();
```

or

```suggestion
    /// let author: Author = boxed.borrow().clone();
```

ranfdev:
I've split the example in two parts: the first part compiles without problems, the second part is ignored because it requires `gio` but that's not a dependency of `glib`, so it will never compile.

Ready to be merged

sdroege (approved):

PR #412, opened 2021-12-11
glib: ignore various unneeded functions

### LINKED ISSUE #43: Remove useless Rust functions
As stated in gtk-rs/glib#464 by @sdroege, we generate a lot of functions which are useless in Rust. We should make a global review about them and remove anything that shouldn't be there.

bilelmoussaoui:
Kind of fixes #43 

There are still some stuff that could be double checked like the various logging functions but in general, it cleans up a huge part of them already


GuillaumeGomez (approved):

sdroege:
There are also various functions in `gio` FWIW :)

glib/Gir.toml
```diff
@@ -51,17 +51,38 @@ manual = [
...
-    name = "set_error_literal"
-    #empty first parameter
-    ignore = true
+    pattern = "(malloc|malloc0|malloc0_n|malloc_n|memdup|memdup2|realloc|realloc_n|try_malloc||try_malloc0|try_malloc0_n|try_malloc_n|try_realloc|try_realloc_n|free)"
+    ignore = true # Not useful allocations functions
+    [[object.function]]
+    pattern = "(slice_get_config|slice_get_config_state|slice_set_config|slice_alloc|slice_alloc0|slice_copy|slice_free1|slice_free_chain_with_offset)"
+    ignore = true # Not useful slices functions
+    [[object.function]]
+    pattern = "(path_is_absolute|build_filename|build_filename_valist|build_filenamev|build_path|build_pathv|canonicalize_filename|chdir)"
```
sdroege, line 66: You might also want to ignore then:
  - `g_path_.*`
  - `g_basename`
  - `g_canonicalize_filename`

glib/Gir.toml
```diff
@@ -51,17 +51,38 @@ manual = [
...
+    ignore = true # Not useful paths functions, use Path/PathBuf
+    [[object.function]]
+    pattern = "(int64_equal|int64_hash|int_equal|int_hash|double_equal|double_hash|direct_equal|direct_hash)"
+    ignore = true # Not useful equal/hash functions, only useful for GHashTable
+    [[object.function]]
+    pattern = "(clear_handle_id|clear_list|clear_pointer|clear_slist|nullify_pointer)"
```
sdroege, line 72: You can fold `g_clear_error` into here

bilelmoussaoui:
> There are also various functions in `gio` FWIW :)

All of them looks useful to me, they are either dbus or content-types

glib/Gir.toml
```diff
@@ -51,17 +51,38 @@ manual = [
...
+    ignore = true # Not useful error handling functions
+    [[object.function]]
+    pattern = "(idle_remove_by_data|qsort_with_data)"
+    ignore = true # Unsafe functions
+    [[object.function]]
+    pattern = "(iconv|convert_with_iconv)"
+    ignore = true # not usable
     [[object.function]]
     pattern = "atomic_.+"
     ignore = true
```
sdroege, line 88: Other functions
 - `g_access` (that's in `std` already)
 - `g_base64_decode` (this should really be manual and return a `glib::Slice<u8>` that's just silly otherwise)
 - `g_base64_.*_(step|inplace|close)` (these need manual bindings if we want such a thing)
 - `g_bit_*` (`std`)
 - `g_compute_.*_for_string` (should be manually implemented because this one just silly)
 - `g_file_open_tmp` (the returned `fd` should be used with the corresponding type from `std`)
 - `g_file_read_link` (`std`)
 - `g_file_test` (`std`)
 - `g_get_current_time` (deprecated and there's a better function)
 - `g_io_.*` (needs manual bindings but it's also awful API)
 - `g_log_.*` (needs manual bindings if we want those to be useful)
 - `g_markup_.*` (these are together with `GMarkupParser` which we don't generate)
 - `g_parse_debug_string` (needs manual bindings but also rather useless)
 - `g_pattern_.*` (pattern matching functions, we have a better API for that in `std`)
 - `g_pointer_.*` (raw pointers...)
 - `g_poll` (manual bindings but there are crates doing this so...)
 - `g_rmdir` (`std`)

While we're at it, also please generate the `GUnicodeScript` enum then we get a few more useful functions here.

glib/Gir.toml
```diff
@@ -51,17 +51,38 @@ manual = [
...
+    pattern = "(propagate_prefixed_error|set_error|file_error_from_errno|clear_error|propagate_error|set_error_literal|set_error)"
+    ignore = true # Not useful error handling functions
+    [[object.function]]
+    pattern = "(idle_remove_by_data|qsort_with_data)"
+    ignore = true # Unsafe functions
+    [[object.function]]
+    pattern = "(iconv|convert_with_iconv)"
+    ignore = true # not usable
     [[object.function]]
     pattern = "atomic_.+"
```
sdroege, line 87: For gio
 - `g_io_scheduler_.*` (dangerous and deprecated)
 - `g_networking_init` (useless, `std` does that already)

Otherwise seems fine but while we're at it you can also generate `GIOModuleScope` to get some more useful API generated.

bilelmoussaoui: > GIOModuleScope 

It can't be auto-generated because it's a boxed type that lacks both a type function and it can't easily be manually implemented because it doesn't provide a copy function.  What should we do here?

glib/Gir.toml
bilelmoussaoui: - [x] `g_access` (that's in `std` already)
 - `g_base64_decode` (this should really be manual and return a `glib::Slice<u8>` that's just silly otherwise)
 - `g_base64_.*_(step|inplace|close)` (these need manual bindings if we want such a thing)

Regarding g_base64, those are rather useful as is, they involve unnecessary copies but they are not fundamentally broken. So I guess let's keep them until someone is interested in improving them in the future?

 - [x] `g_bit_*` (`std`)
 - `g_compute_.*_for_string` (should be manually implemented because this one just silly)

What's not great on the current implementation as is? 

 - [x] `g_file_open_tmp` (the returned `fd` should be used with the corresponding type from `std`)
 
 I've manually implemented it for unix only, no idea how that would look like for windows.
 
 - [x] `g_file_read_link` (`std`)
 - [x] `g_file_test` (`std`)
 - [x] `g_get_current_time` (deprecated and there's a better function)
 - [x] `g_io_.*` (needs manual bindings but it's also awful API)
 - `g_log_.*` (needs manual bindings if we want those to be useful)
 
 As there's a possibility to make a nicer wrapper around them with manual bindings, let's leave those as is so they are still generated as "ignored" on the code?
 
 - [x] `g_markup_.*` (these are together with `GMarkupParser` which we don't generate)
 - `g_parse_debug_string` (needs manual bindings but also rather useless)

 Same as others that require manual bindings

 - [x] `g_pattern_.*` (pattern matching functions, we have a better API for that in `std`)
 - [x] `g_pointer_.*` (raw pointers...)
 
 - `g_poll` (manual bindings but there are crates doing this so...)
 
 Would leave this as is, in case someone wants to fix it someday

 - [x] `g_rmdir` (`std`)

sdroege: > What's not great on the current implementation as is?

`g_compute_.*_for_string` creates a temporary copy of the string in the bindings, while `compute_for_data(a_string.as_bytes())` wouldn't. Having the string function there is misleading because it looks equally cheap but definitely isn't.

> I've manually implemented it for unix only, no idea how that would look like for windows.

I think `std` has no fd API on Windows, you'd have to go through `HANDLE` somehow.

> As there's a possibility to make a nicer wrapper around them with manual bindings, let's leave those as is so they are still generated as "ignored" on the code?

You're generating some of the logging functions but they're useless without the remainder

> Same as others that require manual bindings

Well, `g_parse_debug_string` is also useless :)

glib/Gir.toml
sdroege: Ah, let's ignore that then for now until someone needs it.

glib/Gir.toml
sdroege: `GFileTest` enum can be ignored now

sdroege: @bilelmoussaoui What's the status here?

bilelmoussaoui: Will get back to it in a bit, finishing another "cleanup" merge request for gtk-rs-core first :)

bilelmoussaoui: > GFileTest enum can be ignored now

Done

> Well, g_parse_debug_string is also useless :)

Dropped

> I think std has no fd API on Windows, you'd have to go through HANDLE somehow.

As I've no idea how anything windows related works, I will leave that to someone that needs it.

> You're generating some of the logging functions but they're useless without the remainder

Done

glib/src/auto/functions.rs
```diff
@@ -249,6 +250,17 @@ pub fn file_set_contents_full(
     }
 }
 
+#[doc(alias = "g_file_test")]
+#[allow(dead_code)]
+pub(crate) fn file_test(filename: impl AsRef<std::path::Path>, test: FileTest) -> bool {
```
sdroege, line 255: Isn't this proper public now via the `pub use auto::functions::*` in `lib.rs`?

bilelmoussaoui: Not really, hence the allow dead_code but I will double check

sdroege: Hm I thought that's how you'd change the visibility of items to something else, by re-exporting from a place where it is accessible :)

sdroege (approved):

PR #413, opened 2021-12-14
allow for the modifying GlyphInfo and GlyphGeometry

bvinc:
From what I can understand about this API,  The GlyphGeometry objects inside of a GlyphString should be modifiable.

To make this work, its GlyphInfo's should not be a newly created vector, but rather a slice created from its internal pointer and length.

I believe this change brings it in line with how the API is meant to be used in C.

### DIFF
--- a/pango/src/glyph_info.rs
+++ b/pango/src/glyph_info.rs
@@ -14,7 +14,11 @@ impl GlyphInfo {
     }
 
     pub fn geometry(&self) -> &GlyphGeometry {
-        unsafe { &*(&((self.0).geometry) as *const _ as *const GlyphGeometry) }
+        unsafe { &*(&self.0.geometry as *const _ as *const GlyphGeometry) }
+    }
+
+    pub fn geometry_mut(&mut self) -> &mut GlyphGeometry {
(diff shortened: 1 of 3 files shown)


sdroege:
Can you show some example usage from some application in C where these are modified?

bvinc:
Here's an example.

https://gitlab.gnome.org/GNOME/gtk/-/blob/main/demos/gtk-demo/fontrendering.c#L256

bilelmoussaoui (approved):

pango/src/glyph_info.rs
```diff
@@ -14,7 +14,11 @@ impl GlyphInfo {
     }
 
     pub fn geometry(&self) -> &GlyphGeometry {
-        unsafe { &*(&((self.0).geometry) as *const _ as *const GlyphGeometry) }
+        unsafe { &*(&(self.0.geometry) as *const _ as *const GlyphGeometry) }
```
sdroege, line 17:
```suggestion
        unsafe { &*(&self.0.geometry as *const _ as *const GlyphGeometry) }
```

And also elsewhere

sdroege (dismissed):
Looks good to me otherwise. Would be great to port that example into gtk4-rs/examples :)

sdroege (approved):

PR #415, opened 2021-12-14
Remove thiserror dependency

GuillaumeGomez:


cairo/src/error.rs
```diff
@@ -1,143 +1,156 @@
...
-    #[error("LastStatus")]
     #[doc(alias = "STATUS_LAST_STATUS")]
     LastStatus,
-    #[error("Unknown {0}")]
     #[doc(hidden)]
     __Unknown(i32),
 }
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
sdroege, line 99: This is the default implementation already

gio/src/input_stream.rs
```diff
@@ -525,12 +526,27 @@ impl<T: IsA<InputStream>> InputStreamAsyncBufRead<T> {
...
+#[derive(Debug)]
 enum BufReadError {
-    #[error("Previous read operation failed")]
     Failed,
 }
 
+impl std::error::Error for BufReadError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        match self {
+            Self::Failed => None,
```
sdroege, line 537: This makes no sense, that's the default implementation

sdroege (commented):
So much boilerplate and much harder to review code. The original was obviously correct, yours is not.

GuillaumeGomez:
> The original was obviously correct, yours is not.

Which part is invalid?

gio/src/input_stream.rs
GuillaumeGomez: No, this is needed.

cairo/src/error.rs
GuillaumeGomez: No, this is needed, or I misunderstood what you meant.

gio/src/input_stream.rs
GuillaumeGomez: Ok nevermind, I didn't realize I removed the `derive(Error)`...

GuillaumeGomez:
Apparently not, we have to implement `std::error::Error` by hand.

cairo/src/error.rs
bilelmoussaoui: You don't have to provide the source method if you're going to return None. That's the default implementation

sdroege: Apparently implementing `Error` is complicated enough that you missed this once when implementing it and a second time after I pointed it out :shrug:

sdroege:
> Which part is invalid?

I don't know, I didn't see any further errors that you introduced but the point is that it's absolutely obvious if there's anything wrong in the `thiserror` usage while you now added some 100 lines of code that has to be carefully reviewed.

Hofer-Julian:
@GuillaumeGomez did you measure how much faster this crate compiles with your branch?

cairo/src/error.rs
GuillaumeGomez: Sorry...

GuillaumeGomez:
> @GuillaumeGomez did you measure how much faster this crate compiles with your branch?

A few seconds.

sdroege:
> A few seconds.

You're not measuring correctly then. We still depend on `thiserror` indirectly, check your `Cargo.lock`.

39.91s for a clean build from master, 41.03s with your branch for a clean build here. In theory your branch should be minimally faster, so I guess they're both equally fast and any difference is in the margin of measurement errors.

GuillaumeGomez:
It took 2 min for me, so 3 secs is in the margin.

sdroege:
Yeah my point is that a) we still depend on `thiserror` anyway so you still compile it, b) if you try often enough both builds and average you'll see that they're both approximately the same. This has no effect other than adding more code.

GuillaumeGomez:
Absolutely. However this is more readable and I know what's happening just by reading it unlike `thiserror`.

sdroege:
It's absolutely not more readable. It's just a lot of repetitive code where some silly mistake can easily be hidden :)

GuillaumeGomez:
Let's close it for now then...

PR #421, opened 2021-12-16
gio: impl std::iter::Extend for gio::ListStore

### LINKED ISSUE #417: [FEATURE REQUEST] Implement Extend trait on ListModels
https://doc.rust-lang.org/std/iter/trait.Extend.html

It's nice being able to do
```rust
model.extend(iterator_of_data_from_api);
```

ranfdev:
closes #417

### DIFF
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -59,6 +59,12 @@ impl ListStore {
             );
         }
     }
+
+    // rustdoc-stripper-ignore-next
+    /// Appends all elements in a slice to the `ListStore`.
+    pub fn extend_from_slice(&self, additions: &[impl IsA<glib::Object>]) {
+        self.splice(self.n_items() - 1, 0, additions)
+    }
 }
 
 unsafe extern "C" fn compare_func_trampoline(
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -1,5 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use crate::auto::traits::ListModelExt;
 use crate::ListStore;
 use glib::translate::*;
 use glib::{IsA, Object};
@@ -74,4 +75,24 @@ unsafe extern "C" fn compare_func_trampoline(
     let b = from_glib_borrow(b as *mut glib::gobject_ffi::GObject);
 
     (*func)(&a, &b).into_glib()
+}
+
+impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
+    fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
+        unsafe {
+            let additions = iter
+                .into_iter()
+                .map(|o| o.as_ptr() as *mut glib::gobject_ffi::GObject)
+                .collect::<Vec<_>>();
+            let n_additions = additions.len() as u32;
+
+            ffi::g_list_store_splice(
+                self.to_glib_none().0,
+                self.n_items() - 1,
+                0,
+                mut_override(additions.as_ptr()),
+                n_additions,
+            );
+        }
+    }
 }
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -80,12 +80,9 @@ unsafe extern "C" fn compare_func_trampoline(
     (*func)(&a, &b).into_glib()
 }
 
-impl<A: AsRef<glib::Object>> std::iter::Extend<A> for ListStore {
+impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
     fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
-        let additions = iter
-            .into_iter()
-            .map(|a| a.as_ref().clone())
-            .collect::<Vec<_>>();
+        let additions = iter.into_iter().collect::<Vec<_>>();
         self.splice(self.n_items(), 0, &additions)
     }
 }
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -44,20 +44,28 @@ impl ListStore {
         }
     }
 
+    /// Dumbest possible example just to test this. Will remove.
+    /// ```
+    /// use glib::prelude::*;
+    /// use gio::ListStore;
+    /// use gio::prelude::*;
+    /// let item1 = ListStore::new(ListStore::static_type());
+    /// let item2 = ListStore::new(ListStore::static_type());
+    /// let list = ListStore::new(ListStore::static_type());
+    /// list.splice(0, 0, &[item1, item2.clone()]);
+    /// assert!(list.item(1) == Some(item2.upcast()))
+    /// ```
     #[doc(alias = "g_list_store_splice")]
     pub fn splice(&self, position: u32, n_removals: u32, additions: &[impl IsA<glib::Object>]) {
         let n_additions = additions.len() as u32;
         unsafe {
-            let additions = additions
-                .iter()
-                .map(|o| o.as_ptr() as *mut glib::gobject_ffi::GObject)
-                .collect::<Vec<_>>();
+            let additions = additions.as_ptr() as *mut *mut glib::gobject_ffi::GObject;
 
             ffi::g_list_store_splice(
                 self.to_glib_none().0,
                 position,
                 n_removals,
-                mut_override(additions.as_ptr()),
+                additions,
                 n_additions,
             );
         }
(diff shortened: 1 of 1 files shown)


gio/src/list_store.rs
```diff
@@ -75,3 +76,23 @@ unsafe extern "C" fn compare_func_trampoline(
 
     (*func)(&a, &b).into_glib()
 }
+
+impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
```
sdroege, line 80: Maybe also for `std::iter::Extend<&A>`?

gio/src/list_store.rs
```diff
@@ -75,3 +76,23 @@ unsafe extern "C" fn compare_func_trampoline(
...
+    fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
+        unsafe {
+            let additions = iter
+                .into_iter()
+                .map(|o| o.as_ptr() as *mut glib::gobject_ffi::GObject)
+                .collect::<Vec<_>>();
+            let n_additions = additions.len() as u32;
+
+            ffi::g_list_store_splice(
```
sdroege, line 89: Why via FFI instead of https://gtk-rs.org/gtk-rs-core/git/docs/gio/struct.ListStore.html#method.splice?

gio/src/list_store.rs
```diff
@@ -75,3 +76,23 @@ unsafe extern "C" fn compare_func_trampoline(
 
     (*func)(&a, &b).into_glib()
 }
+
+impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
+    fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
+        unsafe {
+            let additions = iter
+                .into_iter()
+                .map(|o| o.as_ptr() as *mut glib::gobject_ffi::GObject)
```
sdroege, line 85: This is a use-after-free: nothing is keeping the original `o` alive, you're just keeping the pointer

gio/src/list_store.rs
ranfdev: Because splice takes `&[A]`, then collects the items in a `Vec`.
To call `splice` with a `&[A]` I have to collect the iterator's items in a `Vec`.

Calling splice would allocate a `Vec` two times.

sdroege: Yes that's needed actually :) See https://github.com/gtk-rs/gtk-rs-core/pull/421#discussion_r770784002

sdroege: Or rather: you can collect all objects into a `Vec<glib::Object>` and then pass it to the FFI function if you want, without another vec. `your_vec.as_ptr() as *const _ as *const *const glib::gobject_ffi::GObject` is valid.

If that's also valid to do for `splice()`, not sure. I'd have to think about that a bit (please create an issue).

gio/src/list_store.rs
ranfdev: Right, `splice` doesn't have this problems because it takes an array of references, which live longer than the function...
I'll edit the function to take an iterator of references

sdroege: I think both implementations can be useful. See my other comment, it doesn't make much of a difference: you need to collect a single vec in either case.

gio/src/list_store.rs
ranfdev: Yup, I'm now convinced I need to allocate 2 times anyway

ranfdev:
`impl IsA<glib::Object>` is still just a pointer so this works!

gio/src/list_store.rs
```diff
@@ -43,20 +44,28 @@ impl ListStore {
...
+    /// Dumbest possible example just to test this. Will remove.
+    /// ```
+    /// use glib::prelude::*;
+    /// use gio::ListStore;
+    /// use gio::prelude::*;
+    /// let item1 = ListStore::new(ListStore::static_type());
+    /// let item2 = ListStore::new(ListStore::static_type());
+    /// let list = ListStore::new(ListStore::static_type());
+    /// list.splice(0, 0, &[item1, item2.clone()]);
+    /// assert!(list.item(1) == Some(item2.upcast()))
```
sdroege, line 56: Can you put that into a `#[test]` in this module? That way we can make sure it continues to work :)

sdroege:
Looks good to me otherwise, thanks! Please also add the `&A` variant of the trait impl though

gio/src/list_store.rs
```diff
@@ -43,20 +44,28 @@ impl ListStore {
...
+    /// ```
     #[doc(alias = "g_list_store_splice")]
     pub fn splice(&self, position: u32, n_removals: u32, additions: &[impl IsA<glib::Object>]) {
```
sdroege, line 59: It might be worth adding a `fn extend_from_slice(&self, &[impl IsA<Object>])` convenience function as the position/removals usage is not too obvious and requires a few seconds of thinking.

ranfdev:
While implementing extend for `&A`:
```
error[E0119]: conflicting implementations of trait `std::iter::Extend<&'static _>` for type `auto::list_store::ListStore`
  --> gio/src/list_store.rs:83:1
   |
77 | impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
   | ------------------------------------------------------------- first implementation here
...
83 | impl<A: IsA<glib::Object>> std::iter::Extend<&'static A> for ListStore {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `auto::list_store::ListStore`
   |
   = note: downstream crates may implement trait `glib::IsA<glib::Object>` for type `&'static _`
   
```
   
 Well, Rust is right...
 That also means a crate may implement `glib::IsA<glib::Object>` on a `&glib::Object`, therefore making a reference-to-a-pointer `impl glib::IsA<glib::Object>` which then crashes `ListStore::splice`.
 
 `impl glib::IsA<glib::Object>` may be implemented by anyone and it's not enough to affirm that type is a pointer. passing a custom implementation to `ListStore::splice` causes undefined behaviour

sdroege:
> That also means a crate may implement `glib::IsA<glib::Object>` on a `&glib::Object`, therefore making a reference-to-a-pointer `impl glib::IsA<glib::Object>` which then crashes `ListStore::splice`.

No, because `IsA` is an `unsafe trait` and if some code does that it violated the rules of that trait. And I'm quite certain that you actually can't implement the trait on non-`'static` references but the compiler apparently can't know that :)

sdroege:
> but the compiler apparently can't know that

Ah the compiler apparently knows that it could only be implemented on `'static` references. But that implementation would be problematic and invalid regardless.

gio/src/list_store.rs
ranfdev: It's now covered by `extend`. See latest commit.
These now work:
```
list.extend([obj, obj]);
list.extend(&[obj, obj]);
list.extend([&obj, &obj]);
```

gio/src/list_store.rs
jf2048: It would be possible to do this without cloning and without allocating a second `Vec`, if something like this was done instead:

```rust
impl<'a, A: IsA<glib::Object>> std::iter::Extend<&'a A> for ListStore
```

gio/src/list_store.rs
sdroege: API-wise yes, but just like `Vec` has an `extend_from_slice()` for efficiency reasons, it also makes sense here. You wouldn't need to create a temporary `Vec` for that but could directly call into `splice()`.

gio/src/list_store.rs
sdroege: > It would be possible to do this without cloning and without allocating a second Vec

How? If you insert item by item while iterating then this causes a lot of overhead (signals!). Unfortunately this can't be efficiently re-implemented outside of `gliststore.c`.

ranfdev: I can't implement both `Extend<A>` and `Extend<&'a A>`.
```
conflicting implementation for `auto::list_store::ListStore`
note: downstream crates may implement trait `glib::IsA<glib::Object>` for type `&'a _`
```

Also, `splice` needs a slice, so as sdroege said, I have to allocate a `Vec` somewhere. The `ListStore` implementation is in C and it takes a slice.

gio/src/list_store.rs
```diff
@@ -75,3 +79,28 @@ unsafe extern "C" fn compare_func_trampoline(
 
     (*func)(&a, &b).into_glib()
 }
+
+impl<A: AsRef<glib::Object>> std::iter::Extend<A> for ListStore {
+    fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
+        let additions = iter
+            .into_iter()
+            .map(|a| a.as_ref().clone())
+            .collect::<Vec<_>>();
```
sdroege, line 88:
```suggestion
impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
    fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
        let additions = iter
            .into_iter()
            .collect::<Vec<_>>();
```

sdroege (dismissed):

ranfdev:
@sdroege I had left a review of your changes... Please take a look at it.

sdroege:
@ranfdev Where?

PR #428, opened 2021-12-18
gio: Implement `Extend` on `ListStore` for `AsRef<Object>`

sdroege:
This also allows to use iterators over references of objects.

Also add a test for `extend()`.

----

See https://github.com/gtk-rs/gtk-rs-core/commit/277b94f0edecbee063f2ad12093e346868c8f8ef#r62142353

CC @ranfdev

### DIFF
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -80,9 +80,12 @@ unsafe extern "C" fn compare_func_trampoline(
     (*func)(&a, &b).into_glib()
 }
 
-impl<A: IsA<glib::Object>> std::iter::Extend<A> for ListStore {
+impl<A: AsRef<glib::Object>> std::iter::Extend<A> for ListStore {
     fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T) {
-        let additions = iter.into_iter().collect::<Vec<_>>();
+        let additions = iter
+            .into_iter()
+            .map(|o| o.as_ref().clone())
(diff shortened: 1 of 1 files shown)


gio/src/list_store.rs
```diff
@@ -100,4 +104,17 @@ mod tests {
...
         assert_eq!(list.item(1), Some(item1.upcast()));
     }
+
+    #[test]
+    fn extend() {
+        let item0 = ListStore::new(ListStore::static_type());
+        let item1 = ListStore::new(ListStore::static_type());
+        let mut list = ListStore::new(ListStore::static_type());
+        list.extend(&[&item0, &item1]);
+        assert_eq!(list.item(0).as_ref(), Some(item0.upcast_ref()));
```
ranfdev, line 114: You could remove `.as_ref()` and `.upcast_ref()` and it would work equally well

ranfdev, line 114:
```suggestion
        assert_eq!(list.item(0), Some(item0.upcast()));
```

sdroege: No, because `upcast()` consumes the value and then it's gone and can't be used for the following comparison :)

ranfdev: Ups, sorry. 

I don't have an "approve" button, but i approve this PR.

GuillaumeGomez (approved):

bilelmoussaoui (approved):

PR #429, opened 2021-12-19
glib: Add WeakRefNotify

jsparber:
This adds a new weak reference for Objects, so that we can connect a
callback for when the obejct is disposed.

`WeakRefLocal` is needed because `g_object_weak_ref` and `g_object_weak_unref`.
This adds a rust way to use `g_object_weak_ref` `g_object_weak_unref`.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -12,6 +12,7 @@ use std::fmt;
 use std::hash;
 use std::marker::PhantomData;
 use std::mem;
+use std::mem::ManuallyDrop;
 use std::ops;
 use std::pin::Pin;
 use std::ptr;
@@ -2095,6 +2096,20 @@ pub trait ObjectExt: ObjectType {
     fn downgrade(&self) -> WeakRef<Self>;
 
     // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F)
+        -> WeakRefNotify<Self>;
+
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    ///
+    /// This is like [`add_weak_ref_notify`][`crate::OBjectExt::add_weak_ref_notify`]  but doesn't require a `Send+Sync` closure.
+    /// Object dispose will panic if the object is disposed from the wrong thread.
+    #[doc(alias = "g_object_weak_ref")]
+    fn add_weak_ref_notify_local<F: FnOnce() + 'static>(&self, f: F) -> WeakRefNotify<Self>;
+
+    // rustdoc-stripper-ignore-next
     /// Bind property `source_property` on this object to the `target_property` on the `target` object.
     ///
     /// This allows keeping the properties of both objects in sync.
@@ -3151,6 +3166,16 @@ impl<T: ObjectType> ObjectExt for T {
         }
     }
 
+    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F) -> WeakRefNotify<T> {
+        WeakRefNotify::new(self, f)
+    }
+
+    fn add_weak_ref_notify_local<F: FnOnce() + 'static>(&self, f: F) -> WeakRefNotify<T> {
+        let callback = crate::thread_guard::ThreadGuard::new(f);
+
+        WeakRefNotify::new(self, move || callback.into_inner()())
+    }
+
     fn bind_property<'a, O: ObjectType>(
         &'a self,
         source_property: &'a str,
@@ -3373,6 +3398,65 @@ wrapper! {
 
     match fn {
         type_ => || gobject_ffi::g_initially_unowned_get_type(),
+    }
+}
+
+type WeakRefNotifyData = ManuallyDrop<Pin<Box<Option<Box<dyn FnOnce() + 'static>>>>>;
+
+// rustdoc-stripper-ignore-next
+/// A handle to disconnect a weak ref notify closure.
+pub struct WeakRefNotify<T: ObjectType> {
+    object: WeakRef<T>,
+    data: WeakRefNotifyData,
+}
+
+unsafe extern "C" fn notify_func(data: ffi::gpointer, _obj: *mut gobject_ffi::GObject) {
+    let callback: Box<Option<Box<dyn FnOnce()>>> = Box::from_raw(data as *mut _);
+
+    let callback = (*callback).expect("cannot get closure...");
+    callback()
+}
+
+impl<T: ObjectType> WeakRefNotify<T> {
+    fn new<F: FnOnce() + 'static>(obj: &T, f: F) -> WeakRefNotify<T> {
+        unsafe {
+            let data: WeakRefNotifyData = ManuallyDrop::new(Box::pin(Some(Box::new(f))));
+
+            let data_ptr: *const Option<Box<dyn FnOnce()>> = Pin::as_ref(&data).get_ref();
+            gobject_ffi::g_object_weak_ref(
+                obj.as_ptr() as *mut GObject,
+                Some(notify_func),
+                data_ptr as *mut _,
+            );
+
+            WeakRefNotify {
+                object: obj.unsafe_cast_ref::<T>().downgrade(),
+                data,
+            }
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Try to upgrade this weak reference to a strong reference.
+    ///
+    /// If the stored object was already destroyed then `None` is returned.
+    pub fn upgrade(&self) -> Option<T> {
+        self.object.upgrade()
+    }
+
+    #[doc(alias = "g_object_weak_unref")]
+    pub fn disconnect(mut self) {
+        unsafe {
+            if let Some(obj) = self.object.upgrade() {
+                let data_ptr: *const Option<Box<dyn FnOnce()>> = Pin::as_ref(&self.data).get_ref();
+                gobject_ffi::g_object_weak_unref(
+                    obj.as_ptr() as *mut GObject,
+                    Some(notify_func),
+                    data_ptr as *mut _,
+                );
+                ManuallyDrop::drop(&mut self.data);
+            }
+        }
     }
 }
 
@@ -4167,6 +4251,10 @@ pub unsafe trait IsInterface: ObjectType {}
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::cell::Cell;
+    use std::rc::Rc;
+    use std::sync::atomic::{AtomicBool, Ordering};
+    use std::sync::Arc;
 
     #[test]
     fn new() {
@@ -4202,6 +4290,49 @@ mod tests {
 
         drop(obj);
         assert!(weakref.upgrade().is_none());
+    }
+
+    #[test]
+    fn weak_ref_notify() {
+        let obj: Object = Object::new(&[]).unwrap();
+
+        let handle = obj.add_weak_ref_notify(|| {
+            unreachable!();
+        });
+
+        handle.disconnect();
+
+        let called = Arc::new(AtomicBool::new(false));
+        let called_weak = Arc::downgrade(&called);
+        let handle = obj.add_weak_ref_notify(move || {
+            called_weak.upgrade().unwrap().store(true, Ordering::SeqCst);
+        });
+
+        drop(obj);
+        assert!(called.load(Ordering::SeqCst));
+        handle.disconnect();
+
+        let obj: Object = Object::new(&[]).unwrap();
+
+        let called = Arc::new(AtomicBool::new(false));
+        let called_weak = Arc::downgrade(&called);
+        obj.add_weak_ref_notify(move || {
+            called_weak.upgrade().unwrap().store(true, Ordering::SeqCst);
+        });
+
+        drop(obj);
+        assert!(called.load(Ordering::SeqCst));
+
+        let obj: Object = Object::new(&[]).unwrap();
+
+        let called = Rc::new(Cell::new(false));
+        let called_weak = Rc::downgrade(&called);
+        obj.add_weak_ref_notify_local(move || {
+            called_weak.upgrade().unwrap().set(true);
+        });
+
+        drop(obj);
+        assert!(called.get());
     }
 
     #[test]


glib/src/object.rs
```diff
@@ -1981,6 +1981,12 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Downgrade this object to a local weak reference.
+    /// The clousure will be called when the the object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    fn downgrade_local<F: Fn() + 'static>(&self, f: F) -> WeakRefLocal<Self, F>;
```
sdroege, line 1988: This is unsound because you can also call that on actually thread-safe objects and the callback can then be called from any random thread although it's not `Send`.

glib/src/object.rs
```diff
@@ -3267,6 +3279,96 @@ wrapper! {
...
+
+            data.0 = ptr::null_mut();
+
+            let f: &F = &*(&data.1 as *const F);
+
+            f();
+        }
+
+        WeakRefLocal(
+            Box::into_raw(Box::new((ptr::null_mut(), f))),
```
sdroege, line 3314: Why? I think you can also just store the `Box` in your struct :) Also you probably want to make use of `Pin` here because you require that the content of the box is never moved.

glib/src/object.rs
```diff
@@ -3267,6 +3279,96 @@ wrapper! {
...
+            Some(notify_func::<F>),
+            PhantomData,
+        )
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Set this weak reference to the given object.
+    #[doc(alias = "g_weak_ref_set")]
+    pub fn set(&self, obj: Option<&T>) {
```
sdroege, line 3323: This is also unsound for actually thread-safe objects.

I think it would be best if you make the thing immutable after construction and don't allow changing the object it's attached to

glib/src/object.rs
```diff
@@ -3267,6 +3279,96 @@ wrapper! {
...
+            }
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Try to upgrade this weak reference to a strong reference.
+    ///
+    /// If the stored object was already destroyed or no object was set in this weak reference then
+    /// `None` is returned.
+    pub fn upgrade(&self) -> Option<T> {
```
sdroege, line 3346: This is unsound and you really want to use `WeakRef` for this case. I wouldn't provide this feature here :)

glib/src/object.rs
```diff
@@ -1981,6 +1981,12 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Downgrade this object to a local weak reference.
+    /// The clousure will be called when the the object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    fn downgrade_local<F: Fn() + 'static>(&self, f: F) -> WeakRefLocal<Self, F>;
+
```
sdroege, line 1989: Also I would call it something like `add_weak_notify()` or so

sdroege (changes requested):

jsparber:
Thanks for the review. I added the `WeakRefLocal` to add a local variant of `WeakRef`, since it has some overhead compared to simply have a pointer to the object. But i don't actually know how much impact the additional work of `WeakRef`s has.

sdroege:
> But i don't actually know how much impact the additional work of `WeakRef`s has.

None, it's just implemented correctly ;)

jsparber:
Should the callback be disconnected when disconnect handle is dropped? Is fare as I know we generally don't disconnect the callback when the disconnect handle is dropped. But if we want to keep the callback after the disconnect handle I'm not sure how i can use a `Pin<Box<>>` since dropping the handle would also drop the Pin.

sdroege:
> Should the callback be disconnected when disconnect handle is dropped? Is fare as I know we generally don't disconnect the callback when the disconnect handle is dropped.

I think having a `disconnect()` method on the return value could be useful instead (that consumes the value as `self`).

> But if we want to keep the callback after the disconnect handle I'm not sure how i can use a `Pin<Box<>>` since dropping the handle would also drop the Pin.

Well, I have no idea how to do this whole thing in a safe way currently :) It seems like you'd easily get double-frees here.

glib/src/object.rs
jsparber: dropped it.

glib/src/object.rs
jsparber: dropped it as well.

glib/src/object.rs
jsparber: changed the name.

glib/src/object.rs
jsparber: I didn't yet figure out why how i can use Pin<Box<>> in the place of `*mut` in the struct.

jsparber:
> I think having a `disconnect()` method on the return value could be useful instead (that consumes the value as `self`).

I added now the `disconnect()` method.

> Well, I have no idea how to do this whole thing in a safe way currently :) It seems like you'd easily get double-frees here.

I solved the double-free by storing additional information, that tells the callback whether the `WeakRefNotify` handle was dropped and the if the callback should free the data.

I still didn't figure out how i could use Pin<Box<>> but i will look into it.

sdroege:
I'll try to find some time for reviewing this later but

> I solved the double-free by storing additional information, that tells the callback whether the WeakRefNotify handle was dropped and the if the callback should free the data.

Without a mutex that will likely not be correct.

jsparber:
I figured out how to use a Pin<Box<>>. I also wrapped it into a `ManuallyDrop` to manually call `drop()` in the `Drop` implementation or the callback.

>Without a mutex that will likely not be correct.

I used a `Cell` now. Since the `ThreadGuard` will panic  if the `notify_func()` is called from the wrong thread.

jsparber:
Looks like we are leaking 32 bytes when freeing the Pin<Box<>> from the callback.
```
==337860== 
==337860== HEAP SUMMARY:
==337860==     in use at exit: 39,844 bytes in 275 blocks
==337860==   total heap usage: 994 allocs, 719 frees, 133,499 bytes allocated
==337860== 
==337860== 32 bytes in 1 blocks are definitely lost in loss record 110 of 253
==337860==    at 0x484186F: malloc (vg_replace_malloc.c:381)
==337860==    by 0x1BE7DB: alloc::alloc::alloc (alloc.rs:86)
==337860==    by 0x1BE866: alloc::alloc::Global::alloc_impl (alloc.rs:166)
==337860==    by 0x1BEEB9: <alloc::alloc::Global as core::alloc::Allocator>::allocate (alloc.rs:226)
==337860==    by 0x1BE73C: alloc::alloc::exchange_malloc (alloc.rs:315)
==337860==    by 0x294B0A: glib::object::WeakRefNotify<T,F>::new (boxed.rs:250)
==337860==    by 0x294352: <T as glib::object::ObjectExt>::add_weak_ref_notify (object.rs:3056)
==337860==    by 0x192273: glib::object::tests::weak_ref_notify (object.rs:4202)
==337860==    by 0x17E2D9: glib::object::tests::weak_ref_notify::{{closure}} (object.rs:4176)
==337860==    by 0x198FBD: core::ops::function::FnOnce::call_once (function.rs:227)
==337860==    by 0x2CD2E2: call_once<fn(), ()> (function.rs:227)
==337860==    by 0x2CD2E2: test::__rust_begin_short_backtrace (lib.rs:578)
==337860==    by 0x2CBE86: call_once<(), (dyn core::ops::function::FnOnce<(), Output=()> + core::marker::Send), alloc::alloc::Global> (boxed.rs:1638)
==337860==    by 0x2CBE86: call_once<(), alloc::boxed::Box<(dyn core::ops::function::FnOnce<(), Output=()> + core::marker::Send), alloc::alloc::Global>> (unwind_safe.rs:271)
==337860==    by 0x2CBE86: do_call<core::panic::unwind_safe::AssertUnwindSafe<alloc::boxed::Box<(dyn core::ops::function::FnOnce<(), Output=()> + core::marker::Send), alloc::alloc::Global>>, ()> (panicking.rs:403)
==337860==    by 0x2CBE86: try<(), core::panic::unwind_safe::AssertUnwindSafe<alloc::boxed::Box<(dyn core::ops::function::FnOnce<(), Output=()> + core::marker::Send), alloc::alloc::Global>>> (panicking.rs:367)
==337860==    by 0x2CBE86: catch_unwind<core::panic::unwind_safe::AssertUnwindSafe<alloc::boxed::Box<(dyn core::ops::function::FnOnce<(), Output=()> + core::marker::Send), alloc::alloc::Global>>, ()> (panic.rs:133)
==337860==    by 0x2CBE86: run_test_in_process (lib.rs:601)
==337860==    by 0x2CBE86: test::run_test::run_test_inner::{{closure}} (lib.rs:493)
==337860== 
==337860== LEAK SUMMARY:
==337860==    definitely lost: 32 bytes in 1 blocks
==337860==    indirectly lost: 0 bytes in 0 blocks
==337860==      possibly lost: 0 bytes in 0 blocks
==337860==    still reachable: 37,716 bytes in 252 blocks
==337860==         suppressed: 0 bytes in 0 blocks
==337860== Reachable blocks (those to which a pointer was found) are not shown.
==337860== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==337860== 
==337860== For lists of detected and suppressed errors, rerun with: -s
==337860== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

glib/src/object.rs
```diff
@@ -3267,6 +3281,75 @@ wrapper! {
...
+        }
+    }
+}
+impl<T: ObjectType, F: Fn() + 'static> Drop for WeakRefNotify<T, F> {
+    fn drop(&mut self) {
+        let data = &*(self.0);
+        // Drop the data only when the callback was disconnected or when it was already called
+        if data.2.get() {
+            unsafe {
+                ManuallyDrop::drop(&mut self.0);
```
sdroege, line 3345: This is unsound. If the `WeakRefNotify` is dropped from one thread and ends up in this line, and the original thread ends up in the `notify` function and drops it there another time, you have a double-free.

glib/src/object.rs
```diff
@@ -1981,6 +1983,14 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Downgrade this object to a local weak reference.
+    /// The clousure will be called when the the object is disposed.
+    ///
+    /// Object dispose will panic if the object is disposed from the wrong thread.
```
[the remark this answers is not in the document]
jsparber: how would that work? `g_object_weak_ref()` isn't thread-safe

glib/src/object.rs
```diff
@@ -3267,6 +3281,75 @@ wrapper! {
...
+            data: ffi::gpointer,
+            _obj: *mut gobject_ffi::GObject,
+        ) {
+            let data = &mut *(data as *mut (*mut GObject, crate::ThreadGuard<F>, Cell<bool>));
+
+            (data.1.get_ref())();
+
+            // Drop the data when the `WeakRefNotify` handle was already dropped
+            if data.2.get() {
+                std::ptr::drop_in_place(data);
```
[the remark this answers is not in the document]
jsparber: does exactly the same. https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html#method.drop

glib/src/object.rs
jsparber: ah true. But `WeakRefNotify` can't be moved to other threads.

glib/src/object.rs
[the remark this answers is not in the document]
sdroege: It's thread-safe, that's not the problem. Please read the docs again :) What is not thread-safe is to use it to implement weak references. Which also brings us all the problems you have to fix here.

glib/src/object.rs
sdroege: It should be possible to move it to other threads though :) Nothing wrong with that other than your implementation right now.

glib/src/object.rs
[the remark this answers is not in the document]
jsparber: > Note that the weak references created by this method are not thread-safe: they cannot safely be used in one thread if the object’s last g_object_unref() might happen in another thread. Use GWeakRef if thread-safety is required.

Honestly this isn't super clear. So this means that the notify call is thread safe. But the access to the `data` is not.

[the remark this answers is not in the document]
sdroege: The API itself is thread-safe (ref, unref, the callback itself). What is not thread-safe is implementing weak references with it because you have no way of "upgrading" your weak reference that is not racy with the object being destroyed. The callback is called once the object is gone, so even if you have a mutex inside the callback and inside your `upgrade()` function, you can't prevent it from "upgrading" an already destroyed object.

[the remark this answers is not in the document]
jsparber: makes sense. Thanks fro explaining :)

jsparber:
> Looks like we are leaking 32 bytes when freeing the Pin<Box<>> from the callback.

Valgrind still says that we leak the `WeakRefNotifyData` when it's not dropped in `Drop` but could also be a false positive.

glib/src/object.rs
```diff
@@ -3267,6 +3294,86 @@ wrapper! {
...
+    }
+
+    #[doc(alias = "g_object_weak_unref")]
+    pub fn disconnect(self) {
+        unsafe {
+            let data = &*(self.0);
+            let data_ptr: *const _ = Pin::as_ref(&self.0).get_ref();
+
+            if !data.should_drop() {
+                gobject_ffi::g_object_weak_unref(data.ptr, self.1, data_ptr as *mut _);
```
sdroege, line 3360: This will do bad things if the object was destroyed in the meantime. Your usage of the atomic bool is not preventing that

Maybe we just can't provide a way to disconnect safely

sdroege:
> Valgrind still says that we leak the WeakRefNotifyData when it's not dropped in Drop but could also be a false positive.

Well, add some printf debugging :) It's probably not a false positive.

glib/src/object.rs
jsparber: > This will do bad things if the object was destroyed in the meantime. Your usage of the atomic bool is not preventing that

Are you sure? Once the callback is called `should_drop` is set to `true` and `gobject_ffi::g_object_weak_unref()` isn't called anymore.

sdroege: Thread A can see `should_drop == false`, then thread B calls the notify callback, then thread A calls `g_object_weak_unref`.

jsparber: right. I think we should hold instead of the `mut * GObject` a `WeakRef` to the object and then try to `upgrade()`  it before we call `gobject_ffi::g_object_weak_unref()` this way the `ref_count` is definitely >1 and the callback won't be called from any thread.

sdroege: That probably works, yes

jsparber:
> > Valgrind still says that we leak the WeakRefNotifyData when it's not dropped in Drop but could also be a false positive.
> 
> Well, add some printf debugging :) It's probably not a false positive.

found and fixed the memory leak.

jsparber:
I'm wondering if we should impl `Deref<WeakRef> for WeakRefNotify` so that the user doesn't need to hold  onto the `WeakRefNotify` and an additional `WeakRef`.

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
     }
 }
 
+struct WeakRefNotifyData {
+    callback: Box<dyn Fn()>,
+    should_drop: AtomicBool,
+}
+
+impl WeakRefNotifyData {
+    pub fn new<F: Fn() + 'static>(f: F) -> ManuallyDrop<Pin<Box<Self>>> {
```
sdroege, line 3303: This and below does not need to be `pub`

glib/src/object.rs
```diff
@@ -1981,6 +1983,19 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    fn add_weak_ref_notify<F: Fn() + Send + Sync + 'static>(&self, f: F) -> WeakRefNotify;
```
sdroege, line 1989:
```suggestion
    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F) -> WeakRefNotify;
```

Also below

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+impl WeakRefNotify {
+    pub(crate) fn new<T: ObjectType, F: Fn() + 'static>(obj: &T, f: F) -> WeakRefNotify {
+        unsafe extern "C" fn notify_func(data: ffi::gpointer, _obj: *mut gobject_ffi::GObject) {
+            let data = Box::from_raw(data as *mut WeakRefNotifyData);
+
+            (data.callback)();
+
+            // Don't drop the data when the `WeakRefNotify` handle wasn't dropped yet
+            if !data.should_drop() {
+                Box::leak(data);
```
sdroege, line 3334: If your `callback` panics then we'll have a double free :) Maybe use just a reference outside this `if` here and call `from_raw` only inside the `if`?

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+            }
+        }
+
+        unsafe {
+            let mut data = WeakRefNotifyData::new(f);
+            let obj_ptr = obj.as_object_ref().to_glib_none().0;
+
+            let data_ptr: *mut _ = Pin::as_mut(&mut data).get_unchecked_mut();
```
sdroege, line 3342:
```suggestion
            let data_ptr: *mut WeakRefNotifyData = Pin::as_mut(&mut data).get_unchecked_mut().as_ptr();
```

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+
+            // Don't drop the data when the `WeakRefNotify` handle wasn't dropped yet
+            if !data.should_drop() {
+                Box::leak(data);
+            }
+        }
+
+        unsafe {
+            let mut data = WeakRefNotifyData::new(f);
+            let obj_ptr = obj.as_object_ref().to_glib_none().0;
```
sdroege, line 3340:
```suggestion
            let obj_ptr = obj.as_ptr();
```

Don't we have that?

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+                func: Some(notify_func),
+            }
+        }
+    }
+
+    #[doc(alias = "g_object_weak_unref")]
+    pub fn disconnect(self) {
+        unsafe {
+            if !self.data.should_drop() {
+                if let Some(obj) = self.object.upgrade() {
```
sdroege, line 3357: I think this should be before the `should_drop()` or not? Otherwise you might end up swapping the bool but then upgrading fails and then you leak the data?

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+                    let data_ptr: *const _ = Pin::as_ref(&self.data).get_ref();
+                    let obj_ptr = obj.as_object_ref().to_glib_none().0;
```
sdroege, line 3359:
```suggestion
                    let obj_ptr = obj.as_ptr();
```

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+            }
+        }
+    }
+
+    #[doc(alias = "g_object_weak_unref")]
+    pub fn disconnect(self) {
+        unsafe {
+            if !self.data.should_drop() {
+                if let Some(obj) = self.object.upgrade() {
+                    let data_ptr: *const _ = Pin::as_ref(&self.data).get_ref();
```
sdroege, line 3358:
```suggestion
                    let data_ptr: *const WeakRefNotifyData = Pin::as_ref(&self.data).get_ref();
```

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+    /// Checks whether the data should be dropped and marks it to be dropped
+    pub fn should_drop(&self) -> bool {
+        self.should_drop.swap(true, Ordering::SeqCst)
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// A handle to disconnect a weak ref notify closure.
+pub struct WeakRefNotify {
+    object: WeakRef<Object>,
```
sdroege, line 3320: `impl Deref<Target=WeakRef>` would make sense here but then also make make it an `WeakRefNotify<T>` :)

glib/src/object.rs
```diff
@@ -3267,6 +3294,87 @@ wrapper! {
...
+                data,
+                func: Some(notify_func),
+            }
+        }
+    }
+
+    #[doc(alias = "g_object_weak_unref")]
+    pub fn disconnect(self) {
+        unsafe {
+            if !self.data.should_drop() {
```
sdroege, line 3356: So what can happen here is:

- Thread A goes into this branch here and swaps the `bool` to `true`
- Thread B calls the notify, `bool` is `true` so it will drop the data
- Thread A continues here, fails upgrading (no problem), goes to the end of the scope and drops `self`: `bool` is `true` so it drops the data a second time

sdroege: Maybe you want to store the data in an `Arc` and drop that always here, and also either always drop it from the callback or after calling `g_object_weak_unref`?

glib/src/object.rs
jsparber: how would `WeakRef::set()` work? if the user changes the object inside the `WeakRef` it would break `WeakRefNotify::disconnect()`

sdroege: Indeed, you can only provide an `upgrade()` function but not `Deref`

jsparber:
I was thinking about this a little bit. If we add `upgrade()` to `WeakRefNotify` then it becomes  pretty much the same as `WeakRef` I'm wondering if we should add it the `WeakRef::add_notify()` directly.

sdroege:
How would that look like exactly? Currently `WeakRef` is just a pointer and adding a lot more state to it that 99% are not going to use doesn't seem ideal.

jsparber:
> How would that look like exactly? Currently WeakRef is just a pointer and adding a lot more state to it that 99% are not going to use doesn't seem ideal.

It would add the `Option<WeakRefNotifyData>` (probably behind a mutex), a check in `set()` whether we need to disconnect the previous callback and add a new callback when `WeakRefNotifyData` is set.

sdroege:
I would keep it separate :)

glib/src/object.rs
```diff
@@ -2094,6 +2095,20 @@ pub trait ObjectExt: ObjectType {
...
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F)
+        -> WeakRefNotify<Self>;
+
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    ///
+    /// This is like `add_weak_ref_notify` but doesn't require a `Send+Sync` closure.
```
bilelmoussaoui, line 2107:
```suggestion
    /// This is like [`add_weak_ref_notify`][`crate::OBjectExt::add_weak_ref_notify`]  but doesn't require a `Send+Sync` closure.
```
Maybe?

jf2048:
I was thinking about this, it seems like you could maybe get rid of the inner `WeakRef` by storing the notify func in an `AtomicPtr` and then using `compare_exchange`in both `notify_func` and `disconnect`? I think you could also use that to implement `Drop`?

glib/src/object.rs
```diff
@@ -3376,6 +3401,65 @@ wrapper! {
     }
 }
 
+type WeakRefNotifyData = ManuallyDrop<Pin<Box<Option<Box<dyn FnOnce() + 'static>>>>>;
```
sdroege, line 3404: Why the `Pin` and why the double boxing and why the `Option` and why the `ManuallyDrop`? This at least needs some comment explaining this contraption :)

glib/src/object.rs
```diff
@@ -3376,6 +3401,65 @@ wrapper! {
...
+
+// rustdoc-stripper-ignore-next
+/// A handle to disconnect a weak ref notify closure.
+pub struct WeakRefNotify<T: ObjectType> {
+    object: WeakRef<T>,
+    data: WeakRefNotifyData,
+}
+
+unsafe extern "C" fn notify_func(data: ffi::gpointer, _obj: *mut gobject_ffi::GObject) {
+    let callback: Box<Option<Box<dyn FnOnce()>>> = Box::from_raw(data as *mut _);
```
sdroege, line 3414: This is not very clean: you're passing a `ManuallyDrop<Pin<T>>` in here but then handle it like a `T`. It would be better to use the correct type here.

Also this needs a comment explaining why it is safe to drop the closure from the callback here (i.e. why it is guaranteed to be called only once).

glib/src/object.rs
```diff
@@ -3376,6 +3401,65 @@ wrapper! {
...
+
+    let callback = (*callback).expect("cannot get closure...");
```
sdroege, line 3416: You don't really need the `Option` here and it doesn't add much safety

glib/src/object.rs
```diff
@@ -3376,6 +3401,65 @@ wrapper! {
...
+    pub fn disconnect(mut self) {
+        unsafe {
+            if let Some(obj) = self.object.upgrade() {
+                let data_ptr: *const Option<Box<dyn FnOnce()>> = Pin::as_ref(&self.data).get_ref();
+                gobject_ffi::g_object_weak_unref(
+                    obj.as_ptr() as *mut GObject,
+                    Some(notify_func),
+                    data_ptr as *mut _,
+                );
+                ManuallyDrop::drop(&mut self.data);
```
sdroege, line 3457: And also here an explanation why it's safe to drop the data (i.e. why it's guaranteed to be still alive, there are no race conditions, etc).

glib/src/object.rs
```diff
@@ -3376,6 +3401,65 @@ wrapper! {
...
+    ///
+    /// If the stored object was already destroyed then `None` is returned.
+    pub fn upgrade(&self) -> Option<T> {
+        self.object.upgrade()
+    }
+
+    #[doc(alias = "g_object_weak_unref")]
+    pub fn disconnect(mut self) {
+        unsafe {
+            if let Some(obj) = self.object.upgrade() {
```
sdroege, line 3450: Is it actually guaranteed that the weak notify is not currently being called on another thread if `upgrade()` succeeds here?

sdroege (commented):
Needs a lot of documentation about safety, but otherwise is **probably** fine if the weak reference upgrade is guaranteeing that the weak notify is never going to be called (not even from another thread).

glib/src/object.rs
sdroege: Please double check that this is all given and then document this here.

jsparber: ah i miss read your comment. I will check.

sdroege: I checked the code again and I think it's OK, just needs to be documented.

glib/src/object.rs
```diff
@@ -2094,6 +2095,20 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F)
```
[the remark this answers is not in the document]
sdroege: The feature to get a callback when the object is disposed is literally called "weak ref" in GObject: `g_object_weak_ref()`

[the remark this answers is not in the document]
AaronErhardt: Sure, if that's the way GObject calls this feature we can use it. I was just looking at this from the perspective of a "normal" Rust programmer.

Also, I've seen that this PR hasn't seen a lot of activity lately, I'd be happy if I can help to get it across the finish line :)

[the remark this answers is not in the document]
sdroege: I don't like all the names so far, so I would stay with the GObject terminology. But if someone has a better idea. In Rust terms this would be a "drop notify" but that's also sounding strange.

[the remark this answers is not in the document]
sdroege: @AaronErhardt Also I think you can take this up if you want. @jsparber seems to be very busy and there's not much left here anyway.

OOC where do you plan to use this?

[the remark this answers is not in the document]
AaronErhardt: > OOC where do you plan to use this?

The in-development version of Relm4 requires the root of a component to implement a helper trait called `OnDestroy` which is implemented using `WidgetExt::connect_destroy`. This mechanism makes it possible to bind the runtime of the component to the lifetime of its root widget. However, this is problematic for native widgets which don't implement `WidgetExt` because they cannot be the root of a component in this case. The plan is to implement `OnDestroy` for all objects via `ObjectExt` to remove this limitation.

Relevant issue: https://github.com/Relm4/Relm4/issues/143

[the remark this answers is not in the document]
sdroege: You won't get the object itself in the callback though, only a pointer to the already gone object. That's sufficient for you?

Also an alternative to this is to use `ObjectExt::set_data()` and similar and to go via the `Drop` impl of the stored data.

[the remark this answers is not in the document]
AaronErhardt: Yes that should be sufficient. It's just used to send a final message to the component that tells it to shut down its runtime.

The current implementation of `OnDestroy` can be found here: https://github.com/Relm4/Relm4/blob/next/relm4/src/component/mod.rs#L48

sdroege:
Superseded by https://github.com/gtk-rs/gtk-rs-core/pull/666

PR #430, opened 2021-12-20
Implement ToValue on signal::Inhibit

jf2048:
Not particularly sure if this is useful since it can't be a strict requirement for closures, but it can give some consistency with the auto generated `connect` methods.


sdroege (approved):

sdroege:
I think this makes sense.

Issue #422, opened 2021-12-16
[BUG] `glib` must be in scope for #[object_subclass] to work


ranfdev:
**Bug description**

As said, when trying to use `#[object_subclass]` without `glib` in scope:
`failed to resolve: use of undeclared crate or module `glib`  use of undeclared crate or module `glib``

jf2048:
I think this is just changing `crate_ident_new` to put `::` at the start of the token stream?

jf2048:
We could ask for some new environment variable in cargo to detect if we're building the crate itself or if we're in examples/doctests/integration tests?

sdroege:
That could work, yes

PR #432, opened 2021-12-22
glib-macros: Use absolute paths in proc macros

jf2048:
Just some cleanup, also adds use of `crate_ident_new` in a few places where it's missing. Technically this should also be done on the method calls at some point.


sdroege:
Does this fix https://github.com/gtk-rs/gtk-rs-core/issues/422 ?

sdroege (approved):

jf2048:
No I think we're blocked on that one.

PR #434, opened 2021-12-27
glib-macros: Add support for enums to glib::Variant 

### LINKED ISSUE #394: [FEATURE REQUEST] Support enums in Variant derive macro
It should be possible to expand this to support all rust enums as well as `GEnum` types. Inspiration can be taken from [serde's enum serialization](https://serde.rs/enum-representations.html). There is also the (possibly unrelated) question of what to do about ffi enums, which are not currently usable from a struct with `derive(Variant)`.

In the long term, perhaps it would be better to replace this macro with a serde-compatible implementation similar to what [zbus](https://docs.rs/zvariant/2.10.0/zvariant/gvariant/index.html) does.

jf2048:
Fixes #394

Supports a few different serialization methods via a `variant_enum` attribute, which changes the variant type:

- Defaults:
  - `(sv)` for enums with data variants, where `s` is the kebab case tag for the variant and `v` is a tuple variant packed as the equivalent struct would be.
  - `s` for c-style enums, just the kebab case tag.
- `#[variant_enum(repr)]` to be used with `#[repr(u*)]` or `#[repr(i*)]`.
  - `(*v)` or `*` where `*` is the integer type corresponding to the integer type.
- `#[variant_enum(enum)]` to be used with `#[derive(glib::Enum)]`.
  - `s` for the nick string of the value. Equivalent to the GTK builder format.
- `#[variant_enum(flags)]` to be used with `#[glib::flags]`.
  - `s` for the nick strings of the value, separated with `|`. Equivalent to the GTK builder format.
- `#[variant_enum(enum, repr)]` to be used with `#[derive(glib::Enum)]`.
  - `i`, or always serializes as `i32`.
- `#[variant_enum(flags, repr)]` to be used with `#[glib::flags]`.
  - `u`, or always serializes as `u32`.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -437,6 +437,8 @@ pub fn closure_local(item: TokenStream) -> TokenStream {
 ///     ValWithCustomNameAndNick,
 /// }
 /// ```
+///
+/// [`glib::Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
 #[proc_macro_derive(Enum, attributes(enum_type, enum_value))]
 #[proc_macro_error]
 pub fn enum_derive(input: TokenStream) -> TokenStream {
@@ -474,7 +476,7 @@ pub fn enum_derive(input: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
-/// [`glib::Value`]: value/struct.Value.html
+/// [`glib::Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
 #[proc_macro_attribute]
 #[proc_macro_error]
 pub fn flags(attr: TokenStream, item: TokenStream) -> TokenStream {
@@ -501,7 +503,7 @@ pub fn flags(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
-/// [`ErrorDomain`]: error/trait.ErrorDomain.html
+/// [`ErrorDomain`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/error/trait.ErrorDomain.html
 #[proc_macro_derive(ErrorDomain, attributes(error_domain))]
 #[proc_macro_error]
 pub fn error_domain_derive(input: TokenStream) -> TokenStream {
@@ -524,8 +526,8 @@ pub fn error_domain_derive(input: TokenStream) -> TokenStream {
 /// struct MyBoxed(String);
 /// ```
 ///
-/// [`BoxedType`]: subclass/boxed/trait.BoxedType.html
-/// [`glib::Value`]: value/struct.Value.html
+/// [`BoxedType`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/boxed/trait.BoxedType.html
+/// [`glib::Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
 #[proc_macro_derive(Boxed, attributes(boxed_nullable, boxed_type))]
 #[proc_macro_error]
 pub fn boxed_derive(input: TokenStream) -> TokenStream {
@@ -552,8 +554,8 @@ pub fn boxed_derive(input: TokenStream) -> TokenStream {
 /// struct MyShared(std::sync::Arc<MySharedInner>);
 /// ```
 ///
-/// [`SharedType`]: subclass/shared/trait.SharedType.html
-/// [`glib::Value`]: value/struct.Value.html
+/// [`SharedType`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/shared/trait.SharedType.html
+/// [`glib::Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
 #[proc_macro_derive(SharedBoxed, attributes(shared_boxed_nullable, shared_boxed_type))]
 #[proc_macro_error]
 pub fn shared_boxed_derive(input: TokenStream) -> TokenStream {
@@ -590,7 +592,7 @@ pub fn shared_boxed_derive(input: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
-/// [`ObjectSubclass`]: subclass/types/trait.ObjectSubclass.html
+/// [`ObjectSubclass`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/types/trait.ObjectSubclass.html
 #[proc_macro_attribute]
 #[proc_macro_error]
 pub fn object_subclass(_attr: TokenStream, item: TokenStream) -> TokenStream {
@@ -615,7 +617,7 @@ pub fn object_subclass(_attr: TokenStream, item: TokenStream) -> TokenStream {
 /// type Prerequisites = ();
 /// ```
 ///
-/// [`ObjectInterface`]: interface/types/trait.ObjectInterface.html
+/// [`ObjectInterface`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/interface/trait.ObjectInterface.html
 #[proc_macro_attribute]
 #[proc_macro_error]
 pub fn object_interface(_attr: TokenStream, item: TokenStream) -> TokenStream {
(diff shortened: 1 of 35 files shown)


sdroege:
Nice, thanks!

glib-macros/src/lib.rs
```diff
@@ -437,6 +437,8 @@ pub fn closure_local(item: TokenStream) -> TokenStream {
 ///     ValWithCustomNameAndNick,
 /// }
 /// ```
+///
+/// [`glib::Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
```
sdroege, line 441: This is not useful as you'd point to the stable version of the docs. @GuillaumeGomez @bilelmoussaoui how can this be solved better?

glib-macros/src/variant_derive.rs
```diff
@@ -134,7 +137,7 @@ pub fn derive_variant_for_struct(
             let to_variant = quote! {
                 impl #generics #glib::ToVariant for #ident #generics {
                     fn to_variant(&self) -> #glib::Variant {
-                        #glib::Variant::tuple_from_iter(::std::array::IntoIter::new([
+                        #glib::Variant::tuple_from_iter(::std::iter::IntoIterator::into_iter([
```
sdroege, line 140: Why?

glib-macros/src/lib.rs
```diff
@@ -728,8 +728,75 @@ pub fn downgrade(input: TokenStream) -> TokenStream {
...
+/// [`glib::flags`](macro@flags).
+///
+/// # Example
+///
+/// ```
+/// use glib::prelude::*;
+///
+/// #[derive(Debug, PartialEq, Eq, glib::Variant)]
+/// enum Foo {
+///     MyA,
```
sdroege, line 755: How is this serialized? Will it be an `s` for this variant and `sv` for the others? That seems annoying to deserialize :)

glib-macros/src/variant_derive.rs
```diff
@@ -202,3 +205,360 @@ pub fn derive_variant_for_struct(
...
+
+    let tag_type = mode.tag_type();
+
+    let static_variant_type = if has_data {
+        format!("({}v)", tag_type)
+    } else {
+        tag_type.to_string()
+    };
+
+    let (to_variant, from_variant) = if has_data {
```
sdroege, line 257: Maybe move the two branches of this `if` into two separate functions. That's easier to read.

glib/src/enums.rs
```diff
@@ -590,6 +590,45 @@ impl FlagsClass {
...
+    // rustdoc-stripper-ignore-next
+    /// Converts a string of nicks `s` separated by `|` to an integer value.
+    pub fn from_nick_string(&self, mut s: &str) -> Result<u32, ParseFlagsError> {
+        s = s.trim_start();
+        if s.is_empty() {
+            return Ok(0);
+        }
+        let mut value = 0u32;
+        loop {
+            let sep = s.find('|').unwrap_or_else(|| s.len());
```
sdroege, line 619: Why not iterate over `s.split('|').map(|s| s.trim())`?

glib/src/enums.rs
```diff
@@ -625,6 +664,15 @@ impl Clone for FlagsClass {
     }
 }
 
+#[derive(Debug)]
+pub struct ParseFlagsError(pub String);
```
sdroege, line 668: Why a public field?

sdroege:
Seems good to me otherwise

glib-macros/src/lib.rs
bilelmoussaoui: as glib-macros doesn't depend on glib itself, i don't see how this could be done without hardcoding one or another version. Unless the `--extern-html-root-url` could help somehow here, not sure.

glib/src/enums.rs
```diff
@@ -625,6 +664,15 @@ impl Clone for FlagsClass {
...
+
```
bilelmoussaoui, line 669: Shouldn't this impl std::error::Error?

glib/src/object.rs
```diff
@@ -1783,7 +1783,7 @@ pub trait ObjectExt: ObjectType {
     /// Limits the lifetime of `closure` to the lifetime of the object. When
     /// the object's reference count drops to zero, the closure will be
     /// invalidated. An invalidated closure will ignore any calls to
-    /// [`Closure::invoke`](crate::Closure::invoke).
+    /// [`invoke`](crate::closure::RustClosure::invoke).
```
bilelmoussaoui, line 1786: Normally you use only the method name if it's on the same type which is not the case here

glib-macros/src/variant_derive.rs
jf2048: I was getting deprecation warnings on nightly: https://doc.rust-lang.org/nightly/std/array/struct.IntoIter.html#method.new

glib-macros/src/lib.rs
jf2048: It's `(sv)` for all of the enum variants in `Foo`, that's the `StaticVariantType` and it won't violate that. `s` is only used if the enum only has unit variants, i.e. cases where you could put an integer discriminator after any of the variants.

sdroege: Ok, great :)

glib-macros/src/variant_derive.rs
sdroege: That seems problematic. If the crate using the macro is still in the 2018 edition then the `IntoIterator::into_iter()` will have different behaviour (and not work here). For the 2021 edition both have the same behaviour though.

Maybe add an `#[allow]` for the deprecation to the generated code here until we stop supporting the 2018 edition.

glib-macros/src/lib.rs
jf2048: Currently these urls are everywhere in the macro docs, I think it could technically be a link to `../glib/value/struct.Value.html` but I seem to remember there was some issue with that.

bilelmoussaoui: Should normally work I guess, we do that in the book

sdroege:
@jf2048 Do you plan to update this PR some time in the next days? Otherwise I can probably take it over and finish it so we can get these changes merged :)

sdroege:
@jf2048 Ping? :)

glib/src/enums.rs
jf2048: Changed to a `impl From for String`

jf2048:
Currently this is broken in 1.56.0 because we need to do this to support flags:

```rs
#[derive(glib::Variant)]
#[glib::flags(name = "Variant10")]
```

But using an attribute macro after a derive macro was only stabilized on 1.57.0. It is working in stable.

sdroege:
@GuillaumeGomez @bilelmoussaoui OK for you two to update the minimum supported version to 1.57?

glib-macros/src/variant_derive.rs
sdroege: This is still relevant as the generated code would now fail to compile with the 2018 edition

glib/src/enums.rs
```diff
@@ -631,6 +658,23 @@ impl Clone for FlagsClass {
...
+impl std::error::Error for ParseFlagsError {}
+
+impl fmt::Display for ParseFlagsError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Unknown flag: '{}'", self.0)
+    }
+}
+
+impl From<ParseFlagsError> for String {
```
sdroege, line 672: Do we need to have a public constructor of the parse error type?

sdroege (commented):

bilelmoussaoui:
> @GuillaumeGomez @bilelmoussaoui OK for you two to update the minimum supported version to 1.57?

Sounds good to me

GuillaumeGomez:
> @GuillaumeGomez @bilelmoussaoui OK for you two to update the minimum supported version to 1.57?

Sounds good to me as well. Please don't forget to update `minimum_rust_version` in `Cargo.toml` as well.

sdroege:
Yeah, needs update in `Cargo.toml`, `README.md` and `.github/workflows` AFAICS. Not sure if anywhere else too.

glib/src/enums.rs
jf2048: This is the other way around :slightly_smiling_face:  I just wanted a way to get at the underlying string, I changed it to an explicit getter method

glib-macros/src/variant_derive.rs
jf2048: I am not seeing it on CI, the depreciation warning is hidden by the macro anyway

jf2048:
I can bump the version in gtk3-rs and gtk4-rs too when this is merged

sdroege (approved):

sdroege:
> I can bump the version in gtk3-rs and gtk4-rs too when this is merged

Please do, thanks! :)

PR #436, opened 2021-12-28
regenerate with latest pango gir file

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gir-files/pull/90


pango/src/auto/attr_list.rs
```diff
@@ -84,10 +95,26 @@ impl AttrList {
             ffi::pango_attr_list_update(self.to_glib_none().0, pos, remove, add);
         }
     }
+
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    #[doc(alias = "pango_attr_list_from_string")]
+    pub fn from_string(text: &str) -> Option<AttrList> {
```
bilelmoussaoui, line 102: This should probably return a Result & have a FromStr implementation

pango/src/auto/enums.rs
```diff
@@ -1004,6 +1010,8 @@ pub enum FontScale {
     Superscript,
     #[doc(alias = "PANGO_FONT_SCALE_SUBSCRIPT")]
     Subscript,
+    #[doc(alias = "PANGO_FONT_SCALE_SMALL_CAPS")]
+    SmallCaps,
```
bilelmoussaoui, line 1014: This addition is fine because the enum itself was added in 1.50

pango/src/auto/tab_array.rs
```diff
@@ -92,4 +134,37 @@ impl TabArray {
...
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    #[doc(alias = "pango_tab_array_sort")]
+    pub fn sort(&mut self) {
+        unsafe {
+            ffi::pango_tab_array_sort(self.to_glib_none_mut().0);
+        }
+    }
+
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
```
bilelmoussaoui, line 147: those functions probably shouldn't take mut self, but all the functions in this type seems to be doing so already so... I will leave cleaning that up to someone that needs it/knows how it should be used

pango/src/auto/enums.rs
sdroege: Adding variants to enums is also generally not a problem

pango/Gir.toml
```diff
@@ -310,3 +313,36 @@ boxed_inline = true
...
+status = "generate"
+    [[object.member]]
+    name = "right"
+    version = "1.50"
+    [[object.member]]
+    name = "center"
+    version = "1.50"
+    [[object.member]]
+    name = "decimal"
+    version = "1.50"
```
sdroege, line 328: You created an issue/MR to mark these enums correctly in C?

pango/src/auto/font.rs
```diff
@@ -27,6 +33,28 @@ glib::wrapper! {
...
 impl Font {
     pub const NONE: Option<&'static Font> = None;
+
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    #[doc(alias = "pango_font_deserialize")]
+    pub fn deserialize(
+        context: &Context,
+        bytes: &glib::Bytes,
+    ) -> Result<Option<Font>, glib::Error> {
```
sdroege, line 43: This is not actually `nullable`, is it?

pango/src/auto/font.rs
```diff
@@ -74,6 +102,11 @@ pub trait FontExt: 'static {
     #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_44")))]
     #[doc(alias = "pango_font_has_char")]
     fn has_char(&self, wc: char) -> bool;
+
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    #[doc(alias = "pango_font_serialize")]
+    fn serialize(&self) -> Option<glib::Bytes>;
```
sdroege, line 109: And this probably also not `nullable`

pango/src/auto/layout.rs
```diff
@@ -421,6 +427,18 @@ impl Layout {
         }
     }
 
+    #[cfg(any(feature = "v1_50", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+    #[doc(alias = "pango_layout_serialize")]
+    pub fn serialize(&self, flags: LayoutSerializeFlags) -> Option<glib::Bytes> {
```
sdroege, line 433: And neither this

pango/src/auto/tab_array.rs
sdroege: Yes the getters at least shouldn't. Please create an issue :)

pango/src/auto/enums.rs
bilelmoussaoui: my comment was about not having a cfg guard for this but yeah

pango/src/auto/layout.rs
bilelmoussaoui: All of those are caused by trust_nullability :/

pango/src/auto/tab_array.rs
bilelmoussaoui: It seems to break the build, let me fix that :)

pango/Gir.toml
bilelmoussaoui: Well, as there's no proper annotations for enum members; they are kind of documented already https://docs.gtk.org/Pango/enum.Variant.html

pango/src/auto/font.rs
bilelmoussaoui: it seems to be nullable https://gitlab.gnome.org/GNOME/pango/-/blob/main/pango/serializer.c#L1543

pango/src/auto/layout.rs
sdroege: Yes, if we opt in to that we need to keep the upstream annotations correct

pango/src/enums.rs
```diff
@@ -0,0 +1,147 @@
...
+#[cfg(any(feature = "v1_50", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+use std::fmt;
+
+#[cfg(any(feature = "v1_50", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
+#[non_exhaustive]
+#[doc(alias = "PangoLayoutDeserializeError")]
+pub enum LayoutDeserializeError {
```
sdroege, line 20: Why make this manual?

pango/Gir.toml
sdroege: There is, see https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/230 . We use that extensively in GStreamer.

bilelmoussaoui: Oh wow, didn't knew that. So why we always have to set them manually so far? :/

pango/src/enums.rs
bilelmoussaoui: See the commit & todo in the code. gir uses a wrong error quark function name (that doesn't even exists??) and we can't disable only that part of the enum

sdroege: Isn't there a configuration for the error quark?

pango/Gir.toml
sdroege: Because gobject-introspection can only do that since a year :)

pango/src/enums.rs
bilelmoussaoui: I looked the config code but didn't find any, ideally the code in library postprocessing needs to be fixed as the quark is properly defined on the gir file.

bilelmoussaoui:
Hmm i messed up the rebase

gio/src/auto/flags.rs
```diff
@@ -2003,6 +2007,8 @@ bitflags! {
         const STDERR_MERGE = ffi::G_SUBPROCESS_FLAGS_STDERR_MERGE as u32;
         #[doc(alias = "G_SUBPROCESS_FLAGS_INHERIT_FDS")]
         const INHERIT_FDS = ffi::G_SUBPROCESS_FLAGS_INHERIT_FDS as u32;
+        #[doc(alias = "G_SUBPROCESS_FLAGS_SEARCH_PATH_FROM_ENVP")]
+        const SEARCH_PATH_FROM_ENVP = ffi::G_SUBPROCESS_FLAGS_SEARCH_PATH_FROM_ENVP as u32;
```
sdroege, line 2011: Needs a version flag

pango/src/auto/attr_list.rs
```diff
@@ -86,10 +97,26 @@ impl AttrList {
...
```
sdroege, line 104: Should probably be a `Result`

pango/src/auto/font.rs
sdroege, line 43: Can this return `None` without an error?

pango/src/enums.rs
```diff
@@ -0,0 +1,147 @@
...
+    MissingValue,
+    #[doc(hidden)]
+    __Unknown(i32),
+}
+
+#[cfg(any(feature = "v1_50", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_50")))]
+impl LayoutDeserializeError {
+    #[doc(alias = "pango_layout_deserialize_error_quark")]
+    pub fn quark() -> Quark {
```
sdroege, line 35: Should implement https://gtk-rs.org/gtk-rs-core/git/docs/glib/error/trait.ErrorDomain.html instead

pango/src/auto/attr_list.rs
sdroege: And implement `FromStr`?

bilelmoussaoui: I am pretty sure I did this at some point...meh, got some changes lost during the rebase

bilelmoussaoui: Never mind, I did that for TabArray. Will do!

pango/src/auto/font.rs
bilelmoussaoui: As mentioned in prior comments, it can return None yes

sdroege: How very confusing, thanks :)

pango/src/enums.rs
sdroege: You do a few lines below, but having a separate function for the quark is useless

pango/Gir.toml
bilelmoussaoui: https://gitlab.gnome.org/GNOME/pango/-/merge_requests/575

pango/Gir.toml
```diff
@@ -310,3 +324,48 @@ boxed_inline = true
...
+
+[[object]]
+name = "Pango.TabArray"
+status = "generate"
+    [[object.function]]
+    pattern = "(to_string|get_decimal_point|get_positions_in_pixels|get_tab)"
+        [[object.function.parameter]]
+        pattern = "*" # because we can't target self
```
sdroege, line 347: Can't we? It probably works if you use the parameter name from the C function instead of `self`

sdroege (approved):

PR #438, opened 2021-12-28
gio: mark `content_type_guess` parameter as filename

euclio:


bilelmoussaoui:
Ideally this needs to be fixed upstream

euclio:
I see, what would that fix look like?

sdroege:
Like here: https://gitlab.gnome.org/GNOME/glib/-/blob/62a88d9f38c77f140fb7f6d714fb4445277ce5e4/glib/gfileutils.c#L210

sdroege (approved):

euclio:
Upstream MR: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2411

PR #443, opened 2021-12-30
gio: move Task::run_in_thread to an extension trait

pbor:
The trait is implemented only when returning values that implement ToSendValue.

### DIFF
--- a/examples/gio_task/file_size/ffi.rs
+++ b/examples/gio_task/file_size/ffi.rs
@@ -51,7 +51,7 @@ pub unsafe extern "C" fn my_file_size_get_file_size_async(
             .imp();
 
         source_object.size.replace(Some(size));
-        task.return_value(&size.to_value());
+        task.return_value(&size);
     });
 }
 
@@ -70,7 +70,7 @@ pub unsafe extern "C" fn my_file_size_get_file_size_finish(
         .unwrap()
         .propagate_value()
     {
-        Ok(v) => v.get::<i64>().unwrap(),
+        Ok(v) => v,
         Err(e) => {
             *error = e.into_raw();
             0
--- a/gio/src/task.rs
+++ b/gio/src/task.rs
@@ -5,7 +5,7 @@ use crate::Cancellable;
 use crate::Task;
 use glib::object::IsA;
 use glib::translate::*;
-use glib::value::{ToSendValue, ValueType};
+use glib::value::ValueType;
 use libc::c_void;
 use std::boxed::Box as Box_;
 use std::ptr;
@@ -60,9 +60,18 @@ impl Task {
             ffi::g_task_set_priority(self.to_glib_none().0, priority.into_glib());
         }
     }
+}
 
+pub trait TaskExtManual<T: ValueType> {
     #[doc(alias = "g_task_return_value")]
-    pub fn return_value<V: ValueType>(&self, result: &V) {
+    fn return_value(&self, result: &T);
+
+    #[doc(alias = "g_task_propagate_value")]
+    fn propagate_value(&self) -> Result<T, glib::Error>;
+}
+
+impl<T: ValueType> TaskExtManual<T> for Task {
+    fn return_value(&self, result: &T) {
         unsafe extern "C" fn value_free(value: *mut c_void) {
             let _: glib::Value = from_glib_full(value as *mut glib::gobject_ffi::GValue);
         }
@@ -76,8 +85,7 @@ impl Task {
         }
     }
 
-    #[doc(alias = "g_task_propagate_value")]
-    pub fn propagate_value<V: ValueType>(&self) -> Result<V, glib::Error> {
+    fn propagate_value(&self) -> Result<T, glib::Error> {
         unsafe {
             let mut error = ptr::null_mut();
             let value = ffi::g_task_propagate_pointer(self.to_glib_none().0, &mut error);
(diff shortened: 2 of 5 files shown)


pbor:
This seems to address the soundness issue. However

1. It is not ergonomic
2. Is calling the trait `TaskSendManualExt` the right convention? or should it simply be `TaskSend`

I wonder if to improve ergonomics we can make the closure passed to `run_in_thread` return `Result<Value, Error>` and call `return_value` and `return_error` ourselves... would the compiler be able to infer the value type?

sdroege:
That's not how I was thinking of this. I would've done it as `gio::Task<T: ToValue + FromValue>`. That way you could also make sure to only implement `Send+Sync` for `T: Send` and could simplify a lot of the API here to not have to work on `glib::Value`s anymore.

bilelmoussaoui:
Oups sorry I misclicked and marked it as ready for review

pbor:
Actually maybe we do not need neither the extension trait, nor implementing Task manually... How about this approach?

gio/src/task.rs
```diff
@@ -60,11 +61,42 @@ impl Task {
...
+            ffi::g_task_return_pointer(
+                self.to_glib_none().0,
+                result.to_value().to_glib_full() as *mut _,
+                Some(value_free),
+            )
+        }
+    }
+
+    #[doc(alias = "g_task_propagate_value")]
+    pub fn propagate_value<V: ValueType>(&self) -> Result<V, glib::Error> {
```
pbor, line 80: maybe simply `propagate()`?

gio/src/task.rs
```diff
@@ -60,11 +61,42 @@ impl Task {
         }
     }
 
+    #[doc(alias = "g_task_return_value")]
+    pub fn return_value<V: ValueType>(&self, result: &V) {
```
pbor, line 65: Maybe even go with `return_result<V: ValueType>(&self, result: Result<&V, glib:Error>)` and remove `return_error`?

gio/src/task.rs
```diff
@@ -60,11 +61,42 @@ impl Task {
...
+        unsafe {
+            let mut error = ptr::null_mut();
+            let value = ffi::g_task_propagate_pointer(self.to_glib_none().0, &mut error);
+            if !error.is_null() {
+                return Err(from_glib_full(error));
+            }
+            let value =
+                Option::<glib::Value>::from_glib_full(value as *mut glib::gobject_ffi::GValue)
+                    .unwrap_or_else(|| glib::Value::from_type(glib::types::Type::UNIT));
+            Ok(V::from_value(&value))
```
pbor, line 90: This needs to be checked more carefully... can this fail? What happens if we return the UNIT value and V is something else?

gio/src/task.rs
```diff
@@ -60,11 +61,42 @@ impl Task {
...
+    }
+
     #[doc(alias = "g_task_run_in_thread")]
-    pub fn run_in_thread<Q>(&self, task_func: Q)
+    pub fn run_in_thread<Q, V>(&self, task_func: Q)
     where
-        Q: FnOnce(&Self, Option<&glib::Object>, Option<&Cancellable>),
+        Q: FnOnce(&Self, Option<&glib::Object>, Option<&Cancellable>) -> Result<V, glib::Error>,
         Q: Send + 'static,
+        V: ValueType + ToSendValue,
```
sdroege, line 99:
```suggestion
        V: ValueType + Send,
```

That's easier to read and probably also works :)

sdroege:
> Actually maybe we do not need neither the extension trait, nor implementing Task manually... How about this approach?

That would mean that `Task` can't be `Send` though

examples/gio_task/file_size/ffi.rs
```diff
@@ -70,7 +70,7 @@ pub unsafe extern "C" fn my_file_size_get_file_size_finish(
         .unwrap()
         .propagate_value()
     {
-        Ok(v) => v.get::<i64>().unwrap(),
+        Ok(v) => v,
         Err(e) => {
             *error = e.into_raw();
```
pbor, line 75: This should check error is not null (problem was there before this patch)

gio/src/task.rs
```diff
@@ -59,11 +60,58 @@ impl Task {
...
+            }
+            let value =
+                Option::<glib::Value>::from_glib_full(value as *mut glib::gobject_ffi::GValue)
+                    .unwrap_or_else(|| glib::Value::from_type(glib::types::Type::UNIT));
+            Ok(T::from_value(&value))
+        }
+    }
+}
+
+pub trait TaskExtManualSend<T: ValueType + Send>: TaskExtManual<T> {
```
sdroege, line 103: These two traits can be in a single one btw, you can make the functions generic over `T: ValueType` vs `T: ValueType + Send`

gio/src/task.rs
```diff
@@ -59,11 +60,58 @@ impl Task {
...
         }
     }
+}
+
+pub trait TaskExtManual<T: ValueType> {
+    #[doc(alias = "g_task_return_value")]
+    fn return_value(&self, result: &T);
+
+    #[doc(alias = "g_task_propagate_value")]
+    fn propagate_value(&self) -> Result<T, glib::Error>;
```
sdroege, line 70: Are these two functions thread-safe? E.g. can you call safely `return_value()` from the original thread where the `Task` was created and inside the `run_in_thread()` closure at the same time?

sdroege: The answer is: no, for both functions. That's yet another problem with this API then.

pbor:
Closing in favor of https://github.com/gtk-rs/gtk-rs-core/pull/446

PR #444, opened 2021-12-30
glib: manually implement ParamSpec::is_valid_name

bilelmoussaoui:


bilelmoussaoui:
Ok I'm lost here, the `g_param_spec_is_valid_name`  definitely exists on the ffi file and with the correct guard flag.

sdroege:
> Ok I'm lost here, the `g_param_spec_is_valid_name` definitely exists on the ffi file and with the correct guard flag.

See my fixup. The `Cargo.toml` change was missing.

glib/src/param_spec.rs
```diff
@@ -205,6 +211,36 @@ impl ParamSpec {
     //pub fn steal_qdata(&self, quark: /*Ignored*/glib::Quark) -> /*Unimplemented*/Option<Fundamental: Pointer> {
     //    unsafe { TODO: call gobject_ffi::g_param_spec_steal_qdata() }
     //}
+
+    #[doc(alias = "g_param_spec_internal")]
+    pub fn internal(
```
sdroege, line 216: I think exposing it like this is not very useful. You'd have to provide some more infrastructure like for `GObject`s, and also expose `g_param_type_register_static()` / `GParamSpecTypeInfo` in some way (by providing some trait, I guess).

bilelmoussaoui: Right, I don't plan on doing so as I don't really need it. I guess we can keep it and whoever needs that can provide that infra or drop the commit, whatever you prefer

sdroege: I would drop it as whatever we're going to expose here will look different

sdroege:
You dropped the `Cargo.toml` change

sdroege (approved):

PR #445, opened 2022-01-01
Fix new clippy lints

GuillaumeGomez:
I'll need to send a PR to gir too to fix the missing `#[must_use]` in the auto code.

### DIFF
--- a/cairo/src/device.rs
+++ b/cairo/src/device.rs
@@ -157,7 +157,7 @@ impl Device {
             let status = ffi::cairo_device_acquire(self.to_raw_none());
             status_to_result(status)?;
         }
-        Ok(DeviceAcquireGuard { 0: self })
+        Ok(DeviceAcquireGuard(self))
     }
 
     #[doc(alias = "cairo_device_release")]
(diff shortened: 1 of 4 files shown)


cairo/src/device.rs
```diff
@@ -157,7 +157,7 @@ impl Device {
             let status = ffi::cairo_device_acquire(self.to_raw_none());
             status_to_result(status)?;
         }
-        Ok(DeviceAcquireGuard { 0: self })
```
sdroege, line 160: I didn't consider that this was valid syntax :rofl:

gio/src/subclass/action_map.rs
```diff
@@ -101,7 +101,7 @@ unsafe extern "C" fn action_map_lookup_action<T: ActionMapImpl>(
 
         let mut map = wrap
             .steal_qdata::<HashMap<String, Action>>(*ACTION_MAP_LOOKUP_ACTION_QUARK)
-            .unwrap_or_else(HashMap::new);
+            .unwrap_or_else_default();
```
sdroege, line 104:
```suggestion
            .unwrap_or_default();
```

cairo/src/device.rs
GuillaumeGomez: It's very convenient when writing macros. :3

sdroege:
Looks good otherwise, thanks!

GuillaumeGomez:
Once the gir PR is in, I'll send PRs to all repositories (including gstreamer ;) ).

sdroege (approved):

PR #446, opened 2022-01-01
Make gio::Task binding sound and support generics in the `wrapper!` macro

pbor:
The PR covers two things

1. It makes `gio:Task` bindings sound
2. It adds the ability to call the `wrapper!` macro on generic types

The two things are in a single PR because 1 needs 2, but at the same time 2 on its own would not have any caller to prove it works. The two things split into multiple commits so that it at least allows to review and rework one part at a time.


glib/src/object.rs
```diff
@@ -614,37 +614,37 @@ impl FromGlibPtrArrayContainerAsVec<*mut GObject, *const *mut GObject> for Objec
 /// ObjectType implementations for Object types. See `wrapper!`.
 #[macro_export]
 macro_rules! glib_object_wrapper {
-    (@generic_impl [$($attr:meta)*] $visibility:vis $name:ident, $ffi_name:ty, $ffi_class_name:ty, @type_ $get_type_expr:expr) => {
+    (@generic_impl [$($attr:meta)*] $visibility:vis $name:ident $(<$($gen:tt $(: $bound:tt)? ),+>)?, $ffi_name:ty, $ffi_class_name:ty, @type_ $get_type_expr:expr) => {
```
GuillaumeGomez, line 617:
```suggestion
    (@generic_impl [$($attr:meta)*] $visibility:vis $name:ident $(<$($gen:ident $(: $bound:type)+* ),+>)?, $ffi_name:ty, $ffi_class_name:ty, @type_ $get_type_expr:expr) => {
```

Something like this maybe?

pbor:
After discussion in the chat channel, the updated patch:

1. handles multiple bounds (eg <T: A + B>)
2. uses `ident` instead of `tt` for the generic name (it still uses tt for the bounds since they can be both types and lifetimes and we do not care about making the distinction)
3. it is implemented for all wrapper macros (object, boxed, boxed_inline, and shared)

glib/src/boxed.rs
```diff
@@ -15,31 +15,33 @@ use std::ptr;
...
+        $crate::glib_boxed_wrapper!(@generic_impl [$($attr)*] $visibility $name $(<$($generic $(: $($bound)* $bound2)?),+>)?, $ffi_name);
+
         $crate::glib_boxed_wrapper!(
-            @memory_manager_impl $name, $ffi_name,
+            @memory_manager_impl $name $(<$($generic $(: $($bound)* $bound2)?),+>)?, $ffi_name,
             @copy $copy_arg $copy_expr, @free $free_arg $free_expr
         );
-        $($crate::glib_boxed_wrapper!(@value_impl $name, $ffi_name, @type_ $get_type_expr);)?
+
+        $crate::glib_boxed_wrapper!(@value_impl $name $(<$($generic $(: $($bound)* $bound2)?),+>)?, $ffi_name $(, @type_ $get_type_expr)?);
```
pbor, line 28: For this change I very much need review, because I am not sure my workaround is correct.

The old code did: `$($recurse!($params);)?` 
I assume that it meant: if params are specified then recurse, otherwise ignore and expand to nothing.

Unfortunately this cannot work now, because params contains multiple nested meta variables and it gives:

```
error: meta-variable `gen` repeats 0 times, but `get_type_expr` repeats 1 time
  --> glib/src/boxed.rs:28:10
   |
28 |         $($crate::glib_boxed_wrapper!(@value_impl $name $(<$($gen $(: $bound)?),+>)?, $ffi_name, @type_ $get_type_expr);)?
```
The way I worked it around instead is to make only the the last param `$(, @type_ $get_type_expr)?` optional, and below in the `@value_impl` match the case where `@type_` is not specified and expand to nothing.

pbor:
Adding some notes (also for myself):

1. The wrapper macro with generics seems to work (yay!)
2. To make Object work I had to add PhantomData with the generic in the struct. The same will have to be done for Boxed etc
3. It sucks that if you use `T` as the name of the generic type things explode because it conflicts with what used internally in object.rs
4. Object assumes a bunch of traits (Hash, Clone, Debug) are available... I added them for task, not sure it is what we want
5. The patch includes the port of Task to try out the new functionality. However the open questions in https://github.com/gtk-rs/gtk-rs-core/pull/443 with regard to the API and implementation still hold true

bilelmoussaoui:
> It sucks that if you use T as the name of the generic type things explode because it conflicts with what used internally in object.rs

Maybe we could switch to something less common than T?

gio/src/task.rs
```diff
@@ -2,34 +2,59 @@
...
+use std::mem::transmute;
 use std::ptr;
 
-impl Task {
+// Implemented manually to make it generic over the return type to ensure the API
+// is sound when the task is moved across threads.
+
+glib::wrapper! {
+    #[doc(alias = "GTask")]
+    pub struct Task<V: ValueType + Hash + Ord + Clone + Debug>(Object<ffi::GTask, ffi::GTaskClass>) @implements AsyncResult;
```
sdroege, line 25: Why do we need `Hash + Ord + Clone + Debug`? The corresponding impls added by `glib::wrapper!` don't really need those trait bounds as they just operate on the pointer itself. You can implement them  inside the macro via

```rust
impl<V> Clone for Task<V> { ... }
```

etc

gio/src/task.rs
```diff
@@ -84,13 +259,14 @@ impl Task {
...
-        unsafe extern "C" fn trampoline<Q>(
+        unsafe extern "C" fn trampoline<V, Q>(
             task: *mut ffi::GTask,
             source_object: *mut glib::gobject_ffi::GObject,
             user_data: glib::ffi::gpointer,
             cancellable: *mut ffi::GCancellable,
         ) where
-            Q: FnOnce(&Task, Option<&glib::Object>, Option<&Cancellable>),
+            V: ValueType + Hash + Ord + Clone + Debug,
+            Q: FnOnce(&Task<V>, Option<&glib::Object>, Option<&Cancellable>),
```
sdroege, line 269: The source object also has to be of a `Send` type for this to be correct

gio/src/task.rs
pbor: Well, without those it would not compile :)

I considered adding them from the macro, but I was not sure if that was too implicit

sdroege: That's what I'm saying. You should not need them here at all if you implement the traits manually inside the macro instead of using `#[derive]`.

pbor:
> > It sucks that if you use T as the name of the generic type things explode because it conflicts with what used internally in object.rs
> 
> Maybe we could switch to something less common than T?

I now changed the internal use to `OT` since it is an `ObjectType`. I wonder if there is a better workaround to make generic types hygienic rather than "pick an uncommon name"

gio/src/task.rs
```diff
@@ -60,10 +86,156 @@ impl Task {
...
+        unsafe {
+            let mut error = ptr::null_mut();
+            let value = ffi::g_task_propagate_pointer(self.to_glib_none().0, &mut error);
+            if !error.is_null() {
+                return Err(from_glib_full(error));
+            }
+            let value =
+                Option::<glib::Value>::from_glib_full(value as *mut glib::gobject_ffi::GValue)
+                    .unwrap_or_else(|| glib::Value::from_type(glib::types::Type::UNIT));
+            Ok(V::from_value(&value))
```
pbor, line 228: Please double check this... I am not convinced it is correct in the `else` case of the line above

gio/src/task.rs
```diff
@@ -99,44 +273,23 @@ impl Task {
             let task_func: Box_<Q> = Box::from_raw(user_data as *mut _);
             task_func(
                 task.as_ref(),
-                source_object.as_ref().as_ref(),
+                source_object.as_ref().as_ref().map(|s| s.unsafe_cast_ref()),
```
pbor, line 276: double check this, because I am not sure about the use of unsafe cast

pbor:
I think this is now more or less ready, so I am dropping the "Draft" flag. I have updated the title and initial description to be more accurate.

Known things to check before merging:
[ ] two lines in the task part that need double checking (see comments above)
[ ] one change in the macro part that changes the macro logic a bit (see comment above)
[ ] the macro for `boxed`, `boxed_inline`, and `shared` do not yet have the `PhantomData` to hold generics... I would rather do this as a second step once there is a use case to test with
[ ] check if you agree with the API changes for task, some of them are my preference but are not strictly needed for the soundness issue

Clearly any other feedback outside of the list above is also welcome

gio/src/task.rs
```diff
@@ -60,10 +85,188 @@ impl Task {
...
+                Box_::into_raw(f),
+            )
+        }
+    }
+
+    #[doc(alias = "g_task_return_value")]
+    #[doc(alias = "g_task_return_boolean")]
+    #[doc(alias = "g_task_return_int")]
+    #[doc(alias = "g_task_return_pointer")]
+    #[doc(alias = "g_task_return_error")]
```
GuillaumeGomez, line 193: I think we generate too many doc aliases...

glib/src/boxed.rs
```diff
@@ -15,68 +15,82 @@ use std::ptr;
 /// Wrapper implementations for Boxed types. See `wrapper!`.
 #[macro_export]
 macro_rules! glib_boxed_wrapper {
-    ([$($attr:meta)*] $visibility:vis $name:ident, $ffi_name:ty,
+    ([$($attr:meta)*] $visibility:vis $name:ident $(<$($generic:ident $(: $bound:tt $(+ $bound2:tt)*)?),+>)?, $ffi_name:ty,
```
GuillaumeGomez, line 18: Nicely written! :+1:

gio/src/task.rs
sdroege: What problem does it cause?

GuillaumeGomez (approved):
Macros update looks good to me!

gio/src/task.rs
GuillaumeGomez: None, it's just unhelpful.

sdroege: Why? It only shows up if you search for exactly that string, i.e. exactly that C function name. And then you find what you're looking for.

GuillaumeGomez: I guess you're right.

sdroege: If it was a fuzzy search like for the other things then I would agree with you

examples/gio_task/file_size/mod.rs
```diff
@@ -8,6 +8,8 @@ glib::wrapper! {
     pub struct FileSize(ObjectSubclass<imp::FileSize>);
 }
 
+unsafe impl Send for FileSize {}
```
sdroege, line 11:
```suggestion
unsafe impl Send for FileSize {}
unsafe impl Sync for FileSize {}
```

Either a `GObject` is both or neither because of refcounting.

Is `imp::FileSize` both? Otherwise this would be unsound.

gio/src/task.rs
```diff
@@ -99,44 +304,23 @@ impl Task {
...
             ffi::g_task_run_in_thread(self.to_glib_none().0, Some(task_func));
         }
     }
+}
 
-    pub fn return_value(&self, result: &glib::Value) {
-        unsafe extern "C" fn value_free(value: *mut c_void) {
-            let _: glib::Value = from_glib_full(value as *mut glib::gobject_ffi::GValue);
-        }
+unsafe impl<V: ValueType + Send> Send for Task<V> {}
```
sdroege, line 319: Same as above, this also would have to be `Sync` then (is it?). If it is not then also `run_in_task()` needs some changes as that effectively uses the task from a different thread.

AFAIU calling `return_value()` and `propagate_value()` from two threads at once is not thread-safe, and that would be a problem. You will probably need to add some more types here to allow calling these two functions only from the other thread and not from elsewhere.

----

Another problem here is that it would only be `Send` if the store callback (passed to `new()`) is also `Send`.

I think you'll need to duplicate some functions (`new()`, `connect_*()`) into a variant with `ValueType` and `ValueType + Send` so the trait bounds are fitting.

gio/src/task.rs
sdroege: That's correct

glib/src/boxed.rs
sdroege: Indeed, the macro changes look all good

glib/src/object.rs
```diff
@@ -913,66 +967,52 @@ macro_rules! glib_object_wrapper {
...
 
-        impl $crate::types::StaticType for $name {
+        impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? $crate::types::StaticType for $name $(<$($generic),+>)? {
             fn static_type() -> $crate::types::Type {
                 #[allow(unused_unsafe)]
                 unsafe { $crate::translate::from_glib($get_type_expr) }
             }
         }
 
-        impl<T: $crate::object::ObjectType> std::cmp::PartialEq<T> for $name {
```
sdroege, line 923: Why did you remove these?

glib/src/value.rs
```diff
@@ -484,13 +484,14 @@ impl Value {
...
-            ptr::read(&s.0)
+            ptr::read(&s.inner)
         }
     }
 
     pub fn try_into_send_value<T: Send + StaticType>(self) -> Result<SendValue, Self> {
         if self.type_().is_a(T::static_type()) {
-            Ok(SendValue(self.into_raw()))
+            let v = unsafe { SendValue::unsafe_from(self.into_raw()) };
+            Ok(v)
```
sdroege, line 494:
```suggestion
            unsafe { Ok(SendValue::unsafe_from(self.into_raw())) }
```

gio/src/task.rs
```diff
@@ -60,10 +85,188 @@ impl Task {
...
+
+            #[cfg(not(feature = "v2_64"))]
+            {
+                let value = ffi::g_task_propagate_pointer(self.to_glib_none().0, &mut error);
+
+                if error.is_null() {
+                    let value = Option::<glib::Value>::from_glib_full(
+                        value as *mut glib::gobject_ffi::GValue,
+                    )
+                    .unwrap_or_else(|| glib::Value::from_type(glib::types::Type::UNIT));
```
sdroege, line 254: What are you trying to do here? If the return value is `NULL` then this should probably just panic or return an `Err`?

sdroege: I.e. it can only be `NULL` if nothing was returned yet, and then it's a programmer error to call this function so it should probably just panic.

glib/src/object.rs
pbor: I will double check this, could be a mistake when resolving conflicts in a rebase

gio/src/task.rs
pbor: It was handled in the old code, hence my doubts...

glib/src/object.rs
pbor: As far as I can see I just moved them up a few lines to group them together with the other manually implemented traits

gio/src/task.rs
sdroege: You wrote the old code ;)

pbor:
As discussed in chat, this last patch reworks the Task API and splits LocalTask and Task... see commit message for details

gio/src/task.rs
```diff
@@ -2,68 +2,306 @@
...
 
+impl<V: ValueType + Send> Task<V> {
     #[doc(alias = "g_task_run_in_thread")]
-    pub fn run_in_thread<Q>(&self, task_func: Q)
+    pub fn run_in_thread<S, Q>(&self, task_func: Q)
     where
-        Q: FnOnce(&Self, Option<&glib::Object>, Option<&Cancellable>),
+        S: IsA<glib::Object> + Send,
+        Q: FnOnce(Self, Option<&S>, Option<&Cancellable>),
         Q: Send + 'static,
```
sdroege, line 305:
```suggestion
        Q: FnOnce(Self, Option<&S>, Option<&Cancellable> + Send + 'static,
```

sdroege: And also below

sdroege (commented):
Seems good to me apart from those small things. Thanks for working on this :)

sdroege (approved):

PR #453, opened 2022-01-06
gio: require Send also for the source object

pbor:


sdroege (approved):
Seems good, but if you make it a generic parameter on `Task` too then you can also provide the correct target type in the `run_in_task` closure :)

pbor:
Dunno... since you have to explicitly do `task: gio::LocalTask<i64>` etc in callbacks etc, having more generics that are often not used would be annoying imho

sdroege:
True. I guess we can worry about that when someone needs it

sdroege:
Needs a `cargo fmt` run

sdroege (approved):

PR #455, opened 2022-01-07
gio: mark creating Task as unsafe and add public documentation

pbor:
Even with all the reccent improvements to the Task API, it is
still possible to bypass the requirements, for example by
casting LocalTask to Task, since they have the same underlying
GType.

At this point mark the constructors as unsafe and document
which invariants need to be manually esured by the callers.

The goal of this API is to enable callers to implement in
Rust APIs exposed to the C ffi which follow the glib
convention for async calls. Pure Rust applications should
probably use better abstractions to dispatch asynchronous
tasks.

### DIFF
--- a/gio/src/task.rs
+++ b/gio/src/task.rs
@@ -1,5 +1,27 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+// rustdoc-stripper-ignore-next
+//! `Task` bindings.
+//!
+//! The API distinguishes between `LocalTask` and `Task`, where the latter can be used
+//! with threads, but has more strict requirements, in particular it requires its value,
+//! source object, and callbacks must implement the `Send` trait.
+//!
+//! The `Task` and `LocalTask` types provide idiomatic access to glib's `GTask` API, for
+//! instance by being generic over their value type, while not completely departing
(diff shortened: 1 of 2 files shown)


pbor:
Should I drop the unsafe on `return_*` and `propagate ` at this point?

Also let me know if the docs will pick up this content correctly

gio/src/task.rs
```diff
@@ -1,5 +1,27 @@
...
+//! instance by being generic over their value type, while not completely departing
+//! from the underlying C API.
+//! Unfortunately this API does not allow to automatically enforce all the
+//! invariants required to be a completely safe abstraction.
+//!
+//! For this reason, the constructors of `Task` and `LocalTask` are marked as
+//! unsafe and when using these types the caller is responsible to ensure the
+//! following requirements are satisfied
+//!
+//! * You should not crete a `LocalTask`, upcast it to a `glib::Object` and then
```
sdroege, line 20:
```suggestion
//! * You should not create a `LocalTask`, upcast it to a `glib::Object` and then
```

sdroege (dismissed):

sdroege (approved):

PR #458, opened 2022-01-09
gio: generate SimpleProxyResolver

pbor:
Needed to `session.set_proxy_resolver(gio::SimpleProxyResolver::new(proxy_uri, None));` on a `SoupSession`


gio/Gir.toml
```diff
@@ -990,6 +990,20 @@ status = "generate"
...
+status = "generate"
+manual_traits = ["SimpleProxyResolverExtManual"]
+    [[object.function]]
+    name = "new"
+    # takes vec of strings
+    manual = true
+    [[object.function]]
+    name = "set_ignore_hosts"
+    # takes vec of strings
+    ignore = true
```
bilelmoussaoui, line 1004:
```suggestion
    manual = true
```

gio/src/simple_proxy_resolver.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use crate::ProxyResolver;
+use crate::SimpleProxyResolver;
+use glib::object::IsA;
+use glib::translate::*;
+
+impl SimpleProxyResolver {
+    #[doc(alias = "g_simple_proxy_resolver_new")]
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new(default_proxy: Option<&str>, ignore_hosts: Option<&[&str]>) -> ProxyResolver {
```
sdroege, line 11: Are `Some([])` and `None` handled differently for the `ignore_hosts`?

gio/src/simple_proxy_resolver.rs
```diff
@@ -0,0 +1,35 @@
...
+                default_proxy.to_glib_none().0,
+                ignore_hosts.to_glib_none().0,
+            ))
+        }
+    }
+}
+
+pub trait SimpleProxyResolverExtManual: 'static {
+    #[doc(alias = "g_simple_proxy_resolver_set_ignore_hosts")]
+    fn set_ignore_hosts(&self, ignore_hosts: &[&str]);
```
sdroege, line 23: Inconsistency here. This one takes a slice, the constructor an `Option`

gio/Gir.toml
```diff
@@ -990,6 +990,20 @@ status = "generate"
...
     ignore = true
 
+[[object]]
+name = "Gio.SimpleProxyResolver"
+status = "generate"
+manual_traits = ["SimpleProxyResolverExtManual"]
+    [[object.function]]
+    name = "new"
+    # takes vec of strings
+    manual = true
```
sdroege, line 1000: What is autogenerated here?

gio/src/simple_proxy_resolver.rs
pbor: I copied what gir autogenerated and just changed the `&[&str]`...

gio/Gir.toml
pbor: it generates a plain `&str`... not sure why, the annotation for array seems to be there in the C code

gio/src/simple_proxy_resolver.rs
sdroege: If we do it manually, let's no put an `Option` here. From the C code, both are handled exactly the same and all the other APIs around `ignore_hosts` are not `nullable`.

sdroege (approved):

PR #465, opened 2022-01-12
glib: impl From<Infallible> for error types

arcnmx:
This facilitates conversion between potentially-fallible and infallible `Result<>`s, similar to some [std impls](https://doc.rust-lang.org/std/convert/enum.Infallible.html#impl-From%3CInfallible%3E)

### DIFF
--- a/glib/src/error.rs
+++ b/glib/src/error.rs
@@ -6,6 +6,7 @@
 use crate::translate::*;
 use crate::Quark;
 use std::borrow::Cow;
+use std::convert::Infallible;
 use std::error;
 use std::ffi::CStr;
 use std::fmt;
@@ -115,6 +116,12 @@ impl fmt::Debug for Error {
 }
 
 impl error::Error for Error {}
+
+impl From<Infallible> for Error {
+    fn from(e: Infallible) -> Self {
+        match e {}
+    }
+}
 
 // rustdoc-stripper-ignore-next
 /// `GLib` error domain.
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -41,6 +41,7 @@
 //! ```
 
 use libc::{c_char, c_void};
+use std::convert::Infallible;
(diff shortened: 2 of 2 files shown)


sdroege (approved):
Makes sense to me. I guess `thiserror` should also automatically do that :)

sdroege:
Needs a `cargo fmt` run

arcnmx:
> Needs a cargo fmt run

Yeah I wasn't sure since I didn't see a rustfmt toml in the repo. It might be worth introducing an empty one so people don't need to use `cargo fmt -- --config-path /dev/null` or similar, which seems unintuitive.

> I guess thiserror should also automatically do that :)

Though it's generally only particularly important for errors that specifically might be used as conversion errors, mm it would be helpful if it could!
Unrelated question though: why use a `syn` macro (thiserror) in such a foundational crate for something so trivial? Would you accept a PR to remove it?

sdroege:
> Unrelated question though: why use a `syn`

`glib-macros` also depends on `syn` (and `glib` depends on `glib-macros`), so depending on `thiserror` is not really adding anything in addition with regards to build time or dependency tree size.

arcnmx:
> > Unrelated question though: why use a `syn`
> 
> `glib-macros` also depends on `syn` (and `glib` depends on `glib-macros`), so depending on `thiserror` is not really adding anything in addition with regards to build time or dependency tree size.

Hm, maybe I'll open a new issue to discuss this then and perhaps explore making it optional? It's been wreaking havoc on my builds due to `syn` feature flag changes causing massive rebuilds (say I have a workspace with crates X and Y, where X depends on glib, Y depends on X and a different syn macro - building either X or Y causes the entire dependency tree to be thrown out as dirty due to the feature flag changes). But to be fair that is a cargo bug, and I don't know how integral the macros are to glib's internal implementation (rather than downstream crate users).

PR #466, opened 2022-01-13
Update gir files

sdroege:


bilelmoussaoui (approved):

bilelmoussaoui:
I suppose it would be nice to deprecate/not use pango_coverage_ref/pango_coverage_unref if 1_52 feature is enabled?

sdroege:
> I suppose it would be nice to deprecate/not use pango_coverage_ref/pango_coverage_unref if 1_52 feature is enabled?

Yeah but that's not so easy to do because its usage is in `glib::wrapper!`

PR #476, opened 2022-01-15
Update versions to 0.15.1

sdroege:


sdroege:
I also changed all dependency versions so nothing ever tries to pull in 0.15.0. I'll also yank 0.15.0 from crates.io after 0.15.1 is published.

bilelmoussaoui (approved):

PR #483, opened 2022-01-15
implement To/FromVariant for OS strings and paths

euclio:
The impls use the platform-specific byte representation for serialization, so they're `cfg`'d for `unix` and `windows`.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -9,9 +9,18 @@
 //! `Variant` types are described by [`VariantType`](../struct.VariantType.html)
 //! "type strings".
 //!
-//! Although `GVariant` supports arbitrarily complex types, this binding is
-//! currently limited to the basic ones: `bool`, `u8`, `i16`, `u16`, `i32`,
-//! `u32`, `i64`, `u64`, `f64`, `&str`/`String`, and [`VariantDict`](../struct.VariantDict.html).
+//! Although `GVariant` supports arbitrarily complex types, this crate only provides bindings for a
+//! subset of common types. This subset includes primitives, strings, data structures from the
+//! standard library, and [`VariantDict`](crate::VariantDict). See [`ToVariant#foreign-impls`] for
+//! a full list of supported types. You may also implement [`ToVariant`] and [`FromVariant`]
+//! manually, or derive them using the [`Variant`](derive@crate::Variant) derive macro.
+//!
+//! ## Portability Warning
+//!
+//! Variant conversion traits are also implemented for certain platform-specific data types from
+//! the standard library like [`PathBuf`](std::path::PathBuf) and [`OsString`](std::ffi::OsString).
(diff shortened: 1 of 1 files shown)


sdroege (commented):
This seems a bit suboptimal because the variant type is different between platforms, so serializations created on e.g. Linux won't be possible to open on Windows. Not sure if that's acceptable.

@bilelmoussaoui?

bilelmoussaoui:
> This seems a bit suboptimal because the variant type is different between platforms, so serializations created on e.g. Linux won't be possible to open on Windows. Not sure if that's acceptable.
> 
> @bilelmoussaoui?

I think it is okay to merge this if it is properly documented in https://gtk-rs.org/gtk-rs-core/git/docs/glib/variant/index.html

sdroege:
> I think it is okay to merge this if it is properly documented in https://gtk-rs.org/gtk-rs-core/git/docs/glib/variant/index.html

Where would you document it? Both on the `impl` and at the top somewhere?

bilelmoussaoui:
> > I think it is okay to merge this if it is properly documented in https://gtk-rs.org/gtk-rs-core/git/docs/glib/variant/index.html
> 
> Where would you document it? Both on the `impl` and at the top somewhere?

Something like that yes, also maybe add an example on the page i mentioned above

euclio:
Pushed a new revision with the requested documentation updates.

bilelmoussaoui (approved):

glib/src/variant.rs
```diff
@@ -9,9 +9,18 @@
 //! `Variant` types are described by [`VariantType`](../struct.VariantType.html)
 //! "type strings".
 //!
-//! Although `GVariant` supports arbitrarily complex types, this binding is
-//! currently limited to the basic ones: `bool`, `u8`, `i16`, `u16`, `i32`,
-//! `u32`, `i64`, `u64`, `f64`, `&str`/`String`, and [`VariantDict`](../struct.VariantDict.html).
+//! Although `GVariant` supports arbitrarily complex types, this crate only provides bindings for a
```
sdroege, line 12: The caveat at the beginning is not really true anymore. We support arbitrarily complex types in the bindings now.

sdroege (commented):
Looks good to me otherwise

glib/src/variant.rs
```diff
@@ -9,9 +9,18 @@
...
+//! subset of common types. This subset includes primitives, strings, data structures from the
+//! standard library, and [`VariantDict`](crate::VariantDict). See [`ToVariant#foreign-impls`] for
```
sdroege, line 14:
```suggestion
//! `GVariant` supports arbitrarily complex types built from primitives like integers, floating point
//! numbers, strings, arrays, tuples and dictionaries. See [`ToVariant#foreign-impls`] for
```

sdroege:
I'll backport this to 0.15.3 some time in the next days.

PR #488, opened 2022-01-15
gio: move task docs on the structs

pbor:
Module level documentation is not picked up by the gtk-rs doc tool.


bilelmoussaoui:
lgtm, also note the original C docs will be added before your "manual" docs, which is not great. I believe we need a feature in gir to disable docs for a specific type/function if we want to manually document it like here.

sdroege (approved):

Issue #489, opened 2022-01-16
glib: Provide bindings for glib::markup_escape_text


A6GibKm:
The function https://docs.gtk.org/glib/func.markup_escape_text.html is very useful to escape markup, it might take into account different gtk/pango-related details that other escapers might not do.

Looking at `glib/Gir.toml` I see
```toml
    [[object.function]]
    pattern = "markup_.*"
    ignore = true # they are supposed to be used with MarkupParser, which doesn't have safe bindings
```
but I do not think the reasoning applies for this function.

PR #490, opened 2022-01-16
Generate bindings for markup_escape_text

A6GibKm:
Fixes: https://github.com/gtk-rs/gtk-rs-core/issues/489

### DIFF
--- a/glib/src/auto/functions.rs
+++ b/glib/src/auto/functions.rs
@@ -463,6 +463,12 @@ pub fn main_depth() -> i32 {
     unsafe { ffi::g_main_depth() }
 }
 
+#[doc(alias = "g_markup_escape_text")]
+pub fn markup_escape_text(text: &str) -> crate::GString {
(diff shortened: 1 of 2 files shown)


bilelmoussaoui (approved):

glib/src/auto/functions.rs
```diff
@@ -463,6 +463,12 @@ pub fn main_depth() -> i32 {
     unsafe { ffi::g_main_depth() }
 }
 
+#[doc(alias = "g_markup_escape_text")]
+pub fn markup_escape_text(text: &str) -> crate::GString {
+    let length = text.len() as isize;
+    unsafe { from_glib_full(ffi::g_markup_escape_text(text.to_glib_none().0, length)) }
```
sdroege, line 469: How inconvenient that this needs a copy of the string... but that will be fixed properly for the next release together with my other plans :)

sdroege (approved):

PR #493, opened 2022-01-18
Derive class name in glib::object_subclass

andy128k:
Derive boilerplate `const NAME: &'static str = ...;` in `glib::object_subclass` proc macro.

### DIFF
--- a/glib-macros/src/object_subclass_attribute.rs
+++ b/glib-macros/src/object_subclass_attribute.rs
@@ -13,6 +13,7 @@ pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
     let mut has_interfaces = false;
     let mut has_instance = false;
     let mut has_class = false;
+    let mut has_name = false;
     for item in &input.items {
         match item {
             syn::ImplItem::Method(method) => {
@@ -31,6 +32,12 @@ pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
                     has_instance = true;
                 } else if name == "Class" {
                     has_class = true;
+                }
+            }
+            syn::ImplItem::Const(constant) => {
+                let name = &constant.ident;
+                if name == "NAME" {
+                    has_name = true;
                 }
             }
             _ => {}
@@ -86,6 +93,16 @@ pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
         Some(quote!(type Instance = #crate_ident::subclass::basic::InstanceStruct<Self>;))
     };
 
+    let name_opt = if has_name {
+        None
+    } else {
+        base_type_name(&*input.self_ty).map(|name| {
+            quote!(
+                const NAME: &'static str = concat!(module_path!(), "_", stringify!(#name));
+            )
+        })
+    };
+
     let trait_path = match &trait_ {
         Some(path) => &path.1,
         None => abort_call_site!(WRONG_PLACE_MSG),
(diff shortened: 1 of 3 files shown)


glib-macros/src/object_subclass_attribute.rs
```diff
@@ -13,16 +14,17 @@ pub fn impl_object_subclass(input: &syn::ItemImpl) -> TokenStream {
     let mut has_interfaces = false;
     let mut has_instance = false;
     let mut has_class = false;
+    let mut has_name = false;
     for item in &input.items {
         match item {
             syn::ImplItem::Method(method) => {
-                let name = method.sig.ident.to_string();
+                let name = &method.sig.ident;
```
andy128k, line 21: Collateral change. No need to allocate here.

glib-macros/tests/test.rs
```diff
@@ -279,6 +279,59 @@ fn subclassable() {
...
+        use glib::subclass::prelude::*;
+
+        mod imp {
+            use super::*;
+
+            #[derive(Default)]
+            pub struct Bar {}
+
+            #[glib::object_subclass]
+            impl ObjectSubclass for super::imp::Bar {
```
andy128k, line 316: `super::imp::Bar` is redundant but shows how name is derived. It is `SuperImpBar` and not just a `Bar`.

A reason to reject this PR?

ids1024:
I think the concern with this was that just using the type might make it easy to accidentally end up with multiple types using the same name (which is a problem).

I believe Vala prepends something based the source path/module path (not sure exactly). Not currently possible without something like https://doc.rust-lang.org/proc_macro/struct.Span.html#method.source_file being stabilized.

Though in general, I'd definitely like to see `NAME` be be made optional.

glib-macros/tests/test.rs
bilelmoussaoui: No opinion here but that name would be displayed for cases like GtkWidgets in the inspector. Having all the names prefixed with SuperImp or Imp would be confusing. I guess we could strip those two prefixes?

bilelmoussaoui:
If this was to be merged, the other macros that requires a type name should have their name attribute marked as optional and default to something similar to what is done here for `object_subclass` if possible

sdroege:
I don't like this very much. By having it automatically derived it gives the impression that you don't have to worry about namespacing, but unfortunately GObject type names have a flat namespace. Now if you have two types called `Button` in different modules, it would end up with `Button` as GObject type name and things would explode at runtime.

If we can get the whole module path from the crate here somehow plus the crate name as prefix then I wouldn't have any objections as long as the name can still be given manually.

ids1024:
Crate name should be possible by outputting code using `env!("CARGO_PKG_NAME")` and `concat!`. I don't think there would be any (stable) way to get the module path.

andy128k:
I see few ways here:

1. Use `Span::source_file()` when it is stabilized;
2. Use `std::module_path!()` with a runtime penalty of cleaning name in `glib::register_type`. Smth like `sanitize_type_name(T::NAME)`;
3. Use `std::any::TypeId::of` when its `const`antness will be stabilized;
3. Change `const NAME` to `fn name() -> String`;
4. Salt name with a (semi)random string (I know, this is crazy);

I sympathize the idea that default behavior must be robust an error-free. But I am not sure if it is even possible. Someone may use trickery like [package renaming](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?highlight=rename,depende#renaming-dependencies-in-cargotoml),

---- 

Alternatively, there may be an additional parameter added to the existing macro (e.g. `#[object_subclass(namespace = Foo)]`) but it, obviously, does not add any value against a good ol'  `'static str`.

Issue #27, opened 2020-07-23
Add derive macro for structs representing GObject properties


sdroege:
Kind of related to https://github.com/gtk-rs/glib/issues/637 and I would do them together in a few days.

I'll write out the whole plan here once I'm at a computer again, but the basic idea would be a derive macro and corresponding infrastructure that
- registers a GObject property per struct field, incl default, min, max values etc
- provides  a set/get property function that directly does the translation
- provides a notify function for notifying changes

The struct itself would then be stored inside the instance data in a refcell or mutex.

This should allow to bring down the boilerplate for GObject subclasses with properties considerably.

bilelmoussaoui:
I suppose this can be closed now?

Issue #214, opened 2021-07-27
Macro for generating signals() and emit_/connect_ functions for subclass


ids1024:
When defining a GObject type in Rust, it is necessary to define a `ObjectImpl::signals()` function, and useful to define the `emit_` and `connect_` functions like the ones generated by `gir`.

Considering a few things, this seems like a reasonable API for it, which I think would work:

```rust
glib::wrapper! {
    signals {
        fn signal_without_params();
        fn signal_with_params(i32, String);
        fn signal_with_params_and_return(i32, String) -> bool;       
    }
    pub struct FooObject(ObjectSubclass<FooObjectInner>);
}
```

I guess it would be necessary for the caller to define `ObjectImpl::signals()` to call a function generated by the macro. Kind of similar to the boilerplate required with composite templates. I can't think of a way to avoid a slight complication like this.

Alternately to making this part of `glib::wrapper!`, it could be a separate macro, or an attribute macro on something. Given `glib::wrapper!` is necessary, it seems cleaner to add it here. It could also generate the `connect_`/`emit_` functions for types other than ObjectSubclass (code currently generated in gir).

Properties should be handled in a similar way, but that's more complicated since it would be good for it to automatically store a value if there is no custom getter/setter, while also supporting custom getters and setters. And then there's a need to define nick, blurb, and default value.

ids1024:
> Kind of similar to the boilerplate required with composite templates. I can't think of a way to avoid a slight complication like this.

Or maybe that's not a problem. `signals()` could be a method of `ObjectSubclassIs` (the trait implemented when `glib::wrapper!` is used with `ObjectSubclass<...>`).

`ObjectImpl::signals()` could have a default implementation calling `Self::Type::signals()` (or a different function name). Or alternately it could be removed (making this a breaking change) if a manual implementation of `signals` isn't really useful.

jf2048:
I have implemented a (very WIP) version of this in another repo: [gobject-impl](https://github.com/jf2048/gobject-impl)

But let's wait until we figure out what is going on with #546 first before I try to port it upstream

PJungkamp:
I think the macro should take inspiration from the existing `gtk4::template_callbacks` and `glib::properties` macros to keep the developer experience consistent.

From `gtk4::template_callbacks`:
- use a proc-macro on an `impl` block and generate code for all functions tagged with `#[signal(...)]`

From `glib::Properties`:
[...]

PR #494, opened 2022-01-18
Properties macro

ranfdev:
The implementation is a mess... I will clean it up.

Take a look at the newly added test case to see how it can be used.
I'm interested in comments about the API:

```rust
#[derive(Props, Default)]
pub struct Foo {
    #[prop("bar", "bar", "This is a bar", None, glib::ParamFlags::READWRITE)]
    bar: RefCell<String>,
    #[prop("buzz", "buzz", "This is a buzz", 1, 100, 1, glib::ParamFlags::READWRITE)]
    buzz: RefCell<u32>,
}
```
Currently you can't set more than one flag at a time. Everything else seems to work. Oh, and I need to finish implementing the `Parametrize` trait for all the type that have a `ParamSpec`.

Related to #27


glib-macros/tests/props.rs
```diff
@@ -0,0 +1,57 @@
...
+                bar: RefCell<String>,
+                #[prop("buzz", "buzz", "This is a buzz", 1, 100, 1, glib::ParamFlags::READWRITE)]
+                buzz: RefCell<u32>,
+            }
+
+            #[glib::object_subclass]
+            impl ObjectSubclass for Foo {
+                const NAME: &'static str = "MyFoo";
+                type Type = super::Foo;
+            }
```
bilelmoussaoui, line 25: What happens if you want to override ObjectImpl?

glib/src/param_spec.rs
```diff
@@ -1370,6 +1370,77 @@ impl ParamSpecVariant {
...
+        self.to_value()
+    }
+}
+impl<T: Parametrize + ToValue> Parametrize for std::cell::RefCell<T> {
+    type Spec = T::Spec;
+    fn get(&self) -> crate::Value {
+        self.borrow().to_value()
+    }
+}
+impl<T: 'static + Parametrize + for <'b> crate::value::FromValue<'b>> ParametrizeMut for std::cell::RefCell<T> {
```
bilelmoussaoui, line 1436: You will need similar implementations for Cell/Mutex.

glib-macros/src/lib.rs
```diff
@@ -734,3 +734,74 @@ pub fn variant_derive(input: TokenStream) -> TokenStream {
...
+            quote!(<#ty as glib::Parametrize>::Spec::new(#nested))
+        } else {
+            unreachable!("Invalid attribute params")
+        }
+    });
+    let expanded = quote! {
+        use glib::{Parametrize, ParametrizeMut};
+        impl ObjectImpl for #name {
+            fn properties() -> &'static [glib::ParamSpec] {
+                use once_cell::sync::Lazy;
```
bilelmoussaoui, line 771: Can't you use the re-exported one from glib?

glib-macros/tests/props.rs
```diff
@@ -0,0 +1,57 @@
...
+        use glib::subclass::prelude::*;
+        use glib_macros::Props;
+        use std::cell::RefCell;
+
+        mod imp {
+            use super::*;
+
+            #[derive(Props, Default)]
+            pub struct Foo {
+                #[prop("bar", "bar", "This is a bar", None, glib::ParamFlags::READWRITE)]
```
bilelmoussaoui, line 15: I think those should be named attributes and they should have sane defaults if/when possible. Something to note is in general people would want to use a specific function for the setters, so there should be some way to set that.

glib-macros/tests/props.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        glib::wrapper! {
+            pub struct Foo(ObjectSubclass<imp::Foo>);
+        }
+    }
+
+    let myfoo: foo::Foo = glib::object::Object::new(&[]).unwrap();
+
+    let bar: String = myfoo.property("bar");
```
bilelmoussaoui, line 46: Something that would be nice to have is to auto generate the getter/setter/connect_notify by default

glib-macros/tests/props.rs
bilelmoussaoui: I think a way to work around this is to move the properties specific stuff to a separate trait and make it a requirement to implement ObjectImpl. If we ever do that, we might want to do the same thing for signals as well to avoid doing yet another API break once there is some kind of macro to generate signals code.

bilelmoussaoui:
> Currently you can't set more than one flag at a time

I would not use the flags directly but make them separate attributes like (read/write/notify) and map them internally to the corresponding flag

glib-macros/tests/props.rs
bilelmoussaoui: Once they are named attributes, you could make the first three optional and use the field name as the default value

glib-macros/tests/props.rs
BrainBlasted: There should be some way to override the getter/setter at least, since you may want to implement custom behavior.

bilelmoussaoui: Yes, see my comment above.

ids1024:
Vala supports two types of properties (https://wiki.gnome.org/Projects/Vala/Manual/Classes#Properties):

> Properties can either be declared with code that will perform particular actions on get and set, or can simply declare which actions are allowed and allow Vala to implement simple get and set methods. This second pattern (automatic property) will result in fields being added to the class to store values that the property will get and set. If either get or set has custom code, then the other must either be also written in full, or omitted altogether. 

So properties can be backed in something like a `RefCell` with automatic get/set implementations, or manual get/set which may not even have a struct member. I'm honestly not sure what a good Rust API to support both use cases would be, but it seems it should be done in some way.

In addition to properties, GObject has signals, and I think the two should be considered together. (I wrote an idea for signals in https://github.com/gtk-rs/gtk-rs-core/issues/214). Luckily signals are simpler in comparison.

glib/src/param_spec.rs
```diff
@@ -1370,6 +1370,77 @@ impl ParamSpecVariant {
     }
 }
 
+pub trait Parametrize {
+    type Spec;
+    fn get(&self) -> crate::Value;
+}
+pub trait ParametrizeMut {
+    fn set(&self, value: &crate::Value);
```
sdroege, line 1378: Why the `set` / `get` methods here? You could simply require that all the types implement `ValueType` and then you get all that for free :)

glib/src/param_spec.rs
```diff
@@ -1370,6 +1370,77 @@ impl ParamSpecVariant {
...
+}
+
+use crate::ToValue;
+impl Parametrize for String {
+    type Spec = ParamSpecString;
+    fn get(&self) -> crate::Value {
+        self.to_value()
+    }
+}
+impl Parametrize for crate::Object {
```
sdroege, line 1388: This should be moved into the `glib::wrapper!` macro

glib/src/param_spec.rs
sdroege: And `RWLock` and `Box`

glib-macros/tests/props.rs
sdroege: Also it should be possible to manually implement the property/signal related methods on `ObjectImpl`. We shouldn't make the macro mandatory as it won't be able to cover all possible use-cases.

glib/src/param_spec.rs
sdroege: I guess because you also implement that on top of wrapper types like `RefCell`. I think this should be split into two traits somehow. One for the underlying values, and another for the storage

glib-macros/tests/props.rs
bilelmoussaoui: @sdroege What do you think about what I suggested above with splitting the properties/signals into separate traits this way you can still auto-implement them without conflicting with a manual ObjectImpl implementation?

glib/src/param_spec.rs
```diff
@@ -1370,6 +1370,77 @@ impl ParamSpecVariant {
     }
 }
 
+pub trait Parametrize {
+    type Spec;
```
sdroege, line 1374: In my proposal I had multiple traits like this so that you can map the different cases: no default/range, only default, default and range.

And then there are strange cases like valuearray that take another pspec for the element type.

glib-macros/tests/props.rs
```diff
@@ -0,0 +1,57 @@
...
+    mod foo {
+        use glib::prelude::*;
+        use glib::subclass::prelude::*;
+        use glib_macros::Props;
+        use std::cell::RefCell;
+
+        mod imp {
+            use super::*;
+
+            #[derive(Props, Default)]
```
sdroege, line 13: It would be good to allow deriving this on a separate struct. That's the pattern used in all the GStreamer plugins, see e.g. [here](https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/763ad0cb184e9b68bd215c9e5fe63f391a581e2d/audio/audiofx/src/audioecho/imp.rs#L39). Those are then also not using `RefCell` or the like inside the struct but only for storing the whole struct.

The reason there is that you usually want to get a snapshot of all property values at a specific time and then use that snapshot for one round of calculations, instead of possibly changing values.

glib-macros/tests/props.rs
sdroege: Yes, that's what I also suggested in my proposal some months ago.

glib-macros/tests/props.rs
ranfdev: I see, that's useful

glib/src/param_spec.rs
ranfdev: I've seen you proposal, but I still wanted to try reusing the already existing Spec::new constructor. Probably I'll end up implementing things as in your proposal anyway...

glib/src/param_spec.rs
ranfdev: I'm currently writing some bad, dirty code, because it's my first proc macro. The first commits are going to be messy.
Then I'll move everything to the right place.

glib/src/param_spec.rs
ranfdev: Makes sense

glib-macros/tests/props.rs
ranfdev: Ok, I'll do it

glib-macros/tests/props.rs
ranfdev: Makes sense, I'll do it

glib-macros/src/lib.rs
ranfdev: Ok right

glib/src/param_spec.rs
ranfdev: Can't implement ParametrizeMut on Box since it's not mutable. I'll do it for the other ones

ranfdev:
The latest API is:
```rust
#[derive(Props, Default)]
pub struct Foo {
    #[prop(get, set)]
    bar: Mutex<String>,
    #[prop(get = Self::hello_world)]
    _buzz: RefCell<String>,
    #[prop(get, set = Self::set_fizz, name = "fizz")]
    fizz: RefCell<String>,
}
```

glib-macros/tests/props.rs
ranfdev: You can do it with the latest API:

```rust
#[derive(Default)]
struct Author {
    name: String
}


#[derive(Props, Default)]
pub struct Foo {
    author: RefCell<Author>,
    #[prop(get = Self::get_author_name)]
    author_name: PhantomData<String>,
}
```

`PhantomData<T>` is needed because the macro must choose the correct `ParamSpec` for `T`.
By setting a custom getter, you can return anything you want, even another internal field

sdroege: How would that work with multiple properties for the same struct? You would just add multiple phantom data and in the getter/setter function you do the right thing then?

sdroege: Maybe there could be some additional convenience though. You could allow something like `self.author.name` as getter/setter and the macro does the refcell + field access itself?

Maybe something like

```rust
#[derive(Default)]
struct Author {
    first_name: String,
    last_name: String,
}


#[derive(Props, Default)]
pub struct Foo {
    author: RefCell<Author>,
    #[prop(delegate = self.author.first_name)]
    author_first_name: String,
    #[prop(delegate = self.author.last_name)]
    author_last_name: String,
}
```

and whenever you see a `delegate` you would remove the field from the struct after macro expansion. That way you wouldn't need the `PhantomData`. Not sure if you can remove fields from derive macros. The alternative would be to make them comments (you can parse comments from proc macros), e.g.

```rust
#[derive(Props, Default)]
pub struct Foo {
    author: RefCell<Author>,
    #[prop(delegate = self.author.first_name)]
    // author_first_name: String,
    #[prop(delegate = self.author.last_name)]
    // author_last_name: String,
}
```

ranfdev: I don't know... I feel that's a layer of abstraction too high...
To implement `delegate` I need to create a `Monad`-like abstraction over all the types implementing interior mutability, because I need to support `RefCell`, `Mutex`, `RwLock`, [...].

I feel this goes out of the scope "create a glib property easily".

With the current API you can already do this, inline:
`get = |t: &Self| t.author.borrow().first_name.to_value()`

It's long, but I feel that's not gtk-rs fault.

I don't want to add too much "magic" behind the scenes...

Anyway, I'll keep your comments in mind while I try to implement the missing things

glib-macros/tests/props.rs
ranfdev: After searching a bit, I think this is impossible. I would have to make two derive macros communicate, but that's not possible:
```rust
mod imp {
  #[derive(Props)]
  struct Foo{
    #[prop(get, set)]
    name: RefCell<String>
  }
}

// Not possible, I can't get the data from `Props` macro
#[derive(PropsAccess)]
struct Foo{}

// Best we can do
#[derive(PropsAccess)]
#[props[(name: String), (prop_name: PropType), ...]]
struct Foo{}
```

jf2048:
Nice to see someone working on this! :smiley: It's looking good, I was thinking about this for the last few months and I had a few ideas:

- Should it have flags to auto generate getters and setters? Maybe notify methods and pspec accessors too?
   ```rust
  #[prop(auto_get(pub), auto_set)]
  fizz: RefCell<String>,
  ```
  Generates something like:
  ```rust
   impl <Foo as ObjectSubclass>::Type {
       pub fn fizz(&self) -> Ref<String> {
           self.imp().fizz.borrow()
       }
        fn set_fizz(&self, value: String) {
            if self.imp().fizz.borrow() != value {
                self.imp().fizz.replace(value)
                self.fizz_notify();
            }
        }
        fn fizz_notify(&self) {
            self.notify_by_pspec(Self::fizz_pspec());
        }
        fn fizz_pspec() -> &glib::ParamSpec {
            <Self as ObjectSubclass>::properties()[2]
        }
   }
   ```
- Or should it generate a "fields" enum to index the properties array, like is common when writing GObjects in C?
    ```rust
    #[derive(Props, Default)]
    #[glib_props(fields)]
    struct Foo {
        // ...
    }
    ```
  Generates something like:
    ```rust
    enum FooField {
        Bar = 0,
        Buzz = 1,
        Fizz = 2,
    }
    impl Foo {
        fn pspec(field: FooField) -> &glib::ParamSpec {
            <Self as ObjectSubclass>::properties()[field as usize]
       }
    }
    ```
- Should it allow for a custom implementation of `ObjectImpl`, if someone wants to layer functionality on top of the generated code?
    ```rust
    #[derive(Props, Default)]
    #[glib_props(no_trait)]
    struct Foo { /* ... */ }
    ```
    Generates something like:
    ```rust
    impl Foo {
        fn properties() { /* ... */ }
        fn property() { /* ... */ }
        fn set_property() { /* ... */ }
    }
    ```

ranfdev:
> auto generate ...

This is already implemented:
`#[prop(get)]` -> autogenerate
`#[prop(get = func)]` -> custom
`#[prop()]` -> not readable

> should it generate a "fields" enum

Yes, I could do it, makes sense.

> Should it allow for a custom implementation of ObjectImpl, if someone wants to layer functionality on top of the generated code?

The general opinion is "yes". My current idea is to define a trait `AutoProps` and then call that in `ObjectImpl`.

bilelmoussaoui:
> > Should it have flags to auto generate getters and setters?
> 
> Sorry @jf2048, I misread. You actually want to generate getters/setters on the _wrapper_ type. I don't know if this is technically possible. Read the comments above, there's another thread about this in the code comments.

Well, the code suggested by @jf2048 should actually work just fine because it doesn't `impl T`` but instead it does` impl <T as ObjectSubclass>::Type` which is the public type :)

ranfdev:
That syntax is invalid:
```rust
impl <#name as ObjectSubclass>::Type {
    fn hi() {}
}
```
```
error[E0118]: no nominal type found for inherent implementation
  --> glib-macros/tests/props.rs:20:22
   |
20 |             #[derive(Props, Default)]
   |                      ^^^^^ impl requires a nominal type
   |
   = note: either implement a trait on it or create a newtype to wrap it instead
```
Explanation: https://stackoverflow.com/questions/49277733/what-is-a-nominal-type-in-the-context-of-an-inherent-implementation

bilelmoussaoui:
@ranfdev right, but you can generate a trait and implement it for T::Type no?

jf2048:
>you can generate a trait and implement it for T::Type no?

That's what I was getting at, I have been able to do that in some of my experiments. Sorry I should have been clearer, my code above was some really rough pseudocode.

ranfdev:
> but you can generate a trait and implement it for T::Type no?

Yes, I thought about it, but then when you import the type you must import the related trait too, else you don't get access to any method.

It's the best option for now, but it's not perfect

glib-macros/tests/props.rs
sdroege: > To implement delegate I need to create a Monad-like abstraction over all the types implementing interior mutability, because I need to support RefCell, Mutex, RwLock, [...].

Well, you would need a trait like

```rust
trait DelegateStorage {
    type Inner;

    fn set(&self, func: impl FnOnce(&mut T));
    fn get(&self, func: impl FnOnce(&T));
}

impl<T> DelegateStorage for Rc<T> {
    type Inner = T;
    [...]
}
```

Doesn't seem too hard but reduces quite a bit of boilerplate :) But I agree that this could be added in a second step.

sdroege: > But I agree that this could be added in a second step

... but ideally before merging this. Because otherwise updating gst-plugins-rs to the new API would result in more code than with the old one.

glib-macros/tests/props.rs
ranfdev: > I suggested above with splitting the properties/signals...
> I also suggested in my proposal some months ago.

Can you guys provide me with some links about this? I can't find anything.

I've tried splitting the traits, but I haven't found a satisfying solution

glib-macros/tests/props.rs
ranfdev: What do you think of this API?
```rust
#[prop(name = "author-name", get, set, type = String, member = name)]
#[prop(name = "author-nick", get, set, type = String, member = nick)]
author: RefCell<Author>,
```

It doesn't require creating a fake field. I think it's more practical and feels nicer. The two props share the same underlying store, a single `RefCell`, so I like them being defined on the same field

sdroege: That's nice, I like it :) The only sad part is that the type has to be duplicated but I guess that can't really be avoided

glib-macros/tests/props.rs
ranfdev: getters/setters are mostly done, currently implemented on a `{Type}Ext` trait.
Should I also generate `connect_notify_{param}` for each param? I prefer not.

glib-macros/tests/props.rs
bilelmoussaoui: > I've tried splitting the traits, but I haven't found a satisfying solution 

Maybe push your WIP changes to a branch and we can have a look at it? What I was suggesting, is to make a new PropertiesImpl trait that would have the properties related methods in ObjectImpl and make it a requirement of ObjectImpl trait implementation.

glib-macros/src/lib.rs
```diff
@@ -734,3 +735,10 @@ pub fn variant_derive(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     variant_derive::impl_variant(input)
 }
+
+// TODO: Needs cleanup before merge...
+#[proc_macro_derive(Props, attributes(prop))]
```
bilelmoussaoui, line 740: Maybe call the proc macro `Properties`?  and `property` for it attributes

glib-macros/tests/props.rs
ranfdev: Ok, I'll send another WIP PR with only those changes.

glib-macros/tests/props.rs
```diff
@@ -0,0 +1,96 @@
...
+                fizz: RefCell<String>,
+                #[prop(name = "author-name", get, set, type = String, member = name)]
+                #[prop(name = "author-nick", get, set, type = String, member = nick)]
+                author: RefCell<Author>,
+                #[prop(
+                    type = String,
+                    get = |t: &Self| t.author.borrow().name.to_value(),
+                    set = Self::set_author_name)]
+                fake_field: PhantomData<String>,
+                #[prop(get, set, flags(EXPLICIT_NOTIFY, USER_1, DEPRECATED))]
```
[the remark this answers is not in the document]
ranfdev: I'll do it

glib/src/param_spec.rs
ranfdev: Writing this out for feedback...
Once we have a builder for each `ParamSpec`, this macro is going to call the builder with the respective parameters.
API example:
`#[prop(get, set, builder(min = 2, max = 4)]'

Why add `builder(...)`? Because we don't know beforehand what are the possible parameters (eg: GStreamer has custom ParamSpecs), and there shouldn't be a collision with the other macro's attributes.

glib-macros/src/props.rs
```diff
@@ -0,0 +1,498 @@
...
+    let getset_properties_impl = expand_getset_properties_impl(struct_ident, &input.props);
+    let expanded = quote! {
+        use glib::{PropRead, PropWrite};
+        impl ObjectImpl for #struct_ident {
+            #fn_properties
+            #fn_property
+            #fn_set_property
+        }
+
+        pub trait #struct_ident_ext {
```
bilelmoussaoui, line 488: Should also generate a connect_$prop_notify

glib-macros/tests/props.rs
```diff
@@ -0,0 +1,175 @@
...
+                author: RefCell<Author>,
+                #[prop(
+                    type = String,
+                    get = |t: &Self| t.author.borrow().name.to_owned(),
+                    set = Self::set_author_name)]
+                fake_field: PhantomData<String>,
+                #[prop(get, set, user_1, user_2, lax_validation)]
+                custom_flags: RefCell<String>,
+                #[prop(get, set, builder())]
+                simple_builder: RefCell<u32>,
```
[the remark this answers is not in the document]
ranfdev: Nope, I have to fix some things to make it work. But it will work

glib-macros/tests/props.rs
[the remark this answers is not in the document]
ranfdev: In the I've removed `flags()` entirely. Now every flag can be specified directly, by writing the lowercase name of the flag.
Example: `#[prop(get, set, explicit_notify, user_1)]

glib-macros/tests/props.rs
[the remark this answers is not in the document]
ranfdev: It's now supported :)

glib-macros/src/props.rs
ranfdev: But gtk4-rs doesn't have `connect_$prop_notify` methods...

sdroege: That seems like a problem, can you create an issue about that?

ranfdev:
This now also includes #585. This PR is basically done. I only have to add some doc. Is there anything else that I should add? I'd like to get this merged soon

glib-macros/src/properties.rs
```diff
@@ -0,0 +1,527 @@
...
+            let fn_prototype = getter_prototype(&ident, ty);
+            quote!(#fn_prototype { #body })
+        });
+        let setter = p.set.as_ref().map(|mfn| {
+            let body = match (p.member.as_ref(), mfn) {
+                (None, MaybeCustomFn::Default) => quote!(
+                     self.imp().#field_ident.set(move |x| *x = value)
+                ),
+                (Some(member), MaybeCustomFn::Default) => quote!(
+                     self.imp().#field_ident.set(move |x| x.#member = value)
```
jf2048, line 456: These implementations seem to me like they would be incorrect in most cases, `set_property` has to be used here otherwise it will skip the notification and validation.

jf2048 (changes requested):

jf2048:
I've got a bunch more notes from my implementation but it probably makes sense to do them as follow up:

- `abstract` flag to avoid allocating storage for the type
-  capability to use the macro on interfaces, where we have to force every prop to be abstract
- flag(s) to override props, I split this into `override_interface = IfaceType` and `override_class = ClassType` as wrappers around `for_interface` and `for_class`
- `computed` flag to define no storage for the field, requires custom getter/setter
- `delegate` flag to define another storage location for the field
- `borrow` flag meant to be used with `RefCell`, `Mutex`, etc, to generate additional `borrow_*` and `borrow_mut_*` methods, additional changes to the `Property` traits are needed for this
- a flag to force the setter to be inlined into the generated method, implies `explicit_notify` and `lax_validation`. In my implementation I had it so this did an equality check and the notified afterwards
- a convenient shorthand way to automatically use a custom getter/setter with the default name, I liked using `set = _` and `get = _` which imply `set = Self::set_prop` and `get = `Self::prop`
- `Property` implementations for all the atomic types
- a special cell for `construct` types that must be set once before it can be retrieved, to make it so the getter and setter don't have to take an `Option<T>`. I called it `ConstructCell` but it could be anything else
- generated `notify` functions to emit a notify signal
- way to mark this as a final class, then we can avoid generating the ext trait and can just generate methods on the wrapper type itself

And probably a couple things I missed

glib-macros/src/properties.rs
```diff
@@ -0,0 +1,527 @@
...
+            field_ident,
+            member,
+            set,
+            ..
+        } = p;
+
+        let expect = quote!(.expect("Can't convert glib::value to property type"));
+        set.as_ref().map(|set| match (member, set) {
+            (_, MaybeCustomFn::Custom(expr)) => quote!(
+                #name => {
```
jf2048, line 345: Can we use an equality comparison on the `ParamSpec` here, instead of doing a string comparison? That would be faster.

jf2048 (changes requested):

ranfdev:
@jf2048 

> - `abstract` flag to avoid allocating storage for the type, probably needs the type to be `PhantomData` since we can't remove fields with a derive macro.

Yeah, you can already use `PhantomData` and it should work, I don't think we need a specific `abstract` flag for that.

- [ ] capability to use the macro on interfaces, where we have to force every prop to be abstract

- [ ] flag(s) to override props, I split this into `override_interface = IfaceType` and `override_class = ClassType` as wrappers around `for_interface` and `for_class`

> - `computed` flag to define no storage for the field, requires custom getter/setter, also needs `PhantomData`

Yes, It's already supported by using `PhantomData`, no need to use a custom flag.

> - `delegate` flag to define another storage location for the field, also needs `PhantomData`

I don't think it's needed. It's already possible to define multiple properties on a single field (simply by calling `#[prop(name=..., type=...)]` multiple times, each time passing a different name).


> - `borrow` flag meant to be used with `RefCell`, `Mutex`, etc, to generate additional `borrow_*` and `borrow_mut_*` methods, additional changes to the `Property` traits are needed for this

I can't implement methods returning a borrow from a `Mutex` (because of lifetimes), but I can implement an `update_$prop(fn(&T))` method.

- [ ] `update_$prop(fn(&T))` ?

> - a flag to force the setter to be inlined into the generated method, implies `explicit_notify` and `lax_validation`. In my implementation I had it so this did an equality check and the notified afterwards

I prefer not to add another flag and instead simply implement this behavior when both `explicit_notify` and `lax_validation` are present.

 - [ ] a convenient shorthand way to automatically use a custom getter/setter with the default name, I liked using `set = _` and `get = _` which imply `set = Self::set_prop` and `get = Self::prop`

- [ ] `Property` implementations for all the atomic types, and for the other `OnceCell`s

- [ ] a special cell for `construct` types that must be set once before it can be retrieved, to make it so the getter and setter don't have to take an `Option<T>`. I called it `ConstructCell` but it could be anything else

- [ ] generated `notify` (`emit`) functions to emit a notify signal

> - way to mark this as a final class, then we can avoid generating the ext trait and can just generate methods on the wrapper type itself

I don't think it's possible to eliminate the `Ext` trait.

- [ ] a "plain-old-data" flag to make every field `#[property(get, set)]` by default, really useful for types that go in a `ListModel`.

It's useful, but not a priority.


I'll fill the checkboxes as I go along

jf2048:
> Yeah, you can already use `PhantomData` and it should work, I don't think we need a specific `abstract` flag for that.

I don't think that will work, because you have to skip that field in your `set_property`/`property` impl

> I don't think it's needed. It's already possible to define multiple properties on a single field (simply by calling `#[prop(name=..., type=...)]` multiple times, each time passing a different name).

Sorry, this isn't for multiple properties, this is a convenience for being able to put something else in `field_ident`, like a struct path that is multiple structs deep. It was mentioned on the IRC a while ago, I can provide another example if needed

> I can't implement methods returning a borrow from a `Mutex` (because of lifetimes)

I was able to get it to compile with another trait that takes the lifetime parameter, called it `PropertyBorrow<'a>` or something like that

> I prefer not to add another flag and instead simply implement this behavior when both `explicit_notify` and `lax_validation` are present.

That works too but for me it was a bit verbose, I liked the idea of having a shorthand that does both

> I don't think it's possible to eliminate the `Ext` trait.

I was able to do it, it works as long as you pass the type of the wrapper in somewhere and then you can just do a `impl` on it. I think we will want it for a `class` macro eventually :)

> I'll fill the checkboxes as I go along

IMO maybe we put this in another issue? These are just ideas for extra features that I had working out of tree, not anything blocking this PR

glib-macros/src/properties.rs
```diff
@@ -0,0 +1,527 @@
...
+                f(this)
+            })
+        })
+    });
+    quote!(#(#connection_fns)*)
+}
+
+pub fn impl_derive_props(input: PropsMacroInput) -> TokenStream {
+    let struct_ident = &input.ident;
+    let struct_ident_ext = format_ident!("{}Ext", &input.ident);
```
jf2048, line 498: For a signals macro that is separate from this, we are going to have to generate another `Ext` trait for all the `emit`s and `connect`s. So maybe this should be `PropsExt`, or have some way to customize the name of the trait?

jf2048 (changes requested):

glib-macros/src/properties.rs
```diff
@@ -0,0 +1,527 @@
...
+            #fn_properties
+            #fn_property
+            #fn_set_property
+        }
+
+        pub trait #struct_ident_ext {
+            #getset_properties_def
+            #connect_prop_notify_def
+        }
+        impl #struct_ident_ext for #wrapper_type {
```
jf2048, line 520:
```suggestion
        impl<T: #crate_ident::IsA<#wrapper_type>> #struct_ident_ext for T {
```

It needs to be this to work with inherited types, too

jf2048 (changes requested):

glib-macros/tests/properties.rs
```diff
@@ -0,0 +1,221 @@
...
+                builder_with_required_param: RefCell<char>,
+                #[prop(get, set, builder(SimpleBoxedString::static_type()))]
+                boxed: RefCell<SimpleBoxedString>,
+                #[prop(get, set)]
+                optional: RefCell<Option<String>>,
+                #[prop(get, set)]
+                smart_pointer: Rc<RefCell<String>>,
+                #[prop(get, set)]
+                once_cell: OnceCell<u8>,
+            }
```
fortressia, line 65: Heya, @ranfdev! The example and test code doesn't seem to include any examples to set the description of the fields. Can you add them, please?

ranfdev: Check out the latest commit and feel free to add feedback.

glib-macros/tests/properties.rs
```diff
@@ -34,9 +34,13 @@ fn props() {
             pub struct Foo {
                 #[prop(get, set)]
                 bar: Mutex<String>,
+                // The following property doesn't store any data. The value of the property is calculated
+                // when the value is accessed.
                 #[prop(get = Self::hello_world)]
                 _buzz: PhantomData<String>,
-                #[prop(get, set = Self::set_fizz, name = "fizz")]
+                #[prop(get, set = Self::set_fizz, name = "fizz", nick = "fizz-nick",
+                    blurb = "short description stored in the GLib type system"
```
fortressia, line 42: Wouldn't it be nice if we can also use the native Rust way of setting the description of a field?

fortressia: So we can try something like 

```rs
#[derive(Properties, Default)]
pub struct Foo {
 /// The description stored in the Glib type system will use native Rustdoc comments
 #[property(get, set)
 fizz: RefCell<String>,
  
}
```

fortressia: Also, what do you think of making the syntax uses keywords rather than attributes? So we can try something like this:

```rs
#[derive(Properties, Default)]
struct Foo {
// will enable get and set 
rw fizzbar: RefCell<String>;

// will disable set
ro fizz: RefCell<String>
}
```

fortressia: OOohh! I have an idea. 😄

What if we combine this macro with `glib::wrapper!`?


```rs
glib::wrapper! {
pub struct FizzbarBox 
@extends gtk::Widget, gtk::Box,
@implements gtk::Accessible, gtk::Buildable, gtk::ConstraintTarget, gtk::Orientable
{
   rw fizzbar: RefCell<String>,
   ro fizz: RefCell<String>,
   wo fizz_phantom: PhantomData<String>
}

}
```

YaLTeR: Nah, attributes are much better than keywords because they are easier for tooling to understand. They are also easier for newcomers compared to completely new syntax.

fortressia: Okay, then.

ranfdev: >  The description stored in the Glib type system will use native Rustdoc comments

I'm not sure I agree... glib's "blurb" should be a _short_ description, while Rustdoc comments don't have any length constraints. @jf2048 what do you think?

ranfdev: > What if we combine this macro with `glib::wrapper!`?

A future `class` macro will combine everything: subclass, wrapper, properties, signals... See #540.

fortressia: > A future `class` macro will combine everything: subclass, wrapper, properties, signals... See #540.

Ohh. Thanks so much!

fortressia: > 
> I'm not sure I agree... glib's "blurb" should be a _short_ description, while Rustdoc comments don't have any length constraints. @jf2048 what do you think?
> 

I think we can try to get the first line of the comment?

```
The avatar url of the user.

Example:
https://example.com/avatar?id=user_id
```

The Glib blurb will be just `The avatar url of the user.` without the rest of the lines.

ranfdev:
I've ported a gstreamer plugin to use this macro:
https://gitlab.freedesktop.org/ranfdev/gst-plugins-rs/-/commit/b7dfc339f7de3ee457ff5c5a09e2023f78ae1342

CC @slomo, who proposed the idea.

sdroege:
That looks nice except for needing the custom setters for the debug output, but I guess that can't be improved easily :)

MicroJoe:
Hi, I see that you pushed commits to this PR recently. What is the status of this pull request?

glib-macros/tests/properties.rs
jf2048: Upstream has disincentivized the use of blurb and nick: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4717

These don't really have a use case, only the rustdoc comments will matter.

sdroege: That's from a GTK point of view and is otherwise false. For GStreamer we even depend on these for properties provided by plugins.

jf2048: I am not aware of any uses for these besides gstreamer.

ranfdev:
Finally, the ParamSpec's type parameter is set automatically.

Before:
```rust
#[property(get, set, builder(SimpleBoxedString::static_type()))]
boxed: RefCell<SimpleBoxedString>,
```

Now:
```rust
#[property(get, set)]
boxed: RefCell<SimpleBoxedString>,
```


@MicroJoe The feature set is now complete. In the following month I'll keep testing the code, adding more documentation, etc...
Hopefully we can merge this under a feature flag by the end of july...?

glib-macros/tests/properties.rs
ranfdev: I'll leave this as is now => no auto generated blurb

sdroege: @jf2048 GStreamer is the biggest user of the dynamic features of GObject, so that's not very surprising :)

sdroege:
To move this forward, maybe this PR can be split up a bit? Let's first try to get those paramspec traits merged, for example. That way it can all be reviewed in smaller chunks.

glib-macros/src/properties.rs
```diff
@@ -0,0 +1,631 @@
...
+            get,
+            ..
+        } = p;
+
+        let enum_ident = name_to_enum_ident(name.value());
+        let span = p.attrs_span;
+        get.as_ref().map(|get| {
+            let body = match (member, get) {
+                (_, MaybeCustomFn::Custom(expr)) => quote!(
+                    DerivedPropertiesEnum::#enum_ident => Ok((#expr)(&self).to_value())
```
jf2048, line 372: Now that #826 is merged, `to_value` can be replaced with `<#crate_ident::Value as ::std::convert::From>::from` here and anywhere else it is used

jf2048 (changes requested):

glib-macros/src/properties.rs
```diff
@@ -0,0 +1,631 @@
...
+
+pub fn impl_derive_props(input: PropsMacroInput) -> TokenStream {
+    let struct_ident = &input.ident;
+    let crate_ident = crate_ident_new();
+    let wrapper_type = input.wrapper_ty;
+    let fn_properties = expand_properties_fn(&input.props);
+    let fn_property = expand_property_fn(&input.props);
+    let fn_set_property = expand_set_property_fn(&input.props);
+    let getset_properties_impl = expand_getset_properties_impl(&input.props);
+    let connect_prop_notify_impl = expand_connect_prop_notify_impl(&input.props);
```
bilelmoussaoui, line 608: This should also generate `notify_$property` function calls that do `ObjectExt::notify(self, $property)`

sdroege: `notify_by_pspec()` please :)

glib/src/subclass/object.rs
```diff
@@ -130,6 +130,73 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+    ) -> Result<Value, MissingPropertyHandler<'a>> {
+        Err(MissingPropertyHandler(_pspec.name()))
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// Trait containing only the signals related functions of `ObjectImpl`.
+/// Implemented by the `Signals` macro.
+/// When implementing `ObjectImpl` you may want to delegate the function calls to this trait.
+pub trait DerivedObjectSignals: ObjectSubclass {
```
sdroege, line 192: This looks like it should be in a separate PR :)

glib/src/subclass/object.rs
```diff
@@ -130,6 +130,73 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+        Err(MissingPropertyHandler(_pspec.name()))
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a `Result`.
+    fn derived_property<'a>(
+        &self,
+        _id: usize,
+        _pspec: &'a ParamSpec,
+    ) -> Result<Value, MissingPropertyHandler<'a>> {
```
bilelmoussaoui, line 183: Both those functions should have a `unimplemented! as a default and drop the usage of Result. Wdyt @sdroege ?

sdroege: I think so too. They can't really fail in practice, can they?

glib/src/subclass/object.rs
sdroege: @bilelmoussaoui Don't forget to remove this one

glib-macros/src/properties.rs
sdroege: @bilelmoussaoui this one is pending

glib-macros/src/properties.rs
sdroege: @bilelmoussaoui this one is pending

glib/src/subclass/object.rs
```diff
@@ -130,6 +130,44 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+/// properties.
+pub trait DerivedObjectProperties: ObjectSubclass {
+    // rustdoc-stripper-ignore-next
+    /// Properties installed for this type.
+    fn derived_properties() -> &'static [ParamSpec] {
+        &[]
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a `Result`.
```
sdroege, line 147: Does not return a `Result` anymore

glib/src/subclass/object.rs
```diff
@@ -130,6 +130,44 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+    fn derived_set_property(&self, _id: usize, _value: &Value, _pspec: &ParamSpec) {
+        unimplemented!()
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a `Result`.
+    fn derived_property(&self, _id: usize, _pspec: &ParamSpec) -> Value {
+        unimplemented!()
```
sdroege, line 155: Does not return a `Result` anymore

glib/src/subclass/object.rs
```diff
@@ -130,6 +130,44 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
     }
 }
 
+// rustdoc-stripper-ignore-next
+/// Trait containing only the property related functions of `ObjectImpl`.
+/// Implemented by the `Props` macro.
+/// When implementing `ObjectImpl` you may want to delegate the function calls to this trait.
+/// Some functions have a different return type from `ObjectImpl` to handle possible errors caused by missing
```
sdroege, line 137: Does not return a `Result` anymore

glib-macros/src/properties.rs
bilelmoussaoui: It was already there but named emit_impls wrongly,

glib-macros/src/properties.rs
bilelmoussaoui: Did some changes but not  very sure about them

sdroege: Does it compile? Then it's most likely correct ;)

sdroege (approved):

bilelmoussaoui (approved):

PR #504, opened 2022-01-23
glib: Provide `ThreadGuard` and `thread_id()` as public API

sdroege:
This is a preparation for https://github.com/gtk-rs/gir/issues/1305

### DIFF
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -158,77 +158,13 @@ pub use self::source_futures::*;
 mod thread_pool;
 pub use self::thread_pool::ThreadPool;
 
+pub mod thread_guard;
+
 // rustdoc-stripper-ignore-next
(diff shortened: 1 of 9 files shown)


glib/src/lib.rs
```diff
@@ -158,78 +158,14 @@ pub use self::source_futures::*;
 mod thread_pool;
 pub use self::thread_pool::ThreadPool;
 
+pub mod thread_guard;
```
bilelmoussaoui, line 161: Any reason for exposing the whole module instead of just the `ThreadGuard`/  `thread_id()` ?

bilelmoussaoui:
otherwise looks good to me

glib/src/lib.rs
sdroege: I didn't want to pollute the main crate namespace with things that you rarely need, just like `glib::translate`

PR #506, opened 2022-01-23
Add more documentation on how to replace g_signal_connect_object

pbor:


glib/src/object.rs
```diff
@@ -1834,6 +1834,7 @@ pub trait ObjectExt: ObjectType {
 
     // rustdoc-stripper-ignore-next
     /// Similar to [`Self::connect_closure`] but fails instead of panicking.
+    #[doc(alias = "g_signal_connect_object")]
```
bilelmoussaoui, line 1837: should probably be added on connect_closure as well? :)

glib/src/object.rs
```diff
@@ -1858,7 +1865,6 @@ pub trait ObjectExt: ObjectType {
     ) -> SignalHandlerId;
 
     /// Similar to [`Self::connect_closure_id`] but fails instead of panicking.
-    #[doc(alias = "g_signal_connect_closure")]
```
sdroege, line 1861: Why? I would add both aliases here too

pbor: I can add them back, but imho it was weird to have them on the `try_` variant

sdroege: Why? It does exactly the same, just returns a result instead of panicking

pbor: well, then we should add them to all the other `try` variants as well, this was the one-off... I can do that

sdroege: Oh I see. Then let's not do that for now and add them all at once later if needed. Thanks for noticing that :)

sdroege (approved):

PR #511, opened 2022-01-24
Don't require `Send` closures for GIO-style async functions

sdroege:
CC @pbor


gio/src/input_stream.rs
```diff
@@ -227,25 +227,34 @@ impl<O: IsA<InputStream>> InputStreamExtManual for O {
...
+        assert!(
+            main_context.is_owner(),
+            "Async operations only allowed if the thread is owning the MainContext"
+        );
+
         let cancellable = cancellable.map(|c| c.as_ref());
         let gcancellable = cancellable.to_glib_none();
-        let mut user_data: Box<Option<(Q, B)>> = Box::new(Some((callback, buffer)));
+        let mut user_data: Box<(glib::thread_guard::ThreadGuard<Q>, B)> =
+            Box::new((glib::thread_guard::ThreadGuard::new(callback), buffer));
```
sdroege, line 239: Also gets rid of some historical `Option` wrapping that is not needed anymore since quite a few Rust releases

pbor:
The patch makes sense to me, I always assumed that `foo_async` functions would always be called from the thread that owns the main loop

sdroege:
> The patch makes sense to me, I always assumed that `foo_async` functions would always be called from the thread that owns the main loop

@pbor It however means that you must push a thread default main context before calling any async function, or you have to call it from inside the main loop.

sdroege:
In other places (`MainContext::invoke_local()` for example) we don't just check for `is_owner()` but otherwise `acquire()` the main context. I think we'd have to also do that here and then somehow keep it acquired.

pbor:
> > The patch makes sense to me, I always assumed that `foo_async` functions would always be called from the thread that owns the main loop
> 
> @pbor It however means that you must push a thread default main context before calling any async function, or you have to call it from inside the main loop.

At least for my use cases, this is fine: I always call from the main thread

PR #512, opened 2022-01-25
gio: Don't implement `Send` on `GioFuture` anymore

sdroege:
This is unnecessary now because GIO-style async functions don't require
`Send` for their callback anymore, and it also removes the two layers of
checks to enforce at runtime that everything is handled on the same
thread.

----

This should not be backported as it's breaking backwards compatibility.


bilelmoussaoui:
Looks good other than the typo in the commit message

sdroege:
Typo fixed, thanks :)

bilelmoussaoui (approved):

PR #514, opened 2022-01-26
add param_spec_builder macro

ranfdev:
## Motivation
Creating a `Builder` for `ParamSpec*` makes implementing #494 drastically easier.
Also, users will finally be able to create a `ParamSpec*` with some default options, without having to specify `min`, `max` et all...

## About the macro
This PR adds a really powerful macro, capable of generating a Builder from any `impl` block containing a `new` function.

The macro looks for the `new` function, reads the function's parameters and creates a corresponding `Builder` struct containing each of the parameters.

The macro is currently built for `ParamSpec*` but in the future this could be used to make `Builder`s for other types: for example, this should also work for `glib::SimpleAction`.
With enough generalization I could pull this in a new crate. But this is not a priority of mine.

### DIFF
--- a/glib/tests/param_spec.rs
+++ b/glib/tests/param_spec.rs
@@ -1,0 +1,7 @@
+#[test]
+fn generated_builder() {
+    let spec = glib::ParamSpecCharBuilder::new()
+        .name("custom-char")
+        .build();
+    assert_eq!(spec.nick(), "custom-char");
+}
--- a/glib/tests/param_spec.rs
+++ b/glib/tests/param_spec.rs
@@ -1,7 +1,11 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
 #[test]
 fn generated_builder() {
     let spec = glib::ParamSpecCharBuilder::new()
         .name("custom-char")
+        .flags(glib::ParamFlags::READABLE)
         .build();
     assert_eq!(spec.nick(), "custom-char");
+    assert_eq!(spec.flags(), glib::ParamFlags::READABLE);
 }
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -9,6 +9,7 @@ mod error_domain_derive;
 mod flags_attribute;
 mod object_interface_attribute;
 mod object_subclass_attribute;
+mod param_spec_builder;
 mod shared_boxed_derive;
 mod variant_derive;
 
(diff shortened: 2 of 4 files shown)


glib/tests/param_spec.rs
```diff
@@ -0,0 +1,11 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[test]
+fn generated_builder() {
+    let spec = glib::ParamSpecCharBuilder::new()
+        .name("custom-char")
```
bilelmoussaoui, line 6: What happens if you don't provide a name? I believe that one should be passed to the constructor. 

A bit unrelated, but is a proc macro really needed here ?

sdroege: Name, nick, blurb and flags are mandatory so those should really be passed in the constructor.

Also there should be a `ParamSpecFoo::builder()` function that returns the builder.

And I would prefer to not have this as proc macro as that automatically becomes public API that we'll have to maintain.

sdroege: Otherwise this seems like an useful API addition

bilelmoussaoui: > Name, nick, blurb and flags are mandatory so those should really be passed in the constructor.

It currently assumes nick/blurb are equal to name which is what Vala/PyGObject seems to do iirc

sdroege: That makes sense

bilelmoussaoui: Something to note about the auto-generated builder pattern is  there is no way to document it in the future which is imho not great

sdroege: Yeah as there's a limited number of pspecs it doesn't seem worth it

ranfdev: > What happens if you don't provide a name? 

It panics with `Missing parameter name`. The same treatment goes to every parameter (unless there's a default value set in the macro invocation).

> I would prefer to not have this as proc macro as that automatically becomes public API that we'll have to maintain.

Well, yes, but this is also useful for external users. For example, gstreamer has some `ParamSpec`s that could take advantage of this macro.

> there is no way to document it in the future which is imho not great

That's true, but ideally why should a `ParamSpec*Builder` get specific documentation? If the documentation is common between Specs, it can be generated by the macro (for example, documentation about the default values, which I missed).

> [...] there's a limited number of pspecs it doesn't seem worth it.

Well, I could manually write everything, but then you'll get a commit with an addition of ~2000 lines. This only adds ~300 lines, and makes every `ParamSpec*Builder` behave the same way. Maintaining those 2000 lines manually could be tricky.

bilelmoussaoui: You don't have to write everything by hand, the ParamSpec* in glib already make use of macros to generate the common parts. You could something similar here I guess.

ranfdev: Ok, but I'll wait for a final feedback from @sdroege before rewriting everything

sdroege: I agree with @bilelmoussaoui here

PR #515, opened 2022-01-27
gio: add Initable::with_type

pbor:
This corresponds to Object::with_type and it is needed when
the type is available only at runtime.

### DIFF
--- a/gio/src/initable.rs
+++ b/gio/src/initable.rs
@@ -6,9 +6,14 @@ use crate::Initable;
 use glib::object::IsA;
 use glib::object::IsClass;
 use glib::value::ToValue;
-use glib::Object;
+use glib::{Cast, Object, StaticType, Type};
 
 impl Initable {
+    // rustdoc-stripper-ignore-next
+    /// Create a new instance of an initable object with the given properties.
+    ///
+    /// Similar to [`Object::new`] but can fail either because the object
+    /// creation failed or because `Initable::init` failed.
     #[allow(clippy::new_ret_no_self)]
     pub fn new<O: Sized + IsClass + IsA<Object> + IsA<Initable>, P: IsA<Cancellable>>(
         properties: &[(&str, &dyn ToValue)],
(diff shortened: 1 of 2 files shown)


gio/src/initable.rs
```diff
@@ -18,6 +23,21 @@ impl Initable {
...
         Ok(object)
     }
+
+    // rustdoc-stripper-ignore-next
+    /// Create a new instance of an initable object of the given type with the given properties.
+    ///
+    /// Similar to [`Object::with_type`] but can fail either because the object
+    /// creation failed or because `Initable::init` failed.
+    pub fn with_type<P: IsA<Cancellable>>(
+        type_: Type,
```
sdroege, line 33:
```suggestion
        type_: glib::Type,
```

sdroege (dismissed):
Otherwise looks good to me, I'll backport that for the next bugfix release too

sdroege (dismissed):

gio/src/initable.rs
```diff
@@ -18,6 +23,21 @@ impl Initable {
...
+
+    // rustdoc-stripper-ignore-next
+    /// Create a new instance of an initable object of the given type with the given properties.
+    ///
+    /// Similar to [`Object::with_type`] but can fail either because the object
+    /// creation failed or because `Initable::init` failed.
+    pub fn with_type<P: IsA<Cancellable>>(
+        type_: glib::Type,
+        properties: &[(&str, &dyn ToValue)],
+        cancellable: Option<&P>,
```
bilelmoussaoui, line 35:
```suggestion
    pub fn with_type(
        type_: glib::Type,
        properties: &[(&str, &dyn ToValue)],
        cancellable: Option<&impl IsA<Cancellable>>,
```
We use this for auto-generated code nowadays

pbor:
Actually this has another issue once I try it:

```
doesn't satisfy `glib::Object: IsA<Initable>`
```

pbor:
Ok, sorry for all the blind-push-to-ci yesterday evening. This morning I actually compiled it and added a test :)

gio/src/initable.rs
```diff
@@ -18,6 +23,21 @@ impl Initable {
...
+    /// Create a new instance of an initable object of the given type with the given properties.
+    ///
+    /// Similar to [`Object::with_type`] but can fail either because the object
+    /// creation failed or because `Initable::init` failed.
+    pub fn with_type(
+        type_: Type,
+        properties: &[(&str, &dyn ToValue)],
+        cancellable: Option<&impl IsA<Cancellable>>,
+    ) -> Result<Object, InitableError> {
+        let object = Object::with_type(type_, properties)?;
```
sdroege, line 37: You should probably add an assertion here that `type_` is a `Initiable`?

gio/src/initable.rs
```diff
@@ -18,6 +23,21 @@ impl Initable {
...
+    // rustdoc-stripper-ignore-next
+    /// Create a new instance of an initable object of the given type with the given properties.
+    ///
+    /// Similar to [`Object::with_type`] but can fail either because the object
+    /// creation failed or because `Initable::init` failed.
+    pub fn with_type(
+        type_: Type,
+        properties: &[(&str, &dyn ToValue)],
+        cancellable: Option<&impl IsA<Cancellable>>,
+    ) -> Result<Object, InitableError> {
```
bilelmoussaoui, line 36: What is the point of providing the type if you have to downcast later?

sdroege: Why do you have to downcast later? And even if you have to, `type_` and the type you downcast to might not even be the same. The type you might downcast to might be just a parent class because there's no public API for the actual type.

bilelmoussaoui: Ah right, never mind :) I saw that it downcast  later on the test case

gio/src/initable.rs
pbor: Shall I assert of should I return `NewObjectFailed`? I guess the latter, since also `Object::with_type` errors out rather than asserting

sdroege: Return an error, yes

sdroege (approved):

PR #518, opened 2022-01-31
impl `StaticType` and `{From,To}Value` for `gpointer`

### LINKED ISSUE #508: [FEATURE REQUEST] `g_value_{get,set}_pointer` is not exposed
The most obvious solution would come in the form of impl'ing `FromValue`/`ToValue` for `gpointer`/`gconstpointer`, though ideally probably as a blanket `*mut T`/`*const T`.

If there are concerns about memory safety... My assumption is that `GValue` doesn't care whether the pointer is valid or not, and thus it can be treated safely in the same way pointer types can be in rust. If any API accepts a `Value` containing a pointer with the intent of dereferencing it, I'd imagine it would need to be an unsafe API regardless of whether safe code can construct a pointer `Value` intentionally or not.

arcnmx:
Resolves #508, enabling the storage and retrieval of pointers from a `Value`.

Integration with `g_pointer_type_register_static` mentioned in that issue is left for [a future expansion](https://github.com/arcnmx/gtk-rs-core/commit/pointer-subclass)

### DIFF
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -1268,6 +1268,21 @@ impl TryFromGlib<ffi::gpointer> for ptr::NonNull<Pointee> {
     }
 }
 
+impl FromGlib<ffi::gconstpointer> for Pointer {
+    #[inline]
+    unsafe fn from_glib(val: ffi::gconstpointer) -> Self {
+        val as ffi::gpointer
+    }
+}
+
+impl TryFromGlib<ffi::gconstpointer> for ptr::NonNull<Pointee> {
+    type Error = GlibNoneError;
+
+    unsafe fn try_from_glib(val: ffi::gconstpointer) -> Result<Self, Self::Error> {
+        ptr::NonNull::new(val as ffi::gpointer).ok_or(GlibNoneError)
+    }
+}
+
 impl FromGlib<ptr::NonNull<Pointee>> for Pointer {
     #[inline]
     unsafe fn from_glib(val: ptr::NonNull<Pointee>) -> Self {
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -2396,6 +2396,7 @@ mod tests {
     use super::*;
     use crate::GString;
     use std::collections::HashMap;
+    use std::ptr::NonNull;
 
     #[test]
     fn boolean() {
@@ -2404,6 +2405,21 @@ mod tests {
         assert!(unsafe { bool::from_glib(ffi::GTRUE) });
         assert!(!unsafe { bool::from_glib(ffi::GFALSE) });
         assert!(unsafe { bool::from_glib(42) });
+    }
+
+    #[test]
+    fn pointer() {
+        let null: Pointer = ptr::null_mut::<Pointee>();
+        let nonnull: Pointer = NonNull::dangling().as_ptr();
+
+        assert_eq!(null.into_glib(), None::<ptr::NonNull<Pointee>>.into_glib());
+        assert_eq!(
+            nonnull.into_glib(),
+            Some(NonNull::<Pointee>::dangling()).into_glib()
+        );
+
+        assert!(unsafe { Option::<NonNull<Pointee>>::from_glib(null).is_none() });
+        assert!(unsafe { Option::<NonNull<Pointee>>::from_glib(nonnull).is_some() });
     }
 
     #[test]
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -130,6 +130,8 @@
 //!     }
 //! ```
 
+use crate::types::{Pointee, Pointer};
+
 use libc::{c_char, size_t};
 use std::char;
 use std::cmp::{Eq, Ordering, PartialEq};
@@ -351,7 +353,7 @@ impl IntoGlib for Option<char> {
     }
 }
 
-impl IntoGlib for crate::Pointer {
+impl IntoGlib for Pointer {
     type GlibType = ffi::gpointer;
 
     #[inline]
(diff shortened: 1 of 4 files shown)


glib/src/types.rs
```diff
@@ -297,6 +297,33 @@ macro_rules! builtin {
     };
 }
 
+// rustdoc-stripper-ignore-next
+/// A GLib [pointer](ffi::gpointer)
+pub type Pointer = *mut Pointee;
```
sdroege, line 302:
```suggestion
pub type Pointer = ffi::pointer;
```

Why not this?

glib/src/types.rs
```diff
@@ -297,6 +297,33 @@ macro_rules! builtin {
...
+/// use std::ptr::NonNull;
+///
+/// let pointer = NonNull::<Pointee>::dangling();
+/// let value = pointer.to_value();
+/// assert!(value.is::<Pointer>());
+/// assert_eq!(value.get(), Ok(pointer.as_ptr()));
+/// ```
+pub type Pointee = libc::c_void;
+
+impl StaticType for ptr::NonNull<Pointee> {
```
sdroege, line 321: Also for `Pointer` / `ffi::gpointer`

glib/src/translate.rs
```diff
@@ -1229,6 +1253,50 @@ impl FromGlib<ffi::gboolean> for bool {
     }
 }
 
+impl FromGlib<ffi::gpointer> for Pointer {
```
sdroege, line 1256: Are these trait impls useful for something?

glib/src/value.rs
```diff
@@ -830,6 +830,60 @@ impl ToValue for bool {
     }
 }
 
+impl ValueType for Pointer {
```
sdroege, line 833: All the below looks good to me, thanks :)

glib/src/types.rs
arcnmx: I mostly just wanted the docs to be self-descriptive here with a link to `Pointee`, clearly demonstrating the relationship between the two types. It does happen to be equivalent to `ffi::gpointer` so it could go either way, but I was kind of going for an almost "opaque" wrapper despite it not actually being a newtype - which is to say I'm somewhat discouraging its use directly in FFI code as a replacement for `ffi::gpointer`.
If you wanted to be insane, this could actually be defined as `type Pointer = Option<NonNull<Pointee>>` instead :stuck_out_tongue:

glib/src/types.rs
arcnmx: this is done below via the `builtin!()` macro

glib/src/translate.rs
arcnmx: It's partially to be consistent/interchangeable with `NonNull`, but more important when you want to be generic over the pointer types. It's not strictly needed until [you start changing some of these into blanket impls](https://github.com/arcnmx/gtk-rs-core/commit/pointer-subclass#diff-13ade07b8d0274c89030bc91feea353311f502b1d30976ae8c9f37814851779bL837), though.

glib/src/types.rs
sdroege: I'd prefer to go via `ffi::gpointer` then as that's what is used elsewhere for this :)

glib/src/types.rs
sdroege: Indeed

glib/src/translate.rs
sdroege: But that also only works for `c_void`-typed pointer and for anything else you'd still need to cast and do things manually. I'm mostly wondering how useful that is in practice, can it actually be used in any meaningful way in real code?

arcnmx: Hm, well any type that interacts with `g_value_get_pointer` must be castable to/from `gpointer` in some way. These traits seemed the most appropriate, but because `FromGlib<>` is parameterized, you need a concrete instantiation of it like `FromGlib<gpointer>` to use as a common interface.

> I'm mostly wondering how useful that is in practice, can it actually be used in any meaningful way in real code?

Which parts specifically? Sure, it's more useful with the associated blanket impls (that give the same flexibility to all other pointer types, and allow you to newtype them, etc), but the ability to use optional/non-null pointers is still quite useful if they're untyped!

> for anything else you'd still need to cast and do things manually

Blanket impls allow you to skip casting, but I believe there's still use in modelling them more specifically than just a plain `gpointer`, at which point they desired type is just a `.cast::<T>()` away. Often whether a pointer is null or not is relevant to code before it actually needs to work with the type! In other words code often wants to deal with the `None` case in a way where the pointed type itself is irrelevant - then it passes the NonNull (typed or untyped, casted or not) to further logic for actual processing.
One of my goals here is to enable `gir` to deal with pointers better, and be able to model nullable vs non-nullable pointers the same way it can currently model object types. For example something like the following generated code should "just work":
```rust
fn user_data(&self) -> Option<NonNull<Pointee>> {
  unsafe {
    from_glib(ffi::x_get_user_data(self.to_glib_none().0)
  }
}
```

---------

I'll go ahead and kill any extraneous ones for now - it's not like I can't just write `NonNull::new()` instead of `from_glib`. I'll pull the necessary ones back in as part of subclassing and after further experimentation with gpointers in the library I'm wrapping.

For what it's worth, I will be writing code where `value.get::<NonNull<f32>>()` makes sense to want, so I do need to flesh these impls out in interesting ways... But one could argue that the `c_void` variant of `value.get::<NonNull<Pointee>>()?.cast::<f32>()` is more correct, because the type `*mut f32` isn't actually represented by the `value.type_()`.

sdroege (approved):

Issue #507, opened 2022-01-24
[FEATURE REQUEST] Variant constructors are too statically typed


arcnmx:
Certain types of variants are impossible to construct or work with using the public safe interface because they only take type parameters `T: StaticVariantType` rather than accepting a `ty: &VariantTy` argument:
- `Variant::array_from_iter` requires knowing the type in advance at compile-time to create an array of any kind.
- `Variant::from_maybe` has the same problem. It might make make more sense splitting it up into two apis:
  - `Variant::from_some(child: &Variant) -> Self`
  - `Variant::from_none(ty: &VariantTy) -> Self`
- Maybe variants have no equivalent to `variant.as_variant` - however this can at least be worked around using `variant.try_child_value(0)`
- `variant.is::<Ty>()` has no dynamic variant, and it's not clear whether comparison with `variant.type_()` or `is_subtype_of()` or similar would be equivalent?
- `Variant::from_bytes` / `from_data` serialization methods do actually have this via `Variant::from_X_with_type`!

Most of the above methods already panic if passed a value that doesn't match the expected type, so they do dynamic type checking anyway. One could argue that the way `T: StaticVariantType` is used is somewhat of an anti-pattern; a type parameter is being used in place of what otherwise should just be a normal argument, since it isn't even able to be inferred from other arguments it must always be explicitly provided. Sure, it saves you from typing `::static_variant_type()` (just use the `VariantTy::X` consts instead honestly), but it also hinders the usefulness of the API without maintaining duplicates of each method.
It might be nice if these methods actually used `T` as an argument type instead, for example constructing an array from a `IntoIterator<Item=T>`, but that gets messy... It might require a `DynamicVariantType + IntoVariant` set of traits in place of the existing `StaticVariantType + ToVariant` to allow them to still accept a plain `Variant` as an argument... and that wouldn't help when dealing with empty container types like `None` or an empty iter, complicates boxed variant types, and so on :<

Some of the above points may also apply to `Value` APIs as well, though it has a lot less surface area. `value.transform()` and `value.is()` look relevant at a glance?

cc @jf2048 who [has had to deal with this before](https://github.com/jf2048/glib-serde/blob/19e307a86786d385e655183c99212605e866d3dd/src/variant/mod.rs#L85-L97)

sdroege:
I think adding more dynamically typed functions that take the type as argument (like already exist for other functions) and making use of the `T` where possible would make sense. Do you want to provide a PR for that?

[...]

PR #519, opened 2022-01-31
dynamic type function variants for Variants and Values

arcnmx:
Addresses the main part of #507 by providing explicit dynamic type variants of:
- `Variant::array_from_iter` -> `Variant::array_from_iter_with_type`
  (a mouthful but consistent with other existing functions like `from_data_with_type_trusted`)
- `Variant::from_maybe` becomes `Variant::from_none` and `Variant::from_some`
  - and adds an accessor: `Variant::as_maybe`
    (panicking, should there be a `try_as_maybe`? should it return an `Option<Option<Variant>>` to be more in line with `Variant::as_variant`?)
- `Variant::is` -> `Variant::is_type`
  (consider `is_of_type` instead?)
- `Value::is` -> `Value::is_type` (ditto)
- `Value::transform` -> `Value::transform_with_type`
  (consider `transform_into`/`transform_to_type` instead?)

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -236,10 +236,20 @@ impl Variant {
     #[inline]
     #[doc(alias = "g_variant_is_of_type")]
     pub fn is<T: StaticVariantType>(&self) -> bool {
+        self.is_type(&T::static_variant_type())
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Returns `true` if the type of the value corresponds to `type_`.
+    ///
+    /// This is equivalent to [`self.type_().is_subtype_of(type_)`](VariantTy::is_subtype_of).
+    #[inline]
+    #[doc(alias = "g_variant_is_of_type")]
+    pub fn is_type(&self, type_: &VariantTy) -> bool {
         unsafe {
             from_glib(ffi::g_variant_is_of_type(
                 self.to_glib_none().0,
-                T::static_variant_type().to_glib_none().0,
+                type_.to_glib_none().0,
             ))
         }
     }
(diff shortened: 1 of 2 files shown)


sdroege:
CC @cgwalters @lucab in case you have opinions

sdroege (approved):

glib/src/variant.rs
```diff
@@ -471,22 +493,49 @@ impl Variant {
...
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Extract the value of a maybe Variant.
+    #[inline]
+    pub fn as_maybe(&self) -> Option<Variant> {
+        assert!(self.type_().is_maybe());
+
+        self.try_child_value(0)
```
lucab, line 536: I may have misunderstood the logic, but I'm not exactly sure that this matches what the docstring says.
Let's maybe cover this method with a test to avoid any doubts?

lucab (commented):
Generally speaking, these new methods looks useful to have :+1:.

glib/src/variant.rs
arcnmx: A maybe variant is a container with either 1 or 0 values, thus roughly the logic for this function is `if self.n_children() == 0 { None } else { Some(self.child_value(0)) }` - `Variant::try_child_value` already exists as a rough equivalent to rust's `slice.get(0)` for this purpose, so it didn't seem worth duplicating the logic here.
However, I really should just use `ffi::g_variant_get_maybe` instead, since it's clearer - [it implements the exact same logic regardless](https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gvariant.c#L698-703).

Good point on adding a test for this method though, will do!

sdroege (approved):

Issue #526, opened 2022-02-02
[BUG] gio::Task leaks memory


BrainBlasted:
**Bug description**

Due to GTK not providing async callbacks for loading `gdk::Texture`s, I used `gio::Task` in an async function for loading
textures. However, this seemingly caused a massive memory leak in the magnitude of gigabytes when the function was called. Removing the usage of `gio::Task` removed the leak.

Here's the code that leaks:

```rust
async fn load_texture_from_file(file: &gio::File) -> Result<gdk::Texture, glib::Error> {
    unsafe {
        let (sender, receiver) = futures_channel::oneshot::channel();

        let task = gio::Task::new(
            Some(file.upcast_ref()),
            gio::Cancellable::NONE,
            move |task, _: Option<&gio::File>| {
                let res = task.propagate();
                let _ = sender.send(res);
            },
        );

        let task_fn = move |task: gio::Task<gdk::Texture>,
                            source_obj: Option<&gio::File>,
                            _: Option<&gio::Cancellable>| {
            let file = source_obj.unwrap();
            let result = gdk::Texture::from_file(file);
            task.return_result(result);
        };

        task.run_in_thread(task_fn);

        receiver.await.unwrap()
    }
}
```

In comparison, here's code using `std::thread::spawn` instead of `GTask` that does not leak:

```rust
async fn load_texture_from_file(file: &gio::File) -> Result<gdk::Texture, glib::Error> {
    let (sender, receiver) = futures_channel::oneshot::channel();

    std::thread::spawn(clone!(@weak file => move || {
        let result = gdk::Texture::from_file(&file);
        sender.send(result).unwrap()
    }));

    receiver.await.unwrap()
}
```

**Backtrace**

Here are the excerpts from the valgrind log that I thought were the most relevant:

```
 ==19== 336 (216 direct, 120 indirect) bytes in 9 blocks are definitely lost in loss record 28,845 of 32,102
==19==    at 0x4845A83: calloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==19==    by 0x554A3E8: g_malloc0 (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.7101.0)
==19==    by 0x301781: glib::value::copy_value (in /app/bin/loupe)
==19==    by 0x1A5E88: gio::task::Task<V>::return_result (in /app/bin/loupe)
[...]

PR #527, opened 2022-02-02
gtask: fix memory leak when calling g_task_return_value

mbiggio:
The g_task_return_value function does not take ownership of the input
GValue pointer, but just borrows it to make a copy:
https://developer-old.gnome.org/gio/unstable/GTask.html#g-task-return-value
The Rust caller should take care of the deallocation.


mbiggio:
#526

sdroege (approved):

sdroege:
Will backport for the next 0.15 bugfix release at some point

PR #528, opened 2022-02-02
glib: Add function to wrap a future with a timeout

jsparber:


glib/src/future_with_timeout.rs
```diff
@@ -0,0 +1,95 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use futures_util::{
+    future::{self, Either, Future},
+    pin_mut,
+};
+
+// rustdoc-stripper-ignore-next
+/// The error returned when a future times out.
+pub struct FutureWithTimeoutError;
```
sdroege, line 10: Implement `std:error::Error` and `Display` on this please :)

glib/src/future_with_timeout.rs
```diff
@@ -0,0 +1,95 @@
...
+        Either::Left((x, _)) => Ok(x),
+        _ => Err(FutureWithTimeoutError),
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// Add a timeout to a `Future`.
+///
+/// The `Future` must be spawned on an `Executor` backed by a `glib::MainContext`.
+pub async fn future_with_timeout<T>(
```
sdroege, line 34: @bilelmoussaoui @GuillaumeGomez Any suggestions for a better name?

glib/src/future_with_timeout.rs
jsparber: I added `Clone`, `Copy`, `PartialEq`, `Eq`, `Debug`, `Display`, `Error`;

glib/src/future_with_timeout.rs
pbor: maybe we can have a timeout future type and then have its own constructor rather than having a fn?

```
glib::futures::Timeout::new(timeout: std::time::Duration, fut: impl Future<Output = T>)
```
(I proposed a `futures` module in case we have more, eg `Idle` and `Cancellable`, but this is not a requirement)

sdroege: Having a futures module seems useful, yes. The main question still stays: we have "timeout" as the name for the thing that just triggers at some point, and for the new thing that times out another future.

pbor: `Timed`?

GuillaumeGomez: `timeout_future`? Generally it goes like this no?

sdroege: Like https://gtk-rs.org/gtk-rs-core/git/docs/glib/fn.timeout_future.html ? That's exactly the problem :)

GuillaumeGomez: Damn...

sdroege (approved):
Let's go with this name for now then until someone figures out something better

sdroege:
Something similar came up in https://github.com/gtk-rs/gtk4-rs/issues/960 . Maybe we should have a `glib::futures` module with various convenience functions. The one from that issue would be something like `async fn glib::futures::sleep(duration: std::time::Duration)`.

PR #530, opened 2022-02-04
Gir files update

sdroege:


gio/src/auto/file.rs
```diff
@@ -675,7 +675,7 @@ pub trait FileExt: 'static {
 
     #[doc(alias = "g_file_resolve_relative_path")]
     #[must_use]
-    fn resolve_relative_path(&self, relative_path: impl AsRef<std::path::Path>) -> Option<File>;
+    fn resolve_relative_path(&self, relative_path: impl AsRef<std::path::Path>) -> File;
```
sdroege, line 678: Should also make a bugfix release with this. Previous bugfix release changed this to an `Option`, which was apparently wrong.

gio/src/auto/enums.rs
```diff
@@ -213,6 +213,8 @@ pub enum CredentialsType {
     NetbsdUnpcbid,
     #[doc(alias = "G_CREDENTIALS_TYPE_APPLE_XUCRED")]
     AppleXucred,
+    #[doc(alias = "G_CREDENTIALS_TYPE_WIN32_PID")]
```
bilelmoussaoui, line 216: it needs a version guard

glib/gobject-sys/src/lib.rs
```diff
@@ -1869,6 +1895,55 @@ extern "C" {
     pub fn g_binding_get_target_property(binding: *mut GBinding) -> *const c_char;
     pub fn g_binding_unbind(binding: *mut GBinding);
 
+    //=========================================================================
```
bilelmoussaoui, line 1898: Any reason for not generation those new APIs?

sdroege: It needs manual bindings, just like for single signal/property bindings.

gio/src/auto/enums.rs
sdroege: Indeed, it's getting boring. In the next days then :)

glib/gobject-sys/src/lib.rs
bilelmoussaoui: Unfortunate i guess, something for later

sdroege: It's quite low level API :)

gio/src/auto/enums.rs
sdroege: Done. Anything else @bilelmoussaoui ?

PR #533, opened 2022-02-07
Handle empty slices correctly

sdroege:
Passing NULL to `slice::from_raw_parts` is invalid.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/532

### DIFF
--- a/cairo/src/image_surface.rs
+++ b/cairo/src/image_surface.rs
@@ -138,7 +138,12 @@ impl ImageSurface {
                 return Err(BorrowError::from(Error::SurfaceFinished));
             }
             let len = self.height() as usize * self.stride() as usize;
-            f(slice::from_raw_parts(ptr, len));
+            let data = if len == 0 {
+                &[]
+            } else {
+                slice::from_raw_parts(ptr, len)
+            };
+            f(data);
         }
         Ok(())
     }
@@ -186,7 +191,9 @@ impl AsRef<[u8]> for ImageSurfaceDataOwned {
         let len = (self.surface.stride() as usize) * (self.surface.height() as usize);
         unsafe {
             let ptr = ffi::cairo_image_surface_get_data(self.surface.to_raw_none());
-            debug_assert!(!ptr.is_null());
+            if ptr.is_null() || len == 0 {
+                return &[];
+            }
             slice::from_raw_parts(ptr, len)
         }
     }
@@ -197,7 +204,9 @@ impl AsMut<[u8]> for ImageSurfaceDataOwned {
         let len = (self.surface.stride() as usize) * (self.surface.height() as usize);
         unsafe {
             let ptr = ffi::cairo_image_surface_get_data(self.surface.to_raw_none());
-            debug_assert!(!ptr.is_null());
+            if ptr.is_null() || len == 0 {
+                return &mut [];
+            }
             slice::from_raw_parts_mut(ptr, len)
         }
     }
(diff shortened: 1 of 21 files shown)


cairo/src/image_surface.rs
```diff
@@ -187,6 +192,9 @@ impl AsRef<[u8]> for ImageSurfaceDataOwned {
         unsafe {
             let ptr = ffi::cairo_image_surface_get_data(self.surface.to_raw_none());
             debug_assert!(!ptr.is_null());
+            if len == 0 {
```
GuillaumeGomez, line 195: Since it's only a `debug_assert`, shouldn't you check for `ptr.is_null()` in any case?

cairo/src/image_surface.rs
```diff
@@ -198,6 +206,9 @@ impl AsMut<[u8]> for ImageSurfaceDataOwned {
...
```
GuillaumeGomez, line 209: Same.

cairo/src/image_surface.rs
```diff
@@ -235,7 +246,11 @@ impl<'a> ImageSurfaceData<'a> {
             let len = (surface.stride() as usize) * (surface.height() as usize);
             ImageSurfaceData {
                 surface,
-                slice: slice::from_raw_parts_mut(ptr, len),
+                slice: if len == 0 {
```
GuillaumeGomez, line 249: Same.

cairo/src/image_surface_png.rs
```diff
@@ -30,7 +30,11 @@ unsafe extern "C" fn read_func<R: Read>(
         return Error::ReadError.into();
     }
 
-    let buffer = slice::from_raw_parts_mut(data, len as usize);
+    let buffer = if len == 0 {
```
GuillaumeGomez, line 33: Same.

bilelmoussaoui (dismissed):

cairo/src/ps.rs
```diff
@@ -47,7 +47,12 @@ impl PsSurface {
             let mut num_vers = mem::MaybeUninit::uninit();
             ffi::cairo_ps_get_levels(&mut vers_ptr, num_vers.as_mut_ptr());
 
-            std::slice::from_raw_parts(vers_ptr, num_vers.assume_init() as _)
+            let num_vers = num_vers.assume_init();
+            if num_vers == 0 {
```
GuillaumeGomez, line 51: `vers_ptr` could be NULL.

sdroege: No it can't

cairo/src/stream.rs
```diff
@@ -224,7 +224,13 @@ extern "C" fn write_callback<W: io::Write + 'static>(
             // with a W parameter consistent with the box that was unsized to `Box<dyn Any>`.
             let stream = unsafe { stream.downcast_mut_unchecked::<W>() };
             // Safety: this is the callback contract from cairo’s API
-            let data = unsafe { std::slice::from_raw_parts(data, length as usize) };
+            let data = unsafe {
+                if length == 0 {
```
GuillaumeGomez, line 228: Needs to check `data` too.

cairo/src/ps.rs
GuillaumeGomez: Why not?

gdk-pixbuf/src/pixbuf.rs
```diff
@@ -313,6 +313,9 @@ impl Pixbuf {
     pub unsafe fn pixels(&self) -> &mut [u8] {
         let mut len = 0;
         let ptr = ffi::gdk_pixbuf_get_pixels_with_length(self.to_glib_none().0, &mut len);
+        if len == 0 {
```
GuillaumeGomez, line 316: `ptr` could be NULL.

sdroege: No it can't be `NULL` without `len` being `0`.

sdroege: Before adding more comments like this, please check the C API behaviour

GuillaumeGomez (approved):

PR #543, opened 2022-02-11
cairo: Allow writing arbitrary `Surface`s as PNGs

### LINKED ISSUE #541: `cairo_surface_write_to_png_stream` is exposed only on `ImageSurface` without an obvious reason why
As far as I can tell, using cairo in C you'd be able to call `cairo_surface_write_to_png_stream` on any surface, but you'd have to check the returned status to see whether the surface actually supports this operation, at least this is how I interpret the following

> ### Returns
> [CAIRO_STATUS_SUCCESS](https://cairographics.org/manual/cairo-Error-handling.html#CAIRO-STATUS-SUCCESS:CAPS) if the PNG file was written successfully. Otherwise, [CAIRO_STATUS_NO_MEMORY](https://cairographics.org/manual/cairo-Error-handling.html#CAIRO-STATUS-NO-MEMORY:CAPS) is returned if memory could not be allocated for the operation, [CAIRO_STATUS_SURFACE_TYPE_MISMATCH](https://cairographics.org/manual/cairo-Error-handling.html#CAIRO-STATUS-SURFACE-TYPE-MISMATCH:CAPS) if the surface does not have pixel contents, or [CAIRO_STATUS_PNG_ERROR](https://cairographics.org/manual/cairo-Error-handling.html#CAIRO-STATUS-PNG-ERROR:CAPS) if libpng returned an error.

(src: https://cairographics.org/manual/cairo-PNG-Support.html#cairo-surface-write-to-png-stream)

RealKC:
This patch moves `write_to_png` from `ImageSurface` to `Surface` as the
C cairo API supports this usecase. It also adds some documentation for
this method.

Closes #541.

---

Should I also rename the `image_surface_png` module to something else(i.e. `png`) given it contains code that's also related to `Surface` now?


RealKC:
Also is master the right branch to target for breaking changes? This is one as code calling the method like `ImageSurface::write_to_png(surface, reader)` won't compile anymore after this is merged.

cairo/src/image_surface_png.rs
```diff
@@ -112,8 +112,16 @@ impl ImageSurface {
             }
         }
     }
+}
 
+impl Surface {
```
sdroege, line 117: Maybe rename the whole file to `surface_png.rs` but seems good to me otherwise

sdroege (approved):

PR #544, opened 2022-02-11
glib: add trait `StaticTypeExt` with method `type_ensure`

### LINKED ISSUE #542: [FEATURE REQUEST] Add method to register types
Most apps currently call `static_type` within `class_init` to register custom widgets they refer to within the composite template.
That is not a nice way to do it, since we are not interested in the returned type, but rather it's side effect: registering the type.
It would be good to have a method which does only that.
It could be named `type_ensure` corresponding to the function `g_type_ensure`.
Maybe `register_type` would work as well?

@bilelmoussaoui @sdroege

Hofer-Julian:
Fixes #542

### DIFF
--- a/glib/src/types.rs
+++ b/glib/src/types.rs
@@ -246,6 +246,19 @@ impl StaticType for Type {
     }
 }
 
+pub trait StaticTypeExt {
+    // rustdoc-stripper-ignore-next
+    /// Ensures that the type has been registered with the type system.
+    #[doc(alias = "g_type_ensure")]
+    fn ensure_type();
+}
+
+impl<T: StaticType> StaticTypeExt for T {
+    fn ensure_type() {
+        T::static_type();
+    }
+}
+
 #[doc(hidden)]
 impl crate::value::ValueType for Type {
     type Type = Type;
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -36,7 +36,7 @@ pub use self::signal::{
 };
 
 pub use self::enums::{EnumClass, EnumValue, FlagsBuilder, FlagsClass, FlagsValue, UserDirectory};
-pub use self::types::{ILong, Pointer, StaticType, Type, ULong};
+pub use self::types::{ILong, Pointer, StaticType, StaticTypeExt, Type, ULong};
 pub use self::value::{BoxedValue, SendValue, ToSendValue, ToValue, Value};
 pub use self::variant::{
     FixedSizeVariantArray, FixedSizeVariantType, FromVariant, StaticVariantType, ToVariant, Variant,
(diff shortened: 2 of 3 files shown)


glib/src/types.rs
```diff
@@ -246,6 +246,16 @@ impl StaticType for Type {
     }
 }
 
+pub trait StaticTypeExt {
```
sdroege, line 249: Please add it to the prelude and also add some docs :)

sdroege (commented):
seems fine otherwise

glib/src/types.rs
bilelmoussaoui: Also add a doc alias :-)

Hofer-Julian: > Also add a doc alias :-)

I was thinking about that, but does it make sense considering I am not really binding `g_type_ensure`?

sdroege: It does the same, so yes

Hofer-Julian: Should be fine now.

glib/src/types.rs
```diff
@@ -246,6 +246,18 @@ impl StaticType for Type {
...
+    /// Ensures that the type has been registered with the type system.
+    fn type_ensure();
+}
+
+impl<T: StaticType> StaticTypeExt for T {
+    #[doc(alias = "g_type_ensure")]
```
sdroege, line 255: This needs to be on the trait definition, not on the implementation

bilelmoussaoui, line 255: The doc alias should be as part of the trait

glib/src/types.rs
```diff
@@ -246,6 +246,18 @@ impl StaticType for Type {
     }
 }
 
+pub trait StaticTypeExt {
+    /// Ensures that the type has been registered with the type system.
```
bilelmoussaoui, line 250: You need the rustdoc-stripper-ignore-next comment here

glib/src/types.rs
Hofer-Julian: Done

glib/src/types.rs
Hofer-Julian: Done

bilelmoussaoui (dismissed):

glib/src/types.rs
```diff
@@ -246,6 +246,19 @@ impl StaticType for Type {
     }
 }
 
+pub trait StaticTypeExt {
+    // rustdoc-stripper-ignore-next
+    /// Ensures that the type has been registered with the type system.
+    #[doc(alias = "g_type_ensure")]
+    fn type_ensure();
```
sdroege, line 253: `gtk::Button::type_ensure()` doesn't read very nice. Maybe this should be `gtk::Button::ensure_type()`?

sdroege (approved):
@bilelmoussaoui ?

bilelmoussaoui (approved):

PR #545, opened 2022-02-14
Update gir-files

sdroege:


glib/src/auto/functions.rs
```diff
@@ -25,6 +25,27 @@ pub fn access(filename: impl AsRef<std::path::Path>, mode: i32) -> i32 {
     unsafe { ffi::g_access(filename.as_ref().to_glib_none().0, mode) }
 }
 
+//#[cfg(any(feature = "v2_72", feature = "dox"))]
+//#[cfg_attr(feature = "dox", doc(cfg(feature = "v2_72")))]
+//#[doc(alias = "g_aligned_alloc")]
+//pub fn aligned_alloc(n_blocks: usize, n_block_bytes: usize, alignment: usize) -> /*Unimplemented*/Option<Fundamental: Pointer> {
```
bilelmoussaoui, line 31: Are those needed? if not, they should probably be ignored

bilelmoussaoui (dismissed):

glib/src/auto/functions.rs
sdroege: Good point. They only really make sense for FFI

Issue #540, opened 2022-02-09
[PROPOSAL] class! macro


ranfdev:
## Motivation

#494 adds a way to declare properties on the private, implementation struct. In the future, we also want to add signals declaration using macros.
It would be nice being able to generate the corresponding `connect_{signal}` and `set_{property}` methods on the external, wrapper struct.

It's not possible to share data between macros. Right now, (on #494) object properties and signals (still missing) are declared on a `derive` macro. It's not possible to list the properties and signals obtained by that macro and then reuse that list to implement the corresponding `connect_{signal}` and `set_{property}` methods.

## Proposal

I propose adding a `class!` macro which should:
- accept a declaration of properties and signals
- generate the required `ObjectSubclass` implementation
- generate the required `ObjectImpl` on the private struct
- generate the external wrapper struct
- generate the corresponding `connect_` and `set_property_` methods on the wrapper struct.

I propose the following syntax
```rust
        class! {
            const GTYPE_NAME: &'static str = "MyButton";
            class Button extends [gtk::Button], implements [gtk::ActionMap] {
                #[prop(get, set)]
                label: gtk::Label,
            }
            impl Button {
                #[signal]
                pub fn clicked(&self) {};
            }
        }

```

sdroege:
> How would that handle interior mutability? There are many options and that probably should be configurable somehow.

To add to this, how would the implementation access the field values?

ranfdev:
> To add to this, how would the implementation access the field values?

Whoops, that line should have said
```rust
label: RefCell<gtk::Label>,
```

And that works as it does on the current #494 PR

ranfdev:
And yes, I've looked at the work done in https://gitlab.gnome.org/federico/gnome-class. It's a nice work and a great resource.

sdroege:
Where would the `impl` blocks for the private and public struct be put here, and the `impl` blocks for the parent class' `Impl` traits?

Ideally you would have as little code as possible inside the macro to make things easier for tooling.

bilelmoussaoui:
This sounds like a good idea, but overall I believe before trying to build something that would be super complex, having the building blocks handled separately would simplify writing/reviewing/testing such macro. 
Currently, the most verbose part of creating gobjects is 
- the object properties 
- the object signals
- virtual functions. this will probably need separate handling between the two uses cases: applications and libraries. 

[...]

PR #546, opened 2022-02-14
WIP: add ObjectPropsImpl, ObjectSignalsImpl, ObjectContructImpl

ranfdev:
CC: @bilelmoussaoui  
I can't find a solution to this.
The user can't implement just one of `ObjectPropsImpl`, `ObjectSignalsImpl`, `ObjectContructImpl`.
It has to implement them all, or nothing. Which doesn't make sense.

I can't get them to work separately, because `IsSubclassable::class_init` requires a full `ObjectImpl` with every method defined.

I feel we need to edit  `IsSubclassable::class_init` if we want to split the `ObjectImpl`.

### DIFF
--- a/glib/src/subclass/mod.rs
+++ b/glib/src/subclass/mod.rs
@@ -266,7 +266,7 @@ pub mod prelude {
     //! Prelude that re-exports all important traits from this crate.
     pub use super::boxed::BoxedType;
     pub use super::interface::{ObjectInterface, ObjectInterfaceExt, ObjectInterfaceType};
-    pub use super::object::{ObjectClassSubclassExt, ObjectImpl, ObjectImplExt};
+    pub use super::object::{ObjectClassSubclassExt, ObjectImpl, ObjectPropsImpl, ObjectSignalsImpl, ObjectConstructImpl, ObjectImplExt};
     pub use super::shared::{RefCounted, SharedType};
     pub use super::types::{
         ClassStruct, InstanceStruct, IsImplementable, IsSubclassable, IsSubclassableExt,
--- a/glib/src/subclass/object.rs
+++ b/glib/src/subclass/object.rs
@@ -11,6 +11,51 @@ use crate::{Cast, Object, ObjectType, ParamSpec, Value};
 use std::mem;
 use std::ptr;
 
+
+pub trait ObjectPropsImpl: ObjectSubclass {
+    // rustdoc-stripper-ignore-next
+    /// Properties installed for this type.
+    fn properties() -> &'static [ParamSpec];
+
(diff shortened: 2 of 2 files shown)


bilelmoussaoui:
is there a reason for keeping the ObjectImpl trait? also, you could probably provide default implementation of all those functions so you don't have to implement them directly

bilelmoussaoui:
My idea was to actually move the functions from the ObjectImpl trait, not make the ObjectImpl trait require another trait implementation

sdroege:
> is there a reason for keeping the ObjectImpl trait? also, you could probably provide default implementation of all those functions so you don't have to implement them directly

(I didn't review anything yet) It would be good to keep `ObjectImpl` as a "combined" trait so there's a single one in the dependency hierarchy, just like we have for all the other `Impl` traits. For consistency reasons.

Also it would be great if we could try to keep the changes as small as possible so that not a lot of non-trivial changes are needed to all the code out there that is already making use of subclassing :)

ranfdev:
There are many traits that require a full `ObjectImpl`, so I don't know if splitting the traits is a good idea at all. 

What if we discard this PR and instead the `Props` and `Signals` macros simply generate two new traits `imp::Props` and `imp::Signals`?
The user then needs to manually implement `ObjectImpl` (delegating every method to the corresponding generated traits), but this last step will be eliminated once we implement a full `class!` macro (#540). By doing this we are also sure to not break any existing code.

sdroege:
> There are many traits that require a full `ObjectImpl`, so I don't know if splitting the traits is a good idea at all.

There could be a

```rust
pub trait ObjectImpl: All + The + Split + Traits { }
impl<T: All + The + Split + Traits> ObjectImpl for T {}
```

but that's probably more confusing than it helps?

> The user then needs to manually implement `ObjectImpl` (delegating every method to the corresponding generated traits), but this last step will be eliminated once we implement a full `class!` macro (#540).

That was my original idea when thinking about how to provide such APIs for signals/properties. I think that's not very bad: as you say, with a full `class!` macro that's not really a problem anymore and also until then we could provide a convenience macro that does exactly that (either as attribute macro or as function-style macro that just expands to that code).

bilelmoussaoui:
Thinking about this one for a bit, I forgot the use case of GtkEditable which makes use of `delegate_get_property` / `delegate_set_property` functions, see https://github.com/gtk-rs/gtk4-rs/blob/master/examples/custom_editable/custom_editable/imp.rs#L61

So maybe it is not a bad idea after all to generate a new trait called `ObjectPropertiesImpl` and we expect people to call `ObjectPropertiesImpl::property` / ObjectProperties::set_property` there.

PR #549, opened 2022-02-15
Remove remaining public fields

### LINKED ISSUE #495: Avoid public fields in cairo
Currently, cairo is the only crate that has public fields for `#[repr(C)]` structs like `cairo::Rectangle`/ `cairo::RectangleInt`. Instead we should have the same thing as elsewhere and add a constructor/getters/setters.

GuillaumeGomez:
Fixes #495.

### DIFF
--- a/cairo/src/matrices.rs
+++ b/cairo/src/matrices.rs
@@ -5,7 +5,8 @@ use crate::utils::status_to_result;
 
 #[repr(transparent)]
 #[derive(Debug, Clone, PartialEq)]
-pub struct Matrix(ffi::Matrix);
+#[doc(alias = "cairo_matrix_t")]
+pub struct Matrix(ffi::cairo_matrix_t);
 
 impl Default for Matrix {
     fn default() -> Self {
@@ -14,16 +15,16 @@ impl Default for Matrix {
 }
 
 impl Matrix {
-    pub(crate) fn ptr(&self) -> *const ffi::Matrix {
+    pub(crate) fn ptr(&self) -> *const ffi::cairo_matrix_t {
         self as *const Matrix as _
     }
 
-    pub(crate) fn mut_ptr(&mut self) -> *mut ffi::Matrix {
+    pub(crate) fn mut_ptr(&mut self) -> *mut ffi::cairo_matrix_t {
         self as *mut Matrix as _
     }
 
     pub(crate) fn null() -> Self {
-        Self(ffi::Matrix {
+        Self(ffi::cairo_matrix_t {
             xx: 0.0,
             yx: 0.0,
             xy: 0.0,
@@ -34,7 +35,7 @@ impl Matrix {
     }
 
     pub fn identity() -> Self {
-        Self(ffi::Matrix {
+        Self(ffi::cairo_matrix_t {
             xx: 1.0,
             yx: 0.0,
             xy: 0.0,
@@ -45,7 +46,7 @@ impl Matrix {
     }
 
     pub fn new(xx: f64, yx: f64, xy: f64, yy: f64, x0: f64, y0: f64) -> Self {
-        Self(ffi::Matrix {
+        Self(ffi::cairo_matrix_t {
             xx,
             yx,
             xy,
--- a/cairo/src/matrices.rs
+++ b/cairo/src/matrices.rs
@@ -2,20 +2,10 @@
 
 use crate::error::Error;
 use crate::utils::status_to_result;
-use libc::c_double;
 
-#[repr(C)]
-#[derive(Debug, Clone, Copy, PartialEq)]
-pub struct Matrix {
-    pub xx: c_double,
-    pub yx: c_double,
-
-    pub xy: c_double,
-    pub yy: c_double,
-
-    pub x0: c_double,
-    pub y0: c_double,
-}
+#[repr(transparent)]
+#[derive(Debug, Clone, PartialEq)]
+pub struct Matrix(ffi::Matrix);
 
 impl Default for Matrix {
     fn default() -> Self {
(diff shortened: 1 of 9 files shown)


cairo/src/matrices.rs
```diff
@@ -65,6 +65,43 @@ impl Matrix {
         }
     }
 
+    pub fn xx(self) -> c_double {
```
sdroege, line 68:
```suggestion
    pub fn xx(self) -> f64 {
```

And elsewhere

cairo/src/matrices.rs
```diff
@@ -65,6 +65,43 @@ impl Matrix {
...
+        self.xx
+    }
+    pub fn set_xx(&mut self, xx: f64) {
+        self.xx = xx;
+    }
+    pub fn yx(self) -> c_double {
```
sdroege, line 74:
```suggestion
    pub fn yx(&self) -> c_double {
```

would make more sense. It's not super-cheap

cairo/src/rectangle.rs
```diff
@@ -10,10 +10,45 @@ use std::mem;
...
+impl Rectangle {
+    pub fn new(self, x: f64, y: f64, width: f64, height: f64) -> Self {
+        Self {
+            x,
+            y,
+            width,
+            height,
+        }
+    }
+    pub fn x(self) -> f64 {
```
sdroege, line 28:
```suggestion
    pub fn x(&self) -> f64 {
```

cairo/src/rectangle.rs
```diff
@@ -10,10 +10,45 @@ use std::mem;
...
-    pub width: f64,
-    pub height: f64,
+    x: f64,
+    y: f64,
+    width: f64,
+    height: f64,
+}
+
+impl Rectangle {
+    pub fn new(self, x: f64, y: f64, width: f64, height: f64) -> Self {
```
sdroege, line 20:
```suggestion
    pub fn new(x: f64, y: f64, width: f64, height: f64) -> Self {
```

cairo/src/rectangle_int.rs
```diff
@@ -10,10 +10,45 @@ use std::mem;
...
-    pub width: i32,
-    pub height: i32,
+    x: i32,
+    y: i32,
+    width: i32,
+    height: i32,
+}
+
+impl RectangleInt {
+    pub fn new(self, x: i32, y: i32, width: i32, height: i32) -> Self {
```
sdroege, line 20:
```suggestion
    pub fn new(x: i32, y: i32, width: i32, height: i32) -> Self {
```

cairo/src/rectangle_int.rs
```diff
@@ -10,10 +10,45 @@ use std::mem;
...
+        Self {
+            x,
+            y,
+            width,
+            height,
+        }
+    }
+    pub fn x(self) -> i32 {
```
sdroege, line 28:
```suggestion
    pub fn x(&self) -> i32 {
```

cairo/src/matrices.rs
```diff
@@ -7,14 +7,14 @@ use libc::c_double;
 #[repr(C)]
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct Matrix {
```
sdroege, line 9: Should maybe be a `#[repr(transparent)]pub struct Matrix(ffi::cairo_matrix_t)`? Same for the others

cairo/src/matrices.rs
GuillaumeGomez: Should we remove the `Copy` impl then?

cairo/src/rectangle.rs
GuillaumeGomez: Oh wow, I did very bad...

cairo/src/rectangle_int.rs
GuillaumeGomez: Same question: should we remove the `Copy` impl then?

GuillaumeGomez:
I had to manually implement `Debug` on the `Rectangle` types because of this failure:

```
failures:

---- context::tests::clip_rectangle stdout ----
thread 'context::tests::clip_rectangle' panicked at 'assertion failed: `(left == right)`
  left: `"RectangleList([Rectangle(cairo_rectangle_t { x: 0.0, y: 0.0, width: 10.0, height: 10.0 })])"`,
 right: `"RectangleList([Rectangle { x: 0.0, y: 0.0, width: 10.0, height: 10.0 }])"`', cairo/src/context.rs:982:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    context::tests::clip_rectangle
```

cairo/src/matrices.rs
```diff
@@ -2,20 +2,11 @@
...
-    pub yx: c_double,
-
-    pub xy: c_double,
-    pub yy: c_double,
-
-    pub x0: c_double,
-    pub y0: c_double,
-}
+#[repr(transparent)]
+#[derive(Debug, Clone, PartialEq)]
```
sdroege, line 7: @bilelmoussaoui What do you think about implementing `Copy` or not here? For all the graphene types (which are similar) we implement `Copy`, and e.g. a `[f64; 16]` would also be `Copy` (and that would be a `cairo::Matrix` with a different name).

cairo/src/matrices.rs
```diff
@@ -133,21 +161,27 @@ mod tests {
...
+        let ffi_matrix = crate::ffi::cairo_matrix_t {
+            xx: 1.0,
+            yx: 2.0,
+            xy: 3.0,
+            yy: 4.0,
+            x0: 5.0,
+            y0: 6.0,
+        };
+
+        let transmuted: Matrix = unsafe { std::mem::transmute(ffi_matrix) };
```
sdroege, line 173: This test is meaningless now because it's the same type by construction.

cairo/src/rectangle.rs
```diff
@@ -6,14 +6,55 @@ use std::fmt;
...
+    }
+    pub fn height(&self) -> f64 {
+        self.0.height
+    }
+    pub fn set_height(&mut self, height: f64) {
+        self.0.height = height;
+    }
+}
+
+impl fmt::Debug for Rectangle {
```
sdroege, line 49: Please also do a nice `Debug` impl for `cairo::Matrix` while you're at it :)

sdroege (commented):
Seems good otherwise

sdroege:
> I had to manually implement `Debug` on the `Rectangle` types because of this failure:

Why is there a test comparing the output of the `Debug` impl? Can you fix that? :)

cairo/src/matrices.rs
bilelmoussaoui: I guess that is fine, but then will all the Matrix methods take the type by ref or not? because the graphene types do

sdroege: You mean leave it like this or add the `Copy` impl back? And taking any value that is bigger than a pointer by reference makes sense.

bilelmoussaoui: Add the copy impl back

GuillaumeGomez:
> > I had to manually implement `Debug` on the `Rectangle` types because of this failure:
> 
> Why is there a test comparing the output of the `Debug` impl? Can you fix that? :)

What do you mean by fixing? Like removing it?

sdroege:
> What do you mean by fixing? Like removing it?

Doing what the test is doing without relying on the string representation of the `Debug` impl

cairo/src/matrices.rs
GuillaumeGomez: Considering that we need to make our methods take `self` behind a reference, I think it's pretty obvious that `Matrix` shouldn't implement `Copy`.

sdroege: Those are completely independent concerns

GuillaumeGomez: In what way? If we need to be careful about it, our users will need to. At this point, it's simply better to prevent this problem by making it impossible to happen.

sdroege: Copying comment from @ranfdev from Matrix:
> When the CPU fetches some data from the RAM, multiple words are fetched and put in the cache. On a modern computer fetching 64 bits should take mostly the same time as fetching 256 bits (given that those bits are contiguous in memory).

The only real concern here is for passing via functions as passing anything bigger than a pointer can easily require some additional work: making space available on the stack and copying the value in there so that a pointer to the stack can be passed to the function.

sdroege (approved):

PR #551, opened 2022-02-16
define a corresponding `Builder` for each `ParamSpec`

ranfdev:
This came out better than I expected.

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -411,6 +411,96 @@ macro_rules! define_param_spec_numeric {
     };
 }
 
+macro_rules! define_builder {
+    (@constructors $rust_type:ident, $builder_type:ident $(($($req_ident:ident: $req_ty:ty,)*))?) => {
+        impl<'a> $builder_type<'a> {
+            fn new(name: &'a str, $($($req_ident: $req_ty)*)?) -> Self {
+                Self {
+                    name,
+                    $($($req_ident: Some($req_ident),)*)?
+                    ..Default::default()
+                }
+            }
+        }
+
+        impl $rust_type {
+            pub fn builder<'a>(name: &'a str, $($($req_ident: $req_ty),*)?) -> $builder_type<'a> {
+                $builder_type::new(name, $($($req_ident),*)?)
+            }
+        }
+    };
+    (
+        $rust_type:ident, $builder_type:ident {
+            $($field_id:ident: $field_ty:ty $(= $field_expr:expr)?,)*
+        }
+        $(requires $required_tt:tt)?
+    ) => {
+        #[derive(Default)]
+        #[must_use]
+        pub struct $builder_type<'a> {
+            name: &'a str,
+            nick: Option<&'a str>,
+            blurb: Option<&'a str>,
+            flags: Option<crate::ParamFlags>, // `ParamFlags` doesn't implement `Default` so I wrap it in `Option`
+            $($field_id: Option<$field_ty>),*
+        }
+        impl<'a> $builder_type<'a> {
+            /// Default: `self.name`
+            pub fn nick(mut self, nick: &'a str) -> Self {
+                self.nick = Some(nick);
+                self
+            }
+            /// Default: `self.name`
+            pub fn blurb(mut self, blurb: &'a str) -> Self {
+                self.blurb = Some(blurb);
+                self
+            }
+
+            /// Default: `glib::ParamFlags::READWRITE`
+            pub fn flags(mut self, flags: crate::ParamFlags) -> Self {
+                self.flags = Some(flags);
+                self
+            }
+
+            $(
+            $(#[doc = concat!("Default: `", stringify!($field_expr), "`")])?
+            pub fn $field_id(mut self, value: $field_ty) -> Self {
+                self.$field_id = Some(value);
+                self
+            }
+            )*
+
+            #[must_use]
+            pub fn build(self) -> ParamSpec {
+                $rust_type::new(
+                    self.name,
+                    self.nick.unwrap_or(self.name),
+                    self.blurb.unwrap_or(self.name),
+                    $(self
+                        .$field_id
+                        $(.or(Some($field_expr)))?
+                        .expect("impossible: missing parameter in ParamSpec*Builder")
+                    ,)*
+                    self.flags.unwrap_or(crate::ParamFlags::READWRITE),
+                )
+            }
+        }
+        define_builder!(@constructors $rust_type, $builder_type $($required_tt)?);
+    }
+}
+macro_rules! define_builder_numeric {
+    ($rust_type:ident, $builder_type:ident, $n_ty:ty) => {
+        define_builder!(
+            $rust_type,
+            $builder_type {
+                minimum: $n_ty = <$n_ty>::MIN,
+                maximum: $n_ty = <$n_ty>::MAX,
+                default_value: $n_ty = <$n_ty as Default>::default(),
+            }
+        );
+    };
+}
+
 #[track_caller]
 // the default panic formatter will use its caller as the location in its error message
 fn assert_param_name(name: &str) {
@@ -429,6 +519,8 @@ define_param_spec_numeric!(
     "g_param_spec_char"
 );
 
+define_builder_numeric!(ParamSpecChar, ParamSpecCharBuilder, i8);
+
 define_param_spec_numeric!(
     ParamSpecUChar,
     gobject_ffi::GParamSpecUChar,
@@ -437,6 +529,8 @@ define_param_spec_numeric!(
     g_param_spec_uchar,
     "g_param_spec_uchar"
 );
+
+define_builder_numeric!(ParamSpecUChar, ParamSpecUCharBuilder, u8);
 
 define_param_spec!(ParamSpecBoolean, gobject_ffi::GParamSpecBoolean, 2);
 
@@ -465,6 +559,13 @@ impl ParamSpecBoolean {
     }
 }
 
+define_builder!(
+    ParamSpecBoolean,
+    ParamSpecBooleanBuilder {
+        default_value: bool = false,
+    }
+);
+
 define_param_spec_numeric!(
     ParamSpecInt,
     gobject_ffi::GParamSpecInt,
@@ -473,6 +574,8 @@ define_param_spec_numeric!(
     g_param_spec_int,
     "g_param_spec_int"
 );
+
+define_builder_numeric!(ParamSpecInt, ParamSpecIntBuilder, i32);
 
 define_param_spec_numeric!(
     ParamSpecUInt,
@@ -483,6 +586,8 @@ define_param_spec_numeric!(
     "g_param_spec_uint"
 );
 
+define_builder_numeric!(ParamSpecUInt, ParamSpecUIntBuilder, u32);
+
 define_param_spec_numeric!(
     ParamSpecLong,
     gobject_ffi::GParamSpecLong,
@@ -491,6 +596,8 @@ define_param_spec_numeric!(
     g_param_spec_long,
     "g_param_spec_long"
 );
+
+define_builder_numeric!(ParamSpecLong, ParamSpecLongBuilder, libc::c_long);
 
 define_param_spec_numeric!(
     ParamSpecULong,
@@ -501,6 +608,8 @@ define_param_spec_numeric!(
     "g_param_spec_ulong"
 );
 
+define_builder_numeric!(ParamSpecULong, ParamSpecULongBuilder, libc::c_ulong);
+
 define_param_spec_numeric!(
     ParamSpecInt64,
     gobject_ffi::GParamSpecInt64,
@@ -509,6 +618,8 @@ define_param_spec_numeric!(
     g_param_spec_int64,
     "g_param_spec_int64"
 );
+
+define_builder_numeric!(ParamSpecInt64, ParamSpecInt64Builder, i64);
 
 define_param_spec_numeric!(
     ParamSpecUInt64,
@@ -519,8 +630,9 @@ define_param_spec_numeric!(
     "g_param_spec_uint64"
 );
 
-define_param_spec!(ParamSpecUnichar, gobject_ffi::GParamSpecUnichar, 9);
+define_builder_numeric!(ParamSpecUInt64, ParamSpecUInt64Builder, u64);
 
+define_param_spec!(ParamSpecUnichar, gobject_ffi::GParamSpecUnichar, 9);
 define_param_spec_default!(ParamSpecUnichar, Result<char, CharTryFromError>, TryFrom::try_from);
 
 impl ParamSpecUnichar {
@@ -545,6 +657,14 @@ impl ParamSpecUnichar {
         }
     }
 }
+
+define_builder!(
+    ParamSpecUnichar,
+    ParamSpecUnicharBuilder {
+        default_value: char,
+    }
+    requires (default_value: char,)
+);
 
 define_param_spec!(ParamSpecEnum, gobject_ffi::GParamSpecEnum, 10);
 
@@ -587,6 +707,14 @@ impl ParamSpecEnum {
     }
 }
 
+define_builder!(
+    ParamSpecEnum,
+    ParamSpecEnumBuilder {
+        enum_type: crate::Type,
+        default_value: i32 = 0i32,
+    }
+    requires (enum_type: crate::Type,)
+);
 define_param_spec!(ParamSpecFlags, gobject_ffi::GParamSpecFlags, 11);
 
 define_param_spec_default!(ParamSpecFlags, u32, |x| x);
@@ -628,6 +756,15 @@ impl ParamSpecFlags {
     }
 }
 
+define_builder!(
+    ParamSpecFlags,
+    ParamSpecFlagsBuilder {
+        flags_type: crate::Type,
+        default_value: u32 = 0u32,
+    }
+    requires (flags_type: crate::Type,)
+);
+
 define_param_spec_numeric!(
     ParamSpecFloat,
     gobject_ffi::GParamSpecFloat,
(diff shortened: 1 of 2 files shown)


glib/src/param_spec.rs
```diff
@@ -1403,4 +1621,16 @@ mod tests {
             .expect("Not a string param spec");
         assert_eq!(pspec.default_value(), Some("default"));
     }
+
+    #[test]
+    fn test_param_spec_int_builder() {
+        let pspec = ParamSpecIntBuilder::new("name")
```
bilelmoussaoui, line 1627: I would add a builder method to those ParamSpecT types so that you can `ParamSpecInt::builder("name").blurb(...).build()` we do the same things for all the exposed builder patterns

bilelmoussaoui: oh never mind, just saw that

glib/src/param_spec.rs
```diff
@@ -1281,6 +1459,33 @@ impl ParamSpecOverride {
...
+}
+
+impl<'a> ParamSpecOverrideBuilder<'a> {
+    pub fn new(name: &'a str) -> Self {
+        Self {
+            name,
+            overridden: None,
+        }
+    }
+    pub fn overridden(mut self, spec: &'a ParamSpec) -> Self {
```
sdroege, line 1476: This one is mandatory so should be part of the `new()` /  `builder()` function

glib/src/param_spec.rs
```diff
@@ -826,6 +942,13 @@ impl ParamSpecEnum {
     }
 }
 
+define_builder!(
+    ParamSpecEnum,
+    ParamSpecEnumBuilder,
+    enum_type: crate::Type = crate::Type::ENUM,
```
sdroege, line 948: The enum type must be a concrete enum type and is mandatory, so must be given in the `new()` / `builder()` function

glib/src/param_spec.rs
```diff
@@ -876,6 +999,13 @@ impl ParamSpecFlags {
     }
 }
 
+define_builder!(
+    ParamSpecFlags,
+    ParamSpecFlagsBuilder,
+    flags_type: crate::Type = crate::Type::FLAGS,
```
sdroege, line 1005: The flags type must be a concrete flags type and is mandatory, so must be given in the `new()` / `builder()` function

glib/src/param_spec.rs
```diff
@@ -1035,6 +1187,12 @@ impl ParamSpecParam {
     }
 }
 
+define_builder!(
+    ParamSpecParam,
+    ParamSpecParamBuilder,
+    param_type: crate::Type,
```
sdroege, line 1193: The param type must be a concrete param type and is mandatory, so must be given in the `new()` / `builder()` function

glib/src/param_spec.rs
```diff
@@ -1069,6 +1227,12 @@ impl ParamSpecBoxed {
     }
 }
 
+define_builder!(
+    ParamSpecBoxed,
+    ParamSpecBoxedBuilder,
+    default_value: crate::Type = crate::Type::BOXED,
```
sdroege, line 1233: This is not the default value, it's the boxed type that this property can handle.

And it's mandatory and must be a concrete boxed type, so part of the `new()` / `builder()` function

glib/src/param_spec.rs
```diff
@@ -1182,6 +1354,12 @@ impl ParamSpecObject {
     }
 }
 
+define_builder!(
+    ParamSpecObject,
+    ParamSpecObjectBuilder,
+    object_type: crate::Type = crate::Type::OBJECT,
```
sdroege, line 1360: This is mandatory and should be part of the `new()` / `builder()` function

glib/src/param_spec.rs
```diff
@@ -1315,6 +1520,12 @@ impl ParamSpecGType {
     }
 }
 
+define_builder!(
+    ParamSpecGType,
+    ParamSpecGTypeBuilder,
+    is_a_type: crate::Type,
```
sdroege, line 1526: The default can be `glib::Type::NONE` btw. It's not mandatory.

glib/src/param_spec.rs
```diff
@@ -384,6 +384,60 @@ macro_rules! define_param_spec_numeric {
...
+                self.$field_id = Some(value);
+                self
+            }
+            )*
+
+            pub fn build(self) -> ParamSpec {
+                $rust_type::new(self.name, self.nick.unwrap_or(self.name), self.blurb.unwrap_or(self.name),
+                    $(self.$field_id
+                        $(.or(Some($field_expr.into())))?
+                        .expect(concat!("missing `", stringify!($field_id), "` in `", stringify!($builder_type), "`"))
```
sdroege, line 426: Basically, this `expect()` should be gone and the corresponding parameters should be part of the `new()` / `builder()` function.

sdroege: I mean, we can catch this easily at compile time so why only handle it at runtime...

ranfdev: But that's why I wanted to have Builders: to have a common interface for creating a `ParamSpec`.

I need to map `#[prop(...build(pameter = value)]` to
```
Builder::new(name).parameter(value)
```

This mapping also makes the `Props` macro compatible with external `ParamSpec`s types (if they have a builder).
Ff the macro becomes specific for each `ParamSpec*::new` we can't have compatibility with external `ParamSpec`s.

Also, It's not unusual for a `Builder` to panic. The builders in `gtk-rs` behave in the same thing.

sdroege: Maybe there could be a separate constructor that is only used by the macro? For normal usage you really don't want to defer things to runtime that can easily be caught at compile time.

ranfdev: Idea!
What if I add `#[must_use]` on the methods of the required parameters?

sdroege: That doesn't do what you think it does :)

glib/src/param_spec.rs
sdroege: Ah, `UNIT`. No idea why it was renamed :shrug:

glib/src/param_spec.rs
sdroege: And that can't panic unless the builder was misgenerated.

ranfdev: Yes, you are both right. That was a bad example for my argumentation. gtk-rs widgets always have a default value for each parameter, so gtk-rs builders aren't really comparable to these ParamSpecs builders.

ranfdev: So, I've decided to make the builders safe by adding the required parameters in the `new` and `builder` functions.

The `Props` macro API will then "directly" call the builder (the first required argument, `name`, will be automatically passed by the `Props` macro. The other required arguments stay the same):
```rust
#[prop(..., builder(required_params).minimum(10))]
```

ranfdev:
Every builder is safe. I'm done

glib/src/param_spec.rs
```diff
@@ -384,6 +384,97 @@ macro_rules! define_param_spec_numeric {
...
+            $($field_id:ident: $field_ty:ty $(= $field_expr:expr)?,)*
+        }
+        $(requires $required_tt:tt)?
+    ) => {
+        #[derive(Default)]
+        pub struct $builder_type<'a> {
+            name: &'a str,
+            nick: Option<&'a str>,
+            blurb: Option<&'a str>,
+            flags: Option<crate::ParamFlags>, // `ParamFlags` don't implement `Default`, so I wrap them in `Option`
```
bilelmoussaoui, line 416: Should it implement Default as READWRITE?

glib/src/param_spec.rs
```diff
@@ -384,6 +384,97 @@ macro_rules! define_param_spec_numeric {
...
+            }
+            /// Default: `self.name`
+            #[must_use]
+            pub fn blurb(mut self, blurb: &'a str) -> Self {
+                self.blurb = Some(blurb);
+                self
+            }
+
+            #[must_use]
+            pub fn flags(mut self, flags: crate::ParamFlags) -> Self {
```
bilelmoussaoui, line 434: Would mention the default value here as well

glib/src/param_spec.rs
ranfdev: I don't like that. I feel bitflags should have 0, or another "base" state, as the default value. That makes working with bitflags easier.

In this context having `READWRITE` as default makes sense, in other context it doesn't.

sdroege: We could manually implement `Default` like this for `ParamFlags`. I think that would make sense

We currently don't derive/impl `Default` for flags so there's no problem here and this can be considered for other flags on a case-by-base basis.

glib/src/param_spec.rs
```diff
@@ -384,6 +384,98 @@ macro_rules! define_param_spec_numeric {
     };
 }
 
+macro_rules! define_builder {
+    (@constructors $rust_type:ident, $builder_type:ident $(($($req_ident:ident: $req_ty:ty,)*))?) => {
+        impl<'a> $builder_type<'a> {
+            pub fn new(name: &'a str, $($($req_ident: $req_ty)*)?) -> Self {
```
sdroege, line 390: Maybe I missed that in one of the comments, but what's the plan for the macro here? It would use the `Default` impl of the builder and then panic at runtime if required parameters are not provided, or it would have some special syntax for required parameters (say `#[prop(@name = "bla", @other_required_parameter = blabla, optional_parameter = 123]`). The problem with the latter seems that it's not entirely obvious for the user which ones are required and which not, and the compiler errors will probably also not be the most useful.

glib/src/param_spec.rs
```diff
@@ -384,6 +384,98 @@ macro_rules! define_param_spec_numeric {
...
+                Self {
```
sdroege, line 391: I would make the `new()` function of the builder private and only keep the `builder()` function on the pspec. Having both is mildly confusing IMHO

glib/src/param_spec.rs
```diff
@@ -384,6 +384,98 @@ macro_rules! define_param_spec_numeric {
...
+            }
+        }
+    };
+    (
+        $rust_type:ident, $builder_type:ident {
+            $($field_id:ident: $field_ty:ty $(= $field_expr:expr)?,)*
+        }
+        $(requires $required_tt:tt)?
+    ) => {
+        #[derive(Default)]
```
sdroege, line 411: Should add a `#[must_use]` on the type too

glib/src/param_spec.rs
```diff
@@ -384,6 +384,98 @@ macro_rules! define_param_spec_numeric {
...
+            $(
+            $(#[doc = concat!("Default: `", stringify!($field_expr), "`")])?
+            #[must_use]
+            pub fn $field_id(mut self, value: $field_ty) -> Self {
+                self.$field_id = Some(value);
+                self
+            }
+            )*
+
+            pub fn build(self) -> ParamSpec {
```
sdroege, line 449: Should add a `#[must_use]` here too (what's the point of creating a pspec just to not use it? most likely a mistake)

glib/src/param_spec.rs
```diff
@@ -1281,6 +1465,27 @@ impl ParamSpecOverride {
...
+impl<'a> ParamSpecOverrideBuilder<'a> {
+    pub fn new(name: &'a str, overridden: &'a ParamSpec) -> Self {
+        Self { name, overridden }
+    }
+    #[must_use]
+    pub fn overridden(mut self, spec: &'a ParamSpec) -> Self {
+        self.overridden = spec;
+        self
+    }
+    pub fn build(self) -> ParamSpec {
```
sdroege, line 1484: `#[must_use]`

glib/src/param_spec.rs
```diff
@@ -1281,6 +1465,27 @@ impl ParamSpecOverride {
     }
 }
 
+// This builder is not autogenerated because it's the only one that doesn't take
+// `nick`, `blurb` and `flags` as parameters.
+pub struct ParamSpecOverrideBuilder<'a> {
```
sdroege, line 1470: `#[must_use]`

glib/src/param_spec.rs
```diff
@@ -1281,6 +1465,27 @@ impl ParamSpecOverride {
...
+    name: &'a str,
+    overridden: &'a ParamSpec,
+}
+
+impl<'a> ParamSpecOverrideBuilder<'a> {
+    pub fn new(name: &'a str, overridden: &'a ParamSpec) -> Self {
```
sdroege, line 1476: This one private, `builder()` public

sdroege (commented):
Looks good to me apart from those minor things, thanks!

glib/src/param_spec.rs
ranfdev: My current idea is
```
#[prop(get, set, builder(required_params).optional(opt_value))]
name: RefCell<String>
```
which should map to
```
ParamSpecString::builder(name, ...required_params)
  .optional(opt_value)
  .flags(READWRITE) // because the user specified get, set
```


To create a ParamSpecInt with a maximum of 5:
```
#[prop(get, set, builder().maximum(5))]
times: RefCell<i32>
```

ranfdev: It can't panic, because builders have now safe constructors. I like my proposed syntax because it makes it clear you are calling the builder, therefore the user can just lookup the builder docs.

Your proposed solution
`#[prop(@name = "bla", @other_required_parameter = blabla, optional_parameter = 123]`
is not possible, because `optional_parameter = 123` may be a ParamSpec parameter specified by an external crate, conflicting with other attributes from the `Props` macro. Eg: an external crate may create a `ParamSpec` with the `get` parameter. 

Also, specifying the required parameters without order would make the builder implementation more difficult (We would need a ParamSpec*Args struct for each ParamSpec).

bilelmoussaoui: > Eg: an external crate may create a ParamSpec with the get parameter.

I don't think there are that many external ParamSpec types, it would be a good idea to look at what are the available ones and figure out if the current use cases would fit properly there.

sdroege: Makes sense to me but is a bit strange syntax maybe :)

glib/src/param_spec.rs
sdroege: @ranfdev Can you put this into a separate commit, and then squash all the other commits into one?

sdroege:
Apart from the paramflags this seems all good to me now. @bilelmoussaoui ?

glib/src/param_spec.rs
```diff
@@ -898,6 +1079,32 @@ impl ParamSpecOverride {
...
+
+    pub fn builder<'a>(name: &'a str, overridden: &'a ParamSpec) -> ParamSpecOverrideBuilder<'a> {
+        ParamSpecOverrideBuilder::new(name, overridden)
+    }
+}
+
+// This builder is not autogenerated because it's the only one that doesn't take
+// `nick`, `blurb` and `flags` as parameters.
+#[must_use]
+pub struct ParamSpecOverrideBuilder<'a> {
```
bilelmoussaoui, line 1091: ParamSpec has two useful constructors that what people will use almost everytime (that takes either an interface/or class) to override it property. Do we need that integrated somehow as part of the builder?

bilelmoussaoui:
Otherwise looks good to me, make sure the commits are squashed though :)

glib/src/param_spec.rs
ranfdev: I don't know how to add the two constructors and make everything usable from the `Props` macro.

sdroege: This specific pspec can get some special handling in the macro. It's a very special case that doesn't appear anywhere else, and it's a "basic" pspec. You can invent some special syntax for this one if that's necessary.

bilelmoussaoui: What I meant here is whether such builder is really useful to have at all

bilelmoussaoui: As mentioned by slomo we should just sepcial case this one in the proc macro

ranfdev: Reasonable

glib/src/param_spec.rs
```diff
@@ -898,6 +1072,32 @@ impl ParamSpecOverride {
             from_glib_none((*ptr).overridden)
         }
     }
+
+    pub fn builder<'a>(name: &'a str, overridden: &'a ParamSpec) -> ParamSpecOverrideBuilder<'a> {
```
bilelmoussaoui, line 1076: As I mentioned before, I don't see the point of this builder as both params are required

pango/src/glyph_item_iter.rs
```diff
@@ -41,4 +41,23 @@ impl GlyphItemIter {
             }
         }
     }
+
+    pub fn start_char(&self) -> i32 {
```
bilelmoussaoui, line 45: this commit shouldn't be part of this PR :)

ranfdev: I made some errors while rebasing and squashing probably

ranfdev: That's funny... I can't see this change locally:
```
❯ git diff --stat upstream/master
 glib/src/gobject/flags.rs |   6 +++++
 glib/src/param_spec.rs    | 266 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------
 2 files changed, 253 insertions(+), 19 deletions(-)
 ```
 Maybe another force push will fix it

glib/src/param_spec.rs
sdroege: Doesn't hurt IMHO. Maybe it's useful for someone and it's at least consistent.

sdroege:
Looks good to me now except for that spurious pango change

pango/src/glyph_item_iter.rs
sdroege: You have that change in a separate commit now, so getting rid of it should be just a matter of dropping it from `git rebase origin/master -i`.

ranfdev: I don't understand why the change didn't show up while doing `git diff upstream/master`, but indeed, doing `git rebase -i upstream/master` auto-removed it (By the way, it didn't show up even in the interactive rebase... Bah)

sdroege (approved):

PR #552, opened 2022-02-16
 Fix cairo FFI types name

### LINKED ISSUE #550: Fix cairo FFI types name
A lot of cairo FFI types are misnamed (I discovered it while working on https://github.com/gtk-rs/gtk-rs-core/pull/549). It'd be better to use the "real" names instead.

GuillaumeGomez:
Fixes #550.

### DIFF
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -743,43 +743,32 @@ impl Context {
 
     #[doc(alias = "cairo_text_extents")]
     pub fn text_extents(&self, text: &str) -> Result<TextExtents, Error> {
-        let mut extents = TextExtents {
-            x_bearing: 0.0,
-            y_bearing: 0.0,
-            width: 0.0,
-            height: 0.0,
-            x_advance: 0.0,
-            y_advance: 0.0,
-        };
+        let mut extents = MaybeUninit::<TextExtents>::uninit();
 
         unsafe {
             let text = CString::new(text).unwrap();
-            ffi::cairo_text_extents(self.0.as_ptr(), text.as_ptr(), &mut extents);
+            ffi::cairo_text_extents(
+                self.0.as_ptr(),
+                text.as_ptr(),
+                extents.as_mut_ptr() as *mut _,
+            );
+            self.status().map(|_| extents.assume_init())
         }
-        self.status().map(|_| extents)
     }
 
     #[doc(alias = "cairo_glyph_extents")]
     pub fn glyph_extents(&self, glyphs: &[Glyph]) -> Result<TextExtents, Error> {
-        let mut extents = TextExtents {
-            x_bearing: 0.0,
-            y_bearing: 0.0,
-            width: 0.0,
-            height: 0.0,
-            x_advance: 0.0,
-            y_advance: 0.0,
-        };
+        let mut extents = MaybeUninit::<TextExtents>::uninit();
 
         unsafe {
             ffi::cairo_glyph_extents(
                 self.0.as_ptr(),
                 glyphs.as_ptr() as *const _,
                 glyphs.len() as _,
-                &mut extents,
+                extents.as_mut_ptr() as *mut _,
             );
+            self.status().map(|_| extents.assume_init())
         }
-
-        self.status().map(|_| extents)
     }
 
     // paths stuff
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -12,7 +12,6 @@ use crate::{
 };
 #[cfg(feature = "use_glib")]
 use glib::translate::*;
-use libc::c_int;
 use std::ffi::CString;
 use std::fmt;
 use std::mem::MaybeUninit;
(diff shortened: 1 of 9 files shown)


cairo/src/context.rs
```diff
@@ -727,56 +732,40 @@ impl Context {
...
     #[doc(alias = "cairo_font_extents")]
     pub fn font_extents(&self) -> Result<FontExtents, Error> {
-        let mut extents = FontExtents {
-            ascent: 0.0,
-            descent: 0.0,
-            height: 0.0,
-            max_x_advance: 0.0,
-            max_y_advance: 0.0,
-        };
+        let mut extents = FontExtents::new();
```
sdroege, line 735:
```suggestion
        let mut extents = MaybeUninit::uninit();
```

or `::zeroed()` maybe?

I don't think we need a constructor for this type at all, do we?

cairo/src/context.rs
```diff
@@ -727,56 +732,40 @@ impl Context {
...
     pub fn text_extents(&self, text: &str) -> Result<TextExtents, Error> {
-        let mut extents = TextExtents {
-            x_bearing: 0.0,
-            y_bearing: 0.0,
-            width: 0.0,
-            height: 0.0,
-            x_advance: 0.0,
-            y_advance: 0.0,
-        };
+        let mut extents = TextExtents::new();
```
sdroege, line 746: Same

cairo/src/context.rs
```diff
@@ -727,56 +732,40 @@ impl Context {
...
     pub fn glyph_extents(&self, glyphs: &[Glyph]) -> Result<TextExtents, Error> {
-        let mut extents = TextExtents {
-            x_bearing: 0.0,
-            y_bearing: 0.0,
-            width: 0.0,
-            height: 0.0,
-            x_advance: 0.0,
-            y_advance: 0.0,
-        };
+        let mut extents = TextExtents::new();
```
sdroege, line 761: Same

cairo/src/font/scaled_font.rs
```diff
@@ -99,55 +98,39 @@ impl ScaledFont {
...
     #[doc(alias = "cairo_scaled_font_extents")]
     pub fn extents(&self) -> FontExtents {
-        let mut extents = FontExtents {
-            ascent: 0.0,
-            descent: 0.0,
-            height: 0.0,
-            max_x_advance: 0.0,
-            max_y_advance: 0.0,
-        };
+        let mut extents = FontExtents::new();
```
sdroege, line 101: same

cairo/src/font/scaled_font.rs
```diff
@@ -99,55 +98,39 @@ impl ScaledFont {
...
     pub fn text_extents(&self, text: &str) -> TextExtents {
-        let mut extents = TextExtents {
-            x_bearing: 0.0,
-            y_bearing: 0.0,
-            width: 0.0,
-            height: 0.0,
-            x_advance: 0.0,
-            y_advance: 0.0,
-        };
+        let mut extents = TextExtents::new();
```
sdroege, line 110: same

cairo/src/font/scaled_font.rs
```diff
@@ -99,55 +98,39 @@ impl ScaledFont {
...
     pub fn glyph_extents(&self, glyphs: &[Glyph]) -> TextExtents {
-        let mut extents = TextExtents {
-            x_bearing: 0.0,
-            y_bearing: 0.0,
-            width: 0.0,
-            height: 0.0,
-            x_advance: 0.0,
-            y_advance: 0.0,
-        };
+        let mut extents = TextExtents::new();
```
sdroege, line 126: same

cairo/src/font/scaled_font.rs
```diff
@@ -180,9 +163,9 @@ impl ScaledFont {
                 y,
                 text.as_ptr(),
                 text_length,
-                &mut glyphs_ptr,
+                &mut glyphs_ptr as *mut _ as *mut _,
                 &mut glyph_count,
-                &mut clusters_ptr,
+                &mut clusters_ptr as *mut _ as *mut _,
```
sdroege, line 168: Add some more explicit casts here, but also only one cast should be needed

sdroege: Also explicit casts elsewhere

cairo/src/context.rs
GuillaumeGomez: We don't but I found it better this way.

cairo/src/font/scaled_font.rs
GuillaumeGomez: No because we first need to convert from `&mut` to `*mut` and then to the correct pointer type. However I'll update for explicit cast.

cairo/src/context.rs
sdroege: It's unnecessarily noisy. We use `MaybeUninit` elsewhere for out parameters too

sdroege: Also it somewhat hides the fact that this is an out parameter. Could as well be an in-out parameter from your code, but what we really expect here is to provide some memory to the function that is then filled/initialized.

GuillaumeGomez: Fine by me!

GuillaumeGomez:
Updated!

sdroege (approved):

PR #554, opened 2022-02-17
glib: Fix handling of closures that provide a return value pointer th…

sdroege:
…at is initialized to `G_TYPE_INVALID`

Fixes https://github.com/gtk-rs/gtk4-rs/issues/929


sdroege:
For some background, this is what `gobject/gsignal.c` is initializing the return value in various cases if there is no return value (i.e. it zero-initializes it).

bilelmoussaoui (approved):

PR #559, opened 2022-02-17
cairo: fix Glyph::index type

bilelmoussaoui:
it is not a u64, fixes the Windows CI breakage on gtk4-rs

### DIFF
--- a/cairo/src/font/glyph.rs
+++ b/cairo/src/font/glyph.rs
@@ -8,7 +8,7 @@ use std::fmt;
 pub struct Glyph(ffi::cairo_glyph_t);
 
 impl Glyph {
-    pub fn index(&self) -> u64 {
+    pub fn index(&self) -> libc::c_ulong {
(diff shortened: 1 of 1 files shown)


cairo/src/font/glyph.rs
```diff
@@ -8,7 +8,7 @@ use std::fmt;
 pub struct Glyph(ffi::cairo_glyph_t);
 
 impl Glyph {
-    pub fn index(&self) -> u64 {
+    pub fn index(&self) -> libc::c_ulong {
```
sdroege, line 11: `unsigned long` :scream:

bilelmoussaoui: seems like it https://www.cairographics.org/manual/cairo-text.html#cairo-glyph-t, no idea why

sdroege: Historical reasons

PR #561, opened 2022-02-17
translate: Pre-allocate Rust hashmap at correct size

cgwalters:
I am manually doing a hash table translation for
https://github.com/ostreedev/ostree/blob/4066b0ab4578add593c49448fb7553fdee656911/src/libostree/ostree-mutable-tree.c#L599
and went to look at this code and noticed it wasn't doing
the optimization I thought it would be.


sdroege (approved):

sdroege:
Thanks!

PR #563, opened 2022-02-18
glib-macros: Pass a pointer for watched objects into closure macro

jf2048:
This also saves us from doing an extra `ref`/`unref` on watched objects when calling those closures

### DIFF
--- a/glib-macros/tests/test.rs
+++ b/glib-macros/tests/test.rs
@@ -359,7 +359,7 @@ fn closure() {
         assert_eq!(obj.ref_count(), 1);
         let weak_test = glib::closure_local!(@watch obj => move || obj.ref_count());
         assert_eq!(obj.ref_count(), 1);
-        assert_eq!(weak_test.invoke::<u32>(&[]), 3);
+        assert_eq!(weak_test.invoke::<u32>(&[]), 2);
         assert_eq!(obj.ref_count(), 1);
 
         weak_test
@@ -379,7 +379,7 @@ fn closure() {
         }
 
         let obj = glib::Object::new::<glib::Object>(&[]).unwrap();
-        assert_eq!(obj.ref_count_in_closure(), 3);
+        assert_eq!(obj.ref_count_in_closure(), 2);
     }
 
     {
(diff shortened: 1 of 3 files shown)


glib/src/object.rs
```diff
@@ -3168,6 +3168,50 @@ impl<T: ObjectType> ObjectExt for T {
     }
 }
 
+// Helper struct to avoid creating an extra ref on objects inside closure watches. This is safe
+// because `watch_closure` ensures the object has a ref when the closure is called.
+#[doc(hidden)]
+pub struct WatchedObject<T: ObjectType>(*mut T::GlibType);
```
sdroege, line 3174:
```suggestion
pub struct WatchedObject<T: ObjectType>(*mut T::GlibType);

unsafe impl<T: ObjectType + Send + Sync> Send for WatchedObject {}
```

glib/src/object.rs
```diff
@@ -3168,6 +3168,50 @@ impl<T: ObjectType> ObjectExt for T {
...
+    where
+        T: FromGlibPtrBorrow<*mut <T as ObjectType>::GlibType>,
+    {
+        from_glib_borrow(self.0)
+    }
+}
+
+#[doc(hidden)]
+pub trait Watchable<T: ObjectType> {
+    fn get_object(&self) -> WatchedObject<T>;
```
sdroege, line 3191:
```suggestion
    fn get_watched_object(&self) -> WatchedObject<T>;
```

Maybe?

glib-macros/tests/test.rs
```diff
@@ -359,7 +359,7 @@ fn closure() {
         assert_eq!(obj.ref_count(), 1);
         let weak_test = glib::closure_local!(@watch obj => move || obj.ref_count());
         assert_eq!(obj.ref_count(), 1);
-        assert_eq!(weak_test.invoke::<u32>(&[]), 3);
+        assert_eq!(weak_test.invoke::<u32>(&[]), 2);
```
sdroege, line 362: Why? :)

sdroege (commented):
I was trying to get around that without a new trait, but I couldn't find a way. You apparently also couldn't :) Oh well! At least it doesn't have to be in the prelude

glib/src/object.rs
bilelmoussaoui: maybe even `watched_object`

glib/src/object.rs
sdroege: Should probably also use `std::ptr::NonNull` here

sdroege (approved):

PR #564, opened 2022-02-18
Variant::from_dict_entry

arcnmx:
This is a continuation of #519 (addressing #507) for creating a `DictEntry`-typed variant without static types (I missed it in my initial bug report, [even though it was present in the example](https://github.com/jf2048/glib-serde/blob/19e307a86786d385e655183c99212605e866d3dd/src/variant/mod.rs#L89)). The `VariantType` constructor is useful when working with these variants (for example in combination with `Variant::array_from_iter_with_type` to construct a vardict manually), so I included it along with the others.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -493,6 +493,20 @@ impl Variant {
     }
 
     // rustdoc-stripper-ignore-next
+    /// Creates a new dictionary entry Variant.
+    ///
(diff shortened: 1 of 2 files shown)


sdroege:
CC @cgwalters @lucab

sdroege (approved):

cgwalters (commented):
(skimmed) LGTM, but could probably use some unit tests

Issue #382, opened 2021-11-20
Add variant of property binding transform functions that work on concrete types instead of `Value`s


sdroege:
The main challenge here is to get the following to compile as part of the builder:

```rust
    pub fn transform_from_meh<'b, S: FromValue<'b>, T: ToValue,
        F: Fn(&crate::Binding, S) -> Option<T> + Send + Sync + 'static,
    >(
        self,
        func: F,
    ) -> Self {
        Self {
            transform_from: Some(Box::new(move |binding, from_value| {
                let from_value = from_value.get().expect("Wrong value type");
                func(binding, from_value).map(|r| r.to_value())
            })),
            ..self
        }
    }
```

The problem here is the lifetime, as usual. Ideally we'd like to use `&str` etc for the parameters of the closure to avoid unnecessary copies, so working with `for<'b> FromValue<'b> + 'static` is not a solution.

The above fails with

```
error[E0495]: cannot infer an appropriate lifetime for autoref due to conflicting requirements
    --> glib/src/object.rs:3317:45
     |
3317 |                 let from_value = from_value.get().expect("Wrong value type");
     |                                             ^^^
     |
note: first, the lifetime cannot outlive the anonymous lifetime #2 defined on the body at 3316:43...
    --> glib/src/object.rs:3316:43
     |
3316 |               transform_from: Some(Box::new(move |binding, from_value| {
     |  ___________________________________________^
3317 | |                 let from_value = from_value.get().expect("Wrong value type");
3318 | |                 func(binding, from_value).map(|r| r.to_value())
3319 | |             })),
     | |_____________^
note: ...so that reference does not outlive borrowed content
    --> glib/src/object.rs:3317:34
     |
3317 |                 let from_value = from_value.get().expect("Wrong value type");
     |                                  ^^^^^^^^^^
note: but, the lifetime must be valid for the lifetime `'b` as defined on the method body at 3309:31...
    --> glib/src/object.rs:3309:31
     |
3309 |     pub fn transform_from_meh<'b, S: FromValue<'b>, T: ToValue,
     |                               ^^
note: ...so that the types are compatible
    --> glib/src/object.rs:3317:45
     |
3317 |                 let from_value = from_value.get().expect("Wrong value type");
     |                                             ^^^
     = note: expected `FromValue<'_>`
                found `FromValue<'b>`
```

PR #567, opened 2022-02-19
glib: Implement signal connection with static types for the parameter…

sdroege:
…s/return value and automatic `Value` handling

----

CC @jf2048

### DIFF
--- a/glib/src/closure.rs
+++ b/glib/src/closure.rs
@@ -9,11 +9,8 @@ use std::slice;
 use libc::{c_uint, c_void};
 
 use crate::translate::{from_glib_none, mut_override, ToGlibPtr, ToGlibPtrMut, Uninitialized};
-use crate::value::FromValue;
-use crate::StaticType;
-use crate::ToValue;
-use crate::Type;
-use crate::Value;
+use crate::value::{FromValue, ToValue, Value};
+use crate::{StaticType, Type};
 
 wrapper! {
     #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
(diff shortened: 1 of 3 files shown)


glib/src/closure.rs
```diff
@@ -393,6 +390,63 @@ impl<T: for<'a> FromValue<'a> + StaticType + 'static> TryFromClosureReturnValue
...
 
+pub trait SignalClosure<A> {
+    type Output: ToClosureReturnValue;
+
+    fn call(&self, args: &[Value]) -> Option<Value>;
+    fn parameter_types() -> Vec<Type>;
+}
+
+macro_rules! signal_closure_impl ({ $($param:ident)* } => {
+    impl<Func: Fn($($param),*) -> Res, Res: ToClosureReturnValue, $($param: StaticType + for<'a> FromValue<'a>,)*> SignalClosure<($($param,)*)> for Func
```
sdroege, line 401: Because of the `for<'a>` it is currently required that each parameter is `'static`, e.g. `&str` can't be used.

Fixing this requires adding a lifetime parameter to the trait and the parameter of the `call()` function, but I couldn't figure out how to make everything match up then.

glib/src/subclass/object.rs
```diff
@@ -569,14 +569,9 @@ mod test {
 
         let name_changed_triggered = Arc::new(AtomicBool::new(false));
         let name_changed_clone = name_changed_triggered.clone();
-        obj.connect("name-changed", false, move |args| {
-            let _obj = args[0].get::<Object>().expect("Failed to get args[0]");
-            let name = args[1].get::<&str>().expect("Failed to get args[1]");
-
+        obj.connect_static("name-changed", false, move |_obj: Object, name: String| {
```
sdroege, line 572: I.e. the goal is it to make this compile:

```suggestion
        obj.connect_static("name-changed", false, move |_obj: Object, name: &str| {
```

glib/src/object.rs
```diff
@@ -1662,6 +1662,23 @@ pub trait ObjectExt: ObjectType {
...
+    // rustdoc-stripper-ignore-next
+    /// Connect to the signal `signal_name` on this object.
+    ///
+    /// If `after` is set to `true` then the callback will be called after the default class
+    /// handler of the signal is emitted, otherwise before.
+    ///
+    /// # Panics
+    ///
+    /// If the signal does not exist.
+    // FIXME: Add all the other variants, rename this to connect(), rename existing connect() to connect_values()
```
sdroege, line 1674: And once that is done, this FIXME should be implemented

glib/src/closure.rs
sdroege: Also asked on users.rust-lang.org: https://users.rust-lang.org/t/problems-matching-up-lifetimes-between-various-traits-and-closure-parameters/71994

jf2048: There's some promising comments in that thread. I think maybe the same approach would also work for https://github.com/gtk-rs/gtk-rs-core/issues/382

sdroege: Yeah, I'll have to find some time to try this out :)

PR #568, opened 2022-02-19
Smaller spec numeric definition

ranfdev:
Small refactor to reduce code duplication.

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -377,96 +377,66 @@ macro_rules! define_param_spec_min_max {
 }
 
 macro_rules! define_param_spec_numeric {
-    ($rust_type:ident, $ffi_type:path, $value_type:ty, $rust_type_offset:expr, $from_glib:expr) => {
+    ($rust_type:ident, $ffi_type:path, $value_type:ty, $rust_type_offset:expr, $ffi_fun:ident, $alias:literal) => {
         define_param_spec!($rust_type, $ffi_type, $rust_type_offset);
-        define_param_spec_default!($rust_type, $value_type, $from_glib);
-        define_param_spec_min_max!($rust_type, $value_type, $from_glib);
+        define_param_spec_default!($rust_type, $value_type, |x| x);
+        define_param_spec_min_max!($rust_type, $value_type, |x| x);
+
+        impl $rust_type {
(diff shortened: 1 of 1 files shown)


ranfdev:
I don't know that happened with the docs. @bilelmoussaoui any idea?

glib/src/param_spec.rs
```diff
@@ -377,96 +377,66 @@ macro_rules! define_param_spec_min_max {
...
+                        default_value,
+                        flags.into_glib(),
+                    ))
+                }
+            }
+        }
     };
 }
 
+fn assert_param_name(name: &str) {
```
sdroege, line 414: This needs to be a macro or use `#[track_caller]`, otherwise the panic messages are going to point at this very line instead of where things actually failed.

glib/src/param_spec.rs
```diff
@@ -377,96 +377,66 @@ macro_rules! define_param_spec_min_max {
...
+    );
+}
 define_param_spec_numeric!(
     ParamSpecChar,
     gobject_ffi::GParamSpecChar,
     i8,
     0,
-    |x| x
+    g_param_spec_char,
+    "g_param_spec_char"
```
sdroege, line 427: You should be able to use [`stringify!`](https://doc.rust-lang.org/std/macro.stringify.html) for that

sdroege (commented):
Looks good to me otherwise

glib/src/param_spec.rs
ranfdev: I wanted to use `stringify!` as well, but the macro call ends up here
`#[doc(alias = stringify!($ffi_fun))]`
and macro calls in that position are not possible, because `alias` expects a literal.

sdroege: What if you make `alias` expect a `expr` instead?

sdroege: Oh, you can't. You mean the doc alias needs that. Can you create an issue about that for rustc? Seems fine like this here then

sdroege (approved):

glib/src/param_spec.rs
ranfdev: >  Can you create an issue about that for rustc?

https://github.com/rust-lang/rust/issues/94180

PR #569, opened 2022-02-19
remove problematic phantom in glib_shared_wrapper

ranfdev:
While trying to use a generic shared wrapper I got this error:
```
warning: unnecessary parentheses around type                                                                           
   --> /home/lollo/gtk-rs-core/glib/src/shared.rs:34:47                                                           
    |                                                      
34  |               phantom: std::marker::PhantomData<($($($generic),+)?)>,                                            
    |                                                 ^                 ^        
```

But then I found out that field is useless anyway, because the generic type is already used in the field above, so I've removed the phantom field.

### DIFF
--- a/glib/src/boxed.rs
+++ b/glib/src/boxed.rs
@@ -33,7 +33,6 @@ macro_rules! glib_boxed_wrapper {
         #[repr(transparent)]
         $visibility struct $name $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? {
             inner: $crate::boxed::Boxed<$ffi_name, Self>,
-            phantom: std::marker::PhantomData<($($($generic),+)?)>,
         }
 
         impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? std::clone::Clone for $name $(<$($generic),+>)? {
@@ -41,7 +40,6 @@ macro_rules! glib_boxed_wrapper {
             fn clone(&self) -> Self {
                 Self {
                     inner: std::clone::Clone::clone(&self.inner),
-                    phantom: std::marker::PhantomData,
                 }
             }
         }
@@ -145,7 +143,6 @@ macro_rules! glib_boxed_wrapper {
             unsafe fn from_glib_none(ptr: *mut $ffi_name) -> Self {
                 Self {
                     inner: $crate::translate::from_glib_none(ptr),
-                    phantom: std::marker::PhantomData,
                 }
             }
         }
@@ -156,7 +153,6 @@ macro_rules! glib_boxed_wrapper {
             unsafe fn from_glib_none(ptr: *const $ffi_name) -> Self {
                 Self {
                     inner: $crate::translate::from_glib_none(ptr),
-                    phantom: std::marker::PhantomData,
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


jf2048:
Could probably remove it from `Boxed` and `BoxedInline` too?

sdroege:
Can you also give some example code for future reference and explain what you're doing? The `PhantomData` can probably go away here indeed.

sdroege:
> `34  |               phantom: std::marker::PhantomData<($($($generic),+)?)>,`

This is simply missing a `$` in front of the outer `()` or not?

sdroege:
> The `PhantomData` can probably go away here indeed.

And yes it can go away there and in `boxed` and `boxed_inline` (but not in `object`). Can you update the PR accordingly?

ranfdev:
> Can you also give some example code for future reference and explain what you're doing? 

I was just trying some stuff. I've replace the macro code defining a ParamSpec with an "equivalent" implementation using generics.
By the way, the change is backwards compatible, so, if you like it, I can open a PR.
https://github.com/ranfdev/gtk-rs-core/commit/411e00cbeffe9d67a276fcd687da5642ced71a20

sdroege:
Hmm, that seems unnecessarily complicated. What advantage does it have? :)

ranfdev:
> What advantage does it have? :)

Sincerely? Not much.
Since that's "normal" code and not a macro, there's better support from the IDE/editor (primarily, there's better code navigation).
Also, probably, the compiler is able to optimize the code slightly better.
That's it.

I agree this change is unnecessary 😆

sdroege:
OK, let's not do that then :) @ranfdev Can you update this PR so we can get it merged?

ranfdev:
I'd prefer #575 gets merged before updating this PR. Else, the changes may get in conflict, and I don't want to resolve a merge conflict in this macro code.

sdroege:
Ok :)

ranfdev:
> `BoxedInline` too?

I can't remove it from there, because in that struct the panthom field is the only one containing the generics.

By the way, generics in `BoxedInline` don't work at all right now...
For example, the following code doesn't compile at all. Adding the generic parameter creates some errors around lifetimes...
```
wrapper! {
    pub struct DateGenericTest<T>(BoxedInline<ffi::GDate>);

    match fn {
        copy => |ptr| gobject_ffi::g_boxed_copy(ffi::g_date_get_type(), ptr as *const _) as *mut _,
        free => |ptr| ffi::g_date_free(ptr),
        type_ => || ffi::g_date_get_type(),
    }
}
```

sdroege:
Doesn't seem very problematic. Generics support for the wrapper macro is rarely useful/useable anyway. If someone needs it for boxed inline they'll have to fix that then.

ranfdev:
> they'll have to fix that then.

Don't worry, I've fixed it ;)

To me, it's now ready to be merged

sdroege:
Please remove the pspec commit :)

ranfdev:
> Please remove the pspec commit :)
Whoops, that was unintentional. I've added it temporarily to check if the generics worked correctly

sdroege (approved):

sdroege:
Happens :)

PR #575, opened 2022-02-21
use Self where possible in wrapper macros

ranfdev:


glib/src/boxed.rs
```diff
@@ -288,7 +288,7 @@ macro_rules! glib_boxed_wrapper {
             type Checker = $crate::value::GenericValueTypeOrNoneChecker<Self>;
```
sdroege, line 288: One line above there's another `Self`: `&'a $name $(<$($generic),+>)?` can become `&Self`.

Same in the other files

sdroege:
Looks good to me otherwise, thanks. This makes this all a bit more readable :)

glib/src/boxed.rs
ranfdev: Are you sure? I can't find that. By the way, you've commented on the wrong side, the side of upstream/master, which doesn't have the changes.

sdroege: Because GitHub's review interface sucks, I can't comment on the correct line.

See the end of this line: https://github.com/gtk-rs/gtk-rs-core/blob/ca43ffe5e344e06ababdbef533380ca3a2a6d252/glib/src/boxed.rs#L287

ranfdev: I can't replace that with `Self`. I can't say `impl Display for Self`. What's `Self` in there?
I can only use `Self` after the target of the `impl` block has been specified

sdroege: Oh right, that was stupid. Sorry :)

sdroege (approved):

PR #577, opened 2022-02-23
Add new classes from glib 2.72

jf2048:
- `GDebugController`
- `GDebugControllerDBus`
- `GBindingGroup`

`GSignalGroup` can't really get a safe binding until this: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2521


gio/src/auto/debug_controller_dbus.rs
```diff
@@ -0,0 +1,114 @@
...
+                Err(from_glib_full(error))
+            }
+        }
+    }
+}
+
+pub trait DebugControllerDBusExt: 'static {
+    #[cfg(any(feature = "v2_72", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_72")))]
+    fn connection(&self) -> Option<DBusConnection>;
```
bilelmoussaoui, line 57: Probably not nullable

glib/src/gobject/auto/binding_group.rs
```diff
@@ -0,0 +1,124 @@
...
+crate::wrapper! {
+    #[doc(alias = "GBindingGroup")]
+    pub struct BindingGroup(Object<gobject_ffi::GBindingGroup>);
+
+    match fn {
+        type_ => || gobject_ffi::g_binding_group_get_type(),
+    }
+}
+
+impl BindingGroup {
```
bilelmoussaoui, line 26: Should probably manually implement like the current GBinding?

jf2048: Yes, jut realized I forgot that. At the very least `bind` needs it

gio/src/auto/debug_controller_dbus.rs
jf2048: Implemented it manually

glib/src/gobject/auto/binding_group.rs
jf2048: Copied most of the implementation from `BindingBuilder`

gio/src/auto/debug_controller_dbus.rs
sdroege: You can set `nullable = false` in `Gir.toml` instead

gio/Gir.toml
```diff
@@ -553,6 +554,15 @@ status = "generate"
     name = "no_match_rule"
     version = "2.70"
 
+[[object]]
+name = "Gio.DebugControllerDBus"
+status = "generate"
+manual_traits = ["DebugControllerDBusExtManual"]
+    [[object.property]]
+    name = "connection"
```
bilelmoussaoui, line 562: hmm, not sure how it ended up marking the property as nullable :/ maybe it should be but then the constructor isn't so...

sdroege: I guess pointer properties are nullable by default unless overridden in the docs?

jf2048: Sadly using `(not nullable)` does not work on properties, it would be nice if it did

jf2048:
This can wait until closer to the 2.72 release

sdroege (approved):
Looks good to me. Why wait for 2.72 release?

jf2048:
This is broken for me as long as the sysdep version depends on 2.72, the pkg-config version is still only 2.71 until release. But I guess it doesn't matter much because CI is not using that anyway?

sdroege:
> the pkg-config version is still only 2.71 until release

Hrm yes, it should only require 2.71.X right now and not a non-existing 2.72. Can you submit a PR?

jf2048:
Yes, this is ready

PR #578, opened 2022-02-24
glib-macros: Port clone failure tests to trybuild

jf2048:
cc @GuillaumeGomez 

Saves about 65 lines of code


jf2048:
A downside with this approach: Eventually I would like to make the error messages more specific, e.g. the "Can't use self" error should point only to the `self` token. Making those kind of changes will break this approach because of the lack of fuzzy matching.

glib-macros/tests/clone.rs
```diff
@@ -0,0 +1,111 @@
...
+     "Expected `move` after `async`, found `{`"),
+    ("clone!(@strong v => move || async println!(\"foo\");)",
+     "Expected `move` after `async`, found `println`"),
+    ("clone!(@strong v => move || async move println!(\"foo\");)",
+     "Expected block after `| async move`"),
+];
+
+#[test]
+fn clone_failures() {
+    let exe = std::env::current_exe().unwrap();
```
GuillaumeGomez, line 76: It'll create a lot of files which will remain after the tests are done. :-/

GuillaumeGomez:
If you don't mind, I'll cherry-pick your commit and give a try to my `trybuild` extension to have inline tests instead of handling the file part ourselves.

GuillaumeGomez:
By using your commit, I was able to reduce the diff quite a lot. I'll send a PR with your commit in it shortly.

PR #580, opened 2022-02-25
glib-macros: Port clone failure tests to trybuild2

GuillaumeGomez:
This is a simplification of #578.

cc @jf2048


GuillaumeGomez:
CI is failing because of clippy lints. Sending a PR to fix them...

jf2048:
LGTM, if we take this approach I was also thinking about using `quote::quote!` instead of strings

GuillaumeGomez:
You lost me there. Which strings are you referring to?

GuillaumeGomez:
And rebased. Let's see if the CI is happier.

GuillaumeGomez:
CI passed! \o/

jf2048:
Like:
```rust
   (quote! { clone!( => move || {}) },
     "If you have nothing to clone, no need to use this macro!"),
    (quote! { clone!(|| {}) },
     "If you have nothing to clone, no need to use this macro!"),
    (quote! { clone!(|a, b| {}) },
     "If you have nothing to clone, no need to use this macro!"),
```
etc, just to make it clearer this is rust syntax

sdroege:
It gives you proper tooling support for writing more than a single line of code in there.

GuillaumeGomez:
You can write more than one already. I don't see why using `quote` improves anything.

GuillaumeGomez:
There is syntax highlighting in `quote`? Also, you can write that in a real rust file if you prefer.

sdroege:
What's the plan here?

GuillaumeGomez:
For me it's ready as is.

PR #585, opened 2022-02-26
add AutoObjectProps and AutoObjectSignals

ranfdev:
Different take on #546.
These traits are not going to be implemented by the user. These are going to be implemented by macros. The user will then delegate the `ObjectImpl` functions to the ones in these traits

### DIFF
--- a/glib/src/subclass/object.rs
+++ b/glib/src/subclass/object.rs
@@ -133,6 +133,75 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
     }
 }
 
+#[derive(Debug)]
+pub struct MissingPropertyHandler<'a>(&'a str);
+
+impl<'a> MissingPropertyHandler<'a> {
+    fn name(&self) -> &'a str {
+        self.0
+    }
+}
+impl<'a> From<&'a ParamSpec> for MissingPropertyHandler<'a> {
+    fn from(p: &'a ParamSpec) -> Self {
+        Self(p.name())
+    }
+}
+
+impl<'a> std::fmt::Display for MissingPropertyHandler<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "missing handler for property {}", self.name())
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// Trait containing only the property related functions of `ObjectImpl`.
+/// Implemented by the `Props` macro.
+/// When implementing `ObjectImpl` you may want to delegate the function calls to this trait.
+/// Some functions have a different return type from `ObjectImpl` to handle possible errors caused by missing
+/// properties.
+pub trait DerivedObjectProperties: ObjectSubclass {
+    // rustdoc-stripper-ignore-next
+    /// Properties installed for this type.
+    fn derived_properties() -> &'static [ParamSpec] {
+        &[]
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a `Result`.
+    fn derived_set_property<'a>(
+        &self,
+        _obj: &Self::Type,
+        _id: usize,
+        _value: &Value,
+        _pspec: &'a ParamSpec,
+    ) -> Result<(), MissingPropertyHandler<'a>> {
+        Err(MissingPropertyHandler(_pspec.name()))
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a `Result`.
+    fn derived_property<'a>(
+        &self,
+        _obj: &Self::Type,
+        _id: usize,
+        _pspec: &'a ParamSpec,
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
cc @jf2048 as you wanted to look at the signals macro

glib/src/subclass/object.rs
```diff
@@ -133,6 +133,52 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+/// properties.
+pub trait AutoObjectProps: ObjectSubclass {
+    // rustdoc-stripper-ignore-next
+    /// Properties installed for this type.
+    fn properties() -> &'static [ParamSpec] {
+        &[]
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a boolean, `true` in the case
```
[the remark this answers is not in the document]
ranfdev: > ti should fallback to this one which shouldn't fail unless something is very broken

Usually the function `set_property` panics when there isn't an handler for the requested property.
Let's say this is my `ObjectImpl` (with `AutoObjectProps::set_property` *not* returning a bool):
```rust
impl ObjectImpl for MyObject {
  fn set_property(&self, ..., pspec, value) {
    if pspec.name == "author_name" {
      self.author_name = value;
    } else {
      <Self as AutoObjectProps>::set_property(self, ..., pspec, value)
    }
  }
}
```
How do I know if the function should panic? Yes, I could compare the `pspec` parameter against `AutoObjectProps::properties()`, but that's a redundant check, given that the check is already done by `AutoObjectProps::set_property()`.

With a bool the code becomes
```rust
fn set_property(&self, ..., pspec, value) {
    if pspec.name == "author_name" {
      self.author_name = value;
    } else if !<Self as AutoObjectProps>::set_property(self, ..., pspec, value) {
      unreachable!("missing handler for property {pspec.name}")
    }
  }
```

glib/src/subclass/object.rs
```diff
@@ -133,6 +133,52 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+        _value: &Value,
+        _pspec: &ParamSpec,
+    ) -> bool {
+        false
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns an `Option<Value>`,
+    /// `None` in the case the property wasn't found
+    fn property(&self, _obj: &Self::Type, _id: usize, _pspec: &ParamSpec) -> Option<Value> {
```
sdroege, line 165: It should just panic if the property was not found or not? It would've been an error to call it in that case

sdroege (commented):
The general approach makes sense to me but I wouldn't want to merge this without the infrastructure around it being functional

glib/src/subclass/object.rs
```diff
@@ -133,6 +133,52 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
     }
 }
 
+// rustdoc-stripper-ignore-next
+/// Trait containing only the property related functions of `ObjectImpl`.
+/// Implemented by the `Props` macro.
+/// When implementing `ObjectImpl` you may want to delegate the function calls to this trait.
+/// Some functions have a different return type from `ObjectImpl` to handle possible errors caused by missing
+/// properties.
+pub trait AutoObjectProps: ObjectSubclass {
```
sdroege, line 142:
```suggestion
pub trait AutoObjectProperties: ObjectSubclass {
```

and elsewhere? We don't have to pay taxes per character :)

bilelmoussaoui: Also I wouldn't prefix the traits with Auto.

glib/src/subclass/object.rs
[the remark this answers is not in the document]
bilelmoussaoui: I am not happy that you would have to cast `<Self as Trait>` here, maybe we could name those functions slightly differently?

glib/src/subclass/object.rs
ranfdev: `Auto` makes it clear the trait shouldn't be implemented by the user

glib/src/subclass/object.rs
ranfdev: I'd prefer the panic to happen in `ObjectImpl::property`, not here. It's more explicit and more flexible.

jf2048 (changes requested):

glib/src/subclass/object.rs
```diff
@@ -133,6 +133,75 @@ unsafe extern "C" fn dispose<T: ObjectImpl>(obj: *mut gobject_ffi::GObject) {
...
+    fn derived_properties() -> &'static [ParamSpec] {
+        &[]
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`ObjectImpl`](trait.ObjectImpl.html) but returns a `Result`.
+    fn derived_set_property<'a>(
+        &self,
+        _obj: &Self::Type,
+        _id: usize,
```
jf2048, line 174: We could probably just remove the `id` arg here and on `derived_property` since they won't be used

PR #586, opened 2022-02-28
glib: Automatically implement `Send + Sync` on object subclasses if t…

sdroege:
…heir implementation struct implements those traits


glib/src/subclass/object.rs
```diff
@@ -471,6 +471,13 @@ mod test {
         assert_eq!(properties[3].name(), "child");
     }
 
+    #[test]
+    fn test_auto_send_sync() {
+        fn check<T: Send + Sync>() {}
+
+        check::<ChildObject>();
```
sdroege, line 478: @GuillaumeGomez Once your trycompile test stuff is in you could also add a test that `check::<SimpleObject>()` fails to compile

GuillaumeGomez: You can already use it. Or do you mean something else?

sdroege: We don't depend on it yet

GuillaumeGomez: Since it's a `dev-dependences`, you have my blessing. :)

sdroege: Done

glib/tests/subclass_compiletest.rs
```diff
@@ -0,0 +1,8 @@
+#[test]
+pub fn test() {
+    let t = trybuild2::TestCases::new();
+
+    t.pass("tests/subclass_compiletest/01-auto-send-sync.rs");
+    t.compile_fail("tests/subclass_compiletest/02-no-auto-send-sync.rs");
+    t.compile_fail("tests/subclass_compiletest/03-object-no-auto-send-sync.rs");
```
sdroege, line 7: This last one makes sure that none of the bindings around C classes automatically get those traits implemented

GuillaumeGomez (dismissed):

PR #587, opened 2022-03-04
Use glib 2.71 for minimum v2_72 pkg-config versions

jf2048:
Per https://github.com/gtk-rs/gtk-rs-core/pull/577#issuecomment-1058912597

Not sure this makes sense, we don't use odd version numbers for any of the other features?

### DIFF
--- a/gio/sys/Cargo.toml
+++ b/gio/sys/Cargo.toml
@@ -89,4 +89,4 @@ version = "2.68"
 version = "2.70"
 
 [package.metadata.system-deps.gio_2_0.v2_72]
-version = "2.72"
+version = "2.71"
--- a/glib/gobject-sys/Cargo.toml
+++ b/glib/gobject-sys/Cargo.toml
@@ -62,4 +62,4 @@ version = "2.68"
 version = "2.70"
 
 [package.metadata.system-deps.gobject_2_0.v2_72]
-version = "2.72"
+version = "2.71"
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
> Not sure this makes sense, we don't use odd version numbers for any of the other features?

At least in gtk4-rs, we use that a lot. I guess we didn't in the past because we were not updating the bindings that often and were waiting until there is a new release and for the GIR files to land on debian/ubuntu repositories

jf2048:
Hmm, ok. New ubuntu LTS is coming next month so hopefully CI can be updated then.

bilelmoussaoui:
> Hmm, ok. New ubuntu LTS is coming next month so hopefully CI can be updated then.


We no longer Ubuntu for getting the gir files and I plan to update the images we use to build things directly from upstream so we can test new APIs as soon as they land upstream and ensure things are working fine for the rust bindings

jf2048:
Sorry not that, I mean this one: https://github.com/gtk-rs/gtk-rs-core/blob/caa47b15e3ded4ea1c583c8b3b4fbb936e571462/.github/workflows/CI.yml#L12

bilelmoussaoui:
Ah sorry misunderstood, but nonetheless we use an image built from https://github.com/gtk-rs/gtk-rs-core/blob/master/Dockerfile. I will update it this weekend

sdroege:
> At least in gtk4-rs, we use that a lot.

In GStreamer I also do that since a long time, e.g. `v1_20` required 1.19.x to be installed until the 1.20.0 release.

sdroege (approved):

sdroege:
The same also applies to pango btw, but we can worry about that once someone needs it.

Issue #502, opened 2022-01-21
[FEATURE REQUEST] Add `is()' to GParamSpec binding


dabrain34:
Provides an API to tell if an object is a GParamSpec type_()

sdroege:
@bilelmoussaoui Should maybe do the same for other fundamental types (expressions etc) if not already present

PR #588, opened 2022-03-06
glib: Allow using other error types and value type checkers for …

sdroege:
…optional values


jf2048:
Looks like a prereq for #502

sdroege:
> Looks like an additional bit for #502

In what sense?

jf2048:
I think we want to check those (and other fundamental types) when using `FromValue`

sdroege:
Did you link to the wrong issue? Or I'm misunderstanding what you want to do :) Can you give some more details?

sdroege:
@GuillaumeGomez @bilelmoussaoui This one is ready now, I made sure things are working correctly in GStreamer and the new feature can actually be used properly.

GuillaumeGomez (approved):

PR #589, opened 2022-03-06
More gio::Cancellable methods

### LINKED ISSUE #240: API to bridge futures and GCancellable
I came up with this: https://github.com/coreos/rpm-ostree/pull/3083/commits/66b6e7cf4149cc48a96e64455ff83ef1cee51d53#diff-7939b1bb63c982978270ab1c565b42a56f7fea352ce7b2cdb792e1272fe4c6acR16
Does this API exist somewhere already?  It feels useful enough to add as an API on e.g. `GCancellable` perhaps?

jf2048:
Fixes #240, this should work on any async runtime. The oneshot just gets triggered from whatever thread calls `cancel`.

`::connect` and `::disconnect` are maybe not the best choice for method names because they conflict with `ObjectExt`

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -342,7 +342,7 @@ manual_traits = ["CancellableExtManual"]
     ignore = true
     [[object.function]]
     name = "source_new"
-    # Needs manual bindings for setting callback, etc
+    # Only for use with other souce implementations
     ignore = true
     [[object.signal]]
     name = "cancelled"
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -327,6 +327,15 @@ generate_builder = true
 name = "Gio.Cancellable"
 status = "generate"
 concurrency = "send+sync"
+manual_traits = ["CancellableExtManual"]
+    [[object.function]]
+    name = "connect"
+    # Needs manual bindings for closures
+    ignore = true
+    [[object.function]]
+    name = "disconnect"
+    # Needs manual bindings for CancelledHandlerId
+    ignore = true
     [[object.function]]
     name = "reset"
     #undefined behaviour
@@ -334,6 +343,10 @@ concurrency = "send+sync"
     [[object.function]]
     name = "source_new"
     # Needs manual bindings for setting callback, etc
+    ignore = true
+    [[object.signal]]
+    name = "cancelled"
+    # Racy, use 'connect' and 'disconnect' instead
     ignore = true
 
 [[object]]
--- a/gio/src/cancellable.rs
+++ b/gio/src/cancellable.rs
@@ -2,6 +2,8 @@
 
 use crate::prelude::*;
 use crate::Cancellable;
+use futures_channel::oneshot;
+use futures_core::Future;
 use glib::{translate::*, IsA};
 use std::num::NonZeroU64;
 
@@ -62,6 +64,10 @@ pub trait CancellableExtManual {
     /// operation is finished and the signal handler is removed.
     #[doc(alias = "g_cancellable_disconnect")]
     fn disconnect_cancelled(&self, id: CancelledHandlerId);
+    // rustdoc-stripper-ignore-next
+    /// Returns a `Future` that completes when the cancellable becomes cancelled. Completes
+    /// immediately if the cancellable is already cancelled.
+    fn future(&self) -> std::pin::Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>>;
 }
 
 impl<O: IsA<Cancellable>> CancellableExtManual for O {
@@ -101,6 +107,19 @@ impl<O: IsA<Cancellable>> CancellableExtManual for O {
     fn disconnect_cancelled(&self, id: CancelledHandlerId) {
         unsafe { ffi::g_cancellable_disconnect(self.as_ptr() as *mut _, id.as_raw()) };
     }
+    fn future(&self) -> std::pin::Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>> {
+        let cancellable = self.as_ref().clone();
+        let (tx, rx) = oneshot::channel();
+        let id = cancellable.connect_cancelled(move |_| {
+            let _ = tx.send(());
+        });
+        Box::pin(async move {
+            rx.await.unwrap();
+            if let Some(id) = id {
+                cancellable.disconnect_cancelled(id);
+            }
+        })
+    }
 }
 
 #[cfg(test)]
@@ -113,5 +132,36 @@ mod tests {
         let id = c.connect_cancelled(|_| {});
         c.cancel(); // if it doesn't crash at this point, then we're good to go!
         c.disconnect_cancelled(id.unwrap());
+    }
+
+    #[test]
+    fn cancellable_future() {
+        let c = Cancellable::new();
+        c.cancel();
+        glib::MainContext::new().block_on(c.future());
+    }
+
+    #[test]
+    fn cancellable_future_thread() {
+        let cancellable = Cancellable::new();
+        let c = cancellable.clone();
+        std::thread::spawn(move || c.cancel()).join().unwrap();
+        glib::MainContext::new().block_on(cancellable.future());
+    }
+
+    #[test]
+    fn cancellable_future_delayed() {
+        let ctx = glib::MainContext::new();
+        let c = Cancellable::new();
+        let (tx, rx) = oneshot::channel();
+        {
+            let c = c.clone();
+            ctx.spawn_local(async move {
+                c.future().await;
+                tx.send(()).unwrap();
+            });
+        }
+        std::thread::spawn(move || c.cancel()).join().unwrap();
+        ctx.block_on(rx).unwrap();
     }
 }
--- a/gio/src/cancellable.rs
+++ b/gio/src/cancellable.rs
@@ -1,10 +1,117 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::prelude::*;
+use crate::Cancellable;
+use glib::{translate::*, IsA};
+use std::num::NonZeroU64;
 
-#[test]
-fn check_callback() {
-    let c = crate::Cancellable::new();
-    c.connect_cancelled(|_| {});
-    c.cancel(); // if it doesn't crash at this point, then we're good to go!
+// rustdoc-stripper-ignore-next
+/// The id of a cancelled handler that is returned by `CancellableExtManual::connect`. This type is
+/// analogous to [`glib::SignalHandlerId`].
+#[derive(Debug, Eq, PartialEq)]
+#[repr(transparent)]
+pub struct CancelledHandlerId(NonZeroU64);
+
+impl CancelledHandlerId {
+    // rustdoc-stripper-ignore-next
+    /// Returns the internal signal handler ID.
+    pub unsafe fn as_raw(&self) -> libc::c_ulong {
+        self.0.get() as libc::c_ulong
+    }
+}
+
+impl TryFromGlib<libc::c_ulong> for CancelledHandlerId {
+    type Error = GlibNoneError;
+    #[inline]
+    unsafe fn try_from_glib(val: libc::c_ulong) -> Result<Self, GlibNoneError> {
+        NonZeroU64::new(val as u64).map(Self).ok_or(GlibNoneError)
+    }
+}
+
+pub trait CancellableExtManual {
+    // rustdoc-stripper-ignore-next
+    /// Convenience function to connect to the `signal::Cancellable::cancelled` signal. Also
+    /// handles the race condition that may happen if the cancellable is cancelled right before
+    /// connecting. If the operation is cancelled from another thread, `callback` will be called
+    /// in the thread that cancelled the operation, not the thread that is running the operation.
+    /// This may be the main thread, so the callback should not do something that can block.
+    ///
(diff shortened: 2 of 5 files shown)


gio/Gir.toml
```diff
@@ -327,6 +327,17 @@ generate_builder = true
 name = "Gio.Cancellable"
 status = "generate"
 concurrency = "send+sync"
+manual_traits = ["CancellableExtManual"]
+    [[object.function]]
+    name = "connect"
+    # Needs manual bindings for SignalHandlerId
+    manual = true
+    doc_trait_name = "CancellableExtManual"
```
bilelmoussaoui, line 335:
```suggestion
```
you shouldn't need those normally

gio/src/cancellable.rs
```diff
@@ -1,10 +1,153 @@
...
+impl FromGlib<libc::c_ulong> for CancelledHandlerId {
+    #[inline]
+    unsafe fn from_glib(val: libc::c_ulong) -> Self {
+        Self(val)
+    }
+}
+
+pub trait CancellableExtManual {
+    #[doc(alias = "g_cancellable_connect")]
+    fn connect<F: FnOnce(&Self) + Send + Sync + 'static>(&self, f: F) -> CancelledHandlerId;
```
bilelmoussaoui, line 33: would a connect_local variant be needed in some cases?

gio/src/cancellable.rs
```diff
@@ -1,10 +1,153 @@
...
+    #[doc(alias = "g_cancellable_disconnect")]
+    fn disconnect(&self, id: CancelledHandlerId);
+    // rustdoc-stripper-ignore-next
+    /// Returns a `Future` that completes when the cancellable becomes cancelled. Completes
+    /// immediately if the cancellable is already cancelled.
+    fn future(&self) -> std::pin::Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
+    #[doc(alias = "g_cancellable_source_new")]
+    fn source(&self) -> Source;
```
bilelmoussaoui, line 41: should be marked as manual on the gir file as well as rename = "source" otherwise the docs won't be embedded properly

gio/src/cancellable.rs
jf2048: I don't think it's possible to do that, it seems we are bound by the same constraints as `connect_cancelled` since `Cancellable` is `Send + Sync`

sdroege: You could add a `_local()` variant but that would then have to panic if called from a different thread. I don't think that would be very useful.

sdroege, line 33: Maybe call it `connect_cancelled` / `disconnect_cancelled`? It's a better version of the existing `cancelled` signal connector

gio/Gir.toml
```diff
@@ -327,14 +327,23 @@ generate_builder = true
 name = "Gio.Cancellable"
 status = "generate"
 concurrency = "send+sync"
+manual_traits = ["CancellableExtManual"]
+    [[object.function]]
+    name = "connect"
+    # Needs manual bindings for SignalHandlerId
+    manual = true
```
sdroege, line 334: `rename = "connect_cancelled"` etc

And also mark the `cancelled` signal ignored

gio/src/cancellable.rs
```diff
@@ -1,10 +1,153 @@
...
-fn check_callback() {
-    let c = crate::Cancellable::new();
-    c.connect_cancelled(|_| {});
-    c.cancel(); // if it doesn't crash at this point, then we're good to go!
+// rustdoc-stripper-ignore-next
+/// The id of a cancelled handler that is returned by `CancellableExtManual::connect`. This type is
+/// analogous to `glib::SignalHandlerId`.
+#[derive(Debug, Eq, PartialEq)]
+#[repr(transparent)]
+pub struct CancelledHandlerId(libc::c_ulong);
```
sdroege, line 14: I'd make this a `NonZeroU64`. Basically, mirror how `SignalHandlerId` works

gio/src/cancellable.rs
```diff
@@ -1,10 +1,153 @@
...
+    /// Returns the internal signal handler ID.
+    pub unsafe fn as_raw(&self) -> libc::c_ulong {
+        self.0
+    }
+}
+
+impl FromGlib<libc::c_ulong> for CancelledHandlerId {
+    #[inline]
+    unsafe fn from_glib(val: libc::c_ulong) -> Self {
+        Self(val)
```
sdroege, line 27: `assert_ne!(val, 0)`

gio/src/cancellable.rs
```diff
@@ -1,10 +1,153 @@
...
+    }
+    #[doc(alias = "g_cancellable_disconnect")]
+    fn disconnect(&self, id: CancelledHandlerId) {
+        unsafe { ffi::g_cancellable_disconnect(self.as_ptr() as *mut _, id.as_raw()) };
+    }
+    fn future(&self) -> std::pin::Pin<Box<dyn Future<Output = ()> + Send + 'static>> {
+        let cancellable = self.as_ref().clone();
+        let (tx, rx) = oneshot::channel();
+        let id = CancellableExtManual::connect(&cancellable, move |_| {
+            tx.send(()).ok();
```
sdroege, line 87:
```suggestion
            let _ = tx.send(());
```

gio/src/cancellable.rs
```diff
@@ -1,10 +1,153 @@
...
+            );
+
+            from_glib(id)
+        }
+    }
+    #[doc(alias = "g_cancellable_disconnect")]
+    fn disconnect(&self, id: CancelledHandlerId) {
+        unsafe { ffi::g_cancellable_disconnect(self.as_ptr() as *mut _, id.as_raw()) };
+    }
+    fn future(&self) -> std::pin::Pin<Box<dyn Future<Output = ()> + Send + 'static>> {
```
sdroege, line 83:
```suggestion
    fn future(&self) -> std::pin::Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>> {
```

Why is `Sync` not required?

gio/src/cancellable.rs
sdroege: This actually only has to be `Send`. `Sync` is not needed because it's guaranteed to be called only exactly once, and thus never from multiple threads concurrently.

sdroege (commented):
Looks good to me otherwise

gio/src/cancellable.rs
jf2048: > You could add a `_local()` variant but that would then have to panic if called from a different thread. I don't think that would be very useful.

It would probably have to take a `MainContext` as an argument and would just be a wrapper around `MainContext::invoke_local` anyway. IMO, if it's wanted to do that, it would be easier for the user to just use `spawn_local` and await the cancellable's future.

sdroege: I agree, yes

gio/src/cancellable.rs
jf2048: I used to have it as that but it broke the tests, zero is actually a valid return value here: https://docs.gtk.org/gio/method.Cancellable.connect.html#return-value

gio/Gir.toml
```diff
@@ -327,13 +327,26 @@ generate_builder = true
 name = "Gio.Cancellable"
 status = "generate"
 concurrency = "send+sync"
+manual_traits = ["CancellableExtManual"]
+    [[object.function]]
+    name = "connect"
+    # Needs manual bindings for closures
+    ignore = true
```
sdroege, line 334: This should get `manual = true` and `rename = `connect_cancelled`

gio/Gir.toml
```diff
@@ -327,13 +327,26 @@ generate_builder = true
...
+    [[object.function]]
+    name = "disconnect"
+    # Needs manual bindings for CancelledHandlerId
+    ignore = true
```
sdroege, line 338: Same

gio/src/cancellable.rs
```diff
@@ -1,10 +1,238 @@
...
+/// The id of a cancelled handler that is returned by `CancellableExtManual::connect`. This type is
+/// analogous to [`glib::SignalHandlerId`].
+#[derive(Debug, Eq, PartialEq)]
+#[repr(transparent)]
+pub struct CancelledHandlerId(NonZeroU64);
+
+impl CancelledHandlerId {
+    // rustdoc-stripper-ignore-next
+    /// Returns the internal signal handler ID.
+    pub unsafe fn as_raw(&self) -> libc::c_ulong {
```
sdroege, line 28: That should just be `IntoGlib` or not?

gio/src/cancellable.rs
```diff
@@ -1,10 +1,238 @@
...
+    /// Since GLib 2.40, the lock protecting `self` is not held when `callback` is invoked. This
+    /// lifts a restriction in place for earlier GLib versions which now makes it easier to write
+    /// cleanup code that unconditionally invokes e.g.
+    /// [`CancellableExt::cancel()`][crate::prelude::CancellableExt::cancel()].
+    ///
+    /// # Returns
+    ///
+    /// The id of the signal handler or `None` if `self` has already been cancelled.
+    #[doc(alias = "g_cancellable_connect")]
+    fn connect_cancelled<F: FnOnce(&Self) + Send + Sync + 'static>(
```
sdroege, line 61:
```suggestion
    fn connect_cancelled<F: FnOnce(&Self) + Send + 'static>(
```

gio/src/cancellable.rs
```diff
@@ -1,10 +1,238 @@
...
+    // rustdoc-stripper-ignore-next
+    /// Disconnects a handler from a cancellable instance. Additionally, in the event that a signal
+    /// handler is currently running, this call will block until the handler has finished. Calling
+    /// this function from a callback registered with [`Self::connect_cancelled`] will therefore
+    /// result in a deadlock.
+    ///
+    /// This avoids a race condition where a thread cancels at the same time as the cancellable
+    /// operation is finished and the signal handler is removed.
+    #[doc(alias = "g_cancellable_disconnect")]
+    fn disconnect_cancelled(&self, id: Option<CancelledHandlerId>);
```
sdroege, line 89:
```suggestion
    fn disconnect_cancelled(&self, id: CancelledHandlerId);
```

Why would you ever want to disconnect a `None`?

gio/src/cancellable.rs
```diff
@@ -1,10 +1,238 @@
...
+    #[doc(alias = "g_cancellable_connect")]
+    fn connect_cancelled<F: FnOnce(&Self) + Send + Sync + 'static>(
+        &self,
+        callback: F,
+    ) -> Option<CancelledHandlerId>;
+    // rustdoc-stripper-ignore-next
+    /// Same as [`Self::connect_cancelled`] but takes a non-`Send+Sync` closure. If the cancellable
+    /// is cancelled from a different thread than it was connected to then the cancellation will
+    /// panic.
+    fn connect_cancelled_local<F: FnOnce(&Self) + 'static>(
```
sdroege, line 69: Did you find a use case for this?

jf2048: I just got rid of it, i can be added back if anyone asks for it

gio/src/cancellable.rs
jf2048: `g_cancellable_disconnect` lets you pass 0 so you don't have to write the if statement, but it's probably better in Rust to require the if statement to avoid an unnecessary FFI boundary

gio/src/cancellable.rs
jf2048: Not sure, I just copied it from `SignalHandlerId`

gio/src/cancellable.rs
```diff
@@ -1,10 +1,238 @@
...
 
-#[test]
-fn check_callback() {
-    let c = crate::Cancellable::new();
-    c.connect_cancelled(|_| {});
-    c.cancel(); // if it doesn't crash at this point, then we're good to go!
+impl Cancellable {
+    // rustdoc-stripper-ignore-next
+    /// Creates a source that will never trigger.
+    pub fn pending_source() -> Source {
```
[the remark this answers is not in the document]
jf2048: `fn source()` has no way to pass null as the argument.

gio/Gir.toml
jf2048: The only reason for that is the docs right? I ended up rewriting the docs to get rid of the C-isms, so I think `ignore` is appropriate here so it doesn't end up getting the C docs?

gio/src/cancellable.rs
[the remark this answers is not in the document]
jf2048: I just got rid of the source stuff, it shouldn't be used without a lot of other unsafe code anyway

gio/src/cancellable.rs
sdroege: Makes sense. Let's not add something that doesn't seem very useful :)

gio/src/cancellable.rs
sdroege: I guess let's just keep that then

gio/src/cancellable.rs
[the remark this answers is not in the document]
sdroege: Why would you want to have such a `GSource` that never triggers? Just because you can pass `NULL` doesn't mean that it's necessarily useful :)

gio/Gir.toml
sdroege: Yes, that's OK then :)

sdroege (approved):

PR #591, opened 2022-03-07
glib: Only auto-impl `Send+Sync` for subclasses if the parent type does

sdroege:
Previously a `Send+Sync` subclass of a `gtk::Widget` would automatically
be `Send+Sync` too, which is obviously not allowed.

To avoid this, also add the parent type as `PhantomData` to the wrapper
struct so that this is considered too.

Handle `glib::Object` specially here because while it is not `Send+Sync`
itself, subclasses of it can be `Send+Sync` just fine. The only reason
why `glib::Object` is not is because there are some subclasses that are
not and otherwise this could be circumvented by upcasting.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -598,16 +598,18 @@ impl FromGlibPtrArrayContainerAsVec<*mut GObject, *const *mut GObject> for Objec
 }
 
 #[repr(transparent)]
-pub struct TypedObjectRef<T> {
+pub struct TypedObjectRef<T, P> {
     inner: ObjectRef,
-    phantom: PhantomData<T>,
+    imp: PhantomData<T>,
+    parent: PhantomData<P>,
 }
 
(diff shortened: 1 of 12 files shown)


GuillaumeGomez:
So I understand correctly, now we check that the parent is `Send+Sync`, not the child. Or did I misunderstood it?

sdroege:
> So I understand correctly, now we check that the parent is `Send+Sync`, not **just** the child. Or did I misunderstood it?

Yes, we check that both the impl struct and the parent type are `Send+Sync` instead of just the former. Did I write a too detailed commit message? ;)

GuillaumeGomez:
Sorry, I keep forgetting to read the commit messages... >< Please precise it to read it next time for detailed explanations so I don't ask questions you already answered...

GuillaumeGomez (approved):

PR #595, opened 2022-03-09
Structured Logging

jf2048:
Adds bindings for these functions:

- `glib::log_set_writer_func`
- `glib::log_structured_array`
  - Requires a new `GStr` wrapper for a borrowed `GString`, so we can avoid copying the log keys.
- `glib::log_variant`
- `glib::log_writer_default`
- `glib::log_writer_format_fields`
- `glib::log_writer_journald`
- `glib::log_writer_standard_streams`
- `glib::log_writer_is_journald` - unix only
- `glib::log_writer_supports_color` - unix only
- `glib::log_writer_default_set_use_stderr`
- `glib::log_writer_default_would_drop`

And some new macros:
- `glib::gstr` - Turns a string literal into a `&'static GStr`, by appending a nul byte at compile time
- `glib::log_structured` - The main interface. From the docs:

```rust
log_structured!(
    "test",
    LogLevel::Debug,
    {
        // a normal string field
        "MY_FIELD" => "123";
        // fields can also take format arguments
        "MY_FIELD2" => "abc {}", "def";
        // single argument can be a &str or a &[u8] or anything else satsfying AsRef<[u8]>
        "MY_FIELD3" => CString::new("my string").unwrap().to_bytes();
        // field names can also be dynamic, taking an AsRef<GStr>
        GString::from("MY_FIELD4") => b"a binary string".to_owned();
        // the main log message goes in the MESSAGE field
        "MESSAGE" => "test: {} {}", 1, 2, ;
    }
);
```

### DIFF
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -14,6 +14,337 @@ use std::ptr;
 use std::slice;
 use std::string::String;
 
+// rustdoc-stripper-ignore-next
+/// Representaion of a borrowed [`GString`].
+///
+/// This type is very similar to [`std::ffi::CStr`], but with one added constraint: the string
+/// must also be valid UTF-8.
+#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub struct GStr(str);
+
+impl GStr {
+    // rustdoc-stripper-ignore-next
+    /// Creates a GLib string wrapper from a string slice. The string slice must be terminated with
+    /// a nul byte.
+    ///
+    /// This function will cast the provided bytes to a `GStr` wrapper after ensuring
+    /// that the string slice is nul-terminated and does not contain any interior nul bytes.
+    pub fn from_str_with_nul(s: &str) -> Result<&Self, std::ffi::FromBytesWithNulError> {
+        let bytes = s.as_bytes();
+        CStr::from_bytes_with_nul(bytes)?;
+        Ok(unsafe { Self::from_bytes_with_nul_unchecked(bytes) })
+    }
+    // rustdoc-stripper-ignore-next
+    /// Unsafely creates a GLib string wrapper from a byte slice.
+    ///
+    /// This function will cast the provided `bytes` to a `GStr` wrapper without performing any
+    /// sanity checks. The provided slice **must** be valid UTF-8, nul-terminated and not contain
+    /// any interior nul bytes.
+    #[inline]
+    pub const unsafe fn from_bytes_with_nul_unchecked(bytes: &[u8]) -> &Self {
+        debug_assert!(!bytes.is_empty() && bytes[bytes.len() - 1] == 0);
+        mem::transmute(bytes)
+    }
+    // rustdoc-stripper-ignore-next
+    /// Wraps a raw C string with a safe GLib string wrapper. The provided C string **must** be
+    /// valid UTF-8 and nul-terminated. All constraints from [`std::ffi::CStr::from_ptr`] also
+    /// apply here.
+    #[inline]
+    pub unsafe fn from_ptr<'a>(ptr: *const c_char) -> &'a Self {
+        let cstr = CStr::from_ptr(ptr);
+        Self::from_bytes_with_nul_unchecked(cstr.to_bytes_with_nul())
+    }
+    // rustdoc-stripper-ignore-next
+    /// Converts this GLib string to a byte slice containing the trailing 0 byte.
+    ///
+    /// This function is the equivalent of [`GStr::to_bytes`] except that it will retain the
+    /// trailing nul terminator instead of chopping it off.
+    #[inline]
+    pub fn to_bytes_with_nul(&self) -> &[u8] {
+        self.0.as_bytes()
+    }
+    // rustdoc-stripper-ignore-next
+    /// Converts this GLib string to a byte slice.
+    ///
+    /// The returned slice will **not** contain the trailing nul terminator that this GLib
+    /// string has.
+    #[inline]
+    pub fn to_bytes(&self) -> &[u8] {
+        self.as_str().as_bytes()
+    }
+    // rustdoc-stripper-ignore-next
+    /// Returns the inner pointer to this GLib string.
+    ///
+    /// The returned pointer will be valid for as long as `self` is, and points to a contiguous
+    /// region of memory terminated with a 0 byte to represent the end of the string.
+    ///
+    /// **WARNING**
+    ///
+    /// The returned pointer is read-only; writing to it (including passing it to C code that
+    /// writes to it) causes undefined behavior. It is your responsibility to make
+    /// sure that the underlying memory is not freed too early.
+    #[inline]
+    pub fn as_ptr(&self) -> *const c_char {
+        self.0.as_ptr() as *const _
+    }
+    // rustdoc-stripper-ignore-next
+    /// Converts this GLib string to a string slice.
+    #[inline]
+    pub fn as_str(&self) -> &str {
+        // Clip off the nul byte
+        &self.0[0..self.0.len() - 1]
+    }
+    // rustdoc-stripper-ignore-next
+    /// Converts this GLib string to a C string slice.
+    #[inline]
+    pub fn as_c_str(&self) -> &CStr {
+        unsafe { CStr::from_bytes_with_nul_unchecked(self.to_bytes_with_nul()) }
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// Converts a static string literal into a static nul-terminated string.
+///
+/// The expanded expression has type [`&'static GStr`]. This macro will panic if the
+/// string literal contains any interior nul bytes.
+///
+/// # Examples
+///
+/// ```
+/// # fn main() {
+/// use glib::{gstr, GStr, GString};
+///
+/// const MY_STRING: &GStr = gstr!("Hello");
+/// assert_eq!(MY_STRING.to_bytes_with_nul()[5], 0u8);
+/// let owned: GString = MY_STRING.to_owned();
+/// assert_eq!(MY_STRING, owned);
+/// # }
+/// ```
+///
+/// [`&'static GStr`]: crate::GStr
+#[macro_export]
+macro_rules! gstr {
+    ($s:literal) => {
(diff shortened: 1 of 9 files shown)


glib/src/gstring.rs
```diff
@@ -14,6 +14,326 @@ use std::ptr;
...
 use std::string::String;
 
+// rustdoc-stripper-ignore-next
+/// Representaion of a borrowed [`GString`].
+///
+/// This type is very similar to [`std::ffi::CStr`], but with one added constraint: the string
+/// must also be valid UTF-8.
+#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub struct GStr([u8]);
```
sdroege, line 24: Ah nice, I have a commit that adds such a thing locally too :) Why not

```suggestion
pub struct GStr(str);
```

though?

glib/src/gstring.rs
```diff
@@ -14,6 +14,326 @@ use std::ptr;
...
+/// let owned: GString = MY_STRING.to_owned();
+/// assert_eq!(MY_STRING, owned);
+/// # }
+/// ```
+///
+/// [`&'static GStr`]: crate::GStr
+#[macro_export]
+macro_rules! gstr {
+    ($s:literal) => {
+        unsafe { $crate::GStr::from_bytes_with_nul_unchecked($crate::cstr_bytes!($s)) }
```
sdroege, line 128: Why so complicated?

```suggestion
        unsafe { $crate::GStr::from_bytes_with_nul_unchecked(concat!($s, "\0").as_bytes()) }
```

should also work or not?

glib/src/gstring.rs
```diff
@@ -14,6 +14,326 @@ use std::ptr;
...
+impl Default for &GStr {
+    fn default() -> Self {
+        const SLICE: &[c_char] = &[0];
+        unsafe { GStr::from_ptr(SLICE.as_ptr()) }
+    }
+}
+
+impl<'a> From<&'a CStr> for &'a GStr {
+    fn from(s: &'a CStr) -> Self {
+        s.to_str().unwrap();
```
sdroege, line 141: You could directly transmute the return value here btw, but also this should probably rather be a `TryFrom`?

glib/src/gstring.rs
```diff
@@ -14,6 +14,326 @@ use std::ptr;
...
+    }
+}
+
+impl PartialEq<GStr> for str {
+    fn eq(&self, other: &GStr) -> bool {
+        self == other.as_str()
+    }
+}
+
+impl PartialOrd<GStr> for String {
```
sdroege, line 182: Also for `GString` (and the others)

glib/src/gstring.rs
```diff
@@ -14,6 +14,326 @@ use std::ptr;
...
+    }
+}
+
+impl AsRef<[u8]> for GStr {
+    fn as_ref(&self) -> &[u8] {
+        self.to_bytes()
+    }
+}
+
+impl ToOwned for GStr {
```
sdroege, line 236: You probably want to implement `Borrow` too then, and `AsRef<CStr> for GString`

glib/src/gstring.rs
```diff
@@ -14,6 +14,326 @@ use std::ptr;
...
+    }
+}
+
+impl AsRef<GStr> for GStr {
+    fn as_ref(&self) -> &GStr {
+        self
+    }
+}
+
+impl AsRef<str> for GStr {
```
sdroege, line 212: To make it nicer to use you could also implement `Deref<Target=str> for GStr`

glib/src/gstring.rs
```diff
@@ -200,18 +575,36 @@ impl PartialOrd<str> for GString {
 
 impl Eq for GString {}
 
+impl AsRef<GStr> for GString {
+    fn as_ref(&self) -> &GStr {
+        self.as_gstr()
+    }
+}
+
 impl AsRef<str> for GString {
```
sdroege, line 584: Maybe `GString` could `Deref` to `&GStr` and `&GStr` to `&str`?

glib/src/log.rs
```diff
@@ -328,27 +336,131 @@ pub fn log_default_handler(log_domain: Option<&str>, log_level: LogLevel, messag
...
+                length: value.len().try_into().unwrap(),
+            },
+            Default::default(),
+        )
+    }
+    // rustdoc-stripper-ignore-next
+    /// Creates a field from a raw pointer. Can be used to pass custom data into a writer function
+    /// set with [`log_set_writer_func`], where it can be retreived using [`value_ptr`]. Panics if
+    /// `key` contains null bytes.
+    pub fn new_ptr<T>(key: &'a GStr, ptr: *const T) -> Self {
```
sdroege, line 372: This might have to be `unsafe`?

glib/src/log.rs
```diff
@@ -328,27 +336,131 @@ pub fn log_default_handler(log_domain: Option<&str>, log_level: LogLevel, messag
...
+    }
+    // rustdoc-stripper-ignore-next
+    /// Retrieves the raw pointer of the field value. This should only be used on fields created
+    /// with [`new_ptr`] inside writer functions set with [`log_set_writer_func`].
+    pub fn value_ptr<T>(&self) -> *const T {
+        self.0.value as *const T
+    }
+    // rustdoc-stripper-ignore-next
+    /// Retrieves a byte array of the field value. Behavior is undefined if called on a field
+    /// created with [`new_ptr`].
```
sdroege, line 397: ... because of this

glib/src/log.rs
```diff
@@ -328,27 +336,131 @@ pub fn log_default_handler(log_domain: Option<&str>, log_level: LogLevel, messag
...
+        panic!("Writer func can only be set once");
+    }
+    unsafe extern "C" fn writer_trampoline(
+        log_level: ffi::GLogLevelFlags,
+        fields: *const ffi::GLogField,
+        n_fields: libc::size_t,
+        _user_data: ffi::gpointer,
+    ) -> ffi::GLogWriterOutput {
+        let writer_func = WRITER_FUNC.get().unwrap();
+        let fields = std::slice::from_raw_parts(fields as *const LogField<'_>, n_fields);
```
sdroege, line 438: If a field has an arbitrary pointer then usage of it from inside the log function could lead to UB within safe code. That doesn't seem ideal

sdroege (commented):
Looks nice otherwise

glib/src/log.rs
```diff
@@ -328,27 +336,131 @@ pub fn log_default_handler(log_domain: Option<&str>, log_level: LogLevel, messag
...
+    // rustdoc-stripper-ignore-next
+    /// Retrieves the raw pointer of the field value. This should only be used on fields created
+    /// with [`new_ptr`] inside writer functions set with [`log_set_writer_func`].
+    pub fn value_ptr<T>(&self) -> *const T {
+        self.0.value as *const T
+    }
+    // rustdoc-stripper-ignore-next
+    /// Retrieves a byte array of the field value. Behavior is undefined if called on a field
+    /// created with [`new_ptr`].
+    pub fn value_bytes(&self) -> &[u8] {
```
mbiggio, line 398: Maybe instead of having UB in case the object is constructed with `new_ptr`, we could check for `length == 0` and in that case assume that `self.value` is an arbitrary pointer.
The function should have to return `Option<&[u8]>` and return `None` in case of 0-lentgth.

This seems kind of coherent to what the structured logging doc says:
> If a program wants to pass context data from a g_log_structured() call to its log writer function so that, for example, it can use the correct server connection to submit logs to, that user data can be passed as a zero-length GLogField to g_log_structured_array().

glib/src/log.rs
```diff
@@ -328,27 +336,131 @@ pub fn log_default_handler(log_domain: Option<&str>, log_level: LogLevel, messag
...
+    /// Retreives the field key.
+    pub fn key(&self) -> &str {
+        unsafe { std::ffi::CStr::from_ptr(self.0.key as *const i8) }
+            .to_str()
+            .unwrap()
+    }
+    // rustdoc-stripper-ignore-next
+    /// Retrieves the raw pointer of the field value. This should only be used on fields created
+    /// with [`new_ptr`] inside writer functions set with [`log_set_writer_func`].
+    pub fn value_ptr<T>(&self) -> *const T {
```
mbiggio, line 392: Same here: we could return `Option<*const T>`, returning `None` of `lenght != 0`

glib/src/gstring.rs
jf2048: Sure

glib/src/gstring.rs
jf2048: We need to check at compile time if the string literal has any interior nul bytes, because otherwise it breaks the invariant on `GString` where it has to be a valid `CString` internally. The only way to do that is a proc macro. The rest I put in a `macro_rules` to get the hygiene with `$crate`.

glib/src/log.rs
jf2048: I'll do this and also change it to a `usize`. If someone wants to unsafely transmute it back to a pointer that will still work

glib/src/gstring.rs
jf2048: Sure. Only caveat, it would break anything that calls `deref` on `GString` manually

glib/src/gstring.rs
sdroege: Ah I see

glib/src/gstring.rs
jf2048: Yes, a lot of these for `GString` should also be `TryFrom`. Should I change those too?

jf2048: Let's do that in a follow up, it seems a bit disruptive. For now I just changed this one

glib/src/log.rs
jf2048: Removed and replaced with just a `usize`

glib/src/log.rs
jf2048: Changed to an `Option<&[u8]>`. Also made sure that an empty `&[u8]` Doesn't get mistaken for a user data field.

glib/src/gstring.rs
jf2048: I made a PR draft, can make an issue too if you want: https://github.com/gtk-rs/gtk-rs-core/pull/600

sdroege: Draft PR is even better :)

sdroege (approved):

Issue #594, opened 2022-03-09
gobject-sys: license file missing from published crates


decathorpe:
Looks like https://github.com/gtk-rs/gtk-rs-core/issues/172 was not fixed entirely.
The published gobject-sys crates still do not contain a license file as of 0.15.7.

sdroege:
This only affects gobject-sys, correct? That one is missing the symlink to the license file, the others seem to have it unless I missed one.

decathorpe:
Yes, I think gobject-sys was the only one that was missed. Thanks!

sdroege:
0.15.9 has it now

PR #598, opened 2022-03-10
glib: Add license file to gobject-sys

sdroege:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/594


GuillaumeGomez (approved):

sdroege:
Backported directly into 0.15 and doing a new release with that now

Issue #10, opened 2021-03-27
Panic if GString contains null byte


sophie-h:
```rust
glib::GString::from("test\0");
```

> thread 'main' panicked at 'CString::new failed: NulError(4, [116, 101, 115, 116, 0])'

Not surprising, but can we do something about it? I was thinking about dropping null bytes in the `from()` function. But maybe I'm just to obsessed with having code that cannot panic :laughing:

JAicewizard:
Could it be an option to implement `ToGlibPtr` for `CString` so that the users of glib have more control over the panic without doing double work??

Then all functions that use this trait could also have a version that takes a `CString` which would allow all usesr to handle the errors gracefully. Or instead of taking a CString directly, make the functions generic over `ToGlibPtr`.

PR #600, opened 2022-03-10
GString refactor

jf2048:
Update: See comment https://github.com/gtk-rs/gtk-rs-core/pull/600#issuecomment-1302872708 for the general status of this PR. 

Original comment: All of these `TryFrom` are panicking when they don't really need to. This is probably going to break some other things, I'd like to check this against gtk-rs and gstreamer-rs at least. Another option is to leave the panicking `From` traits in and also support `TryFrom`.

Also makes `GString` deref as a `GStr`

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.63.0"
+          - "1.64.0"
         conf:
           - { name: "cairo", features: "png,pdf,svg,ps,use_glib,v1_18,freetype,script,xcb,xlib,win32-surface", nightly: "--features 'png,pdf,svg,ps,use_glib,v1_18,freetype,script,xcb,xlib,win32-surface'", test_sys: true }
           - { name: "gdk-pixbuf", features: "v2_42", nightly: "--all-features", test_sys: true }
@@ -100,7 +100,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.63.0"
+          - "1.64.0"
     steps:
       - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -186,13 +186,13 @@ impl PartialEq<str> for GStr {
     }
 }
 
-impl<'a> PartialEq<&'a str> for GStr {
-    fn eq(&self, other: &&'a str) -> bool {
+impl PartialEq<&str> for GStr {
+    fn eq(&self, other: &&str) -> bool {
         self.as_str() == *other
     }
 }
 
-impl<'a> PartialEq<GStr> for &'a str {
+impl PartialEq<GStr> for &str {
     fn eq(&self, other: &GStr) -> bool {
         *self == other.as_str()
     }
@@ -485,18 +485,21 @@ impl IntoGlibPtr<*mut c_char> for GString {
 }
 
 impl Clone for GString {
+    #[inline]
     fn clone(&self) -> GString {
         self.as_str().into()
     }
 }
 
 impl fmt::Debug for GString {
+    #[inline]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         <&str as fmt::Debug>::fmt(&self.as_str(), f)
     }
 }
 
 impl Drop for GString {
+    #[inline]
     fn drop(&mut self) {
         if let Inner::Foreign { ptr, .. } = self.0 {
             unsafe {
@@ -507,138 +510,161 @@ impl Drop for GString {
 }
 
 impl fmt::Display for GString {
+    #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str(self.as_str())
     }
 }
 
 impl hash::Hash for GString {
+    #[inline]
     fn hash<H: hash::Hasher>(&self, state: &mut H) {
         self.as_str().hash(state)
     }
 }
 
 impl Borrow<GStr> for GString {
+    #[inline]
     fn borrow(&self) -> &GStr {
         self.as_gstr()
     }
 }
 
 impl Borrow<str> for GString {
+    #[inline]
     fn borrow(&self) -> &str {
         self.as_str()
     }
 }
 
 impl Ord for GString {
+    #[inline]
     fn cmp(&self, other: &GString) -> Ordering {
         self.as_str().cmp(other.as_str())
     }
 }
 
 impl PartialOrd for GString {
+    #[inline]
     fn partial_cmp(&self, other: &GString) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
 
 impl PartialEq for GString {
+    #[inline]
     fn eq(&self, other: &GString) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialEq<GString> for String {
+    #[inline]
     fn eq(&self, other: &GString) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialEq<GStr> for GString {
+    #[inline]
     fn eq(&self, other: &GStr) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialEq<&GStr> for GString {
+    #[inline]
     fn eq(&self, other: &&GStr) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialEq<str> for GString {
+    #[inline]
     fn eq(&self, other: &str) -> bool {
         self.as_str() == other
     }
 }
 
 impl PartialEq<&str> for GString {
+    #[inline]
     fn eq(&self, other: &&str) -> bool {
         self.as_str() == *other
     }
 }
 
 impl PartialEq<GString> for &GStr {
+    #[inline]
     fn eq(&self, other: &GString) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialEq<GString> for &str {
+    #[inline]
     fn eq(&self, other: &GString) -> bool {
         *self == other.as_str()
     }
 }
 
 impl PartialEq<String> for GString {
+    #[inline]
     fn eq(&self, other: &String) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialEq<GString> for str {
+    #[inline]
     fn eq(&self, other: &GString) -> bool {
         self == other.as_str()
     }
 }
 
 impl PartialEq<GString> for GStr {
+    #[inline]
     fn eq(&self, other: &GString) -> bool {
         self.as_str() == other.as_str()
     }
 }
 
 impl PartialOrd<GString> for String {
+    #[inline]
     fn partial_cmp(&self, other: &GString) -> Option<Ordering> {
         Some(self.cmp(&String::from(other.as_str())))
     }
 }
 
 impl PartialOrd<String> for GString {
+    #[inline]
     fn partial_cmp(&self, other: &String) -> Option<Ordering> {
         Some(self.as_str().cmp(other.as_str()))
     }
 }
 
 impl PartialOrd<GString> for GStr {
+    #[inline]
     fn partial_cmp(&self, other: &GString) -> Option<Ordering> {
         Some(self.as_str().cmp(other))
     }
 }
 
 impl PartialOrd<GStr> for GString {
+    #[inline]
     fn partial_cmp(&self, other: &GStr) -> Option<Ordering> {
         Some(self.as_str().cmp(other.as_str()))
     }
 }
 
 impl PartialOrd<GString> for str {
+    #[inline]
     fn partial_cmp(&self, other: &GString) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
 
 impl PartialOrd<str> for GString {
+    #[inline]
     fn partial_cmp(&self, other: &str) -> Option<Ordering> {
         Some(self.as_str().cmp(other))
     }
@@ -647,12 +673,14 @@ impl PartialOrd<str> for GString {
 impl Eq for GString {}
 
 impl AsRef<GStr> for GString {
+    #[inline]
     fn as_ref(&self) -> &GStr {
         self.as_gstr()
     }
 }
 
 impl AsRef<str> for GString {
+    #[inline]
     fn as_ref(&self) -> &str {
         self.as_str()
     }
@@ -665,18 +693,21 @@ impl AsRef<CStr> for GString {
 }
 
 impl AsRef<OsStr> for GString {
+    #[inline]
     fn as_ref(&self) -> &OsStr {
         OsStr::new(self.as_str())
     }
 }
 
 impl AsRef<Path> for GString {
+    #[inline]
     fn as_ref(&self) -> &Path {
         Path::new(self.as_str())
     }
 }
 
 impl AsRef<[u8]> for GString {
+    #[inline]
     fn as_ref(&self) -> &[u8] {
         self.as_str().as_bytes()
     }
@@ -685,6 +716,7 @@ impl AsRef<[u8]> for GString {
 impl Deref for GString {
     type Target = str;
 
+    #[inline]
     fn deref(&self) -> &str {
         self.as_str()
     }
@@ -1070,6 +1102,7 @@ impl GlibPtrDefault for GString {
 }
 
 impl StaticType for GString {
+    #[inline]
     fn static_type() -> Type {
         String::static_type()
     }
@@ -1084,22 +1117,26 @@ impl crate::value::ValueTypeOptional for GString {}
 unsafe impl<'a> crate::value::FromValue<'a> for GString {
     type Checker = crate::value::GenericValueTypeOrNoneChecker<Self>;
 
+    #[inline]
     unsafe fn from_value(value: &'a Value) -> Self {
         Self::from(<&str>::from_value(value))
     }
 }
 
 impl crate::value::ToValue for GString {
+    #[inline]
     fn to_value(&self) -> Value {
         <&str>::to_value(&self.as_str())
     }
 
+    #[inline]
     fn value_type(&self) -> Type {
         String::static_type()
     }
 }
 
 impl crate::value::ToValueOptional for GString {
+    #[inline]
     fn to_value_optional(s: Option<&Self>) -> Value {
         <str>::to_value_optional(s.as_ref().map(|s| s.as_str()))
     }
@@ -1116,6 +1153,7 @@ impl From<GString> for Value {
 }
 
 impl StaticType for Vec<GString> {
+    #[inline]
     fn static_type() -> Type {
         <Vec<String>>::static_type()
     }
@@ -1144,6 +1182,7 @@ impl ToValue for Vec<GString> {
         }
     }
 
+    #[inline]
     fn value_type(&self) -> Type {
         <Vec<GString>>::static_type()
     }
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -83,7 +83,7 @@ impl GStr {
     /// This function is the equivalent of [`GStr::to_bytes`] except that it will retain the
     /// trailing nul terminator instead of chopping it off.
     #[inline]
-    pub fn to_bytes_with_nul(&self) -> &[u8] {
+    pub fn as_bytes_with_nul(&self) -> &[u8] {
         self.0.as_bytes()
     }
     // rustdoc-stripper-ignore-next
@@ -92,7 +92,7 @@ impl GStr {
     /// The returned slice will **not** contain the trailing nul terminator that this GLib
     /// string has.
     #[inline]
-    pub fn to_bytes(&self) -> &[u8] {
+    pub fn as_bytes(&self) -> &[u8] {
         self.as_str().as_bytes()
     }
     // rustdoc-stripper-ignore-next
@@ -121,7 +121,7 @@ impl GStr {
     /// Converts this GLib string to a C string slice.
     #[inline]
     pub fn as_c_str(&self) -> &CStr {
-        unsafe { CStr::from_bytes_with_nul_unchecked(self.to_bytes_with_nul()) }
+        unsafe { CStr::from_bytes_with_nul_unchecked(self.as_bytes_with_nul()) }
     }
 
     #[doc(alias = "g_utf8_collate")]
@@ -144,7 +144,7 @@ impl GStr {
 /// use glib::{gstr, GStr, GString};
 ///
 /// const MY_STRING: &GStr = gstr!("Hello");
-/// assert_eq!(MY_STRING.to_bytes_with_nul()[5], 0u8);
+/// assert_eq!(MY_STRING.as_bytes_with_nul()[5], 0u8);
 /// let owned: GString = MY_STRING.to_owned();
 /// assert_eq!(MY_STRING, owned);
 /// # }
@@ -273,7 +273,7 @@ impl AsRef<Path> for GStr {
 
 impl AsRef<[u8]> for GStr {
     fn as_ref(&self) -> &[u8] {
-        self.to_bytes()
+        self.as_bytes()
     }
 }
 
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -273,6 +273,7 @@ impl GStr {
             Ok(())
         }
     }
+    pub const NONE: Option<&'static GStr> = None;
 }
 
 // rustdoc-stripper-ignore-next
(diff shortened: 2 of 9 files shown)


sdroege:
> Also needs to be updated with `GStr` bits if #595 is merged

It's merged :)

jf2048:
More PRs to address breakage caused by this PR:

- https://github.com/gtk-rs/gtk3-rs/pull/739
- https://github.com/gtk-rs/gir/pull/1327
- https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/989

Probably they should be merged all at once

sdroege:
Next step: change all functions that take `&str` to instead take `impl AsRef<GStr>` or so to avoid all the unnecessary memory copies :)

glib/src/gstring.rs
```diff
@@ -617,10 +617,10 @@ impl AsRef<[u8]> for GString {
 }
 
 impl Deref for GString {
-    type Target = str;
+    type Target = GStr;
```
sdroege, line 620: I guess we should think about the advantages / disadvantages of this. What does a `&GStr` bring in comparison to just having a `&str`?

glib/src/gstring.rs
```diff
@@ -678,50 +678,48 @@ impl From<&GStr> for GString {
...
-            ptr::copy_nonoverlapping(s.as_ptr() as *const c_char, copy, s.len() + 1);
-            ptr::write(copy.add(s.len()), 0);
-
-            GString(Inner::Foreign {
-                ptr: ptr::NonNull::new_unchecked(copy),
-                len: s.len(),
-            })
-        }
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        s.to_owned().try_into()
```
sdroege, line 685: This now creates a `CString`-based `GString`, so doesn't use the GLib allocator anymore (also in some other places). The whole point here was that if we have to copy anyway then we'd use the GLib allocator.

The idea was that `GString` will contain a string allocated with the GLib allocator whenever a copy has to happen and only otherwise contain something based on the `std` allocator. If that makes sense, I don't know and this needs some more thinking.

The goal would be to reduce the number of copies as much as possible, and e.g. later being able to pass a `GString` to a `transfer full` C function without any copies (and a `&GStr` to a `transfer none` C function).

glib/src/gstring.rs
GuillaumeGomez: Do we have functions taking `GStr` only?

sdroege: Not yet, see my other comments

GuillaumeGomez: Maybe better to use `str` for the time being until we make the switch then.

jf2048: I'll take this out, but https://github.com/gtk-rs/gtk3-rs/pull/739 and https://github.com/gtk-rs/gir/pull/1327 can still be merged without this if you want. They don't break anything.

sdroege: Sounds like a good idea regardless, yes. More future-proof.

glib/src/gstring.rs
jf2048: That makes sense, I've changed it back and updated it to check for interior nuls

jf2048:
Now added `impl Default for GString`

jf2048:
> Next step: change all functions that take `&str` to instead take `impl AsRef<GStr>` or so to avoid all the unnecessary memory copies :)

That was my thought too. We can test it out later with some of the manually implemented functions, and then afterwards work it into gir.

bilelmoussaoui:
While you are updating GString traits, maybe we should have Default impls here. See the gstreamer-rs PR for context

sdroege (commented):
Seems OK to me apart from that final comment

sdroege:
> While you are updating GString traits, maybe we should have Default impls here. See the gstreamer-rs PR for context

It's here already

glib/src/gstring.rs
```diff
@@ -660,23 +660,23 @@ impl From<GString> for Box<str> {
     }
 }
 
-impl From<String> for GString {
+impl TryFrom<String> for GString {
```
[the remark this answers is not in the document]
sdroege: @bilelmoussaoui What do you think?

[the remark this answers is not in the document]
sdroege: Any opinions? :)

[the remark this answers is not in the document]
bilelmoussaoui: I guess it is fine to also keep a From implementation, but those are not expected to panic, aren't they?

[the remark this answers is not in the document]
sdroege: Yes but then you're also not expecting `label.set_text("bla\0bla")` to panic :)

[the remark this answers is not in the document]
sdroege: @jf2048 Any opinions from your side here?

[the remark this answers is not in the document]
jf2048: Is there any harm in passing through a string with nul bytes? That would be faster if we could just skip the check altogether. Sure the C API will truncate it, but is that a worse behavior than panicking?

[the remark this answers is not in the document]
sdroege: > but is that a worse behavior than panicking?

Depends on what the string is used for, I guess :)

> Otherwise I don't really see a problem with implementing a panicking `From`

Let's go with that then for now?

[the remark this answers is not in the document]
jf2048: Let's wait a bit on this, I want to test out this one first: https://github.com/gtk-rs/gir/pull/1184

sdroege:
What should we do about this one?

jf2048:
Need to get some time to look through this one and try to update it with another trait that can handle `GStr`: https://github.com/gtk-rs/gir/pull/1184

sdroege:
@jf2048 My ideal plan here would be to have something that allows passing in a `&str` (etc), `&GStr`, `String`, `GString` into `transfer none` string parameters (and in case of the already-NUL-terminated types just doesn't many a copy), and `impl Into<GString>` for `transfer full` string parameters.

I assume for the former you need a new trait as none of the existing ones really fits the different cases here. Is that what you meant?

(Also I'd like to do basically the same for arrays and basic containers like `GList`)

jf2048:
Yes but I am going back and forth on this. Now I feel like that trait would actually be bad and it should just be `GStr` everywhere. It seems like it is not idiomatic Rust to implicitly create a bunch of string copies. But that would mean requiring a manual conversion to a `GString` everywhere if you want to use `String`, or encourage people to use `GStringBuilder` wherever possible.

sdroege:
That seems fine. We already return `GString` anyway. Going from there to a `String` if needed is a single `into()`.

The bigger problem is IMHO `&str` for transfer none arguments: it would be annoying to require passing `gstr!("blabla")` everywhere instead of allowing to pass `"blabla"`. Doing implicit copies here is not nice but seems like the lesser problem, and is also what we already do anyway. The difference is that if someone actually passes a `&GStr` or `&GString` it wouldn't require a copy.

For transfer full string arguments requiring a `GString` and allowing to pass `String` (no copy) and `&str` (copy) seems fine to me too.

sdroege:
My idea would be to have the following situation in the end for strings:

 - `transfer none` / `transfer full` return values would always use `GString` as they do now
 - `transfer none` parameters would use a new `IntoGStr` trait, which is implemented for `&GStr` and `&GString` (no copy, or maybe just for `AsRef<GStr>` if we can make that work), `&CStr` and `&CString` (no copy), and `&str` and `&String` (copy)
 - `transfer full` parameters would use `Into<GString>`, which is implemented for `GString` (no copy) and `String` (no copy!), and also `&str` and `&GStr` (both copy)

From a usability point of view this would work the same way as it does right now (and would have implicit copies like it has now, under some conditions), except that you can avoid any copies as a user if you do the right thing. One day Rust will hopefully have zero-terminated string literals (something like `z"blabla"`) and then we can revisit this, but not being able to pass string literals seems rather suboptimal.

----

```rust
pub trait IntoGStr {
    type Stash: AsRef<GStr>;

    fn into_gstr(&self) -> Self::Stash;
}
```

would be approximately my idea for the new trait FWIW.

jf2048:
> it would be annoying to require passing `gstr!("blabla")` everywhere instead of allowing to pass `"blabla"`

I think you should have to use `gstr!`. Any use of `&str` is going to create unnecessary copies...

sdroege:
Sure there will be unnecessary copies but for 99% of the cases it doesn't really matter and requiring the macro is a huge usability burden in comparison.

sdroege:
Yeah some improvement on the gettext situation would be nice too but that seems on top of this here :) @bilelmoussaoui @jf2048 @BrainBlasted Can one of you create an issue about that with the summary of the discussion yesterday?

jf2048:
It seems a `IntoGStr` trait is unnecessary, I was thinking about this and it actually has exactly the same semantics as `Into<Cow<GStr>>`. So we can just implement traits for that.

glib/src/gstring.rs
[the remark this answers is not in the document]
jf2048: My idea for the ideal approach:
- Implement both `From` and `TryFrom` so users can get the fallible constructor, if they want.
- In debug builds, the `From` impl will scan the string during construction and panic on interior nuls. The check will be skipped in production builds. Also do this in `impl ToGlibPtr for str`, etc.
- Since the `CString` invariant will no longer hold in all cases, change `gstring::Inner::Native` to just contain a `String`. Also remove `AsRef<CStr>` and add a `to_cstr` method that does the same thing as [`CStr::from_bytes_until_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul).

jf2048:
Actually never mind, we can't implement `From<String> for Cow<GStr>`

glib/src/gstring.rs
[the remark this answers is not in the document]
sdroege: Sounds good to me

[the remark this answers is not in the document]
sdroege: This would also solve https://github.com/gtk-rs/gtk-rs-core/issues/10

jf2048:
From my tests, a downside of using `IntoGStr` is that callers must fully qualify `None` types:

```rust
fn do_something(a_string: Option<impl IntoGStr>) { ... }
do_something(None); // compile error
do_something(None::<&str>); // ok
```

But otherwise, seems to be usable?

sdroege:
Hrm that problem again... that seems suboptimal :) We had that with the `IsA` trait already and there it causes still a lot of friction

glib/src/gstring.rs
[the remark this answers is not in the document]
jf2048: A couple more notes.

- Since it's immutable, it should be `Native(Option<Box<str>>)` to preserve the same memory layout.
- We can support `Native(None)` as a valid case to optimize an empty string, that way it can skip an allocation when calling `to_glib_none` on an empty `GString`.

jf2048:
I wouldn't mind `glib::NONE_STR`

glib/src/gstring.rs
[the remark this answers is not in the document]
sdroege: `NULL` and `""` are different things though, and passing the former instead of the latter can potentially cause some C code to crash

sdroege:
To reduce allocations for `to_glib_none()` strings, we could also do the same trick as in https://github.com/rust-lang/rust/pull/93668 but this currently conflicts a bit with how our `Stash` works (the stack allocated memory would have to be in the stash, so the string pointer will change its location when returning from `to_glib_none()`).

jf2048:
I think the only way to do that correctly would be to replicate the same approach as that PR, i.e. make a helper function taking a `FnOnce` and wrap the FFI calls in it

sdroege:
Yeah, I guess we could model `to_glib_none()` like this. That also makes the stash unnecessary then.

jf2048:
But even more than that, since `GString` is already an enum and an immutable string, then we could do what `smartstring` does and have a variant that stores small strings inline (up to 22 chars)

jf2048:
Updates:

- `GString` now contains `Box<str>` and got a lot of new constructors. Checks for interior nul now mostly happen only in debug mode. Checking still available in release mode when using a constructor that asks for it.
- Lots of new trait impls added: `Default`, `Display`, and `From` for various other string types.
- Optimizations in `ToGlibPtr` to avoid interior nul checking or avoid allocations in certain cases.
- `gformat!` macro for formatting a `GString` directly.
- `IntoGStr` traits for avoiding copies in function arguments when passing to ffi.
- Bumps required Rust version to 1.64 so we can use const `slice::from_raw_parts`.

glib/src/gstring.rs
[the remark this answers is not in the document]
jf2048: It won't return `NULL`, it will always produce a static `""`.

glib/src/gstring.rs
```diff
@@ -344,37 +540,269 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GStr {
     }
 }
 
+// size_of::<Inner>() minutes two bytes for length and enum discriminant
```
tristan957, line 543: minus instead of minutes

glib/src/gstring.rs
```diff
@@ -537,6 +542,10 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GStr {
     }
 }
 
+// size_of::<Inner>() minus two bytes for length and enum discriminant
+const INLINE_LEN: usize =
+    mem::size_of::<Option<Box<str>>>() + mem::size_of::<usize>() - mem::size_of::<u8>() * 2;
```
sdroege, line 545: Can we get a test for making sure this doesn't unnecessarily increase the size of `GString`?

glib/src/gstring.rs
```diff
@@ -989,12 +1014,15 @@ impl IntoGlibPtr<*mut c_char> for GString {
     /// Transform into a nul-terminated raw C string pointer.
     unsafe fn into_glib_ptr(self) -> *mut c_char {
         match self.0 {
-            Inner::Native(None) => ffi::g_malloc0(1) as *mut _,
+            Inner::Native(None) => unreachable!(),
```
sdroege, line 1015: As this case is impossible now, maybe we can get rid of the `Option` instead?

glib/src/gstring.rs
```diff
@@ -573,6 +573,63 @@ impl GString {
...
+            }
+        }
+
+        impl Drop for GStringInline {
+            fn drop(&mut self) {
+                unsafe { ffi::g_free(self.0.str as *mut _) };
+            }
+        }
+
+        let mut s = GStringInline(ffi::GString {
```
sdroege, line 629: Why not go via `GStringBuilder`?

glib/src/gstring.rs
```diff
@@ -573,6 +573,63 @@ impl GString {
...
+            str: ptr::null_mut(),
+            len: 0,
+            allocated_len: 0,
+        });
+        fmt::Write::write_fmt(&mut s, args).unwrap();
+
+        if s.0.str.is_null() {
+            debug_assert_eq!(s.0.allocated_len, 0);
+            Self(Inner::Native(None))
```
sdroege: Why not `Self::new()`?

Also `GString` should probably get a `Default` impl

glib/src/gstring.rs
sdroege: Ah because you inline-allocate it here. Should maybe do the same for `GStringBuilder` though?

glib/src/translate.rs
```diff
@@ -501,69 +502,116 @@ impl<'a, P: Ptr, T: ?Sized + ToGlibPtr<'a, P>> ToGlibPtr<'a, P> for &'a T {
...
-            CString::new(self).expect("str::ToGlibPtr<*const c_char>: unexpected '\0' character");
-        Stash(tmp.as_ptr(), tmp)
+        let bytes = if self.is_empty() {
+            Cow::Borrowed([0u8].as_slice())
+        } else {
+            if cfg!(debug_assertions) {
+                crate::GStr::check_interior_nuls(self).unwrap();
+            }
+            let mut bytes = self.as_bytes().to_owned();
+            bytes.reserve_exact(1);
```
sdroege, line 552: This potentially requires a re-allocation. Can we avoid that by pre-allocating the bytes with enough capacity?

glib/src/translate.rs
```diff
@@ -501,69 +502,116 @@ impl<'a, P: Ptr, T: ?Sized + ToGlibPtr<'a, P>> ToGlibPtr<'a, P> for &'a T {
...
-    type Storage = CString;
+    type Storage = Cow<'static, [u8]>;
 
-    #[inline]
     fn to_glib_none(&'a self) -> Stash<'a, *const c_char, Self> {
-        let tmp =
-            CString::new(self).expect("str::ToGlibPtr<*const c_char>: unexpected '\0' character");
-        Stash(tmp.as_ptr(), tmp)
+        let bytes = if self.is_empty() {
+            Cow::Borrowed([0u8].as_slice())
```
sdroege, line 546: Is the address of this guaranteed to be stable? The stash is returned here, so copied into the stack of the caller, and that would also move around the slice or not?

.github/workflows/CI.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.63.0"
+          - "1.64.0"
```
sdroege, line 21: Can you add a comment to the commit message why it's needed?

glib/src/gstring.rs
```diff
@@ -226,12 +392,6 @@ impl AsRef<str> for GStr {
     }
 }
 
-impl AsRef<CStr> for GStr {
```
sdroege, line 222: Why do you remove this one?

glib/src/gstring.rs
sdroege: Ah you added a `Default` impl somewhere :) Other question still stands

glib/src/gstring.rs
```diff
@@ -355,11 +530,18 @@ impl<'a> ToGlibPtr<'a, *mut c_char> for GStr {
...
-/// must also be valid UTF-8.
+/// This type is similar to [`std::ffi::CString`], but with some special behavior. When debug
+/// assertions are enabled, <code>[From]&lt;[String]></code> will panic if there are interior
+/// nul-bytes. In production builds, no checks will be made for interior nul-bytes, and strings
+/// that contain interior nul-bytes will simply end at first nul-byte when converting to a C
+/// string.
+///
+/// The constructors beginning with `from_utf8` `and `from_string` can also be used to further
+/// control how interior nul-bytes are handled.
+#[repr(transparent)]
```
sdroege, line 561: Why? This only really makes sense if you need guarantees about the representation, which we don't really here?

glib/src/gstring.rs
```diff
@@ -392,44 +730,187 @@ impl GString {
 
     // rustdoc-stripper-ignore-next
     /// Extracts the [`GStr`] containing the entire string.
+    #[inline]
     pub fn as_gstr(&self) -> &GStr {
         let bytes = match self.0 {
-            Inner::Native(ref cstr) => cstr.as_ref().unwrap().to_bytes_with_nul(),
+            Inner::Native(None) => return <&GStr>::default(),
+            Inner::Native(Some(ref s)) => s.as_bytes(),
             Inner::Foreign { len, .. } if len == 0 => &[0],
```
sdroege, line 823: Could combine this and the first branch I guess? But maybe we can forbid `Native(None)` for simplification and only have one possible representation of an empty string

glib/src/gstring.rs
```diff
@@ -640,12 +1128,6 @@ impl AsRef<str> for GString {
     }
 }
 
-impl AsRef<CStr> for GString {
```
sdroege, line 610: Why?

glib/src/gstring.rs
```diff
@@ -892,63 +892,32 @@ impl FromGlibPtrBorrow<*mut i8> for GString {
     }
 }
 
-#[doc(hidden)]
-impl<'a> ToGlibPtr<'a, *const u8> for GString {
```
sdroege, line 859: Why? Also you then probably want to do the same for the impls for `String`, `&str` etc

glib/src/gstring.rs
```diff
@@ -1757,6 +1758,90 @@ impl crate::value::ToValue for Vec<GString> {
 impl_from_glib_container_as_vec_string!(GString, *const c_char);
 impl_from_glib_container_as_vec_string!(GString, *mut c_char);
 
+// rustdoc-stripper-ignore-next
+/// A trait to accept both <code>&[str]</code> or <code>&[GStr]</code> as an argument.
```
sdroege, line 1797: Why `<code>` instead of just backticks?

glib/src/gstring.rs
```diff
@@ -1757,6 +1758,90 @@ impl crate::value::ToValue for Vec<GString> {
...
+}
+
+impl IntoGStr for &GString {
+    #[inline]
+    fn run_with_gstr<T, F: FnOnce(&GStr) -> T>(self, f: F) -> T {
+        f(self.as_gstr())
+    }
+}
+
+const MAX_STACK_ALLOCATION: usize = 384;
```
sdroege, line 1823: Maybe add a comment where this number comes from :)

sdroege:
Looks generally nice to me. Are you aware of any drawbacks here or does it all seem nice to you?

Also this should probably get a change to `gir` at some point to
  - Use `impl IntoGStr` for transfer none
  - Use `impl Into<GString>` for transfer full
parameters.

glib/src/gstring.rs
jf2048: This looks confusing, these are not arrays. The `[str]` is a link and the `&` is not. So it will look like <code>&[str](https://doc.rust-lang.org/std/primitive.str.html)</code>

glib/src/gstring.rs
jf2048: I think this should at least give us some assurance it remains the same size as `String`?

glib/src/gstring.rs
sdroege: Oh I see, that's why.

glib/src/gstring.rs
sdroege: `String` is also not `#[repr(transparent)]` though :) I think if that's your goal then the only thing that you can really do here would be to add a test that asserts this, and if that ever fails we need to do some thinking.

glib/src/translate.rs
jf2048: Replaced with a static slice

glib/src/gstring.rs
jf2048: I seem to remember we replaced this with `c_char` somewhere else? I can add the `u8` and `i8` back, not sure it makes a difference

glib/src/gstring.rs
jf2048: Added layout test

glib/src/gstring.rs
jf2048: Removed the `Option`, any time we need to take the `Box` it uses `mem::replace(b, "".into())` which always gets optimized to avoid allocation

glib/src/gstring.rs
jf2048: Typo, fixed

glib/src/gstring.rs
jf2048: Replaced with some `to_cstr` functions since `GStr` can now contain interior nuls and `AsRef` implementations should probably not re-scan the string or panic

glib/src/gstring.rs
jf2048: `GStringBuilder` now uses `BoxedInline`

glib/src/gstring.rs
jf2048: Removed the `repr` and added a test

glib/src/gstring.rs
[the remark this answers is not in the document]
jf2048: Strings now using `From` again,.`TryFrom` is only used for conversion that could have invalid UTF-8.

glib/src/utils.rs
```diff
@@ -244,16 +258,19 @@ mod tests {
...
         assert_eq!(unescaped, Some(GString::from("&foo")));
 
         assert_eq!(
-            crate::uri_unescape_segment(Some("/foo"), None, None),
+            crate::uri_unescape_segment(Some("/foo"), crate::NONE_STR, crate::NONE_STR),
             Some(GString::from("/foo"))
         );
-        assert_eq!(crate::uri_unescape_segment(Some("/foo%"), None, None), None);
+        assert_eq!(
+            crate::uri_unescape_segment(Some("/foo%"), crate::NONE_STR, crate::NONE_STR),
```
bilelmoussaoui, line 272: why do we provide both `NONE_STR` & `GStr::NONE`?

jf2048: Just for consistency. `NONE_STR` is not really necessary

glib/src/gstring.rs
sdroege: I think it's mostly for convenience. `c_char` is sometimes `u8`, sometimes `i8` (depending on platform), and simply having both trait impls available here reduces the potential of mismatches.

glib/src/utils.rs
sdroege: I think we should move all these `NONE_BLA` constants into `Bla::NONE` for better namespacing, but that's something for a separate PR.

glib/src/gstring.rs
sdroege: ack

glib/src/gstring.rs
jf2048: I just took that commit out

jf2048:
Some drawbacks that still persist:

- Unfortunate that it is not possible anymore to safely convert from `GStr` to `CStr` without re-scanning the string. But I think it is more "Rusty" to make paying that cost optional.
- Still have to check all strings coming from C for valid UTF-8. No way to fix that right now as C has no real guarantees about strings.
- The previously mentioned issue with needing `None::<&str>` or `GStr::NONE`.
- There are a ton of `GString` constructors and error types now, could be confusing. Not sure if all are strictly necessary but in any case it was trivial to implement them in terms of each other.
- The `Inline` variant is limited to 22 chars, it could be optimized further to hold an additional byte with some unsafe hackery to combine the length (5 bits) and the discriminant (2 bits).

jf2048:
I think those issues are all design issues or can be fixed later, so this is ready now IMO

PR #601, opened 2022-03-11
Use Send+Sync for more Boxed types

jf2048:
These should work provided the type is not refcounted, contains no refcounted types inside it, and contains no thread local data. It would be nice to find a way to test this...

### DIFF
--- a/pango/Gir.toml
+++ b/pango/Gir.toml
@@ -29,7 +29,6 @@ generate = [
     "Pango.FontsetSimple",
     "Pango.Glyph",
     "Pango.GlyphItem",
-    "Pango.GlyphString",
     "Pango.GlyphUnit",
     "Pango.Gravity",
     "Pango.GravityHint",
@@ -123,6 +122,7 @@ status = "generate"
 [[object]]
 name = "Pango.Attribute"
 status = "generate"
+concurrency = "send+sync"
     [[object.function]]
     name = "init"
     # irrelevant with all the constructors and unusable like this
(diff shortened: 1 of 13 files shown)


sdroege:
Also immutable refcounted types are usually fine (if they only contain exposed `Send` things themselves) FWIW

pango/Gir.toml
```diff
@@ -282,6 +281,11 @@ status = "generate"
         name = "language"
         const = true
 
+[[object]]
+name = "Pango.GlyphItem"
```
sdroege, line 285: This has a `PangoItem`, which has a `PangoAnalysis`, which has a `PangoEngineShape`, `PangoEngineLang`, `PangoFont` and `PangoLanguage`. Are these all OK (language seems so, I didn't check the others)? If so you probably want to mark them here too, otherwise this needs some explanation why it is OK :)

pango/Gir.toml
```diff
@@ -290,6 +294,16 @@ status = "generate"
     # converted to proper constructors
     manual = true
 
+[[object]]
+name = "Pango.GlyphString"
+status = "generate"
+concurrency = "send+sync"
+
+[[object]]
+name = "Pango.Item"
```
sdroege, line 303: See above, the `PangoAnalysis` is potentially problematic

sdroege (commented):
Looks good otherwise

pango/Gir.toml
jf2048: Good catch, `PangoFont` isn't safe. Removed this one

bilelmoussaoui (approved):

PR #602, opened 2022-03-12
gio: add AsyncInitable

pbor:


gio/src/async_initable.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::Cancellable;
+use glib::object::IsA;
+use glib::object::IsClass;
+use glib::value::ToValue;
+use glib::{Cast, Object, StaticType, Type};
+
+impl AsyncInitable {
+    #[doc(alias = "g_async_initable_new_async")]
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new_async<
```
bilelmoussaoui, line 14: Should probably have a _future variant as well?

gio/Gir.toml
```diff
@@ -307,6 +307,19 @@ status = "generate"
         name = "name"
         string_type = "os_string"
 
+[[object]]
+name = "Gio.AsyncInitable"
```
bilelmoussaoui, line 311: Should also be added to the other crates in the `manual` list so that it is implemented for the types that uses it

pbor: Which ones do you have in mind? as far as I know gtk etc does not have any async initable, no?

gio/src/async_initable.rs
pbor: It should... but that always makes my head spin :)

gio/Gir.toml
bilelmoussaoui: Never mind, i didn't found any in gtk-rs-core/gtk3-rs/gtk4-rs

gio/src/auto/async_initable.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+impl AsyncInitable {
+    pub const NONE: Option<&'static AsyncInitable> = None;
+}
+
+pub trait AsyncInitableExt: 'static {
+    #[doc(alias = "g_async_initable_init_async")]
+    fn init_async<P: FnOnce(Result<(), glib::Error>) + 'static>(
```
sdroege, line 29: This needs to be `unsafe` as you are not allowed to call it multiple times on an object. Also the future variant below

sdroege (commented):
Otherwise seems good to me

gio/src/auto/async_initable.rs
pbor: If I mark `init_async` as unsafe in Gir.toml, not only `init_future` is not automatically marked as unsafe, but it fails to compile. This seems to be a bug in gir.

sdroege (approved):

PR #603, opened 2022-03-13
gio: backport AsyncInitable

pbor:
This includes an updated gir version to properly generate unsafe async bindings.


sdroege:
Please first backport your gir change to the 0.15 branch of gir (and maybe more things if there are more things that are useful to backport and don't change APIs), and then regenerate with that here

pbor:
Okay, as far as I can see there are no breaking changes on master, so I backported everything: https://github.com/gtk-rs/gir/pull/1329

PR #608, opened 2022-03-14
Structured logging backport

mbiggio:


sdroege:
@mbiggio Can you cherry pick the individual commits you want to backport?

sdroege (approved):
Seems ok to backport to me

sdroege:
Unfortunately the `gstr!` macro requires a newer Rust version: https://github.com/gtk-rs/gtk-rs-core/runs/5535196508?check_suite_focus=true . So that's not acceptable. Can we do without for the backport?

mbiggio:
> Unfortunately the `gstr!` macro requires a newer Rust version: https://github.com/gtk-rs/gtk-rs-core/runs/5535196508?check_suite_focus=true . So that's not acceptable. Can we do without for the backport?

The problem comes from this assert:
https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/gstring.rs#L46
that is not allowed in a `const` function in rust 1.56.0.

Maybe we can drop the assert? - the necessary preconditions are listed in the comments above the functions, so in a sense the caller is supposed to make sure they are satisfied, but this opens the door to UB ...

pbor:
> The problem comes from this assert: https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/gstring.rs#L46 that is not allowed in a `const` function in rust 1.56.0.
> 
> Maybe we can drop the assert? - the necessary preconditions are listed in the comments above the functions, so in a sense the caller is supposed to make sure they are satisfied, but this opens the door to UB ...

The function is called `unchecked`, so imho dropping the assert is ok

sdroege:
@jf2048 What do you think?

jf2048:
Yes, you can take that assert out, it was copied from `CStr` nightly and it was only added last month.

PR #611, opened 2022-03-15
glib: Rework signal class handlers and accumulators

### LINKED ISSUE #365: Add closure variant of `SignalBuilder::class_handler()` and use `Value` traits for `accumulator()`

jf2048:
Fixes #365

I came up with a new approach using `g_signal_get_invocation_hint`, this could use some tests but I think this should work for getting rid of `SignalClassHandlerToken` and still retaining safety.


glib/src/subclass/signal.rs
```diff
@@ -592,27 +595,31 @@ impl Signal {
...
                         }
-                    }
-                }
 
-                res
-            })
+                        res
+                    })
+                },
+                SignalClassHandler::Closure(closure) => closure.as_ref().clone(),
```
sdroege, line 621: Where does this check for a valid return type now?

glib/src/subclass/signal.rs
```diff
@@ -471,12 +464,22 @@ impl<'a> SignalBuilder<'a> {
     /// This is called if multiple signal handlers are connected to the signal for accumulating the
     /// return values into a single value.
     pub fn accumulator<
-        F: Fn(&SignalInvocationHint, &mut Value, &Value) -> bool + Send + Sync + 'static,
+        R: ValueType,
+        F: Fn(&SignalInvocationHint, R, &R) -> ControlFlow<R, R> + Send + Sync + 'static,
```
sdroege, line 468: This now required you to take a copy of the accumulator if you just want to forward it to the return value

glib/src/subclass/signal.rs
```diff
@@ -471,12 +464,22 @@ impl<'a> SignalBuilder<'a> {
...
     >(
         mut self,
         func: F,
     ) -> Self {
-        self.accumulator = Some(Box::new(func));
+        self.accumulator = Some(Box::new(move |hint, accu, value| {
+            let curr_accu = accu.get::<R>().unwrap();
+            let value = value.get::<R>().unwrap();
```
sdroege, line 475: This doesn't allow getting the values by reference so would cause needless copies

glib/src/subclass/object.rs
```diff
@@ -216,20 +282,97 @@ impl<T: ObjectImpl> ObjectImplExt for T {
...
+    let instance = values
+        .get(0)
+        .and_then(|v| v.get::<&Object>().ok())
+        .expect("First value in array passed to `signal_chain_from_overridden` must be a GObject");
+    let hint: super::SignalInvocationHint = unsafe {
+        from_glib(gobject_ffi::g_signal_get_invocation_hint(
+            instance.to_glib_none().0,
+        ))
+    };
+    let query = hint.signal_id().query();
```
sdroege, line 320: Part of the reason for the token was also to be able to pass around the signal query for type checking purposes. Is this going to be a performance problem if its called for every signal invocation another time?

sdroege (commented):
Seems good to me otherwise

glib/src/subclass/signal.rs
jf2048: Hmm, the only signature I could think that wouldn't create additional copies is something like:

```rust
A: ValueType,
V: FromValue<'_>,
F: Fn(&SignalInvocationHint, A, V) -> (Continue, Option<A>) + Send + Sync + 'static,

Comment (jf2048) [reply glib/src/subclass/signal.rs:468]: Actually that's not so good, maybe this?
```rust
pub fn accumulator<
    T: FromValue<'_> + StaticType,
    R: ToValue + StaticType,
    F: Fn(&SignalInvocationHint, T, T) -> ControlFlow<Option<R>, Option<R>> + Send + Sync + 'static,
>(mut self, func: F) -> Self {
    assert_eq!(T::static_type(), R::static_type());
    // ...
}
```

glib/src/subclass/object.rs
jf2048: I don't like there being two ways to do this. If we use the token, that should be required for everything. But that means we can't allow closures to be passed directly as override handlers. Instead, if we want to use value unwrapping like in `closure!` we would have to add another macro:

```rust
class.override_signal_class_handler(
    "my-signal",
    glib::override_handler!(
        |token, this: &MyObject, param: &str| -> String {
            // ...
            token.chain(&[&this, &param])
        }
    )
)
```

jf2048: Ok it always uses the token now, callers don't have to pass the original value array but if they do pass it then we can skip the type checks and take the fast path

glib/src/subclass/signal.rs
jf2048: I just took out the `class_closure` bit, I don't think it's possible to do that safely and also do a check for the return type

glib/src/subclass/signal.rs
jf2048: Changed it to the above, `None` can be returned to avoid the copy

PR #615, opened 2022-03-24
glib: Fix build with `v2_72` and require 2.72.0 now

sdroege:


GuillaumeGomez (approved):

gio/sys/Cargo.toml
```diff
@@ -89,4 +89,4 @@ version = "2.68"
 version = "2.70"
 
 [package.metadata.system-deps.gio_2_0.v2_72]
-version = "2.71"
```
bilelmoussaoui, line 92: is this change actually needed?

sdroege: In what sense? The API of various things added between 2.70 and 2.72 changed a couple of times during the 2.71 development cycle, so we can only really guarantee that it works with 2.72.0 for sure.

sdroege:
Apparently ring requires a newer Rust version now, and we use ring indirectly in one of our examples. What to do?

PR #616, opened 2022-03-24
Fix a couple of new clippy warnings

sdroege:


glib/tests/structured_log.rs
```diff
@@ -1,3 +1,6 @@
+#![allow(clippy::unnecessary_to_owned)]
```
GuillaumeGomez, line 1: This one seems surprising.

GuillaumeGomez (approved):

glib/tests/structured_log.rs
sdroege: The test is explicitly testing that things work with a `String` as well as with a `&str`. So technically the `to_owned()` is unnecessary indeed but that would make the test useless.

PR #621, opened 2022-04-03
Implement ValueTypeOptional for BoxedInline and cairo types

jf2048:
Follow up to #420

### DIFF
--- a/cairo/src/lib.rs
+++ b/cairo/src/lib.rs
@@ -29,6 +29,8 @@ macro_rules! gvalue_impl {
             type Type = Self;
         }
 
+        impl glib::value::ValueTypeOptional for $name {}
+
         unsafe impl<'a> glib::value::FromValue<'a> for $name {
             type Checker = glib::value::GenericValueTypeOrNoneChecker<Self>;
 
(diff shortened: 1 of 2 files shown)


sdroege (approved):

cairo/src/lib.rs
```diff
@@ -29,6 +29,8 @@ macro_rules! gvalue_impl {
             type Type = Self;
         }
 
+        impl glib::value::ValueTypeOptional for $name {}
```
sdroege, line 32: For `cairo` this is wrong (and was already wrong before) as the same macro also covers enums for example, which are not nullable at all. Those should also not use the `GenericValueTypeOrNoneChecker`: we'd now map the `0` variant to `None` instead of the actual enum value.

We should probably add two distinct macros for these cases here.

jf2048: Confusingly, it looks like the enums already have their own macro with the same name? https://github.com/gtk-rs/gtk-rs-core/blob/36eb57054496722e769af288ccf40e7a281612b5/cairo/src/enums.rs#L12

sdroege: Ugh how confusing. Sorry for missing that :)

PR #624, opened 2022-04-04
glib: Don't serialize 0 value flags as variants

jf2048:
Also to skip things like `G_APPLICATION_FLAGS_NONE`, etc

### DIFF
--- a/glib/src/enums.rs
+++ b/glib/src/enums.rs
@@ -602,7 +602,7 @@ impl FlagsClass {
         let mut s = String::new();
         for val in self.values() {
             let v = val.value();
-            if (value & v) == v {
+            if v != 0 && (value & v) == v {
                 value &= !v;
                 if !s.is_empty() {
                     s.push('|');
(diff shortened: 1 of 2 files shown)


sdroege (approved):

sdroege:
It should probably also skip flags that have multiple bits set or not?

sdroege:
@jf2048 Or do I misunderstand this? :)

glib/src/enums.rs
```diff
@@ -602,7 +602,7 @@ impl FlagsClass {
         let mut s = String::new();
         for val in self.values() {
             let v = val.value();
-            if (value & v) == v {
+            if v != 0 && (value & v) == v {
                 value &= !v;
                 if !s.is_empty() {
                     s.push('|');
```
sdroege, line 608: Elsewhere we're using `+` for that btw

sdroege:
https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/999/diffs#7d16025f9d459b7ccbb3a024a658ece9c4a694b5_1267_1283 for some other code doing such things

PR #625, opened 2022-04-04
glib: Remove ending NUL when converting Variant to OsString.

### LINKED ISSUE #620: [BUG] Convert a GVariant to OsString adds a NUL byte
**Bug description**

I'm trying to use `gtk::Application::connect_handle_local_options` to add an option to open a file. Converting the `GVariant` to a `std::string` works fine, but if I try the properer way of setting `glib::OptionArg::Filename` and getting a `std::ffi::OsString` or a `std::path::PathBuf`, it gets an extra NUL byte at the end.

```
use gtk::prelude::*;

fn main() {
    let app = gtk::Application::new(None,
        gio::ApplicationFlags::HANDLES_OPEN | gio::ApplicationFlags::NON_UNIQUE
    );
    app.add_main_option("import", glib::Char::from(b'I'),
        glib::OptionFlags::NONE, glib::OptionArg::Filename,
        "Imports a file", None);
    app.connect_handle_local_options(
        move |_app, dict| {
            let s: Option<std::ffi::OsString> = dict.lookup("import").unwrap();
            if let Some(s) = s {
               std::fs::read(dbg!(s)).unwrap(); // Panics!
            }
            -1
        }
    );
    app.run();
}
```
Run it with `cargo run -- --import foo`:
```
[src/main.rs:12] s = "foo\u{0}"
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { kind: InvalidInput, message: "data provided contains a nul byte" }', src/main.rs:12:39
```

I'm not sure if it is a bug in glib itself, because maybe it should not add the NUL to the GVariant in the first place, or a bug in the Rust binding. In C that NUL byte is invisible so not a big deal, but in Rust Rust it makes my program crash and it is not so easy to remove.

rodrigorc:
Fixes #620.

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -14,13 +14,6 @@
 //! a full list of supported types. You may also implement [`ToVariant`] and [`FromVariant`]
 //! manually, or derive them using the [`Variant`](derive@crate::Variant) derive macro.
 //!
-//! ## Portability Warning
-//!
-//! Variant conversion traits are also implemented for certain platform-specific data types from
-//! the standard library like [`PathBuf`](std::path::PathBuf) and [`OsString`](std::ffi::OsString).
-//! These types will be serialized to their platform-specific representation and **should not** be
-//! deserialized on a different machine from where they were serialized.
-//!
 //! # Examples
 //!
 //! ```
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (dismissed):

sdroege:
Makes sense to me but now I'm wondering if maybe we should add `NUL`s when serializing those types into variants. I wonder what C code making use of paths in variants is expecting.

Might make sense bringing that up as a GLib issue if this is not documented anywhere.

rodrigorc:
> ...but now I'm wondering if maybe we should add `NUL`s when serializing those types into variants.

I was just thinking about that!

>  I wonder what C code making use of paths in variants is expecting. Might make sense bringing that up as a GLib issue if this is not documented anywhere.

I think it is documented if you look deep enough:
In [GApplication](https://github.com/GNOME/glib/blob/main/gio/gapplication.c#L709):

> * for G_OPTION_ARG_FILENAME, use `^&ay`

With `ay` being an array of bytes, that would be read with [g-variant-get-bytestring](https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/glib/glib-GVariant.html#g-variant-get-bytestring), that says:
> If the array does not end with a nul terminator character, the empty string is returned. For this reason, you can always trust that a non-NULL nul-terminated string will be returned by this function.

So my guess is that the NUL byte is mandatory. Should I update this pull-request or create a new one, as this is technically a different issue?

What I don't quite understand here is why the Windows version stores `WCHAR` (`u16`) values instead of bytes. Nothing in the GLib source code that I saw suggests that an array of `WCHAR` should be used there. Maybe I'm missing something?

I don't have a Windows machine around to try, maybe I'll try with Wine and let you know.

jf2048:
That does look like it should probably always get converted to utf8, to match what glib always does. Exposing the utf16 is a rust thing

sdroege:
> That does look like it should probably always get converted to utf8, to match what glib always does. Exposing the utf16 is a rust thing

The "local options" in `GApplication` are using a non-UTF-8 string for filename options though. It would be in local encoding (except on Windows where it's always UTF-8?).

> What I don't quite understand here is why the Windows version stores `WCHAR` (`u16`) values instead of bytes. Nothing in the GLib source code that I saw suggests that an array of `WCHAR` should be used there. Maybe I'm missing something?

That was because `OsString` uses `u16` on Windows, but probably does not make much sense considering the above, and should probably always use UTF-8 on Windows. (This means it would have to work similar to `ToGlibPtr` / `FromGlibPtr` for `OsString` / `Path` which also special-cases Windows).

> With `ay` being an array of bytes, that would be read with [g-variant-get-bytestring](https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/glib/glib-GVariant.html#g-variant-get-bytestring), that says:

`ay` is a byte array that doesn't have to end on `NUL` but `get_bytestring()` is a helper function that assumes that for "byte strings" (i.e. not byte arrays), AFAIU.

I think your analysis is correct in that regard and it should indeed be assuming / putting a `NUL` at the end for `Path` / `OsString` conversions.

jf2048:
Hmm, yes it does look like the "local encoding" on windows is supposed to be utf8: [g_win32_get_command_line](https://developer-old.gnome.org/glib/stable/glib-Windows-Compatibility-Functions.html#g-win32-get-command-line)

rodrigorc:
Ok, I'm trying to fix everything in one go. I've just added a new commit and squashed the PR.

I've added changes to `OsString::from_variant()` and `OsStr::to_variant()`, both for Unix and Windows variants, pushing and popping the NUL byte appropriately.

I'm not totally sure about the lossy conversions in Windows... But then, if you have a Windows file name that is not UTF-8 valid, I think that you have already lost: Rust can handle that with the internal WTF-8 encoding of OsString, but AFAIK glib uses pure UTF-8. The alternative to lossy conversion would be to replace them with an empty string, as Variant does with the missing NUL.

I have tested my example from the original bug report, both in Linux and in Windows (tricky!) and now it seems to work great. (previously Windows failed with something about the type of variant being "aq" instead of "ay").

sdroege:
I would go with how it's implemented in `glib/src/translate.rs` in `c_to_path_buf` and `path_to_c`. I mean, you can even call those functions from here after making them `pub(crate)` :)

glib/src/variant.rs
```diff
@@ -1230,8 +1234,13 @@ impl FromVariant for std::ffi::OsString {
...
 impl FromVariant for std::ffi::OsString {
     fn from_variant(variant: &Variant) -> Option<Self> {
-        let wide = <Vec<u16>>::from_variant(variant)?;
-        Some(std::os::windows::ffi::OsStringExt::from_wide(&wide))
+        // Remove the ending NUL like with the Unix variant.
+        let mut bytes = <Vec<u8>>::from_variant(variant)?;
+        while bytes.last() == Some(&0) {
+            bytes.pop();
+        }
+        let s = String::from_utf8_lossy(&bytes);
```
sdroege, line 1242: I.e. here you would panic instead

glib/src/variant.rs
```diff
@@ -1276,9 +1288,11 @@ impl StaticVariantType for std::ffi::OsStr {
 #[cfg(windows)]
 impl ToVariant for std::ffi::OsStr {
     fn to_variant(&self) -> Variant {
-        use std::os::windows::ffi::OsStrExt;
-        let wide = self.encode_wide().collect::<Vec<u16>>();
-        wide.to_variant()
+        // Just like the Unix variant, except that the internal [u8] of OsSTr is not visible
+        // and the wchar array ([u16]) is not useful, because it should be utf-8 encoded.
+        let mut bytes = self.to_string_lossy().into_owned().into_bytes();
```
sdroege, line 1293: And here too

glib/src/variant.rs
```diff
@@ -1218,7 +1218,11 @@ impl ToVariant for std::ffi::OsString {
 #[cfg(unix)]
```
sdroege, line 1218: The docs above here and also below need to be updated

glib/src/variant.rs
```diff
@@ -1218,7 +1218,11 @@ impl ToVariant for std::ffi::OsString {
...
 impl FromVariant for std::ffi::OsString {
```
sdroege, line 1219: And also further above the `Path` conversions need to consider long paths on Windows like in `path_to_c()`

sdroege:
Otherwise this looks all good to me!

glib/src/variant.rs
rodrigorc: Makes sense, I'll use `c_to_path_buf` and `c_to_os_string` instead `PathBuf::FromVariant` calling `OsString::FromVariant`.

glib/src/variant.rs
rodrigorc: But these docs only want about portability. That comment is there even for `any(unix,windows)`, so I guess these now should be just that.

glib/src/variant.rs
rodrigorc: Calling `path_to_c` literally should just work. I guess. I'll use `os_str_to_c` to convert `OsStr` to avoid that when not requested.

rodrigorc:
I have force-pushed a new version of this PR.

Funnily, using `translate.rs` makes the code shorter: (23 insertions(+), 54 deletions(-)).

glib/src/variant.rs
```diff
@@ -1163,7 +1163,10 @@ impl ToVariant for std::path::PathBuf {
 #[cfg(any(unix, windows))]
```
sdroege, line 1163: This `cfg`s can go away now. `glib::translate` always provides these helper functions and they're unconditionally used (indirectly) from elsewhere too.

glib/src/variant.rs
```diff
@@ -1185,7 +1188,8 @@ impl StaticVariantType for std::path::Path {
...
 impl ToVariant for std::path::Path {
     fn to_variant(&self) -> Variant {
-        self.as_os_str().to_variant()
+        let tmp = crate::translate::path_to_c(self);
+        return unsafe { from_glib_none(ffi::g_variant_new_bytestring(tmp.as_ptr() as *const u8)) };
```
sdroege, line 1192:
```suggestion
        unsafe {
            from_glib_none(ffi::g_variant_new_bytestring(tmp.as_ptr() as *const u8))
        }
```

glib/src/variant.rs
```diff
@@ -1250,35 +1243,11 @@ impl StaticVariantType for std::ffi::OsStr {
...
-/// deserialized on the same machine.
-#[cfg(windows)]
+#[cfg(any(unix, windows))]
 impl ToVariant for std::ffi::OsStr {
     fn to_variant(&self) -> Variant {
-        use std::os::windows::ffi::OsStrExt;
-        let wide = self.encode_wide().collect::<Vec<u16>>();
-        wide.to_variant()
+        let tmp = crate::translate::os_str_to_c(self);
+        return unsafe { from_glib_none(ffi::g_variant_new_bytestring(tmp.as_ptr() as *const u8)) };
```
sdroege, line 1250:
```suggestion
        unsafe {
            from_glib_none(ffi::g_variant_new_bytestring(tmp.as_ptr() as *const u8))
        }
```

glib/src/variant.rs
sdroege: They're not really platform-specific anymore but according to the GLib rules. Before there was the `u8` / `u16` difference, now they're just byte arrays in either case and work the same like everything else in GLib on all platforms.

So I think either these portability notes should be removed, or maybe there should be a note of the above (GLib rules, local encoding on UNIX, UTF-8 on Windows).

sdroege:
The implementation looks all good to me, just needs a little bit of cleanup and then this can go in.

> Funnily, using translate.rs makes the code shorter: (23 insertions(+), 54 deletions(-)).

That's nice :)

glib/src/variant.rs
rodrigorc: Ops! Done.

rodrigorc:
I've just force-pushed a new version.
Also, now that these functions call into `translate.rs` I've fully separated the impls of `Path/PathBuf` and `OsStr/OsString`.

sdroege (approved):

glib/src/variant.rs
```diff
@@ -1134,151 +1134,75 @@ impl ToVariant for str {
...
-///
-/// This impl is for a platform-specific type. Any variants serialized from this type should be
-/// deserialized on the same machine.
-#[cfg(any(unix, windows))]
 impl FromVariant for std::path::PathBuf {
     fn from_variant(variant: &Variant) -> Option<Self> {
-        Some(std::ffi::OsString::from_variant(variant)?.into())
+        unsafe {
+            let ptr = ffi::g_variant_get_bytestring(variant.to_glib_none().0);
+            Some(crate::translate::c_to_path_buf(ptr as *const i8))
```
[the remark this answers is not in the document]
rodrigorc: That `c_char` thing is madness!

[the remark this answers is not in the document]
sdroege: It is indeed. Bad choices made many decades ago that are still haunting us :)

Issue #628, opened 2022-04-06
[BUG] glib::mkdtemp produces a crash


melix99:
**Bug description**

Running a simple program with `glib::mkdtemp` makes the app crash with the following output:

```
free(): double free detected in tcache 2
Aborted (core dumped)
```

This is the code I used:

```rust
fn main() {
    glib::mkdtemp("test-XXXXXX");
}
```

sdroege:
Can you get a backtrace with gdb or lldb?

melix99:
```
#0  0x00007ffff7c8988c in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007ffff7c3c6a6 in raise () at /lib64/libc.so.6
#2  0x00007ffff7c267d3 in abort () at /lib64/libc.so.6
#3  0x00007ffff7c7da07 in __libc_message () at /lib64/libc.so.6
#4  0x00007ffff7c9387c in  () at /lib64/libc.so.6
#5  0x00007ffff7c9579f in _int_free () at /lib64/libc.so.6
#6  0x00007ffff7c97c05 in free () at /lib64/libc.so.6
#7  0x000055555555d38b in alloc::alloc::dealloc (ptr=0x5555555b1b20, layout=...)
    at /rustc/c84f39e6c08397d4ab9b51b472002a8bfc3b5b59/library/alloc/src/alloc.rs:105
#8  0x000055555555d52b in alloc::alloc::{impl#1}::deallocate (self=0x7fffffffdba0, ptr=..., layout=...)
    at /rustc/c84f39e6c08397d4ab9b51b472002a8bfc3b5b59/library/alloc/src/alloc.rs:242
#9  0x000055555555d4a9 in alloc::alloc::box_free<[u8], alloc::alloc::Global> (ptr=..., alloc=...)
    at /rustc/c84f39e6c08397d4ab9b51b472002a8bfc3b5b59/library/alloc/src/alloc.rs:340
        layout = core::alloc::layout::Layout {size_: 12, align_: core::num::nonzero::NonZeroUsize (1)}
        align = 1
        size = 12
#10 0x000055555555d2c0 in core::ptr::drop_in_place<alloc::boxed::Box<[u8], alloc::alloc::Global>> ()
    at /rustc/c84f39e6c08397d4ab9b51b472002a8bfc3b5b59/library/core/src/ptr/mod.rs:188
[...]

PR #630, opened 2022-04-07
glib: Fix `mkdtemp()` and `mkdtemp_full()`

sdroege:
These functions modify the passed in string in-place and either return it directly or return `NULL`. This kind of behaviour can only be handled manually.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/628

### DIFF
--- a/glib/Gir.toml
+++ b/glib/Gir.toml
@@ -232,7 +232,16 @@ status = "generate"
     manual = true
     [[object.function]]
     name = "mkstemp"
-    #manual is_windows_utf8
+    #manual is_windows_utf8 and also modifies the string passed in
+    manual = true
+    [[object.function]]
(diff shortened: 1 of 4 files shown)


glib/Gir.toml
```diff
@@ -232,7 +232,16 @@ status = "generate"
     manual = true
     [[object.function]]
     name = "mkstemp"
-    #manual is_windows_utf8
+    #manual is_windows_utf8 and also modifies the string passed in
```
GuillaumeGomez, line 235: `#manual`?

sdroege: It's manually implemented. This is the same comment as in the 50 lines above and IIRC you added that back then ;)

Issue #622, opened 2022-04-04
[FEATURE REQUEST] Add Variant::is_object_path()


lucab:
It looks like [g_variant_is_object_path()](https://docs.rs/glib-sys/latest/glib_sys/fn.g_variant_is_object_path.html) does not currently have a safe wrapper in `glib`.

Among several possible approaches, there is consensus to expose it as `Variant::is_object_path(&str)->bool` so that it can be easily found.

lucab:
These are the ones I've seen, aggregating similar ones by signatures (the crossed ones already have wrappers):
 1. string -> bool: `g_variant_is_object_path`, `g_variant_is_signature`
 2. Variant -> bool: `g_variant_is_floating`, `g_variant_is_normal_form`, ~~`g_variant_is_container`~~
 3. (Variant, Type) -> bool: ~~`g_variant_is_of_type`~~

I'll start tackling the first bucket, which is the one I cared about initially.

sdroege:
> 2\. `g_variant_is_floating`

This will always return `false` for us :)

jf2048:
Would it be wanted to have some types for `ObjectPath` and `ObjectPathBuf`? They could be similar to `VariantTy` and `VariantType`, i.e. do the checking during construction, and then could also implement `StaticVariantType` of `o`

And then some would also be needed for dbus signatures because those are a subset of GVariant types.

lucab:
@slomo good point! Then I guess it doesn't make much sense to wrap that.

I added wrappers for all the other three methods in https://github.com/gtk-rs/gtk-rs-core/pull/632.

PR #632, opened 2022-04-07
glib/variant: add some more safe wrappers

lucab:
This adds manual wrappers for the following functions:
 - `g_variant_is_normal_form`
 - `g_variant_is_object_path`
 - `g_variant_is_signature`

It completes wrapping all the existing `g_variant_is_*` methods,
except for `g_variant_is_floating` which is not useful in Rust.

Closes: https://github.com/gtk-rs/gtk-rs-core/issues/622

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -859,10 +859,31 @@ impl Variant {
     }
 
     // rustdoc-stripper-ignore-next
-    /// Variant has a container type.
+    /// Return whether this Variant is a container type.
     #[doc(alias = "g_variant_is_container")]
     pub fn is_container(&self) -> bool {
-        unsafe { ffi::g_variant_is_container(self.to_glib_none().0) != ffi::GFALSE }
+        unsafe { from_glib(ffi::g_variant_is_container(self.to_glib_none().0)) }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Return whether this Variant is in normal form.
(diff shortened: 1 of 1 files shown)


glib/src/variant.rs
```diff
@@ -859,11 +859,32 @@ impl Variant {
...
+    #[doc(alias = "g_variant_is_object_path")]
+    pub fn is_object_path(string: &str) -> bool {
+        unsafe { ffi::g_variant_is_object_path(string.to_glib_none().0) != ffi::GFALSE }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Return whether input string is a valid `Variant::Signature`.
+    #[doc(alias = "g_variant_is_signature")]
+    pub fn is_signature(string: &str) -> bool {
+        unsafe { ffi::g_variant_is_signature(string.to_glib_none().0) != ffi::GFALSE }
```
sdroege, line 886:
```suggestion
        unsafe { from_glib(ffi::g_variant_is_signature(string.to_glib_none().0)) }
```

also for everything else here

lucab: Indeed it seemed a bit odd, but I was following the existing code in `is_container()`. I'll take the liberty of tweaking that one too. Thanks for the prompt feedback!

lucab:
Rebased according to review suggestions, CI is green.

bilelmoussaoui (approved):

sdroege (approved):

PR #633, opened 2022-04-08
error: add `matches` method

pbor:
This is handy when in a match you want to:

```
    Err(e) if e.matches(IoErrorEnum::Cancelled) => ...
```

### DIFF
--- a/glib/src/error.rs
+++ b/glib/src/error.rs
@@ -52,6 +52,13 @@ impl Error {
     }
 
     // rustdoc-stripper-ignore-next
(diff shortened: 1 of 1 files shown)


pbor:
`is` and `matches` imho would make more sense swapped, with `is` taking the exact error and `matches` just matching the domain... is that too much API churn?

sdroege:
I would agree with you but what you just said, and also this is how the C version of the function is named so... :)

sdroege (approved):

PR #636, opened 2022-04-09
glib: Add new `ObjectImplRef` and `ObjectImplWeakRef` types

sdroege:
These provide strong/weak reference-counted wrappers around the
implementation type of GObject subclasses and can be used for passing
into closures, for example.

----

@bilelmoussaoui Do we know any places where this can be used in the gtk4-rs examples?

CC @piegamesde

### DIFF
--- a/glib/src/subclass/mod.rs
+++ b/glib/src/subclass/mod.rs
@@ -261,6 +261,9 @@ pub mod shared;
 
 pub mod signal;
 
+mod object_impl_ref;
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

bilelmoussaoui:
> @bilelmoussaoui Do we know any places where this can be used in the gtk4-rs examples?

Can't think of anything on top of my head, but will give it a look later today

PR #637, opened 2022-04-10
Update gir files

sdroege:


gio/src/list_store.rs
```diff
@@ -65,6 +65,48 @@ impl ListStore {
     pub fn extend_from_slice(&self, additions: &[impl IsA<glib::Object>]) {
         self.splice(self.n_items() - 1, 0, additions)
     }
+
+    #[cfg(any(feature = "v2_74", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_74")))]
+    #[doc(alias = "g_list_store_find_with_equal_func_full")]
+    #[doc(alias = "g_list_store_find_with_equal_func")]
+    pub fn find_with_equal_func<F: FnMut(&glib::Object, &glib::Object) -> bool>(
```
sdroege, line 73: This one is a bit strange. From a Rust point of view we would probably like to have a closure `FnMut(&glib::Object) -> bool` as the second object is literally just `item` that is passed through every time.

In Rust we have proper closures, such hacks are not really needed.

sdroege: Changed it accordingly, but opinions here? @bilelmoussaoui @jf2048 ?

gio/Gir.toml
```diff
@@ -833,7 +833,12 @@ generate_builder = true
     manual = true
     [[object.function]]
     name = "find_with_equal_func"
-    ignore = true # See https://gitlab.gnome.org/GNOME/glib/-/issues/2447
+    # calls find_with_equal_func_full
```
bilelmoussaoui, line 836: I would mark this one as ignored as mark the other one as renamed instead

gio/sys/src/lib.rs
```diff
@@ -12195,6 +12195,13 @@ extern "C" {
         equal_func: glib::GEqualFunc,
         position: *mut c_uint,
     ) -> gboolean;
+    pub fn g_list_store_find_with_equal_func_full(
```
bilelmoussaoui, line 12198: Needs a feature flag

gio/Gir.toml
sdroege: I did it this way because the docs are more useful then

GuillaumeGomez: The FFI calls aren't renamed?

sdroege: How do you mean?

sdroege: `find_with_equal_func` calls `g_list_store_find_with_equal_func_full` because the non-`_full` variant is not usable for bindings and the difference between the two functions is irrelevant for the bindings user (i.e. the extra sentence in the docs of the `_full` variant are just confusing for the bindings user).

GuillaumeGomez: Oh I see. I thought the code generated by gir was invalid because it was badly renamed or something like that.

gio/src/list_store.rs
jf2048: Seems fine to me, definitely more in line with `std::iter::Iterator::find`

PR #644, opened 2022-04-11
glib: Add bindings for GSignalGroup

### LINKED ISSUE #638: gobject: Add bindings for SignalGroup
CC @jf2048 who had some initial version of this IIRC. `g_signal_group_connect_closure()` is added by https://github.com/gtk-rs/gtk-rs-core/pull/637

jf2048:
Fixes #638, could use a unit test or two


glib/src/gobject/auto/signal_group.rs
```diff
@@ -0,0 +1,76 @@
+// This file was generated by gir (https://github.com/gtk-rs/gir)
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use crate::object::IsA;
+use crate::object::ObjectType as ObjectType_;
```
jf2048, line 6: This is unused and is causing a warning. Is there a way to suppress this warning, or stop this statement from being generated?

sdroege: Can you create an issue about that? It's a problem in gir. In other places there's already an `#[allow(unused_imports)]` on top of `mod auto` in `lib.rs` for the very same reason.

glib/Gir_GObject.toml
```diff
@@ -103,6 +103,44 @@ concurrency = "send+sync"
...
+    [[object.function]]
+    name = "connect_closure"
+    rename = "connect"
+    manual = true
+    [[object.function]]
+    name = "dup_target"
+    rename = "target"
+    [[object.property]]
+    name = "target"
+    # notify signal is racy, use connect_bind/connect_unbind instead
```
sdroege, line 135: In what sense?

glib/Gir_GObject.toml
```diff
@@ -103,6 +103,44 @@ concurrency = "send+sync"
     name = "source"
     generate = ["notify"]
 
+[[object]]
+name = "GObject.SignalGroup"
+status = "generate"
+final_type = true
+concurrency = "send+sync"
```
sdroege, line 110:
```suggestion
concurrency = "send+sync"
version = "2.74"
```

Probably? It's unusable from the bindings without the `connect_closure()` function

glib/src/gobject/signal_group.rs
```diff
@@ -0,0 +1,185 @@
...
+impl SignalGroup {
+    #[cfg(any(feature = "v2_74", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_74")))]
+    #[doc(alias = "g_signal_group_connect_closure")]
+    pub fn connect(&self, signal_name: &str, after: bool, closure: RustClosure) {
+        unsafe {
+            gobject_ffi::g_signal_group_connect_closure(
+                self.to_glib_none().0,
+                signal_name.to_glib_none().0,
+                closure.as_ref().to_glib_none().0,
```
sdroege, line 21: Is there anything that has to be done here to ensure that the parameter/return types of the closure are matching with the signal?

sdroege (commented):

glib/Gir_GObject.toml
jf2048: The `bind` signal holds the internal mutex during emission and also emits the target as an argument to the handler, so is safe to reference the target from another thread. The docs mention it but don't elaborate: https://docs.gtk.org/gobject/signal.SignalGroup.bind.html

glib/src/gobject/signal_group.rs
jf2048: The type checking is handled by `RustClosure`, but I can't write a unit test for it because catching a panic from within the signal handler is UB and corrupts the internal gobject state

sdroege: Ok!

jf2048: Filed another issue for that here, btw: https://github.com/gtk-rs/gtk-rs-core/issues/645

glib/src/gobject/signal_group.rs
```diff
@@ -0,0 +1,185 @@
...
+use crate::ObjectType;
+use crate::RustClosure;
+use crate::SignalGroup;
+use std::mem::transmute;
+
+impl SignalGroup {
+    #[cfg(any(feature = "v2_74", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_74")))]
+    #[doc(alias = "g_signal_group_connect_closure")]
+    pub fn connect(&self, signal_name: &str, after: bool, closure: RustClosure) {
```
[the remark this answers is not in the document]
jf2048: I made it `connect_closure` and then added `connect` and `connect_local`, do we care that the names conflict with `ObjectExt`?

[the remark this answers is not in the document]
sdroege: Uh right, this is also a `GObject`... not too bad here but still ugly :) We could name it `connect_signal_group_XXX()`?

sdroege (approved):

PR #647, opened 2022-04-12
glib: Remove redundant null-checks before `g_free`

RealKC:
`g_free`'s docmentation explicitly says that calling it with a null
pointer is simply a noop:

    If mem is NULL it simply returns, so there is no need to check mem
    against NULL before calling this function.

### DIFF
--- a/glib/src/collections.rs
+++ b/glib/src/collections.rs
@@ -202,9 +202,7 @@ impl<
 
         PtrSlice {
             ptr: if len == 0 {
-                if !ptr.is_null() {
-                    ffi::g_free(ptr as ffi::gpointer);
-                }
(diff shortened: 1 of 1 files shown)


sdroege:
> `g_free`'s docmentation explicitly says that calling it with a null
> pointer is simply a noop:

The same is also true for the C `free` function btw, at least on a C99 compatible implementation (which GLib requires).

sdroege (approved):

sdroege:
> The same is also true for the C `free` function btw, at least on a C99 compatible implementation (which GLib requires).

Which is extra funny because I, being aware of that and that `g_free` also behaves like that, added those `NULL` checks in that file some weeks ago.

PR #648, opened 2022-04-13
Remove more redundant null checks

RealKC:
I missed these last night ^^

### DIFF
--- a/gio/src/file_attribute_info.rs
+++ b/gio/src/file_attribute_info.rs
@@ -19,9 +19,7 @@ glib::wrapper! {
             copy
         },
         free => |ptr| {
-            if !(*ptr).name.is_null() {
-                glib::ffi::g_free((*ptr).name as *mut _);
-            }
+            glib::ffi::g_free((*ptr).name as *mut _);
             glib::ffi::g_free(ptr as *mut _);
         },
         init => |ptr| {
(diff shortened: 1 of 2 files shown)


gio/src/file_attribute_info.rs
```diff
@@ -13,30 +13,22 @@ glib::wrapper! {
...
-                (*copy).name = glib::ffi::g_strdup((*ptr).name);
-            }
+            (*copy).name = glib::ffi::g_strdup((*ptr).name);
             copy
         },
         free => |ptr| {
-            if !(*ptr).name.is_null() {
-                glib::ffi::g_free((*ptr).name as *mut _);
-            }
+            glib::ffi::g_free((*ptr).name as *mut _);
```
RealKC, line 20: Is `ptr` here guaranteed to be nonnull?

sdroege: Yes

RealKC: That's good then, I looked at the docs and wasn't sure

sdroege: It's only called internally in `glib/src/{shared,boxed,boxed_inline,object}.rs` for the corresponding type, and mostly from the `Drop` impl. All the pointers in use there are required to be non-`NULL`.

PR #650, opened 2022-04-15
gio: Add local variant for connect_cancelled

SeaDve:
I'm not sure if I got it right

### DIFF
--- a/gio/src/cancellable.rs
+++ b/gio/src/cancellable.rs
@@ -55,6 +55,13 @@ pub trait CancellableExtManual {
         callback: F,
     ) -> Option<CancelledHandlerId>;
     // rustdoc-stripper-ignore-next
+    /// Local variant of [`Self::connect_cancelled`].
+    #[doc(alias = "g_cancellable_connect")]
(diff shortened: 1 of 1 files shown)


sdroege:
We discussed this when adding the non-local variants and didn't consider adding it because why would anybody need that :) What's your use-case?

sdroege (approved):

SeaDve:
> We discussed this when adding the non-local variants and didn't consider adding it because why would anybody need that :) What's your use-case?

I use it currently to update some GTK UI state directly through the cancelled callback

sdroege:
How do you make sure that cancellation only happens from the main thread?

SeaDve:
I'm not sure how to explain it correctly, but here is my current usage (used a custom cancellable there):

https://github.com/SeaDve/Mousai/blob/53f2afbadbb89151c7afac120fa45cae3d20f12a/src/recognizer/mod.rs#L199

It was made sure that the cancellation happens at the main thread because it was not sent at all to other threads.

Issue #646, opened 2022-04-12
List of functions with missing safety docs


RealKC:
During the discussion on the matrix channel which prompted me to open #629, I mentioned I'd take a look for more functions that were unsafe yet lacked a safety section in their documentation, and open an issue. I was told this would be welcome, so here's that issue.

Methodology: I used VSCode's `Search` panel to search for `pub unsafe fn` and manually inspected each function.

# Cairo

The following do not include safety docs:

<details>

 - `Context::from_raw_none`
 - `Context::from_raw_borrow`
 - `Context::from_raw_full`
 - `Device::from_raw_none`
 - `Device::from_raw_borrow`
 - `Device::from_raw_full`
 - `Path::from_raw_full`
 - `Pattern::from_raw_none`
 - `Pattern::from_raw_full`
 - `Region::from_raw_none`
 - `Region::from_raw_borrow`
 - `Region::from_raw_full`
 - Generated surface `from_raw_full` and `from_raw_none` through the `declare_surface!` macro
 - `Surface::from_raw_none`
 - `Surface::from_raw_borrow`
 - `Surface::from_raw_full`
 - `Surface::mime_data_raw`
 - `utils::debug_reset_static_data`
 - `XCBConnection::from_raw_none`
 - `XCBConnection::from_raw_borrow`
 - `XCBConnection::from_raw_full`
 - `XCBRenderPictFormInfo::from_raw_none`
 - `XCBRenderPictFormInfo::from_raw_borrow`
 - `XCBRenderPictFormInfo::from_raw_full`
 - `XCBScreen::from_raw_full`
 - `XCBScreen::from_raw_none`
 - `XCBScreen::from_raw_borrow`
 - `XCBVisualType::from_raw_full`
 - `XCBVisualType::from_raw_none`
 - `XCBVisualType::from_raw_borrow`
 - `FontFace::from_raw_full`
 - `FontFace::from_raw_none`
 - `FontOptions::from_raw_full`
 - `FontOptions::from_raw_none`
 - `ScaledFont::from_raw_full`
 - `ScaledFont::from_raw_none`

</details>

The following have safety comments where the function is defined, but as it's written like a normal comment, it won't be shown on the generated docs (or on hover by IDEs):
 - `FontFace::create_from_ft`
 - `FontFace::create_from_ft_with_flags`

# Gio

The following do not include safety docs:

[...]

PR #651, opened 2022-04-17
glib: WatchedObject improvements

jf2048:
Adds support for watching a `Sync` object, adds a test for that, and adds safety docs as per #646

### DIFF
--- a/glib-macros/tests/test.rs
+++ b/glib-macros/tests/test.rs
@@ -652,4 +652,48 @@ fn closure() {
         };
         cast_test.invoke::<()>(&[]);
     }
+
+    {
+        use glib::subclass::prelude::*;
+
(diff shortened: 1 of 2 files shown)


glib-macros/tests/test.rs
```diff
@@ -652,4 +652,48 @@ fn closure() {
         };
         cast_test.invoke::<()>(&[]);
     }
+
+    {
```
sdroege, line 656: Why is this test in glib-macros?

sdroege (approved):
Looks good to me otherwise

glib-macros/tests/test.rs
jf2048: No particular reason, all the proc macro tests were there before

PR #657, opened 2022-04-22
0.15 backports

sdroege:


GuillaumeGomez (approved):

sdroege:
@GuillaumeGomez You broke things in the checker :)

```
--- a/cairo/src/matrices.rs
+++ b/cairo/src/matrices.rs
@@ -7,6 +7,7 @@ use libc::c_double;
 // checker-ignore-item
 #[repr(C)]
+#[doc(alias = "Matrix")]
 pub struct Matrix {
     pub xx: c_double,
     pub yx: c_double,
Error: Process completed with exit code 1.
```

a) it's supposed to be ignored, see the comment above, b) if I would apply that change it would not compile because alias and struct name are the same

GuillaumeGomez:
Ah interesting. Sending fix to the checker shortly.

Issue #664, opened 2022-04-29
[BUG] glib::BindingGroup bindings are broken


A6GibKm:
The following operation crashes
```rust
        let binding_group = glib::BindingGroup::new();
        binding_group
            .bind("some_prop", &some_target_object, "other_prop")
            .build();

        binding_group.set_source(Some(&source_object));
```

with message  [(full backtrace)](https://gitlab.gnome.org/-/snippets/3411)
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: BoolError { message: "Binding group does not have a source set"...
```
But the binding groups are meant to be used without a source object. The bindings are connected and disconnected whenever a new source object is set.

The other problem is that in C, the bindings are by default have G_BINDING_BIDIRECTIONAL by default.

The expected behavior is stated in the [docs](https://docs.gtk.org/gobject/class.BindingGroup.html) and can be verified to work as advertised.

sdroege:
CC @jf2048

jf2048:
It sounds good to me to have the flags in `BindingGroupBuilder` set to `BIDIRECTIONAL` by default

jf2048:
If the test is doing the same thing that C is doing, then it seems you would verify that the flag is needed?

PR #665, opened 2022-04-29
Fix GBindingGroup bindings

A6GibKm:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/664.

### DIFF
--- a/glib/src/gobject/binding_group.rs
+++ b/glib/src/gobject/binding_group.rs
@@ -173,7 +173,8 @@ impl<'a> BindingGroupBuilder<'a> {
             let _ = Box::from_raw(data as *mut (TransformFn, TransformFn, String, ParamSpec));
         }
 
-        let source_property = if let Some(source) = self.group.source() {
+        let mut _source_propery_name_cstr = None;
+        let source_property_name = if let Some(source) = self.group.source() {
             let source_property = source.find_property(self.source_property).ok_or_else(|| {
                 bool_error!(
                     "Source property {} on type {} not found",
@@ -181,15 +182,17 @@ impl<'a> BindingGroupBuilder<'a> {
                     source.type_()
                 )
             })?;
-            Some(source_property)
+
+            // This is NUL-termianted from the C side
+            source_property.name().as_ptr()
         } else {
-            None
+            // This is a Rust &str and needs to be NUL-terminated first
+            let source_propery_name = std::ffi::CString::new(self.source_property).unwrap();
+            let source_property_name_ptr = source_propery_name.as_ptr() as *const u8;
+            _source_propery_name_cstr = Some(source_propery_name);
+
+            source_property_name_ptr
         };
-        let source_property_name = source_property
-            .as_ref()
-            .map(crate::ParamSpec::name)
-            .unwrap_or_else(|| self.source_property)
-            .as_ptr();
 
         unsafe {
             let target: Object = from_glib_none(self.target.clone().to_glib_none().0);
--- a/glib/src/gobject/binding_group.rs
+++ b/glib/src/gobject/binding_group.rs
@@ -213,7 +213,7 @@ impl<'a> BindingGroupBuilder<'a> {
                 Box::into_raw(Box::new((
                     self.transform_to,
                     self.transform_from,
-                    source_property_name,
+                    String::from_glib_none(source_property_name as *const _),
                     target_property,
                 )))
             } else {
--- a/glib/src/gobject/binding_group.rs
+++ b/glib/src/gobject/binding_group.rs
@@ -170,7 +170,7 @@ impl<'a> BindingGroupBuilder<'a> {
         }
 
         unsafe extern "C" fn free_transform_data(data: ffi::gpointer) {
-            let _ = Box::from_raw(data as *mut (TransformFn, TransformFn, ParamSpec, ParamSpec));
+            let _ = Box::from_raw(data as *mut (TransformFn, TransformFn, String, ParamSpec));
         }
 
         let source_property = if let Some(source) = self.group.source() {
@@ -246,7 +246,7 @@ impl<'a> BindingGroupBuilder<'a> {
     }
 
     // rustdoc-stripper-ignore-next
-    /// Similar to `try_build` but fails instead of panicking.
+    /// Similar to `try_build` but panics instead of failing.
     pub fn build(self) {
         self.try_build().unwrap()
     }
--- a/glib/src/gobject/binding_group.rs
+++ b/glib/src/gobject/binding_group.rs
@@ -251,3 +251,188 @@ impl<'a> BindingGroupBuilder<'a> {
         self.try_build().unwrap()
     }
 }
+
+#[cfg(test)]
+mod test {
+    use crate::prelude::*;
+    use crate::subclass::prelude::*;
+
+    #[test]
+    fn binding_without_source() {
+        let binding_group = crate::BindingGroup::new();
+
+        let source = TestObject::default();
+        let target = TestObject::default();
+
+        assert!(source.find_property("name").is_some());
+        binding_group.bind("name", &target, "name").build();
+
+        binding_group.set_source(Some(&source));
+
+        source.set_name("test_source_name");
+        assert_eq!(source.name(), target.name());
+
+        target.set_name("test_target_name");
+        assert_eq!(source.name(), target.name());
+    }
+
+    #[test]
+    fn binding_with_source() {
+        let binding_group = crate::BindingGroup::new();
+
+        let source = TestObject::default();
+        let target = TestObject::default();
+
+        binding_group.set_source(Some(&source));
+
+        binding_group.bind("name", &target, "name").build();
+
+        source.set_name("test_source_name");
+        assert_eq!(source.name(), target.name());
+    }
+
+    #[test]
+    fn binding_to_transform() {
+        let binding_group = crate::BindingGroup::new();
+
+        let source = TestObject::default();
+        let target = TestObject::default();
+
+        binding_group.set_source(Some(&source));
+        binding_group
+            .bind("name", &target, "name")
+            .flags(crate::BindingFlags::SYNC_CREATE)
+            .transform_to(|_binding, value| {
+                let prev_name = value.get::<String>().unwrap();
+                assert_eq!(prev_name, "Hello");
+
+                Some("World".to_value())
+            })
+            .build();
+
+        source.set_name("Hello");
+        assert_eq!(target.name(), "World");
+    }
+
+    #[test]
+    fn binding_from_transform() {
+        let binding_group = crate::BindingGroup::new();
+
+        let source = TestObject::default();
+        let target = TestObject::default();
+
+        binding_group.set_source(Some(&source));
+        binding_group
+            .bind("name", &target, "name")
+            .flags(crate::BindingFlags::SYNC_CREATE | crate::BindingFlags::BIDIRECTIONAL)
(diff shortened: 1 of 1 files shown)


sdroege:
CC @jf2048

glib/src/gobject/binding_group.rs
```diff
@@ -140,21 +140,23 @@ impl<'a> BindingGroupBuilder<'a> {
             user_data: ffi::gpointer,
         ) -> ffi::gboolean {
             let transform_data =
-                &*(user_data as *const (TransformFn, TransformFn, ParamSpec, ParamSpec));
+                &*(user_data as *const (TransformFn, TransformFn, Option<ParamSpec>, ParamSpec));
```
sdroege, line 143: I don't think this will work like this. You need the source pspec here for checking that the value types are matching, but if there is none duration construction then you can't really get it in here easily.

jf2048:
This should probably get a unit test

glib/src/gobject/binding_group.rs
jf2048: This should probably just get rid of the `ParamSpec`s and use strings and `find_property`, that is what the `BindingGroup` does internally

sdroege: Sounds like a plan

sdroege:
What's the status here?

A6GibKm:
> What's the status here?

I been busy this week, I hope to return to it today or tomorrow.

glib/src/gobject/binding_group.rs
```diff
@@ -233,8 +249,157 @@ impl<'a> BindingGroupBuilder<'a> {
...
+        pub struct TestObject {
+            pub name: RefCell<String>,
+            pub enabled: RefCell<bool>,
+        }
+
+        #[crate::object_subclass]
+        impl ObjectSubclass for TestObject {
+            const NAME: &'static str = "Test";
+            type Type = super::TestObject;
+            type ParentType = crate::Object;
```
bilelmoussaoui, line 319:
```suggestion
```

sdroege:
@A6GibKm The problem was in your code: you were not NUL-terminating the string if there was no source object.

Now the test is still failing but that's something plainly on the Rust side of things :)

glib/src/gobject/binding_group.rs
```diff
@@ -233,8 +252,159 @@ impl<'a> BindingGroupBuilder<'a> {
...
+
+        source.set_name("test_source_name");
+        assert_eq!(source.name(), target.name());
+
+        target.set_name("test_target_name");
+        assert_eq!(source.name(), target.name());
+
+        let err = binding_group.bind("enabled", &target, "name").try_build();
+
+        assert!(err.is_err())
```
sdroege, line 289: This fails because the binding group is **not** checking if the types are matching, and it also really can't because the transform functions might make it work for this set of types.

glib/src/gobject/binding_group.rs
```diff
@@ -120,7 +120,10 @@ impl<'a> BindingGroupBuilder<'a> {
                 None => false,
                 Some(res) => {
                     assert!(
-                        res.type_().is_a(transform_data.3.value_type()),
+                        crate::Value::type_transformable(
+                            res.type_(),
+                            transform_data.3.value_type()
+                        ),
```
sdroege, line 126: Why this change?

glib/src/gobject/binding_group.rs
```diff
@@ -111,7 +111,7 @@ impl<'a> BindingGroupBuilder<'a> {
             user_data: ffi::gpointer,
         ) -> ffi::gboolean {
             let transform_data =
-                &*(user_data as *const (TransformFn, TransformFn, ParamSpec, ParamSpec));
```
sdroege, line 114: It would be good to have a test with the transform functions to make sure this actually works properly.

glib/src/gobject/binding_group.rs
```diff
@@ -140,19 +143,26 @@ impl<'a> BindingGroupBuilder<'a> {
...
 
             match (transform_data.1.as_ref().unwrap())(
-                &from_glib_borrow(binding),
+                &binding,
                 &*(from_value as *const Value),
             ) {
                 None => false,
                 Some(res) => {
+                    let pspec_name = transform_data.2.clone();
+                    let source = binding.source().unwrap();
```
sdroege, line 156: This is not thread-safe I think. The source might be unset while the closure is running or not? CC @jf2048

jf2048: Yes, it seems the real source is kept on the stack: https://gitlab.gnome.org/GNOME/glib/-/blob/91407cd7efec344f6bb5fd88fd7f771afb4fed25/gobject/gbinding.c

glib/src/gobject/binding_group.rs
A6GibKm: I think the check is more correct, as it should accept something that is transformable to the actual type and not exactly the same one (it was what it it was suggested for the other closure too), but it needs a test

glib/src/gobject/binding_group.rs
jf2048: This should be thread safe actually. `GBinding:source` is construct only, and the notify handler will always hold a reference while the transform function is running

sdroege: How is this construct-only if there's `binding_group.set_source(Some(&source))` (in the test)? :)

jf2048: This is in the `GBinding`, not the `GBindingGroup`. The binding group creates a new `GBinding` for each property, each `GBinding` holds a construct only weak ref to the source.

sdroege: Uh... right. Seems indeed fine then :)

glib/src/gobject/binding_group.rs
sdroege: Yes please. Is something actually going to do that transform at some point ;)

Also this probably means that `GBinding` bindings need to be updated the same way?

sdroege: Also unclear if we want to allow that. This means that you can pass a `0u32` and it would become a `false`, for example. For properties/signals we disallow automatic conversions as that seems out of place in Rust.

jf2048: Let's leave it as `type_().is_a`, that should disallow numeric conversions but still allow for object/interface types to be passed without casting them

A6GibKm:
> @A6GibKm The problem was in your code: you were not NUL-terminating the string if there was no source object.
> 
> Now the test is still failing but that's something plainly on the Rust side of things :)

Is this Ok now? what does NULL-terminating a string even means in rust.

glib/src/gobject/binding_group.rs
sdroege: @A6GibKm This is still pending

glib/src/gobject/binding_group.rs
sdroege: @A6GibKm This is still pending

glib/src/gobject/binding_group.rs
sdroege: @A6GibKm This is still pending

sdroege:
> Is this Ok now?

See comments I added to the still pending discussions above.

> what does NULL-terminating a string even means in rust.

These are strings that are provided to C functions. They need to have a 0-byte at the end.

glib/src/gobject/binding_group.rs
A6GibKm: Removed the assert.

glib/src/gobject/binding_group.rs
A6GibKm: Ok leaved as it was.

glib/src/gobject/binding_group.rs
A6GibKm: Added tests but they fail here locally with 
```
Caused by:
  process didn't exit successfully: `/var/home/user/Projects/gtk-rs-core/target/debug/deps/glib-73fbdc0a479b0288` (signal: 11, SIGSEGV: invalid memory reference)
  ```

Comment (sdroege) [reply glib/src/gobject/binding_group.rs:114]: Good that we added a test then :)

Comment (jf2048) [reply glib/src/gobject/binding_group.rs:114]: I think the crash is because this should be `String, ParamSpec`?

Comment (A6GibKm) [reply glib/src/gobject/binding_group.rs:114]: It is like that now

File: glib/src/gobject/binding_group.rs [hunk 7]
```diff
@@ -166,20 +173,30 @@ impl<'a> BindingGroupBuilder<'a> {
             let _ = Box::from_raw(data as *mut (TransformFn, TransformFn, ParamSpec, ParamSpec));
```
jf2048, line 173:
```suggestion
            let _ = Box::from_raw(data as *mut (TransformFn, TransformFn, String, ParamSpec));
```

jf2048 (changes requested):

glib/src/gobject/binding_group.rs
jf2048: Sorry not this one, it's in another spot. See my other review

glib/src/gobject/binding_group.rs
A6GibKm: I think it does, but i have crashes locally on smething else

sdroege: The problem there was that you pass a `*const u8` and not a `String`. And that `*const u8` would even be freed already in various cases.

glib/src/gobject/binding_group.rs
```diff
@@ -140,19 +140,26 @@ impl<'a> BindingGroupBuilder<'a> {
...
                 Some(res) => {
+                    let pspec_name = transform_data.2.clone();
+                    let source = binding.source().unwrap();
+                    let pspec = source.find_property(&pspec_name);
+                    assert!(pspec.is_some(), "Source object does not have a property {}", pspec_name);
+                    let pspec = pspec.unwrap();
+
                     assert!(
-                        res.type_().is_a(transform_data.2.value_type()),
+                        crate::Value::type_transformable(res.type_(), pspec.value_type()),
```
sdroege, line 159: This should stay as before. We don't want to loosen the type checks, as discussed.

glib/src/gobject/binding_group.rs
```diff
@@ -233,8 +249,197 @@ impl<'a> BindingGroupBuilder<'a> {
...
+                Some(format!("{} World", value).to_value())
+            })
+            .build();
+
+        source.set_name("Hello");
+        assert_eq!(target.name(), "Hello World");
+    }
+
+    #[test]
+    fn binding_from_transform() {
```
sdroege, line 322: Would be good to have transform tests where the source/target property have different types.

sdroege (commented):
Apart from those two comments this should be OK now.

glib/src/gobject/binding_group.rs
A6GibKm: give me an example of different types and ill code it, I tried using the current 2 properties (bool and string) and of course it was not possible due to the checks we do, even if we can do the conversion in the GtkExpression.

glib/src/gobject/binding_group.rs
A6GibKm: Done.

sdroege (approved):

glib/src/gobject/binding_group.rs
sdroege: See https://github.com/gtk-rs/gtk-rs-core/issues/700

PR #666, opened 2022-04-30
 glib: Add WeakRefNotify (v2)

AaronErhardt:
Successor to https://github.com/gtk-rs/gtk-rs-core/pull/429

I hope all concerns have been addressed :)

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -2204,13 +2204,13 @@ pub trait ObjectExt: ObjectType {
     /// Add a callback to be notified when the Object is disposed.
     #[doc(alias = "g_object_weak_ref")]
     #[doc(alias = "connect_drop")]
-    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F)
+    fn add_weak_ref_notify<F: FnOnce() + Send + 'static>(&self, f: F)
         -> WeakRefNotify<Self>;
 
     // rustdoc-stripper-ignore-next
     /// Add a callback to be notified when the Object is disposed.
     ///
-    /// This is like [`add_weak_ref_notify`][`ObjectExt::add_weak_ref_notify`] but doesn't require the closure to be [`Send`] and [`Sync`].
+    /// This is like [`add_weak_ref_notify`][`ObjectExt::add_weak_ref_notify`] but doesn't require the closure to be [`Send`].
     /// Object dispose will panic if the object is disposed from the wrong thread.
     #[doc(alias = "g_object_weak_ref")]
     #[doc(alias = "connect_drop")]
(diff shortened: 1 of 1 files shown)


glib/src/object.rs
```diff
@@ -2199,6 +2200,22 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    #[doc(alias = "connect_drop")]
+    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F)
```
sdroege, line 2207: @bilelmoussaoui @GuillaumeGomez Do you maybe have a better suggestion for the name? The C API is called `g_object_weak_ref()` but that's even worse :)

glib/src/object.rs
```diff
@@ -3231,6 +3248,16 @@ impl<T: ObjectType> ObjectExt for T {
         }
     }
 
+    fn add_weak_ref_notify<F: FnOnce() + Send + Sync + 'static>(&self, f: F) -> WeakRefNotify<T> {
```
sdroege, line 3251:
```suggestion
    fn add_weak_ref_notify<F: FnOnce() + Send + 'static>(&self, f: F) -> WeakRefNotify<T> {
```

Is actually enough here. It's an `FnOnce`, it's not accessed from multiple threads at the same time

glib/src/object.rs
```diff
@@ -2199,6 +2200,22 @@ pub trait ObjectExt: ObjectType {
     /// Downgrade this object to a weak reference.
     fn downgrade(&self) -> WeakRef<Self>;
 
+    // rustdoc-stripper-ignore-next
+    /// Add a callback to be notified when the Object is disposed.
+    #[doc(alias = "g_object_weak_ref")]
+    #[doc(alias = "connect_drop")]
```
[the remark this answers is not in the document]
AaronErhardt: I thought it might be nice to have something that someone might type into the search bar without any knowledge of glib. Maybe someone has a better idea for such a name.

sdroege (commented):
Seems good to me otherwise, thanks!

sdroege (approved):
Looks good to me apart from the naming but I don't have any better suggestions!

So unless @bilelmoussaoui or @GuillaumeGomez have any better suggestions let's get this in?

sdroege (dismissed):

sdroege (approved):

sdroege (approved):

glib/src/object.rs
GuillaumeGomez: `add_weak_ref_callback`?

sdroege: The current one is more descriptive than that: notification vs. callback is clearer

sdroege:
I guess let's go with this then.

Issue #204, opened 2021-07-22
Implement IntoIterator for all ListModels


aledomu:
Just wondering if `IntoIterator` could be implemented for any type that is a `ListModel`, so we get access to all the iterator goodies of Rust. Something like this:

```rust
struct ListModelIter<'a, T: IsA<ListModel>> {
    index: u32,
    list_model: &'a T,
}

impl<'a, T: IsA<ListModel>> Iterator for ListModelIter<'a, T> {
    type Item = Object;

    fn next(&mut self) -> Option<Self::Item> {
        let item = self.list_model.item(self.index);
        self.index += 1;

        item
    }

    // Maybe implement optimized versions of other trait methods.
}

impl<I: IsA<ListModel>> IntoIterator for &'a I {
    type Item = Self::IntoIter::Item;
    // We can avoid creating this data structure by calling `iter::from_fn` instead,
    // but we'd lose any ability to optimize any trait methods.
    type IntoIter = ListModelIter<'a>;

    fn into_iter(self) -> Self::IntoIter {
        ListModelIter {
            index: 0,
            list_model: self,
        }
    }
}
```

GuillaumeGomez:
So, just had an idea about this: we could go around the unsafe iteration issue by creating an unsafe method: `fn iter() -> IteratorOverListModel`. Like that, people are forced to know that using this iterator is at their own risk. :)

bilelmoussaoui:
This can be closed

PR #668, opened 2022-05-01
impl fallible IntoIterator for ListModel,ListStore

ranfdev:
It's me with another take on #204.

### DIFF
--- a/gio/src/list_model.rs
+++ b/gio/src/list_model.rs
@@ -40,10 +40,10 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
         let changed = Rc::new(Cell::new(false));
 
         let changed_clone = changed.clone();
-        let signal_id = Cell::new(Some(self.connect_items_changed(move |_, pos, _, _| {
+        let signal_id = Some(self.connect_items_changed(move |_, pos, _, _| {
             let old = changed_clone.get();
             changed_clone.replace(old || pos < len);
-        })));
+        }));
 
         Ok(ListModelIter {
             ty: Default::default(),
@@ -74,7 +74,7 @@ pub struct ListModelIter<'a, T: IsA<glib::Object>> {
     reverse_pos: u32,
     model: &'a ListModel,
     changed: Rc<Cell<bool>>,
-    signal_id: Cell<Option<SignalHandlerId>>,
+    signal_id: Option<SignalHandlerId>,
 }
 impl<'a, T: IsA<glib::Object>> Iterator for ListModelIter<'a, T> {
     type Item = Result<T, ListModelMutatedDuringIter>;
--- a/gio/src/list_model.rs
+++ b/gio/src/list_model.rs
@@ -62,6 +62,7 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
 #[error("the list model was mutated during iteration")]
 pub struct ListModelMutatedDuringIter;
 
+// rustdoc-stripper-ignore-next
 /// Iterator of `ListModel`'s items.
 /// This iterator will always give `n = initial_model.n_items()` items, even if the `ListModel`
 /// is mutated during iteration.
--- a/gio/src/list_model.rs
+++ b/gio/src/list_model.rs
@@ -44,7 +44,8 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
 
         let changed_clone = changed.clone();
         let signal_id = Cell::new(Some(self.connect_items_changed(move |_, pos, _, _| {
-            changed_clone.replace(pos < len);
+            let old = changed_clone.get();
+            changed_clone.replace(old || pos < len);
         })));
 
         Ok(ListModelIter {
@@ -164,14 +165,26 @@ fn list_model_iter_err() {
     let m1 = crate::Menu::new();
     let m2 = crate::Menu::new();
     let m3 = crate::Menu::new();
+    let m4 = crate::Menu::new();
 
     list.append(&m1);
     list.append(&m2);
     list.append(&m3);
+    list.append(&m4);
 
     let mut iter = list.iter::<crate::Menu>().unwrap();
-    assert_eq!(iter.next_back(), Some(Ok(m3)));
-    list.remove_all();
+    assert_eq!(iter.next_back(), Some(Ok(m4)));
+
+    // These two don't affect the iter
+    list.append(&m2);
+    list.append(&m2);
+
+    assert_eq!(iter.next(), Some(Ok(m1)));
+
+    // Does affect the iter
+    list.remove(2);
+    // Doesn't affect the iter, but the iter should stay tainted.
+    list.remove(4);
     assert_eq!(iter.next(), Some(Err(ListModelMutatedDuringIter)));
     assert_eq!(iter.next(), Some(Err(ListModelMutatedDuringIter)));
     // Returned n items
--- a/gio/src/list_model.rs
+++ b/gio/src/list_model.rs
@@ -1,6 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use std::cell::Cell;
+use std::iter::FusedIterator;
 use std::marker::PhantomData;
 use std::rc::Rc;
 
@@ -17,7 +18,7 @@ pub trait ListModelExtManual: Sized {
     fn snapshot(&self) -> Vec<glib::Object>;
 
     // rustdoc-stripper-ignore-next
-    /// If `T::static_type() == self.item_type()` then it returns an iterator over the `ListModel` elements,
+    /// If `T::static_type().is_a(self.item_type())` then it returns an iterator over the `ListModel` elements,
     /// else the types are not compatible and returns `Err(&Self)`.
     fn iter<T: IsA<glib::Object>>(&self) -> Result<ListModelIter<T>, &Self>;
 }
@@ -34,7 +35,7 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
     where
         Self: Sized + Clone,
     {
-        if self.item_type() != LT::static_type() {
+        if !self.item_type().is_a(LT::static_type()) {
             return Err(self);
         }
         let rc = Rc::new(Cell::new(false));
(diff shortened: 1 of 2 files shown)


ranfdev:
My only question is: once the error `ListModelMutatedDuringIter` is returned, when should the iterator stop? Should we keep returning `Some(Err(...))` for the remaining items (`initial_model_size - current_i`), ad infinitum, or just one time?

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+}
+impl<T: IsA<glib::Object>> Iterator for ListModelIter<T> {
+    type Item = Result<T, ListModelMutatedDuringIter>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let res = match self.changed.get() {
+            true => Some(Err(ListModelMutatedDuringIter)),
+            false => self
+                .model
+                .item(self.i)
```
sdroege, line 76: This can't possibly return `None` here so I would `unwrap()` / `expect()` right after `.item()` the first time

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+                .map(|x| Ok(x.downcast::<T>().unwrap())),
```
sdroege, line 77: ... and then a second time after `downcast()` because that can't really fail either

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+    i: u32,
+    model: ListModel,
+    changed: Rc<Cell<bool>>,
+    signal_id: Cell<Option<SignalHandlerId>>,
+}
+impl<T: IsA<glib::Object>> Iterator for ListModelIter<T> {
+    type Item = Result<T, ListModelMutatedDuringIter>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let res = match self.changed.get() {
```
sdroege, line 72: You should get the number of items in the model when creating the iterator and then work based on that. The number of items can't change duration iteration, and by doing that you can implement
  - `Iterator::size_hint()` and `ExactSizeIterator`
  - `FusedIterator`
  - `DoubleEndedIterator`

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
         }
```
sdroege, line 30: You can implement ^---- this one via `self.iter().collect::<Result<Vec<T>, _>>>().unwrap()` now

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
         res
     }
+    fn iter<LT: IsA<glib::Object>>(&self) -> Result<ListModelIter<LT>, &Self>
+    where
+        Self: Sized + Clone,
+    {
+        if self.item_type() != LT::static_type() {
```
sdroege, line 37:
```suggestion
        if !self.item_type().is_a(LT::static_type()) {
```

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+
+        let rcc = rc.clone();
+        let signal_id = Cell::new(Some(self.connect_items_changed(move |_, _, _, _| {
+            rcc.replace(true);
+        })));
+
+        Ok(ListModelIter {
+            ty: Default::default(),
+            i: 0,
+            model: self.clone().upcast(),
```
sdroege, line 50: Why not just store a reference in here instead of cloning?

gio/src/list_model.rs
```diff
@@ -19,15 +30,86 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+        res
+    }
+}
+impl<T: IsA<glib::Object>> Drop for ListModelIter<T> {
+    fn drop(&mut self) {
+        self.model.disconnect(self.signal_id.take().unwrap());
+    }
 }
 
 impl std::iter::IntoIterator for ListModel {
```
sdroege, line 89:
```suggestion
impl std::iter::IntoIterator for &ListModel {
```

makes more sense as you don't actually consume the model (compare: `IntoIterator` on `Vec` and `&Vec`).

glib/src/object.rs
```diff
@@ -2263,16 +2263,13 @@ impl<T: ObjectType> ObjectExt for T {
...
-            Some(pspec) => pspec,
-            None => {
-                return Err(bool_error!(
-                    "property '{}' of type '{}' not found",
-                    property_name,
-                    self.type_()
-                ));
-            }
-        };
+        let pspec = self.find_property(property_name).ok_or_else(|| {
```
sdroege, line 2266: Can you put all this cleanup into a separate PR? That's going to be merged faster and keeps this PR smaller then :)

sdroege (commented):
The general idea of having an iterator that returns a `Result` seems good to me. That seems like the simplest and most sensible implementation for how people would actually use this.

gio/src/list_model.rs
ranfdev: because `upcast` consumes `self`:
`pub fn upcast<T>(self)`

gio/src/list_model.rs
ranfdev: I'd keep `snapshot` as it is, because now it's faster and the implementation is really small anyway

gio/src/list_model.rs
sdroege: You could do `model: self.upcast_ref(),` and just store a reference. That also makes it less strange as iterators usually keep a normal reference to their underlying data

gio/src/list_model.rs
```diff
@@ -19,15 +31,149 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
         }
         res
     }
+    fn iter<LT: IsA<glib::Object>>(&self) -> Result<ListModelIter<LT>, &Self>
+    where
+        Self: Sized + Clone,
+    {
+        if !self.item_type().is_a(LT::static_type()) {
+            return Err(self);
```
sdroege, line 39: Would be nicer to return a proper error type here. Returning `Self` only really makes sense for functions that take it by value

gio/src/list_model.rs
```diff
@@ -19,15 +31,149 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+        let n = (self.reverse_pos - self.i) as usize;
+        (n as usize, Some(n))
+    }
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.i >= self.reverse_pos {
+            return None;
+        }
+        let res = match self.changed.get() {
+            true => Err(ListModelMutatedDuringIter),
```
sdroege, line 93: You probably want to return `None` forever after returning `Err` once

gio/src/list_model.rs
```diff
@@ -19,15 +31,149 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+/// will return `Some(Err(...))` for the remaining items.
+/// Mutations to the `ListModel` in position >= `initial_model.n_items()` are allowed.
+pub struct ListModelIter<T: IsA<glib::Object>> {
+    ty: PhantomData<T>,
+    i: u32,
+    // it's > i when valid
+    reverse_pos: u32,
+    model: ListModel,
+    changed: Rc<Cell<bool>>,
+    signal_id: Cell<Option<SignalHandlerId>>,
```
sdroege, line 79:
```suggestion
    signal_id: Option<SignalHandlerId>,
```

Doesn't have to be a cell. You only mutate it when you have a mutable reference anyway

gio/src/list_model.rs
```diff
@@ -19,15 +31,149 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+        if !self.item_type().is_a(LT::static_type()) {
+            return Err(self);
+        }
+
+        let len = self.n_items();
+        let changed = Rc::new(Cell::new(false));
+
+        let changed_clone = changed.clone();
+        let signal_id = Cell::new(Some(self.connect_items_changed(move |_, pos, _, _| {
+            changed_clone.replace(pos < len);
```
sdroege, line 47:
```suggestion
            let old = changed_clone.get();
            changed_clone.set(old || pos < len);
```

gio/src/list_model.rs
```diff
@@ -19,15 +31,149 @@ impl<T: IsA<ListModel>> ListModelExtManual for T {
...
+}
+impl<T: IsA<glib::Object>> Iterator for ListModelIter<T> {
+    type Item = Result<T, ListModelMutatedDuringIter>;
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let n = (self.reverse_pos - self.i) as usize;
+        (n as usize, Some(n))
+    }
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.i >= self.reverse_pos {
```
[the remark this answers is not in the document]
ranfdev: They both work equally well. In this case I read `==` as "If both are equal, return early. The code below shouldn't be executed".
I prefer `>=` because it also says "The code below should be executed ONLY if `self.i < self.reverse_pos`".

But I see why you may like `==`: "`self.i` can't never be greater than `self.reverse_pos`, so why check?"

Whatever... :)

gio/src/list_model.rs
ranfdev: Wouldn't this break the trait `ExactIterator`? I can't return `None` earlier than expected. I'm ok with returning `None`, but now while `ExactIterator` is implemented

gio/src/list_model.rs
ranfdev: True, but this keeps `iter` similar to `downcast`. This type check feels similar to what downcast is doing. I know downcast returns `Err(Self)`... I try to return the most similar thing I can

gio/src/list_model.rs
sdroege: That's a good point, yes you're right :)

gio/src/list_model.rs
sdroege: `downcast_ref()` returns an `Option` :) The only reason why `downcast()` returns `Self` in the error case is because it takes `Self` by value and it would otherwise be gone on error.

ranfdev: > The only reason why downcast() returns Self in the error case is because it takes Self by value and it would otherwise be gone on error.

I hadn't thought of that 😅

sdroege (approved):

PR #670, opened 2022-05-04
Move `gio::compile_resources` to its own crate

### LINKED ISSUE #574: [FEATURE REQUEST] Move `gio::compile_resources` to a seperate crate
Currently my build script exists just to run this function, and that requires depending on the entirety of glib and gio for the build script (build scripts do _not_ share dependencies with the rest of your project). Additionally, it doesn't really make sense to need to update the gio version for the build script to keep in sync with your project's gio, when nothing with this one function changed.

I think it would make sense to split this into a separate crate. As far as I can tell, the function doesn't depend on anything from glib/gio, just the rust stdlib.

RealKC:
Closes #574

---

Some things to consider:

* naming: Should we name this `gio-build-helpers`, `glib-build-helpers` or something else entirely? This is (to me) made confusing by the fact that the executable is named `glib-compile-resources` but the resources machinery lives in Gio
* versioning: Should this start from `0.1.0` or be in sync with the other crates? I don't think this would need version bumps at the same time as the other crates do, so I imagine it would end up desyncing sooner than later anyway

NOTE: If this PR is alright, I'll send PRs to gtk4-rs and gtk3-rs to move the examples to this crate


RealKC:
I think I need to update the CI workflows too? Not sure what I need to do exactly though

sdroege:
This will either have to be moved to a separate crate or be removed completely, but keeping it inside the gio crate does not make much sense.

sdroege:
@Hofer-Julian mentioned that the book is using this before introducing meson, so it shouldn't go away at least

gio-build-helpers/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "gio-build-helpers"
```
sdroege, line 2: Can you rename it to `gio-macros`?

gio-build-helpers/LICENSE
```diff
@@ -0,0 +1,18 @@
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
```
sdroege, line 2: Just make this a symlink like the others

glib-build-tools/README.md
```diff
@@ -0,0 +1,41 @@
...
+Currently, the minimum supported Rust version is `1.57.0`.
+
+## Documentation
+
+* [Stable API](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib-build-tools)
+* [Development API](https://gtk-rs.org/gtk-rs-core/git/latest/docs/glib-build-tools)
+
+## Using
+
+We recommend using [crates from crates.io](https://crates.io/keywords/gtk-rs) like this:
```
bilelmoussaoui, line 16: Is any of this relevant to this crate? I don't think so as It doesn't link to glib/gio at all

RealKC: fair point, I simply used an existing readme as template, should I just remove the entire section then?

RealKC: I've removed it

sdroege (approved):

PR #671, opened 2022-05-04
cairo+gio+glib: Add safety docs for low hanging fruits

RealKC:
ref #646


cairo/src/font/font_face.rs
```diff
@@ -55,7 +55,11 @@ impl FontFace {
         Ok(font_face)
     }
 
-    // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
+    // rustdoc-stripper-ignore-next
+    /// Creates a new font face for the FreeType backend from an already opened FreeType face.
+    ///
+    /// Safety: the FT_Face must be valid and not be freed until the last reference to the
```
sdroege, line 61:
```suggestion
    /// Safety: the `FT_Face` must be valid and not be freed until the last reference to the
```

cairo/src/font/font_face.rs
```diff
@@ -71,7 +75,12 @@ impl FontFace {
         Ok(font_face)
     }
 
-    // Safety: the FT_Face must be valid and not be freed until the `FontFace` is dropped.
+    // rustdoc-stripper-ignore-next
+    /// Creates a new font face for the FreeType backend from an already opened FreeType face,
+    /// additionally allowing you to pass flags to the underlying C API.
+    ///
+    /// Safety: the FT_Face must be valid and not be freed until the last reference to the
```
sdroege, line 82:
```suggestion
    /// Safety: the `FT_Face` must be valid and not be freed until the last reference to the
```

cairo/src/utils.rs
```diff
@@ -4,6 +4,13 @@ use crate::error::Error;
 use std::ffi::CStr;
 use std::fmt;
 
+// rustdoc-stripper-ignore-next
+/// Resets all static data within cairo to its original state (i.e. identical to the state at program
+/// invocation). For example, all caches within cairo will be flushed empty.
+///
+/// # Safety
+/// It is only safe to call this function when there are no active cairo objects remaining (the
+/// appropriate destroy functions have been called as necessary).
```
sdroege, line 13: There are other `destroy` functions? Or do you mean the objects need to be dropped?

Also I assume this must not be called simultaneously from multiple threads?

gio/src/unix_input_stream.rs
```diff
@@ -39,7 +49,13 @@ pub trait UnixInputStreamExtManual: Sized {
     #[doc(alias = "g_unix_input_stream_get_fd")]
     #[doc(alias = "get_fd")]
     fn fd<T: FromRawFd>(&self) -> T;
-    #[allow(clippy::missing_safety_doc)]
+
+    // rustdoc-stripper-ignore-next
+    /// Sets whether the fd of this stream will be closed when the stream is closed.
+    ///
+    /// # Safety
+    /// If you pass in `false` as the parameter, you may only clsose the fd if the all references
```
sdroege, line 57:
```suggestion
    /// If you pass in `false` as the parameter, you may only close the fd if the all references
```

gio/src/unix_output_stream.rs
```diff
@@ -38,6 +50,12 @@ pub trait UnixOutputStreamExtManual: Sized {
     #[doc(alias = "get_fd")]
     fn fd<T: FromRawFd>(&self) -> T;
 
+    // rustdoc-stripper-ignore-next
+    /// Sets whether the fd of this stream will be closed when the stream is closed.
+    ///
+    /// # Safety
+    /// If you pass in `false` as the parameter, you may only clsose the fd if the all references
```
sdroege, line 57:
```suggestion
    /// If you pass in `false` as the parameter, you may only close the fd if the all references
```

gio/src/win32_input_stream.rs
```diff
@@ -40,17 +40,27 @@ impl fmt::Display for Win32InputStream {
 impl Win32InputStream {
     pub const NONE: Option<&'static Win32InputStream> = None;
 
+    // rustdoc-stripper-ignore-next
+    /// Creates a new [`Self`] that takes ownership of the passed in handle.
+    ///
+    /// # Safety
+    /// You must not close the fd unless you've previously called [`Win32InputStreamExtManual::set_close_fd`]
```
sdroege, line 47:
```suggestion
    /// You must not close the handle unless you've previously called [`Win32InputStreamExtManual::set_close_fd`]
```

gio/src/win32_input_stream.rs
```diff
@@ -70,8 +80,13 @@ pub trait Win32InputStreamExtManual: Sized {
     #[doc(alias = "get_handle")]
     fn handle<T: FromRawHandle>(&self) -> T;
 
+    // rustdoc-stripper-ignore-next
+    /// Sets whether the handle of this stream will be closed when the stream is closed.
+    ///
+    /// # Safety
+    /// If you pass in `false` as the parameter, you may only clsose the handle if the all references
```
sdroege, line 87:
```suggestion
    /// If you pass in `false` as the parameter, you may only close the handle if the all references
```

gio/src/win32_output_stream.rs
```diff
@@ -40,17 +40,27 @@ impl fmt::Display for Win32OutputStream {
 impl Win32OutputStream {
     pub const NONE: Option<&'static Win32OutputStream> = None;
 
+    // rustdoc-stripper-ignore-next
+    /// Creates a new [`Self`] that takes ownership of the passed in handle.
+    ///
+    /// # Safety
+    /// You must not close the fd unless you've previously called [`Win32InputStreamExtManual::set_close_fd`]
```
sdroege, line 47:
```suggestion
    /// You must not close the handle unless you've previously called [`Win32InputStreamExtManual::set_close_fd`]
```

gio/src/win32_output_stream.rs
```diff
@@ -70,8 +80,13 @@ pub trait Win32OutputStreamExtManual: Sized {
     #[doc(alias = "get_handle")]
     fn handle<T: FromRawHandle>(&self) -> T;
 
+    // rustdoc-stripper-ignore-next
+    /// Sets whether the handle of this stream will be closed when the stream is closed.
+    ///
+    /// # Safety
+    /// If you pass in `false` as the parameter, you may only clsose the handle if the all references
```
sdroege, line 87:
```suggestion
    /// If you pass in `false` as the parameter, you may only close the handle if the all references
```

glib/src/log.rs
```diff
@@ -1006,6 +1006,16 @@ pub fn log_writer_default(log_level: LogLevel, fields: &[LogField<'_>]) -> LogWr
     }
 }
 
+// rustdoc-stripper-ignore-next
+/// Sets whether Glib log functions output to stderr or stdout.
```
sdroege, line 1010:
```suggestion
/// Sets whether GLib log functions output to stderr or stdout.
```

glib/src/log.rs
```diff
@@ -1006,6 +1006,16 @@ pub fn log_writer_default(log_level: LogLevel, fields: &[LogField<'_>]) -> LogWr
...
+///
+/// By default, log messages of level [`LogLevel::Info`] and [`LogLevel::Debug`] are sent to stdout,
+/// and other log messages are sent to stderr. Passing `true` will send all messages to stderr.
+///
+/// # Safety
+///
+/// This function sets global state and is not thread-aware, as such it should be called before any
+/// threads may try to use Glib logging.
```
sdroege, line 1018:
```suggestion
/// threads may try to use GLib logging.
```

cairo/src/font/font_face.rs
bilelmoussaoui, line 61: While you are doing this, i will keep those consistent. 
```suggestion
    /// # Safety
    /// The FT_Face must be valid and not be freed until the last reference to the
```

cairo/src/utils.rs
RealKC: Got too immersed in the Cairo C docs :sweat_smile: 

> Also I assume this must not be called simultaneously from multiple threads?

The docs don't mention anything about this, but I looked at the source and it seems to use a mutex, I think it's safe to expose this detail in our documentation since just not using that mutex anymore would be a breaking change. We could just be more overly restrictive and tell people to not do this though

sdroege (approved):

PR #672, opened 2022-05-06
cairo: Add missing User Fonts

songww:


GuillaumeGomez:
Only adding the sys calls, not their bindings?

songww:
> 

That's all I noticed, what else am I missing?

sdroege:
It would make sense to also add safe bindings for these, not just the FFI bindings.

cairo/src/font/user_fonts.rs
```diff
@@ -0,0 +1,205 @@
...
+        let font_face = unsafe { FontFace::from_raw_full(ffi::cairo_user_font_face_create()) };
+        let status = unsafe { ffi::cairo_font_face_status(font_face.to_raw_none()) };
+        status_to_result(status)?;
+        Ok(font_face)
+    }
+
+    #[doc(alias = "cairo_user_font_face_set_init_func")]
+    pub fn set_init_func<F>(&self, func: F)
+    where
+        F: Fn(&mut ScaledFont, &Context) -> Result<FontExtents, Error> + Send + Sync + 'static,
```
sdroege, line 21: Does this have to be a mutable reference? All the API on `ScaledFont` is using `&self`

cairo/src/font/user_fonts.rs
```diff
@@ -0,0 +1,206 @@
...
+            extents: *mut ffi::cairo_font_extents_t,
+        ) -> ffi::cairo_status_t {
+            let init_func = INIT_FUNC.get().unwrap();
+            match init_func(
+                &mut ScaledFont::from_raw_none(scaled_font),
+                &Context::from_raw_none(cr),
+            ) {
+                Err(err) => err.into(),
+                Ok(font_extents) => {
+                    *extents = std::mem::transmute(font_extents);
```
sdroege, line 41:
```suggestion
                    *extents = font_extents.0;
```

Don't use `transmute` if it can be avoided

cairo/src/font/user_fonts.rs
```diff
@@ -0,0 +1,206 @@
...
+        ) -> ffi::cairo_status_t {
+            let render_glyph_func = RENDER_GLYPH_FUNC.get().unwrap();
+            match render_glyph_func(
+                &mut ScaledFont::from_raw_none(scaled_font),
+                glyph,
+                &Context::from_raw_none(cr),
+            ) {
+                Err(err) => err.into(),
+                Ok(text_extents) => {
+                    *extents = std::mem::transmute(text_extents);
```
sdroege, line 76:
```suggestion
                    *extents = text_extents.0;
```

cairo/src/font/user_fonts.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
+    #[doc(alias = "cairo_user_font_face_set_render_glyph_func")]
+    pub fn set_render_glyph_func<F>(&self, func: F)
+    where
+        F: Fn(&mut ScaledFont, u64, &Context) -> Result<TextExtents, Error> + Send + Sync + 'static,
+    {
+        static RENDER_GLYPH_FUNC: once_cell::sync::OnceCell<
+            Box<dyn Fn(&mut ScaledFont, u64, &Context) -> Result<TextExtents, Error> + Send + Sync>,
```
sdroege, line 57: Doesn't have to be mutable AFAICS

cairo/src/font/user_fonts.rs
```diff
@@ -0,0 +1,206 @@
...
+                self.to_raw_none(),
+                Some(render_glyph_trampoline),
+            );
+        }
+    }
+
+    #[doc(alias = "cairo_user_font_face_set_unicode_to_glyph_func")]
+    pub fn set_unicode_to_glyph_func<F>(&self, func: F)
+    where
+        F: Fn(&mut ScaledFont, u64) -> Result<u64, Error> + Send + Sync + 'static,
```
sdroege, line 92: Doesn't have to be mutable AFAICS

cairo/src/font/user_fonts.rs
```diff
@@ -0,0 +1,206 @@
...
+                Some(unicode_to_glyph_trampoline),
+            );
+        }
+    }
+
+    #[doc(alias = "cairo_user_font_face_set_text_to_glyph_func")]
+    pub fn set_text_to_glyph_func<F>(&self, func: F)
+    where
+        F: Fn(
+                &mut ScaledFont,
```
sdroege, line 126: Doesn't have to be mutable AFAICS

cairo/Cargo.toml
```diff
@@ -27,7 +27,8 @@ freetype = ["ffi/freetype", "freetype-crate"]
 script = ["ffi/script"]
 xcb = ["ffi/xcb"]
 xlib = ["ffi/xlib"]
-dox = ["ffi/dox", "glib/dox", "freetype"]
+dox = ["ffi/dox", "glib/dox", "freetype", "user-fonts"]
+user-fonts = ["ffi/user-fonts", "once_cell"]
```
sdroege, line 31: Isn't this always available in cairo? In that case it would make sense to not have a feature for that but always include it

sdroege (commented):
Looks mostly good to me, thanks a lot!

cairo/src/font/user_fonts.rs
songww: .0 is private, can not access at there.

songww: Dose this could work?
```
*extents = font_extents as _;
```

sdroege: It would make sense to add an `impl From<FontExtends> for ffi::cairo_font_extends_t`

sdroege (dismissed):

sdroege (approved):

Issue #50, opened 2017-05-25
ToGlibPtrMove trait and "move" function parameter configuration


sdroege:
It would be good to have a ToGlibPtrMove, that consumes the Rust wrapper struct and just returns the C pointer. Without any copying or change of reference count (if possible: e.g. for strings it is not).

In addition to that, a function parameter configuration to declare that parameter as "move" (i.e. using ↑, taking the Rust wrapper per value instead of as a reference) would be useful.

GuillaumeGomez:
Now remains to add the gir parameter.

bilelmoussaoui:
It would be nice to also implement `ToGlibPtrMove` for shared types as currently it is implemented for all the types generated with glib::wrapper! except shared ones.

sdroege:
What's also missing is to do this automatically for non-refcounted types on `transfer full` parameters (and configurably for other types too, I guess: I want this for various refcounted types in GStreamer).

sdroege:
> It would be nice to also implement `ToGlibPtrMove` for shared types as currently it is implemented for all the types generated with glib::wrapper! except shared ones.

See https://github.com/gtk-rs/gtk-rs-core/pull/756

bilelmoussaoui:
Now that we have landed the automatically generated changes, the latest remaining step would be having a look at all the manual code and see where IntoGlibPtr could be used

sdroege:
This has happened now.

PR #673, opened 2022-05-06
Add ToGlibPtrMove trait

GuillaumeGomez:
Part of https://github.com/gtk-rs/gtk-rs-core/issues/50.

### DIFF
--- a/glib/src/variant_type.rs
+++ b/glib/src/variant_type.rs
@@ -90,12 +90,14 @@ impl VariantType {
         VariantTy::new(&type_string)?;
 
         let len = type_string.len();
-        let ptr = type_string.into_raw();
+        unsafe {
+            let ptr = type_string.into_glib_ptr();
 
-        Ok(VariantType {
-            ptr: ptr as *mut ffi::GVariantType,
-            len,
-        })
+            Ok(VariantType {
+                ptr: ptr as *mut ffi::GVariantType,
+                len,
+            })
+        }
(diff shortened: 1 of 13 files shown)


GuillaumeGomez:
Updated.

glib/src/variant_type.rs
```diff
@@ -90,12 +90,14 @@ impl VariantType {
...
-        let ptr = type_string.into_raw();
+        unsafe {
+            let ptr = type_string.into_glib_ptr();
 
-        Ok(VariantType {
-            ptr: ptr as *mut ffi::GVariantType,
-            len,
-        })
+            Ok(VariantType {
+                ptr: ptr as *mut ffi::GVariantType,
```
sdroege, line 97: Hm why doesn't this use `ptr::NonNull`?

sdroege (commented):
Seems good to me

glib/src/variant_type.rs
GuillaumeGomez: Aren't you the one who wrote this code? :p 

If there is a problem with it, please an open an issue so it's not forgotten.

sdroege: I'll update it once this here is merged :) I didn't write it though

bilelmoussaoui (approved):

GuillaumeGomez:
Sending follow-up PRs soon.

PR #674, opened 2022-05-07
glib: Use `ptr::NonNull` for `VariantType` pointer storage

sdroege:
And add some assertions against `NULL` pointers in a few places.


glib/src/variant_type.rs
```diff
@@ -137,8 +137,7 @@ impl Deref for VariantType {
     type Target = VariantTy;
     fn deref(&self) -> &VariantTy {
         unsafe {
-            assert!(self.len > 0);
```
sdroege, line 140: This is already guaranteed by construction, just like that the stored pointer is always not `NULL`

GuillaumeGomez (approved):

Issue #484, opened 2022-01-15
Cleanup the images


bilelmoussaoui:
We should simplify the various images we use across gtk-rs here.

- Make a generic image that contains the building blocks in gtk-rs-core
- Inherit gtk3-rs/gtk4-rs from that image and only install the necessary stuff on top of it.

This would allow us to move to a common parent image (maybe fedora, or something else) and reduce the weird issues that could happen due to the differences between the base images

jf2048:
Bumping to fedora latest will bring it up to glib 2.70: https://packages.fedoraproject.org/pkgs/glib2/glib2/fedora-35.html

It would probably be good to build the latest git for each package (glib, cairo, pango, gdk-pixbuf, graphene), but I can see a problem with this: In fedora, there is a circular dependency with `cairo-devel` -> `freetype-devel` -> `harfbuzz-devel` -> `cairo`

bilelmoussaoui:
> Bumping to fedora latest

We could always use fedora rawhide for now that should give us glib 2.72 at least. We can figure try to figure out how to build glib ourselves in the future.

jf2048:
This does seem to work correctly from git if using the meson subprojects and if built in the correct order

bilelmoussaoui:
This is now done

PR #676, opened 2022-05-08
Port Dockerfile to fedora rawhide and use system deps from git

jf2048:
Fixes part of #484

This is not using xlib/xft/rsvg, but we probably need those to build the gtk-rs images later? Or we can leave the gtk-rs images and not base them on this image? The image is already pretty large as-is, around 1.6GB on my system because of all the debug symbols. Another option is to maybe use `--buildtype release`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,33 @@
-FROM ubuntu:20.10
+FROM fedora:rawhide
 
-RUN apt update -y
-RUN apt install -y \
-    libgtk-3-dev \
-    libglib2.0-dev \
-    libgraphene-1.0-dev \
-    git \
-    xvfb \
-    curl \
-    libcairo-gobject2 \
-    libcairo2-dev \
-    wget
+RUN dnf update -y && \
+    dnf install wget git meson cmake gcc gcc-c++ \
+    libpng-devel turbojpeg-devel libXext-devel libXrender-devel -y && \
+    dnf clean all -y
+
+RUN git clone https://gitlab.gnome.org/GNOME/glib.git --depth=1 && \
+    (cd /glib && \
+        meson setup builddir --prefix=/usr --buildtype release -Dtests=false && \
+        meson install -C builddir) && \
+    git clone https://gitlab.freedesktop.org/cairo/cairo.git --depth=1 && \
+    (cd /cairo && \
+        meson setup builddir --prefix=/usr --buildtype release -Dglib=enabled -Dtests=disabled && \
+        meson install -C builddir) && \
+    git clone https://github.com/harfbuzz/harfbuzz.git --depth=1 && \
+    (cd /harfbuzz && \
+        meson setup builddir --prefix=/usr --buildtype release -Dintrospection=disabled -Dtests=disabled -Ddocs=disabled && \
+        meson install -C builddir) && \
+    git clone https://gitlab.gnome.org/GNOME/pango.git --depth=1 && \
+    (cd /pango && \
+        meson setup builddir --prefix=/usr --buildtype release -Dfreetype=enabled -Dintrospection=disabled -Dxft=disabled && \
+        meson install -C builddir) && \
+    git clone https://gitlab.gnome.org/GNOME/gdk-pixbuf.git --depth=1 && \
+    (cd /gdk-pixbuf && \
+        meson setup builddir --prefix=/usr --buildtype release -Dintrospection=disabled -Dinstalled_tests=false -Dgio_sniffing=false && \
+        meson install -C builddir) && \
+    git clone https://github.com/ebassi/graphene.git --depth=1 && \
+    (cd /graphene && \
+        meson setup builddir --prefix=/usr --buildtype release -Dintrospection=disabled -Dtests=false -Dinstalled_tests=false && \
+        meson install -C builddir) && \
+    rm -rf /glib /cairo /harfbuzz /pango /gdk-pixbuf /graphene
+
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -8,15 +8,18 @@ on:
 
 jobs:
   docker:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Push to GitHub Packages
-        uses: docker/build-push-action@v1
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
         with:
-          push: true
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
           registry: ghcr.io
-          repository: ${{ github.repository }}/core
-          tags: latest
+      - name: Push to GitHub Packages
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          allow: security.insecure
+          tags: ${{ github.repository }}/core:latest


GuillaumeGomez (approved):

bilelmoussaoui:
>  all the debug symbols.

I don't think we should include the debug symbols at it is mostly meant to be used for CI, I would go with a release build type here.

bilelmoussaoui:
> This is not using xlib/xft/rsvg, but we probably need those to build the gtk-rs images later?

I don't think we need those, do we ? at least rsvg is a runtime dependency for loading svg icons, so it is fine to not include it.

jf2048:
The tests in this repository don't use them, but gtk will need them and a lot of other packages: https://gitlab.gnome.org/GNOME/gtk/-/blob/main/.gitlab-ci/fedora.Dockerfile

bilelmoussaoui:
Do they? they are only useful for GTK_DEBUG or inspector related stuff. For bindings CI, I don't think so

jf2048:
It looks like rsvg is not needed, but xlib is needed to run those gtk tests in xvfb. I got that to work, but building gtk using this image now fails with some errors about missing harfbuzz symbols, because of a missing pkg config require: https://gitlab.gnome.org/GNOME/pango/-/blob/a2ccd36a42e039d3600b04fe37fdc47f267d90c7/pango/meson.build#L173-176

jf2048:
I put harfbuzz in this to stop gtk from failing. In a follow up PR we can bump the versions in CI.yml

GuillaumeGomez (approved):

bilelmoussaoui:
Looks good to me! Something we should ensure that is the same across all the images is setting the default shell to bash as otherwise the docs that executes random shell commands would fail. That is what motivated the issue #484 as the docs were failing for gtk3/gtk-rs-core but not gtk4-rs.

jf2048:
The default shell in the fedora image is bash and all the fedora containers seem to have it: https://pagure.io/fedora-kickstarts/blob/main/f/fedora-container-common.ks#_35

Let me do a full local test with the docs though

bilelmoussaoui (dismissed):
Once the image is built and published, it would be nice to enable tests for the newest features

Dockerfile
```diff
@@ -1,13 +1,53 @@
...
-    xvfb \
-    curl \
-    libcairo-gobject2 \
-    libcairo2-dev \
-    wget
+FROM fedora:rawhide
+
+RUN dnf update -y
+RUN dnf install wget git meson cmake gcc gcc-c++ \
+    libXext-devel libXrender-devel -y
```
sdroege, line 5: Shouldn't we also install libpng-dev etc so that we can also build cairo's PNG/etc backends on the CI?

Also `harfbuzz`/`pango` don't need freetype/fontconfig?

jf2048: The way this is right now, libpng, freetype and fontconfig are all built and installed as subprojects of cairo

Dockerfile
```diff
@@ -1,13 +1,53 @@
...
+
+# build glib
+RUN git clone https://gitlab.gnome.org/GNOME/glib.git --depth=1
+WORKDIR glib
+RUN meson setup builddir --prefix=/usr --buildtype release -Dtests=false
+RUN meson install -C builddir
+WORKDIR /
+RUN rm -rf glib
```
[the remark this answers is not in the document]
jf2048: I put them all in one `RUN` instruction, the size is down to 532MB now

Dockerfile
sdroege: It would seem better to take these dependencies that we don't directly care about from the distro packages, and to only build those ourselves that we a) directly bind, or b) where the distro versions are simply too old / non-existing.

Dockerfile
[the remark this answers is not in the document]
sdroege: That also works :) For `gstreamer-rs` I have one shell script per project that is installed into the image.

Dockerfile
sdroege: And now you have to hope that the different meson subprojects for the same library are building the correct and compatible version for all projects. Installing meson subprojects and sharing the results with other projects (libpng is used at least by cairo and gdk-pixbuf, for example) is super fragile and I'd rather have us build those things completely manually than depending on meson subprojects.

bilelmoussaoui: I guess we could just install whatever is available in rawhide for now and switch to building manually if needed

sdroege: It's also not like libpng, libjpeg, freetype, fontconfig, etc are projects that are moving fast. The versions we need here at least are probably 10 years old or older :)

sdroege: Why rawhide though? The versions of dependencies we need here are ancient and this Fedora release should have new enough versions for the next few years of GNOME releases.

bilelmoussaoui: I agree that there is no need to use rawhide if we are building everything we need ourselves anyway...

sdroege: Maybe what I'm writing is confusing, so again a bit clearer :)

  - We should build the libraries we directly bind ourselves, i.e. glib, cairo, graphene, gdk-pixbuf, pango, atk, gtk3/4, etc.
  - We should use system dependencies for basic system libraries like libpng, libjpeg, X11 (already done), Wayland, fontconfig/freetype, etc. instead of relying on the meson subprojects to magically match between projects
  - If system dependencies are too old (this shouldn't really be a problem for anything?) then those should ideally be built separately instead of relying on meson subprojects

Dockerfile
```diff
@@ -1,13 +1,32 @@
-FROM ubuntu:20.10
+FROM fedora:rawhide
```
alatiera, line 1: Is rawhide necessary since you bootstrap almost everything?

Dockerfile
jf2048: It can be done for libpng and libjpeg, but fontconfig and freetype cause a problem on fedora: https://github.com/gtk-rs/gtk-rs-core/issues/484#issuecomment-1120282621

jf2048: Out of all of these the only one I think is active enough to need as a subproject for gtk4 is wayland-protocols, for example:

- xdg-activation published Oct 15, 2020: [032a0bdc](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/commit/032a0bdc9b293bfa04a691f92a85ab88d4907447)
- Released in wayland-protocols on Apr 30, 2021: [1.21](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/tags/1.21)
- gtk merges support May 3, 2021: [7b735b88](https://gitlab.gnome.org/GNOME/gtk/-/commit/7b735b88b4a1d47cdb65e4c23ef5a8dfa70d3ff6)
- rawhide gets the 1.21 update on May 5, 2021: [wayland-protocols-devel](https://packages.fedoraproject.org/pkgs/wayland-protocols/wayland-protocols-devel/fedora-35.html)
- The release channel only gets 1.23 on November 2, 2021 in [f35](https://docs.fedoraproject.org/en-US/fedora/f35/release-notes/)

Dockerfile
jf2048: Hmm, it seems so, on fedora:latest it fails due to outdated meson: https://github.com/jf2048/gtk-rs-core/runs/6371974889#step:3:1407

alatiera: Could install meson manually and avoid rawhide

bilelmoussaoui: The image is only built manually, I wouldn't bother with that now

PR #678, opened 2022-05-10
Implement `FusedIterator` for various custom iterators

sdroege:


bilelmoussaoui (dismissed):

sdroege:
@bilelmoussaoui @GuillaumeGomez Any opinions on this? :)

GuillaumeGomez (approved):

GuillaumeGomez:
It extends the API so seems like a good idea to me.

sdroege:
> It extends the API so seems like a good idea to me.

Not really. These functions have default implementations but we can implement them much more efficiently in our case.

GuillaumeGomez:
Then it's even better!

PR #679, opened 2022-05-10
cairo: Fix user font callbacks.

songww:


cairo/src/font/user_fonts.rs
```diff
@@ -41,12 +45,14 @@ impl UserFontFace {
             extents: *mut ffi::cairo_font_extents_t,
         ) -> ffi::cairo_status_t {
             let init_func = INIT_FUNC.get().unwrap();
+            let mut font_extents = FontExtents::from(*extents);
```
sdroege, line 48:
```suggestion
            let font_extents = &mut *(extents as *mut FontExtents);
```

Would also work FWIW (because `#[repr(transparent)]`). No need to copy it around

cairo/src/font/font_extents.rs
```diff
@@ -27,6 +27,26 @@ impl FontExtents {
     pub fn max_y_advance(&self) -> f64 {
         self.0.max_y_advance
     }
+
+    pub fn set_ascent(&mut self, ascent: f64) {
```
sdroege, line 31: Can you move the addition of the setters into a separate commit (in the same PR)?

sdroege (approved):

sdroege:
Thanks!

PR #685, opened 2022-05-13
gio: Use guard objects for `Application::hold()` and `mark_busy()`

sdroege:
These would automatically release/unmark_busy the application once
they're dropped and ensure that the release/unmark_busy operation is
called exactly the number of times the hold/mark_busy operation was
called.

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -284,6 +284,26 @@ generate_builder = true
     name = "run"
     manual = true
     doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "hold"
+    # Returns a value that releases on Drop.
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "release"
+    # Implemented via Drop on ApplicationHoldGuard.
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "mark_busy"
+    manual = true
+    # Returns a value that unmarks-busy on Drop.
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "unmark_busy"
+    # Implemented via Drop on ApplicationBusyGuard.
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
 
 [[object]]
 name = "Gio.ApplicationCommandLine"
--- a/gio/src/application.rs
+++ b/gio/src/application.rs
@@ -2,8 +2,7 @@
 
 use crate::Application;
 use crate::File;
-use glib::object::Cast;
-use glib::object::IsA;
+use glib::prelude::*;
 use glib::signal::{connect_raw, SignalHandlerId};
 use glib::translate::*;
 use glib::GString;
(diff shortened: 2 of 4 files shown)


gio/Gir.toml
```diff
@@ -284,6 +284,24 @@ generate_builder = true
     name = "run"
     manual = true
     doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "hold"
+    # Returns an value that releases on Drop
```
GuillaumeGomez, line 289:
```suggestion
# Returns a value that releases on `Drop`.
```

gio/Gir.toml
```diff
@@ -284,6 +284,24 @@ generate_builder = true
...
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "release"
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "mark_busy"
+    manual = true
+    # Returns an value that unmarks-busy on Drop
```
GuillaumeGomez, line 299:
```suggestion
    # Returns a value that unmarks-busy on `Drop`.
```

gio/Gir.toml
```diff
@@ -284,6 +284,24 @@ generate_builder = true
...
     manual = true
     doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "hold"
+    # Returns an value that releases on Drop
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "release"
+    manual = true
```
GuillaumeGomez, line 294: What's the reason for this one?

gio/Gir.toml
```diff
@@ -284,6 +284,24 @@ generate_builder = true
...
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "mark_busy"
+    manual = true
+    # Returns an value that unmarks-busy on Drop
+    doc_trait_name = "ApplicationExtManual"
+    [[object.function]]
+    name = "unmark_busy"
+    manual = true
```
GuillaumeGomez, line 303: What's the reason for this one?

gio/Gir.toml
sdroege: See 4 lines above

gio/Gir.toml
sdroege: See 4 lines above

gio/Gir.toml
GuillaumeGomez: And how someone without this PR context can guess it? ;)

gio/Gir.toml
GuillaumeGomez: And how someone without this PR context can guess it? ;)

gio/Gir.toml
sdroege: What does that have to do with the PR? The explanation is 4 lines above in the .toml file

GuillaumeGomez: Because they don't seem linked to each other just from taking a look at it.

sdroege:
Done

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #688, opened 2022-05-14
bump versions in CI tests, fix CI

jf2048:


bilelmoussaoui (approved):

.github/workflows/CI.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Check doc aliases
         run: |
           python3 doc_aliases.py ../${{matrix.conf.name}}
-          cd .. && git diff --exit-code
+          cd .. && git config --global --add safe.directory `pwd` && git diff --exit-code
```
bilelmoussaoui, line 62: I am not sure why this would be needed though

sdroege (approved):

.github/workflows/CI.yml
jf2048: The second checkout seems to overwrite `safe.directory`, I'll see if there is a more idiomatic way to stop this

Issue #693, opened 2022-05-18
cairo: Is it okay to migrate freetype to freetype-rs or use freetype-sys directly?


songww:

sdroege:
What would be the rationale for this? We can change to whatever makes more sense :)

songww:
The freetype bindings has not been updated for 5 years and do not contain secure bindings.
The crate that contains safe bindings is freetype-rs, which has been actively maintained, but is not compatible with freetype.

Sorry my english is poor.

sdroege:
> The freetype bindings has not been updated for 5 years and do not contain secure bindings.
> The crate that contains safe bindings is freetype-rs, which has been actively maintained, but is not compatible with freetype.

Thanks, that perfectly makes sense. Do you want to provide a PR for that?

> Sorry my english is poor.

Don't worry, your English seems perfectly fine to me :)

sdroege:
Thanks for remembering to close this issue

PR #694, opened 2022-05-18
cairo: Use freetype-sys instead of freetype.

songww:
#693

### DIFF
--- a/cairo/src/font/font_face.rs
+++ b/cairo/src/font/font_face.rs
@@ -62,15 +62,14 @@ impl FontFace {
 
     // rustdoc-stripper-ignore-next
     /// Creates a new font face for the FreeType backend from an already opened FreeType face.
-    ///
-    /// # Safety
-    ///
-    /// The`FT_Face` must be valid and not be freed until the last reference to the
-    /// `FontFace` is dropped.
     #[cfg(any(feature = "freetype", feature = "dox"))]
     #[doc(alias = "cairo_ft_font_face_create_for_ft_face")]
     pub fn create_from_ft(face: &freetype::face::Face) -> Result<FontFace, Error> {
+        // Increase reference count of `FT_Face`.
         let mut face = face.clone();
+
+        // SAFETY: The user data entry keeps `freetype::face::Face` alive
+        // until the FontFace is dropped.
         let font_face = unsafe {
             FontFace::from_raw_full(ffi::cairo_ft_font_face_create_for_ft_face(
                 face.raw_mut() as freetype::ffi::FT_Face as *mut _,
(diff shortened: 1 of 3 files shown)


cairo/src/font/font_face.rs
```diff
@@ -64,9 +64,7 @@ impl FontFace {
     /// `FontFace` is dropped.
     #[cfg(any(feature = "freetype", feature = "dox"))]
     #[doc(alias = "cairo_ft_font_face_create_for_ft_face")]
-    pub unsafe fn create_from_ft(
```
bilelmoussaoui, line 67: I am not sure if it is a great idea to use the ffi bindings here

songww: I'm not sure that changing to a safe method is an appropriate change, it's a big change.

sdroege:
Using the safe freetype-rs instead of freetype-sys is probably nicer, or not?

cairo/src/font/font_face.rs
```diff
@@ -87,15 +94,19 @@ impl FontFace {
...
-        ));
-        let status = ffi::cairo_font_face_status(font_face.to_raw_none());
+        let mut face = face.clone();
+        let font_face = unsafe {
+            FontFace::from_raw_full(ffi::cairo_ft_font_face_create_for_ft_face(
+                face.raw_mut() as freetype::ffi::FT_Face as *mut _,
+                load_flags,
+            ))
+        };
+        font_face.set_user_data(&FT_FACE_KEY, Rc::new(face))?;
```
sdroege, line 108: Why is the `Rc` needed here?

sdroege (commented):
Looks good to me generally

jf2048:
this is only running 11 checks right now, please re-run the CI on this after https://github.com/gtk-rs/gtk-rs-core/pull/695

cairo/src/font/font_face.rs
```diff
@@ -64,14 +69,16 @@ impl FontFace {
     /// `FontFace` is dropped.
     #[cfg(any(feature = "freetype", feature = "dox"))]
     #[doc(alias = "cairo_ft_font_face_create_for_ft_face")]
-    pub unsafe fn create_from_ft(
```
RealKC, line 67: If you're turning this into a safe method, you should remove the `# Safety` section from the docs

cairo/src/font/font_face.rs
```diff
@@ -64,14 +69,16 @@ impl FontFace {
...
-        face: freetype_crate::freetype::FT_Face,
-    ) -> Result<FontFace, Error> {
-        let font_face = FontFace::from_raw_full(ffi::cairo_ft_font_face_create_for_ft_face(
-            face as *mut _,
-            0,
-        ));
-        let status = ffi::cairo_font_face_status(font_face.to_raw_none());
+    pub fn create_from_ft(face: &freetype::face::Face) -> Result<FontFace, Error> {
+        let mut face = face.clone();
+        let font_face = unsafe {
```
RealKC, line 74: Can you add a safety comment here? I missed the `.clone()` call at first and was worried about potential use-after-frees

RealKC (changes requested):

cairo/src/font/font_face.rs
sdroege: Better would be to use `// SAFETY: blablabla` but this also works :)

cairo/src/font/font_face.rs
```diff
@@ -87,15 +94,19 @@ impl FontFace {
...
+        face: &freetype::face::Face,
         load_flags: libc::c_int,
     ) -> Result<FontFace, Error> {
-        let font_face = FontFace::from_raw_full(ffi::cairo_ft_font_face_create_for_ft_face(
-            face as *mut _,
-            load_flags,
-        ));
-        let status = ffi::cairo_font_face_status(font_face.to_raw_none());
+        let mut face = face.clone();
+        let font_face = unsafe {
```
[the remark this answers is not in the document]
sdroege: Same here about the format, should ideally be `// SAFETY: blablabla`

cairo/src/font/font_face.rs
sdroege: Indeed, thanks :)

sdroege (approved):
Looks good to me apart from the comments. Want to fix those @songww or otherwise I'll just do it before merging.

sdroege:
Uh, new beta with new clippy warnings happened. Let's worry about those separately.

sdroege:
Thanks for the improvements :)

songww:
Thank you all for your advice！

PR #699, opened 2022-05-20
Update gir

sdroege:
See https://github.com/gtk-rs/gir/pull/1357


filnet:
There are two `#[allow(clippy::let_and_return)]` that might now be unnecessary.

sdroege:
Good point, thanks!

GuillaumeGomez (approved):

PR #705, opened 2022-05-31
SocketAddressEnumerator: `next` return value should be nullable

mbiggio:
The `next` method (both the sync and async variants) relies on the
pointer being NULL to indicate the end of the address stream.
See also the corresponding MR to add the annotation in the glib source
code:

https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2701

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -116,7 +116,6 @@ generate = [
     "Gio.SettingsSchemaSource",
     "Gio.SimpleActionGroup",
     "Gio.SimplePermission",
-    "Gio.SocketAddressEnumerator",
     "Gio.SocketClient",
     "Gio.SocketClientEvent",
     "Gio.SocketConnectable",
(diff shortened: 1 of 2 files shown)


gio/Gir.toml
```diff
@@ -1122,6 +1121,18 @@ name = "Gio.SocketAddress"
 status = "generate"
 concurrency = "send+sync"
 
+[[object]]
+name = "Gio.SocketAddressEnumerator"
+status = "generate"
+    [[object.function]]
+    name = "next"
```
GuillaumeGomez, line 1128: Please add a comment to both to tell why it's needed.

GuillaumeGomez:
Apart from the missing comments, looks good to me, thanks!

gio/Gir.toml
sdroege: Please add a link to the GLib MR here. The comment is optional, but having the MR link here allows us to get rid of this once we imported updated XML files.

mbiggio: Sure, here it is:
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2701

GuillaumeGomez: I think he meant in the comment? (please correct me if I'm wrong @sdroege)

sdroege: Yes, inside the `Gir.toml`

sdroege: So I can simply grep for things when updating the XML files

PR #707, opened 2022-06-01
ProcessLauncher: `close` method should be available on unix only

mbiggio:
According to the gio public header, `g_subprocess_launcher_close` should
be available on unix platforms only:
https://gitlab.gnome.org/GNOME/glib/-/blob/glib-2-72/gio/gsubprocesslauncher.h#L81-115

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -1206,6 +1206,9 @@ status = "generate"
     [[object.function]]
     name = "set_child_setup"
     cfg_condition = "unix"
+    [[object.function]]
+    name = "close"
+    cfg_condition = "unix"
 
 [[object]]
 name = "Gio.ThemedIcon"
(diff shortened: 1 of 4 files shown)


gio/Gir.toml
```diff
@@ -1206,6 +1206,9 @@ status = "generate"
     [[object.function]]
     name = "set_child_setup"
     cfg_condition = "unix"
+    [[object.function]]
```
bilelmoussaoui, line 1209: we don't do the same for the sys crate?

mbiggio: as far as I can see, we are generating all of the `SubprocessLauncher` raw bindings in the `sys` crate, independently if they are unix-specific or not:
https://github.com/gtk-rs/gtk-rs-core/blob/master/gio/sys/Gir.toml

sdroege: The UNIX-specific one should be marked as such in the sys and non-sys crate.

mbiggio: yep, will add those to the sys one, thanks

bilelmoussaoui (approved):

sdroege (approved):

PR #643, opened 2022-04-11
Remove is_windows_utf8

jim4067:

PR #708, opened 2022-06-02
glib: bind iconv functions

jf2048:
- Is there a better name for `CvtError`?
- Maybe some of these should be in utils.rs?

### DIFF
--- a/glib/Gir.toml
+++ b/glib/Gir.toml
@@ -10,6 +10,7 @@ deprecate_by_min_version = true
 trust_return_value_nullability = true
 
 generate = [
+    "GLib.ConvertError",
     "GLib.DateDay",
     "GLib.DateMonth",
     "GLib.DateWeekday",
@@ -91,8 +92,14 @@ status = "generate"
     pattern = "(idle_remove_by_data|qsort_with_data)"
     ignore = true # Unsafe functions
     [[object.function]]
-    pattern = "(iconv|convert_with_iconv)"
-    ignore = true # not usable
+    pattern = "iconv"
+    doc_struct_name = "IConv"
+    manual = true
+    [[object.function]]
+    pattern = "convert_with_iconv"
+    doc_struct_name = "IConv"
+    manual = true
+    rename = "convert"
     [[object.function]]
     pattern = "pointer_.*"
     ignore = true # Not usable, uses raw pointers
@@ -122,8 +129,7 @@ status = "generate"
     ignore = true # Deprecated
     [[object.function]]
     name = "get_filename_charsets"
-    #string
-    ignore = true
+    manual = true
     [[object.function]]
     pattern = "str.+v"
     #string
@@ -181,10 +187,6 @@ status = "generate"
     [[object.function]]
     pattern = "[gs]et_prgname"
     #manual pathbuf
-    ignore = true
-    [[object.function]]
-    name = "filename_from_utf8"
-    #not in 64bit lib for Windows
     ignore = true
     [[object.function]]
     name = "get_user_name"
@@ -251,8 +253,7 @@ status = "generate"
     manual = true
     [[object.function]]
     name = "convert_with_fallback"
-    #out param not in .gir
-    ignore = true
+    manual = true
     [[object.function]]
     name = "unicode_canonical_decomposition"
     #out param not in .gir
@@ -263,20 +264,19 @@ status = "generate"
     ignore = true
     [[object.function]]
     name = "convert"
-    #unusable
-    ignore = true
+    manual = true
+    [[object.function]]
+    name = "filename_from_utf8"
+    manual = true
     [[object.function]]
     name = "filename_to_utf8"
-    #unusable
-    ignore = true
+    manual = true
     [[object.function]]
     name = "locale_from_utf8"
-    #unusable
-    ignore = true
+    manual = true
     [[object.function]]
     name = "locale_to_utf8"
-    #unusable
-    ignore = true
+    manual = true
     [[object.function]]
     name = "child_watch_source_new"
     # Need manual bindings to be useful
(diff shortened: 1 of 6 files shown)


glib/src/convert.rs
```diff
@@ -0,0 +1,331 @@
...
+    };
+    if result.is_null() {
+        Err(CvtError::new(unsafe { from_glib_full(error) }, bytes_read))
+    } else {
+        Ok((unsafe { from_glib_full(result) }, bytes_read))
+    }
+}
+
+#[doc(alias = "g_filename_from_utf8")]
+pub fn filename_from_utf8(utf8string: &GStr) -> Result<(Slice<c_char>, usize), CvtError> {
```
sdroege, line 241:
```suggestion
pub fn filename_from_utf8(utf8string: &GStr) -> Result<(Slice<u8>, usize), CvtError> {
```

Would probably be more usable

glib/src/convert.rs
```diff
@@ -0,0 +1,331 @@
...
+    };
+    if result.is_null() {
+        Err(CvtError::new(unsafe { from_glib_full(error) }, bytes_read))
+    } else {
+        Ok((unsafe { from_glib_full(result) }, bytes_read))
+    }
+}
+
+#[doc(alias = "g_filename_to_utf8")]
+pub fn filename_to_utf8(opsysstring: &[u8]) -> Result<(GString, usize), CvtError> {
```
sdroege, line 219: Not `&OsStr`?

sdroege (commented):
Seems good otherwise, thanks!

glib/src/convert.rs
jf2048: I suppose it should be `Path`, but also this is inconsistent with the rest of these functions unless this one is finished: https://github.com/gtk-rs/gtk-rs-core/pull/643

jf2048: That PR makes it look like this could be autogenerated though, so maybe just remove this one for now?

sdroege: I guess I should finish that PR first then :)

sdroege: Done

jf2048: Fixed but now it needs https://github.com/gtk-rs/gir/pull/1367

glib/src/convert.rs
jf2048: Now it's autogenerated as `PathBuf`

glib/src/convert.rs
sdroege: Merged, can you update to that version?

glib/src/convert.rs
```diff
@@ -0,0 +1,264 @@
...
+        offset: usize,
+    },
+}
+
+#[derive(Debug)]
+#[repr(transparent)]
+#[doc(alias = "GIConv")]
+pub struct IConv(ffi::GIConv);
+
+unsafe impl Send for IConv {}
```
sdroege, line 114: All functions are taking `&self`, so is this actually also `Sync`? Otherwise having the conversion functions take a `&mut self` would seem useful.

jf2048: Reference is the man page, suggests it is not `Sync`: https://www.man7.org/linux/man-pages/man3/iconv.3.html#ATTRIBUTES

jf2048: Yeah, changed it

glib/Gir.toml
```diff
@@ -262,20 +263,13 @@ status = "generate"
     ignore = true
     [[object.function]]
     name = "convert"
-    #unusable
-    ignore = true
+    manual = true
     [[object.function]]
     name = "filename_to_utf8"
-    #unusable
-    ignore = true
```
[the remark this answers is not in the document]
jf2048: I was just seeing if it passes the tests, now with `AsRef<Path>` gir needs another change to make it generate `.as_ref().to_glib_none().1.as_bytes().len()`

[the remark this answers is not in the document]
sdroege: Ok :) Some other time, I guess

glib/src/auto/functions.rs
```diff
@@ -293,6 +293,34 @@ pub fn filename_from_uri(
...
+            len,
+            bytes_read.as_mut_ptr(),
+            bytes_written.as_mut_ptr(),
+            &mut error,
+        );
+        if error.is_null() {
+            Ok((
+                from_glib_full(ret),
+                bytes_read.assume_init(),
+                bytes_written.assume_init(),
```
sdroege, line 316: Isn't this equivalent to the length of the returned `PathBuf`?

glib/src/auto/functions.rs
```diff
@@ -539,6 +567,32 @@ pub fn listenv() -> Vec<std::ffi::OsString> {
...
```
sdroege, line 588: Isn't this equivalent to the length of the returned `GString`? And I assume it can be used also to give a size hint to it when creating the `GString`

glib/src/convert.rs
```diff
@@ -0,0 +1,292 @@
...
+            path.1.as_bytes().len() as isize,
+            bytes_read.as_mut_ptr(),
+            bytes_written.as_mut_ptr(),
+            &mut error,
+        );
+        if error.is_null() {
+            Ok((
+                from_glib_full(ret),
+                bytes_read.assume_init(),
+                bytes_written.assume_init(),
```
sdroege, line 252: Isn't this equivalent to the length of the returned `GString`? And I assume it can be used also to give a size hint to it when creating the `GString`

sdroege:
Ping?

sdroege:
@jf2048 Ping?

glib/src/convert.rs
jf2048: Only bytes_read works as a size hint, the bytes_written param does not have to be returned though.

jf2048:
Sorry for delay, this is updated now

glib/src/auto/functions.rs
jf2048: Yes, `FromGlibContainer` is now implemented for `PathBuf` so the scan for the nul byte can be avoided

sdroege (approved):

PR #709, opened 2022-06-03
translate.rs: free the container memory also when the container is empty

mbiggio:
When transferring ownership of the container to the caller, the
container should be freed also when it is empty, since it might contain
a terminating NULL element. Calling g_free on a NULL container is
harmless, since in this case g_free is a no-op.

### DIFF
--- a/glib/src/boxed.rs
+++ b/glib/src/boxed.rs
@@ -219,6 +219,7 @@ macro_rules! glib_boxed_wrapper {
 
             unsafe fn from_glib_full_num_as_vec(ptr: *mut *mut $ffi_name, num: usize) -> Vec<Self> {
                 if num == 0 || ptr.is_null() {
+                    $crate::ffi::g_free(ptr as *mut _);
                     return Vec::new();
                 }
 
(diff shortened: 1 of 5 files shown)


sdroege (dismissed):
Did you check if there are other instances of the same mistake?

mbiggio:
I will double check for other instances and update the PR

mbiggio:
There were indeed a few more instances of the problem, thanks for pointing that out.

I don't think [this implementation](https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/translate.rs#L2110) using a GList  and [this implementation](https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/translate.rs#L2301) using a PointerArray need the fix above

sdroege:
> I don't think [this implementation](https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/translate.rs#L2110) using a GList

That doesn't need changes

> and [this implementation](https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/translate.rs#L2301) using a PointerArray need the fix above

That needs an `unref()` in the `num == 0` case but not in the `ptr.is_null()` case

sdroege:
Thanks!

PR #713, opened 2022-06-06
Implement Borrow on object types

### LINKED ISSUE #711: [FEATURE REQUEST] Borrow on Object types
An object subclass already implements `AsRef<$ParentObject>`. Could it also implement `Borrow<$ParentObject>`?

The Borrow trait has different requirements and I don't know if this is possible.
From the rust doc:
> AsRef has the same signature as [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html), but [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) is different in few aspects:
    - Unlike AsRef, [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) has a blanket impl for any T, and can be used to accept either a reference or a value.
    - [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) also requires that [Hash](https://doc.rust-lang.org/std/hash/trait.Hash.html), [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html) and [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html) for borrowed value are equivalent to those of the owned value. For this reason, if you want to borrow only a single field of a struct you can implement AsRef, but not [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html).
    
I would use the `Borrow` trait to accept parameters in setter functions in the Properties macro (#494). `AsRef` isn't enough because it doesn't have a blanket impl for any T

ranfdev:
The [`Borrow`](https://doc.rust-lang.org/std/borrow/trait.Borrow.html) trait is more restrictive than
`AsRef`, but since `PartialOrd`, `Hash`, `PartialEq`, (...), are all implemented on the object pointer,
it's safe, and I believe correct, to implement `Borrow`.

Closes #711. Covers my use case in the Properties macro

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -745,6 +745,7 @@ macro_rules! glib_object_wrapper {
             }
         }
 
+
         impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? std::fmt::Debug for $name $(<$($generic),+>)? {
             fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
                 f.debug_struct(stringify!($name)).field("inner", &self.inner).finish()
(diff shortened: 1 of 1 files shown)


jf2048:
This looks fine to me

sdroege (approved):

PR #726, opened 2022-07-04
glib: fix ParamSpec with null blurb and nick

zhangyuannie:
ParamSpec can be created with blurb and nick set to NULL. In fact, gtk has [removed](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4717) all blurbs and nicks from their objects recently.


glib/src/param_spec.rs
```diff
@@ -475,14 +481,14 @@ macro_rules! define_builder {
             $($field_id: Option<$field_ty>),*
         }
         impl<'a> $builder_type<'a> {
-            /// Default: `self.name`
-            pub fn nick(mut self, nick: &'a str) -> Self {
```
bilelmoussaoui, line 479: Why would you pass an Option<T> to a builder type.. This change doesn't make sense

glib/src/param_spec.rs
```diff
@@ -1225,15 +1236,15 @@ mod tests {
     fn test_param_spec_string() {
         let pspec = ParamSpecString::new(
             "name",
-            "nick",
-            "blurb",
+            None,
+            Some("blurb"),
```
bilelmoussaoui, line 1240: why you dropped the nick & not the blurb?

zhangyuannie: Arbitrary choice as it's in tests. I can set all to None or Some if you wish.

bilelmoussaoui: Right, never mind then.

glib/src/param_spec.rs
zhangyuannie: Fixed

glib/src/param_spec.rs
```diff
@@ -198,14 +198,11 @@ impl ParamSpec {
 
     #[doc(alias = "g_param_spec_get_nick")]
     #[doc(alias = "get_nick")]
-    pub fn nick(&self) -> Option<&str> {
+    pub fn nick(&self) -> &str {
```
bilelmoussaoui, line 201: this one should stay nullable though :)

zhangyuannie: Yeah I thought that too. But glib apparently returns `name` if nick is not set.

https://gitlab.gnome.org/GNOME/glib/-/blob/main/gobject/gparam.c#L331

bilelmoussaoui: Ah ok, thanks for checking! Could use a doc comment.

zhangyuannie: Done!

glib/src/param_spec.rs
```diff
@@ -475,12 +480,12 @@ macro_rules! define_builder {
             $($field_id: Option<$field_ty>),*
         }
         impl<'a> $builder_type<'a> {
-            /// Default: `self.name`
+            /// Default: `None`
```
sdroege, line 483: Technically the default is still `self.name` as that's what will be used as default from C

zhangyuannie: It is actually a bit more subtle than that, I have updated the docs. Let me know if you think this is better.

glib/src/param_spec.rs
```diff
@@ -417,8 +422,8 @@ macro_rules! define_param_spec_numeric {
             #[doc(alias = $alias)]
             pub fn new(
                 name: &str,
-                nick: &str,
-                blurb: &str,
```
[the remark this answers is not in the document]
sdroege: Not providing nick/blurb only really makes sense for GTK. GTK users can use the builder if they care :)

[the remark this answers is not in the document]
bilelmoussaoui: > Not providing nick/blurb only really makes sense for GTK.

Isn't only gstreamer that make use of them nowadays?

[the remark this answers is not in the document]
sdroege: At least, yes

glib/src/param_spec.rs
```diff
@@ -175,7 +178,7 @@ impl ParamSpec {
 
     #[doc(alias = "g_param_spec_get_name")]
     #[doc(alias = "get_name")]
-    pub fn name<'a>(&self) -> &'a str {
+    pub fn name(&self) -> &str {
```
sdroege, line 181: Sorry, I missed this part. This change is wrong and breaks existing code.

The returned string here is an interned string that is valid for the remaining runtime of the process. It is not bound by the `&self` reference.

sdroege: Solved in https://github.com/gtk-rs/gtk-rs-core/pull/728

PR #728, opened 2022-07-11
glib: Fix `ParamSpec::name()` return value lifetime

sdroege:
The string is an interned string that is valid for the remaining runtime
of the process and is not bound by the `&self` reference.

----

See https://github.com/gtk-rs/gtk-rs-core/pull/726/files#r917705349


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #734, opened 2022-07-21
glib: Simplify SignalBuilder constructor

bilelmoussaoui:
The return_type & param_types are optional, no need to enforce those

### DIFF
--- a/glib/src/subclass/signal.rs
+++ b/glib/src/subclass/signal.rs
@@ -4,6 +4,7 @@ use crate::translate::*;
 use crate::utils::is_canonical_pspec_name;
 use crate::Closure;
 use crate::SignalFlags;
+use crate::StaticType;
 use crate::Type;
 use crate::Value;
 
@@ -15,10 +16,10 @@ use std::{fmt, num::NonZeroU32};
 /// Builder for signals.
 #[allow(clippy::type_complexity)]
 #[must_use = "The builder must be built to be used"]
-pub struct SignalBuilder<'a> {
-    name: &'a str,
+pub struct SignalBuilder {
+    name: String,
     flags: SignalFlags,
-    param_types: &'a [SignalType],
+    param_types: Vec<SignalType>,
     return_type: SignalType,
     class_handler: Option<
         Box<dyn Fn(&SignalClassHandlerToken, &[Value]) -> Option<Value> + Send + Sync + 'static>,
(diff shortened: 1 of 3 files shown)


sdroege (approved):
Looks good to me. Can you send MRs for gstreamer-rs / gst-plugins-rs for this?

bilelmoussaoui:
> Looks good to me. Can you send MRs for gstreamer-rs / gst-plugins-rs for this?

Nothing to change on gstreamer-rs / https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/810. I will look into updating gtk3-rs/gtk4-rs later

sdroege:
Thanks, just merge when you're ready :)

glib/src/subclass/signal.rs
```diff
@@ -377,7 +378,24 @@ enum SignalRegistration {
     },
 }
 
-impl<'a> SignalBuilder<'a> {
+impl SignalBuilder {
+    // rustdoc-stripper-ignore-next
+    /// The signal parameters.
+    pub fn param_types(mut self, param_types: &[Type]) -> Self {
```
sdroege, line 384: This doesn't allow for marking the type `G_SIGNAL_TYPE_STATIC_SCOPE` anymore

glib/src/subclass/signal.rs
```diff
@@ -377,7 +378,24 @@ enum SignalRegistration {
...
+        self.param_types = param_types
+            .iter()
+            .map(|t| SignalType::from(*t))
+            .collect::<Vec<_>>();
+        self
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// The signal is returned value type.
+    pub fn return_type<T: StaticType>(mut self) -> Self {
```
sdroege, line 394: This doesn't allow for marking the type `G_SIGNAL_TYPE_STATIC_SCOPE` anymore

bilelmoussaoui: What do  you suggest we do here ? Revert the change or add another variant that automatically wraps things with `SignalType::with_static_type`?

sdroege: I would here do

```fn
fn param_types(mut self, param_types: impl IntoIterator<Item = Into<SignalType>>) -> Self;
fn return_type(mut self, type_: impl Into<SignalType>) -> Self;
```

The latter might also be in addition, having something like `fn return_type_with_type()` (but a better name please :P )

sdroege: Also IMHO having all functions take an actual `type` parameter is fine instead of using `T: StaticType`. The latter doesn't bring much improvement

bilelmoussaoui: > Also IMHO having all functions take an actual type parameter is fine instead of using T: StaticType. The latter doesn't bring much improvement

It avoid having to write `T::static_type()` everywhere but yeah that can work. I will submit a patch later today

sdroege: But how often do you create signals with non-unit return types? :)

bilelmoussaoui: For gui apps, a lot of times :)

sdroege: Oh really? I thought that's just something people do in GStreamer often

PR #735, opened 2022-07-21
glib: Add flag setters to ParamSpec builders

bilelmoussaoui:
cc @ranfdev 

Similar to what we have in `SignalBuilder`

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -449,6 +449,90 @@ macro_rules! define_param_spec_numeric {
     };
 }
 
+/// A trait implemented by the various [`ParamSpec`] builder types.
+///
+/// It is useful for providing a builder pattern for [`ParamSpec`] defined
+/// outside of GLib like in GStreamer or GTK 4.
+pub trait ParamSpecBuilderExt<'a>: Sized {
+    /// Implementation detail.
+    fn set_nick(&mut self, nick: Option<&'a str>);
+    /// Implementation detail.
+    fn set_blurb(&mut self, blurb: Option<&'a str>);
+    /// Implementation detail.
+    fn set_flags(&mut self, flags: crate::ParamFlags);
+    /// Implementation detail.
+    fn current_flags(&self) -> crate::ParamFlags;
+
+    /// By default, the nickname of its redirect target will be used if it has one.
+    /// Otherwise, `self.name` will be used.
+    fn nick(mut self, nick: &'a str) -> Self {
+        self.set_nick(Some(nick));
+        self
+    }
+
+    /// Default: `None`
+    fn blurb(mut self, blurb: &'a str) -> Self {
+        self.set_blurb(Some(blurb));
+        self
+    }
+
+    /// Default: `glib::ParamFlags::READWRITE`
+    fn flags(mut self, flags: crate::ParamFlags) -> Self {
+        self.set_flags(flags);
+        self
+    }
+
+    /// Mark the property as read only and drops the READWRITE flag set by default.
+    fn read_only(self) -> Self {
+        let flags =
+            (self.current_flags() - crate::ParamFlags::WRITABLE) | crate::ParamFlags::READABLE;
+        self.flags(flags)
+    }
+
+    /// Mark the property as write only and drops the READWRITE flag set by default.
+    fn write_only(self) -> Self {
+        let flags =
+            (self.current_flags() - crate::ParamFlags::READABLE) | crate::ParamFlags::WRITABLE;
+        self.flags(flags)
+    }
+
+    /// Mark the property as readwrite, it is the default value.
+    fn readwrite(self) -> Self {
+        let flags = self.current_flags() | crate::ParamFlags::READWRITE;
+        self.flags(flags)
+    }
+
+    /// Mark the property as construct
+    fn construct(self) -> Self {
+        let flags = self.current_flags() | crate::ParamFlags::CONSTRUCT;
+        self.flags(flags)
+    }
+
+    /// Mark the property as construct only
+    fn construct_only(self) -> Self {
+        let flags = self.current_flags() | crate::ParamFlags::CONSTRUCT_ONLY;
+        self.flags(flags)
+    }
+
+    /// Mark the property as lax validation
+    fn lax_validation(self) -> Self {
+        let flags = self.current_flags() | crate::ParamFlags::LAX_VALIDATION;
+        self.flags(flags)
+    }
+
+    /// Mark the property as explicit notify
+    fn explicit_notify(self) -> Self {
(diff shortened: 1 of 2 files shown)


glib/src/param_spec.rs
```diff
@@ -501,6 +501,48 @@ macro_rules! define_builder {
...
+                self.flags |= crate::ParamFlags::READABLE;
+                self
+            }
+
+            /// Mark the property as writable
+            pub fn writable(mut self) -> Self {
+                self.flags |= crate::ParamFlags::WRITABLE;
+                self
+            }
+
```
bilelmoussaoui, line 515: Those two should probably be dropped as the default is set to readwrite, in case you want to overwrite those you would go with a `T::builder(...).flags()`?

bilelmoussaoui: I am talking about readable/writable here

ranfdev: Yeah, having readwrite as default while having the readable/writable methods is  confusing. I would remove the readable/writable methods.

In any case, in the Properties macro I will keep using .flags(), so that it stays compatible with custom, external flags.

bilelmoussaoui: Dropped those already, thanks

sdroege: You could add `readwrite()`, `read_only()` and `write_only()` that set those two flags accordingly. I think that would be a nicer API than having to call `flags()` for these cases.

glib/src/param_spec.rs
```diff
@@ -501,6 +501,36 @@ macro_rules! define_builder {
                 self
             }
 
+            /// Mark the property as construct
+            pub fn construct(mut self) -> Self {
+                self.flags |= crate::ParamFlags::CONSTRUCT;
```
sdroege, line 506: I think `CONSTRUCT` implies `WRITE`, so you should probably also set that one here. Similarly, `CONSTRUCT_ONLY` implies `CONSTRUCT`.

Or we could leave it to the constructor to do a critical warning. Not sure what's nicer.

sdroege (commented):
Looks good to me except for those two comments

glib/src/param_spec.rs
bilelmoussaoui: > readwrite()

That is the default, I don't see why that would be needed.  If I add a `read_only` / `write_only`, I suppose those must be called before any flags function, as those might override future flags setters?

glib/src/param_spec.rs
```diff
@@ -501,6 +501,36 @@ macro_rules! define_builder {
...
+            }
+
+            /// Mark the property as explicit notify
+            pub fn explicit_notify(mut self) -> Self {
+                self.flags |= crate::ParamFlags::EXPLICIT_NOTIFY;
+                self
+            }
+
+            /// Mark the property as deprecated
+            pub fn deprecated(mut self) -> Self {
```
sdroege, line 529: Maybe the builders should all implement a trait so that in gstreamer-rs I can implement an extension trait on all the builders that adds builder methods for all the GStreamer pspec flags.

bilelmoussaoui: Yeah that would be great indeed, I wanted to port the gst-plugins-rs to the builder patterns, but as the flags has to be set manually, I dropped that idea for now.

glib/src/param_spec.rs
bilelmoussaoui: I kept that as is, by assuming `READWRITE` is the default, I think leaving to the constructor to throw a critical for now is fine

glib/src/param_spec.rs
sdroege: `readwrite()` would always set both flags (that might be unset by `read_only()` before, for example). It would just be for consistency :)

`read_only()` would set `READ` and unset `WRITE`. You can call that at any time, it doesn't overwrite any other flags.

glib/src/param_spec.rs
sdroege: That trait would have all those common builder functions like `flags()`, the ones you just added, etc. so that I can do a `impl<T: ThatTrait> GStreamerPSpecBuilderExt for T { ... }`

ranfdev: That works indeed, but Imho that's too much complexity for something that's already doable by OR-ing the ParamFlags.

bilelmoussaoui: it is not as you would have to use the TBuilder::flags() instead of using one of the setter methods above. We were already thinking of moving all the builder patterns to traits anyway, so it would make sense to have that here as well for the sake of consistency.

sdroege: Also it's nicer to read compared to the flags.

bilelmoussaoui: I implemented the trait split

sdroege:
What's the status here? Ready for another review, open questions, ...? :)

bilelmoussaoui:
> What's the status here? Ready for another review, open questions, ...? :)

ready for review

glib/src/param_spec.rs
```diff
@@ -466,6 +550,22 @@ macro_rules! define_builder {
                 $builder_type::new(name, $($($req_ident),*)?)
             }
         }
+
+        impl <'a>crate::prelude::ParamSpecBuilderExt<'a> for $builder_type<'a> {
```
sdroege, line 554:
```diff
        impl <'a>crate::prelude::ParamSpecBuilderExt<'a> for $builder_type<'a> {
        impl<'a> crate::prelude::ParamSpecBuilderExt<'a> for $builder_type<'a> {
```

glib/src/param_spec.rs
```diff
@@ -501,6 +501,56 @@ macro_rules! define_builder {
                 self
             }
 
+            /// Mark the property as read only and drops the READWRITE flag set by default.
+            pub fn read_only(mut self) -> Self {
+                self.flags -= crate::ParamFlags::default();
```
sdroege: That seems like a bad idea. I think you want to remove only the `WRITABLE` flag here.

Or maybe work with a constant that has the flags you care about. But relying on the default value here seems like a bug waiting to happen.

sdroege (commented):
Seems good to me otherwise.

Did you check if this can be used nicely in gtk4?

glib/src/param_spec.rs
bilelmoussaoui: Right, I just assumed the default would always be readwrite

bilelmoussaoui:
> Seems good to me otherwise.
> 
> Did you check if this can be used nicely in gtk4?

We have a ParamSpecExpression that can be used at, but no specific flags like in gstreamer

sdroege:
> We have a ParamSpecExpression that can be used at, but no specific flags like in gstreamer

The flags should work fine like this. So if everything else works fine in gtk4 then it should be ok :)

sdroege (approved):

PR #737, opened 2022-07-23
fix glyph string methods that don't need to be &mut

bvinc:


pango/Gir.toml
```diff
@@ -309,6 +309,26 @@ status = "generate"
...
+    [[object.function]]
+    name = "index_to_x"
+        [[object.function.parameter]]
+        name = "glyphs"
+        const = true
+    [[object.function]]
+    name = "x_to_index"
+        [[object.function.parameter]]
+        name = "glyphs"
+        const = true
```
bilelmoussaoui, line 331:
```suggestion
    [[object.function]]
    pattern = "(get_logical_widths|get_width|index_to_x|x_to_index)"
        [[object.function.parameter]]
        name = "glyphs"
        const = true
```

sdroege: Or just get_.*

bvinc:
fixed

sdroege (approved):

PR #738, opened 2022-07-25
glib: add doc alias for g_type_class_ref

pbor:


bilelmoussaoui (approved):

glib/src/object.rs
```diff
@@ -4142,6 +4142,7 @@ impl<T: IsClass> Class<T> {
     /// Gets the class struct for `Self` of `type_`.
     ///
     /// This will return `None` if `type_` is not a subclass of `Self`.
+    #[doc(alias = "g_type_class_ref")]
```
sdroege, line 4145: And g_type_class_peek()?

pbor: I do not think we should add it, because `from_type` has the side effect of registering the type, and if someone is using peek then he's intentionally trying not to do that... maybe we need a `try_from_type` that returns an `Option` and calls peek

PR #743, opened 2022-08-10
gio: Manually implement ActionEntry

bilelmoussaoui:
It is pretty useful for ActionMap implementors like GtkApplication
and GtkWindow. It avoids usually the usage of the clone! macro
as we pass the ActionMap to the activate/change_state callbacks.

Especially that Gtk doesn't provide a
gtk_application_class_install_action
alternative to the gtk_widget_class_install_action.

### DIFF
--- a/gio/src/action_entry.rs
+++ b/gio/src/action_entry.rs
@@ -1,0 +1,120 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{ActionMap, SimpleAction};
+use glib::{IsA, Variant};
+
+#[doc(alias = "GActionEntry")]
+pub struct ActionEntry<O>
+where
+    O: IsA<ActionMap>,
+{
+    name: String,
+    parameter_type: Option<String>,
+    state: Option<String>,
+    pub(crate) activate: Option<Box<dyn Fn(&O, &SimpleAction, Option<&Variant>) + 'static>>,
+    pub(crate) change_state: Option<Box<dyn Fn(&O, &SimpleAction, Option<&Variant>) + 'static>>,
+}
+
+impl<O> ActionEntry<O>
+where
+    O: IsA<ActionMap>,
+{
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn parameter_type(&self) -> Option<&str> {
+        self.parameter_type.as_deref()
+    }
+
+    pub fn state(&self) -> Option<&str> {
+        self.state.as_deref()
+    }
+
+    pub fn builder(name: &str) -> ActionEntryBuilder<O> {
+        ActionEntryBuilder::new(name)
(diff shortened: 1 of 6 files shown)


gio/src/action_entry.rs
```diff
@@ -0,0 +1,120 @@
...
+            .field("parameter_type", &self.parameter_type())
+            .field("state", &self.state())
+            .finish()
+    }
+}
+
+#[derive(Debug)]
+pub struct ActionEntryBuilder<O>(ActionEntry<O>)
+where
+    O: IsA<ActionMap>;
```
sdroege, line 55: All these `where` clauses are not actually needed, are they?

gio/src/action_map.rs
```diff
@@ -0,0 +1,50 @@
...
+
+pub trait ActionMapExtManual {
+    #[doc(alias = "g_action_map_add_action_entries")]
+    fn add_action_entries(&self, entries: Vec<ActionEntry<Self>>) -> Result<(), glib::BoolError>
+    where
+        Self: IsA<ActionMap>;
+}
+
+impl<O: IsA<ActionMap>> ActionMapExtManual for O {
+    fn add_action_entries(&self, entries: Vec<ActionEntry<Self>>) -> Result<(), glib::BoolError> {
```
sdroege, line 14:
```suggestion
    fn add_action_entries(&self, entries: impl IntoIterator<Item = ActionEntry<Self>>) -> Result<(), glib::BoolError> {
```

maybe?

sdroege (commented):
Seems good to me otherwise

gio/src/action_entry.rs
bilelmoussaoui: From my tests, either we had to add those where clauses or you will have to specify the type when constructing the ActionEntry itself.

gio/src/action_map.rs
bilelmoussaoui: Doesn't help much because you can't pass in a `&[some_action_entry]` for example

sdroege: Of course not, but you can pass in a `[some_action_entry]` :) `&[T]` `IntoIterator` gives you `&T`.

gio/src/action_entry.rs
sdroege: Ah I would've expected the compiler to figure that out when the action entries are added to some actionmap

gio/src/action_map.rs
bilelmoussaoui: Oh right, i didn't knew I could pass `[T]`.

sdroege: Since Rust 2022 this works nicely. I'm using that in many places in GStreamer now, makes things nicer quite often.

sdroege (approved):

Issue #740, opened 2022-07-31
[BUG] `set_user_data` might have a memory leak if the function errors


SabrinaJewson:
**Bug description**

[In `set_user_data`](https://github.com/gtk-rs/gtk-rs-core/blob/faed1d691ecc513d13869d20bf0a26c4f0a8fd54/cairo/src/user_data.rs#L61), if the function `$ffi_set_user_data` were to fail I don’t believe Cairo will ever run the destructor, so therefore there would be a memory leak of the `Rc` (the [example code for `cairo_ft_font_face_create_for_ft_face`](https://www.cairographics.org/manual/cairo-FreeType-Fonts.html#cairo-ft-font-face-create-for-ft-face) seems to imply this). To fix this I believe it just needs to check if an error occurred and if so call `destructor(ptr)`. However, I’m not certain about this so I’m only opening an issue and not a pull request.

sdroege:
You're probably right but this is not very clear from the docs unfortunately. Needs some checking of the actual implementation.

Note that AFAIU this function can only fail when running out of memory.

sdroege:
Yes, in all error cases the data would be leaked.

@SabrinaJewson Do you want to submit a PR that fixes this leak?

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/745

SabrinaJewson:
Thank you for handling this! I was away from my computer for a couple days.

PR #745, opened 2022-08-12
cairo: Don't leak user data if setting it failed

sdroege:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/740


GuillaumeGomez (approved):

GuillaumeGomez:
You're calling an unsafe function outside an unsafe block apparently.

PR #748, opened 2022-08-17
glib: Allow passing `SignalType`s and plain arrays to the signal buil…

sdroege:
…der for the types

### DIFF
--- a/glib/src/subclass/signal.rs
+++ b/glib/src/subclass/signal.rs
@@ -380,19 +380,29 @@ enum SignalRegistration {
 
 impl SignalBuilder {
     // rustdoc-stripper-ignore-next
-    /// The signal parameters.
-    pub fn param_types(mut self, param_types: &[Type]) -> Self {
+    /// The signal's parameters.
+    pub fn param_types(
+        mut self,
+        param_types: impl IntoIterator<Item = impl Into<SignalType>>,
+    ) -> Self {
         self.param_types = param_types
-            .iter()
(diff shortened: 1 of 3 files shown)


glib/src/subclass/signal.rs
```diff
@@ -396,6 +399,13 @@ impl SignalBuilder {
         self
     }
 
+    // rustdoc-stripper-ignore-next
+    /// The signal is returned value type.
+    pub fn return_type_from_type(mut self, type_: impl Into<SignalType>) -> Self {
```
bilelmoussaoui, line 404:
```suggestion
    pub fn return_type_from(mut self, type_: impl Into<SignalType>) -> Self {
```
maybe?

sdroege: From what? :)

bilelmoussaoui: Or maybe return_from_type?

sdroege: Seems inconsistent and you wouldn't find it when searching for `return_type`

fengalin: Except if there could be other `return_type_from_*`, I think `return_type_from` is fine because it reads ok and explicit in expressions such as: `builder.return_type_from(Type::BOOL)`.

Function comment seems wrong btw :) (same for existing function `return_type`).

bilelmoussaoui (approved):

PR #749, opened 2022-08-17
glib: Variant wrapper types for handles, object paths and signatures

sdroege:
CC @SeaDve

### DIFF
--- a/glib/src/variant.rs
+++ b/glib/src/variant.rs
@@ -1797,6 +1797,148 @@ impl FromVariant for Handle {
     }
 }
 
+/// A wrapper type around `Variant` object paths.
+///
+/// Values of these type are guaranteed to be valid object paths.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ObjectPath(String);
+
+impl ObjectPath {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl std::ops::Deref for ObjectPath {
+    type Target = str;
+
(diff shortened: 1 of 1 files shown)


SeaDve:
Thanks for looking at the issue! LGTM

sdroege:
Any other opinions here @bilelmoussaoui @GuillaumeGomez ?

glib/src/variant.rs
```diff
@@ -1757,6 +1757,188 @@ impl<A: AsRef<[T]>, T: FixedSizeVariantType> ToVariant for FixedSizeVariantArray
     }
 }
 
+/// A wrapper type around `Variant` handles.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Handle(pub i32);
```
bilelmoussaoui, line 1762: Any reason why the inner value is public?

sdroege: Just because "why not" :) There's no problem giving direct access to it

bilelmoussaoui: I guess I have no idea in which context you would use this. Making it public allows people to construct a new `Handle` that might contain a garbage value but then we also have a From implementation so :)

sdroege: It's a plain integer that is an index into an out of band transmitted array (of which you know the size). No problem here

sdroege: Also you can easily parse any kind of integer as a handle from a string anyway.

bilelmoussaoui (approved):

PR #754, opened 2022-08-22
glib/object: add track_caller annotation in a few places

fengalin:
This helps user localize panics resulting from their code.

###  Before

thread 'main' panicked at 'called `Result::unwrap()` on an `Err`
value: BoolError { message: "property 'max-buffers' of type
[...]
**/home/francois/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/4ca23ca/glib/src/object.rs:2372:53**

### After

thread 'main' panicked at 'called `Result::unwrap()` on an `Err`
value: BoolError { message: "property 'max-buffers' of type
[...]
**generic/threadshare/examples/standalone/main.rs:148:18**

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -261,6 +261,7 @@ pub trait Cast: ObjectType {
     ///
     /// If not running with `debug_assertions` enabled, the caller is responsible
     /// for ensuring that the instance implements `T`
+    #[track_caller]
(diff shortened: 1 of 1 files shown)


sdroege (approved):
This makes sense to me. There might be other functions where this would be useful, but those can be added on demand then.

@GuillaumeGomez @bilelmoussaoui what do you think?

bilelmoussaoui (approved):

PR #755, opened 2022-08-22
glib/types: use rustc-hash for HashMap

fengalin:
While profiling the standalone threadshare benchmark in
gst-plugins-rs, callgrind showed 2% cycles spent in
`ElementImplExt::panicked`. This function retrieves the
panicking status in trampolines and makes use of
`instance_data` from `ObjectSubclass`, which is stored in
a `HashMap`.

Rust's `std::collections::HashMap` is known to make use of
a ["hashing algorithm selected to provide resistance against
HashDoS attacks"](https://doc.rust-lang.org/std/collections/struct.HashMap.html). Since the `intance_data` is indexed
by a `glib::Type`, I thought it would make sense to test a
less strict but quicker hasher.

I first tested [`fnv`](https://crates.io/crates/fnv) which is an implementation of the
Fowler-Noll-Vo hash function which is used in major crates
such as `hashbrown`. The crate [`twox-hash`](https://crates.io/crates/twox-hash) features a
benchmark showing larger throughputs for `fnv` with small
data. Callgrind reports a 50% reduction in cycle estimation
for `ElementImplExt::panicked`. I also observe a slight
reduction in CPU usage with `ts-standalone`.

[`rustc-hash`](https://crates.io/crates/rustc-hash) is used by `rustc` where it "consistently
out-performs adn FNV-based hash 5 [...] -- the collision
rate is similar or slightly worse then FNV, but the speed
of the has function itself is much higher". Callgrind
reports 70% reduction in cycle estimation for `panicked`
compared to the default hasher. CPU usage and stats seem
similar to those obtained with `fnv` for `ts-standalone`.

`rustc-hash` was adopted.

### DIFF
--- a/glib/Cargo.toml
+++ b/glib/Cargo.toml
@@ -31,6 +31,7 @@ ffi = { package = "glib-sys", path = "sys" }
 gobject_ffi = { package = "gobject-sys", path = "gobject-sys" }
 glib-macros = { path = "../glib-macros" }
 rs-log = { package = "log", version = "0.4", optional = true }
+rustc-hash = "1.1.0"
 smallvec = "1.0"
 thiserror = "1"
 
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -6,10 +6,11 @@
 use crate::object::{Cast, IsClass, IsInterface, ObjectSubclassIs, ObjectType, ParentClassIs};
 use crate::translate::*;
 use crate::{Closure, Object, StaticType, Type, Value};
+use rustc_hash::FxHashMap as HashMap;
+use std::any::Any;
 use std::marker;
 use std::mem;
 use std::ptr;
-use std::{any::Any, collections::HashMap};
 
 use super::SignalId;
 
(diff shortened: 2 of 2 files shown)


glib/Cargo.toml
```diff
@@ -22,6 +22,7 @@ name = "glib"
 once_cell = "1.0"
 libc = "0.2"
 bitflags = "1.0"
+fnv = "1.0.7"
```
sdroege, line 25: Maybe https://crates.io/crates/rustc-hash would be a better choice? Claims to be faster than FVN, is used in rustc and maintained by the Rust team.

fengalin: Oh! Totally missed it. Giving it a try now. Thanks.

GuillaumeGomez: nnethercote wrote a very nice blog about why it was faster [here](https://nnethercote.github.io/2021/12/08/a-brutally-effective-hash-function-in-rust.html). I'm not sure if it's "valid" for all cases.

fengalin: Interesting blog post. Thanks @GuillaumeGomez

fengalin: With `rustc-hash`, callgrind shows 70% reduction in cycle estimation for `panicked` compared to the hasher from `std` (50% for `fnv`). I observe similar CPU usage and stats to those obtained with `fnv` for `ts-standalone`.

sdroege: Let's go with that one then, thanks!

sdroege (approved):

sdroege:
@fengalin I was wondering, did you also compare performance to `BTreeMap`? I'd expect it to be faster for the small sizes we're dealing with here.

fengalin:
@sdroege good idea! I'll give it a try shortly.

PR #759, opened 2022-08-31
Update gir

sdroege:


sdroege:
> `error: package `dashmap v5.4.0` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.57.0`

Need to update the minimum supported Rust version.

GuillaumeGomez (approved):

PR #760, opened 2022-09-03
Don't checkout submodules by default

sdroege:
This makes sure that cargo does not clone and checkout all the
submodules if pointing to this repository as a git dependency.

To checkout the submodules `git submodule update --checkout` can be
used.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -145,11 +145,13 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
+          set-safe-directory: true
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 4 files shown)


sdroege:
Any ideas how this github actions stuff works?

```
Run actions/checkout@v3
[...]
Adding repository directory to the temporary git global config as a safe directory
/usr/bin/git config --global --add safe.directory /__w/gtk-rs-core/gtk-rs-core
[...]
Run git submodule update --checkout
fatal: unsafe repository ('/__w/gtk-rs-core/gtk-rs-core' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/gtk-rs-core/gtk-rs-core
Error: Process completed with exit code 128.
```

PR #765, opened 2022-09-12
gstring: Implement Default

sophie-h:
Convenient for `unwrap_or_default()`

### DIFF
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -368,17 +368,17 @@ impl GString {
     // rustdoc-stripper-ignore-next
     /// Return the `GString` as string slice.
     pub fn as_str(&self) -> &str {
-        unsafe {
-            let (ptr, len) = match self.0 {
-                Inner::Native(ref cstr) => {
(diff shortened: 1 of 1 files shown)


glib/src/gstring.rs
```diff
@@ -368,17 +368,17 @@ impl GString {
     // rustdoc-stripper-ignore-next
     /// Return the `GString` as string slice.
     pub fn as_str(&self) -> &str {
-        unsafe {
```
sdroege, line 371: Why this change?

glib/src/gstring.rs
```diff
@@ -445,6 +445,12 @@ impl fmt::Debug for GString {
     }
 }
 
+impl Default for GString {
+    fn default() -> Self {
+        Self(Inner::Native(Some(CString::default())))
```
sdroege, line 450: You probably want to create a "foreign" empty string here. The way you'd potentially have fewer copies when actually using it with C APIs and it's in line with the other code here.

sdroege:
This has happened now

PR #770, opened 2022-09-22
glib: Reduce number of allocations when writing non-structured logs v…

sdroege:
…ia the `GlibLogger`

Directly construct a `NUL`-terminated string instead of going via temporary Rust `String`s.

### DIFF
--- a/cairo/src/surface.rs
+++ b/cairo/src/surface.rs
@@ -421,7 +421,7 @@ mod tests {
         /* Initially the data for any mime type has to be none */
         assert!(data.is_none());
 
-        assert!(surface.set_mime_data(MIME_TYPE_PNG, &[1u8, 10u8]).is_ok());
+        assert!(surface.set_mime_data(MIME_TYPE_PNG, [1u8, 10u8]).is_ok());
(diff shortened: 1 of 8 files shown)


sdroege:
@GuillaumeGomez We should really switch to either use stable clippy or having explicitly updated image versions. It's annoying to have to update for beta clippy all the time, especially if the lints are not properly stable.

bilelmoussaoui:
> @GuillaumeGomez We should really switch to either use stable clippy or having explicitly updated image versions. It's annoying to have to update for beta clippy all the time, especially if the lints are not properly stable.

We could also make beta clippy not fail the build

sdroege:
> We could also make beta clippy not fail the build

We're only running it on beta though. We could make that not fail, and also run it on stable where it would fail?

GuillaumeGomez (approved):

bilelmoussaoui:
> > We could also make beta clippy not fail the build
> 
> We're only running it on beta though. We could make that not fail, and also run it on stable where it would fail?

We actually do run it on stable as well, https://github.com/gtk-rs/gtk-rs-core/actions/runs/3105858901/jobs/5031996358

PR #772, opened 2022-09-27
gio: `CancellableFuture` ergonomic for cancelling tasks

BiagioFesta:
An ergonomic to work with `gio::Cancellable` and tipical Rust task-cooperative programming.

`CancellableFuture` is a wrapper for a `Future` which can be cancelled (via `gio::Cancellable`).


gio/src/cancellable_future.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn is_cancelled(&self) -> bool {
+        self.cancellable.is_cancelled()
+    }
+}
+
+impl<F> Future for CancellableFuture<F>
+where
+    F: Future,
+{
+    type Output = Result<<F as Future>::Output, glib::Error>;
```
sdroege, line 61: As the only possible error here is `Cancelled`, maybe it should get its own error type

gio/src/cancellable_future.rs
```diff
@@ -0,0 +1,143 @@
...
+            return Poll::Ready(Err(glib::Error::new(
+                IOErrorEnum::Cancelled,
+                "Task cancelled",
+            )));
+        }
+
+        match self.as_mut().project().future.poll(cx) {
+            Poll::Ready(out) => Poll::Ready(Ok(out)),
+            Poll::Pending => {
+                self.cancellable.connect_cancelled({
```
sdroege, line 74: This should probably only happen once. It's absolutely valid to call `poll()` over and over again after `Pending` is returned without waiting for the waker.

Also I think if the waker changes (and it can change), you will want to disconnect for the older waker

ids1024: Yeah, the waker can change and it should only use the most recent one.

> Note that on multiple calls to poll, only the [Waker](https://doc.rust-lang.org/std/task/struct.Waker.html) from the [Context](https://doc.rust-lang.org/std/task/struct.Context.html) passed to the most recent call should be scheduled to receive a wakeup.

https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll

gio/src/cancellable_future.rs
BiagioFesta: Done. I've added a conversion trait for `glib::Error` tho. I think it might nice to have a integration with the rest of the library

gio/src/cancellable_future.rs
BiagioFesta: Done. Thanks for the catch

gio/src/cancellable_future.rs
sdroege: Sounds like a good idea, yes

gio/src/cancellable_future.rs
```diff
@@ -0,0 +1,184 @@
...
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::Context;
+use std::task::Poll;
+
+// rustdoc-stripper-ignore-next
+/// Indicator that the [`CancellableFuture`] was cancelled.
+pub struct Cancelled;
```
sdroege, line 18:
```suggestion
#[derive(Debug)]
pub struct Cancelled;
```

gio/src/cancellable_future.rs
```diff
@@ -0,0 +1,184 @@
...
+        glib::Error::new(IOErrorEnum::Cancelled, "Task cancelled")
+    }
+}
+
+impl Debug for Cancelled {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Task cancelled")
+    }
+}
+
```
sdroege, line 126: Also `std::error::Error`

BiagioFesta: Done

gio/src/cancellable_future.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Checks whether the future has been cancelled.
+    ///
+    /// Note that all this method indicates is whether [`cancel`](CancellableExt::cancel)
+    /// was called. This means that it will return true even if:
+    ///   * `cancel` was called after the future had completed.
+    ///   * `cancel` was called while the future was being polled.
+    #[inline]
+    pub fn is_cancelled(&self) -> bool {
+        self.cancellable.is_cancelled()
+    }
```
sdroege, line 73: Should there maybe be a getter for the `&Cancellable`?

sdroege (approved):

gio/src/cancellable_future.rs
BiagioFesta: I agree having an accessor to the inner cancellable might be useful as generic API. 

However, I would keep the function `is_cancelled()` because I think it gives a handy shortcut and it allows the possibility to add a minimal documentation about the usage with the polling (see the current documentation).

Don't you think?

sdroege: Yes, I meant in addition to that function :)

sdroege (approved):

sdroege (approved):

PR #774, opened 2022-09-29
glib: Don't allow passing a construct(-only) property twice during ob…

sdroege:
…ject construction

GLib also does not allow this and would give a critical warning.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1365,18 +1365,33 @@ impl Object {
                 .ok_or_else(|| bool_error!("Can't retrieve class for type '{}'", type_))?;
             let pspecs = klass.list_properties();
 
-            properties
-                .iter()
-                .map(|(name, value)| {
-                    let pspec = pspecs.iter().find(|p| p.name() == *name).ok_or_else(|| {
-                        bool_error!("Can't find property '{}' for type '{}'", name, type_)
-                    })?;
+            let mut validated_properties =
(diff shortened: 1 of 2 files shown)


sdroege:
CC @MathieuDuponchelle

sdroege:
Sure, done

glib/src/object.rs
```diff
@@ -1365,18 +1365,33 @@ impl Object {
...
+                })?;
+
+                if (pspec.flags().contains(crate::ParamFlags::CONSTRUCT)
+                    || pspec.flags().contains(crate::ParamFlags::CONSTRUCT_ONLY))
+                    && properties[0..idx]
+                        .iter()
+                        .any(|(other_name, _)| name == other_name)
+                {
+                    return Err(bool_error!(
+                        "Can't set property '{}' for type '{}' twice",
```
MathieuDuponchelle, line 1382: Probably worth explaining why ("because it is marked as construct / construct-only")

glib/src/object.rs
```diff
@@ -1395,18 +1410,33 @@ impl Object {
...
```
MathieuDuponchelle, line 1427: and here :)

glib/src/object.rs
sdroege: Good point, done!

MathieuDuponchelle:
patch lgtm

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #775, opened 2022-09-30
glib-macros: Add thread_local_object macro

AaronErhardt:
In some cases, it is useful to have access to an object (or widget) as a static variable. Unfortunately, the thread_local API of the standard library is pretty inconvenient and requires a couple of lines of boilerplate to setup a convenient wrapper function.

This new macro simplifies this process. Of course it's really not the most important thing to have, but I wanted to submit a PR here first instead of making this only available in Relm4.

### DIFF
--- a/glib-macros/src/thread_local_object.rs
+++ b/glib-macros/src/thread_local_object.rs
@@ -1,0 +1,72 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use proc_macro2::TokenStream;
+use quote::{quote, quote_spanned, ToTokens};
+use syn::{
+    parse::{Parse, ParseStream},
+    spanned::Spanned,
+    token, Expr, Ident, TypePath,
+};
+
+pub(crate) struct ThreadLocalObjectTokens {
+    name: Ident,
+    ty: TypePath,
+    init_expr: Option<Expr>,
+}
+
+impl Parse for ThreadLocalObjectTokens {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let name: Ident = input.parse()?;
+        let _comma: token::Comma = input.parse()?;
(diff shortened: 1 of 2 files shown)


glib-macros/src/thread_local_object.rs
```diff
@@ -0,0 +1,72 @@
...
+            name.span(),
+        );
+
+        tokens.extend(quote! {
+            mod #mod_name {
+                use super::*;
+                ::std::thread_local!(static THREAD_LOCAL_OBJ: #ty = #init_stream);
+
+                pub fn #name() -> #ty {
+                    THREAD_LOCAL_OBJ.with(|w| w.clone())
```
sdroege, line 65: Why not use something from `once_cell` here? Then you wouldn't have to export a function but could export a value that derefs to the object

glib-macros/src/thread_local_object.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        Ok(Self {
+            name,
+            ty,
+            init_expr,
+        })
+    }
+}
+
+impl ToTokens for ThreadLocalObjectTokens {
```
sdroege, line 38: This could also just be a declarative macro in the glib crate or not? :) Having it as a proc-macro just complicates things

glib-macros/src/thread_local_object.rs
AaronErhardt: Wow, that actually works :sweat_smile:. Because `unsync::Lazy::new` is const you can use it in global const variables. I wasn't aware that you could use non-`Send` values in global constants, but that seems to work pretty well actually. I guess this makes this PR rather obsolete then...

sdroege: Yay, less code :)

PR #777, opened 2022-10-05
glib-macros: main macro for simple async applications

BiagioFesta:
I am interested into https://github.com/gtk-rs/gtk-rs-core/issues/31.
I would like to contribute if/as I can.

This PR is more a request for comment, and it might go in that direction of having more ergonomic constructs for async code rust/glib.

It is deliberately inspired by https://docs.rs/tokio/latest/tokio/attr.main.html

### DIFF
--- a/examples/gio_async_tls/main.rs
+++ b/examples/gio_async_tls/main.rs
@@ -7,13 +7,8 @@ use std::pin::Pin;
 
 use std::error::Error;
 
-fn main() -> Result<(), Box<dyn Error>> {
-    // Get the default main context and run our async function on it
-    let main_context = glib::MainContext::default();
-    main_context.block_on(run())
-}
-
-async fn run() -> Result<(), Box<dyn Error>> {
+#[glib::main]
+async fn main() -> Result<(), Box<dyn Error>> {
     // Connect to https://www.rust-lang.org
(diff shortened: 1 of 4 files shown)


BiagioFesta:
Honestly, I did not think about that case. 

Maybe the macro can be expanded with `args` to build a wrapper for `gio::Application` too. Or maybe it is not worth it.

If any, please leave your suggestion.

Otherwise I think we can reject and close this PR.

sdroege:
I think this needs some thinking for how to handle that or if that's possible at all.

jf2048:
How about:

```rust
#[gio::main(
    app(
        class = "MyApplication",
        id = "org.example.MyApp",
        flags = "gio::ApplicationFlags::HANDLES_OPEN",
    )
)]
// leave fields blank to defaults to using gio::Application with empty flags
#[gio::main(app(id = "org.example.MyApp"))]
```

Would expand to this:

```rust
gio::prelude::ApplicationExtManual::run(
    &glib::Object::new::<MyApplication>(&[
        ("application-id", &"org.example.MyApp"),
        ("flags", &(gio::ApplicationFlags::HANDLES_OPEN)),
    ])
)
```

sdroege:
How should we move forward here?

PR #778, opened 2022-10-06
Activate dox feature wherever available

pentamassiv:
Whenever a crate is published on crates.io, its documentation is automatically built on docs.rs regardless of the documentation field in the Cargo.toml file. On docs.rs it is possible to look at all previous versions. This is why there is value in having working builds on docs.rs. Having broken builds on docs.rs also prevents crates using that library from providing their documentation on docs.rs. Analogous to https://gitlab.gnome.org/World/Rust/libadwaita-rs/-/merge_requests/50, I enabled the dox feature wherever it is available to ensure the docs always build, even if the dependencies are not available.

I also fixed a broken link in the documentation

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1365,18 +1365,33 @@ impl Object {
                 .ok_or_else(|| bool_error!("Can't retrieve class for type '{}'", type_))?;
             let pspecs = klass.list_properties();
 
-            properties
-                .iter()
-                .map(|(name, value)| {
-                    let pspec = pspecs.iter().find(|p| p.name() == *name).ok_or_else(|| {
-                        bool_error!("Can't find property '{}' for type '{}'", name, type_)
-                    })?;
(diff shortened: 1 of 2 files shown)


gdk-pixbuf/sys/Cargo.toml
```diff
@@ -21,7 +21,7 @@ shell-words = "1.0.0"
 tempfile = "3"
 
 [features]
-dox = []
+dox = ["glib/dox", "gio/dox", "gobject/dox"]
```
bilelmoussaoui, line 24: You will have to teach gir as it is what generates the sys/Cargo.toml to do this automatically.

pentamassiv: Okay, I'll try that instead.

pentamassiv:
Fixed, since https://github.com/gtk-rs/gtk-rs-core/pull/798 was merged

PR #780, opened 2022-10-07
Expose some fields of GDBusNodeInfo

abergmeier:
This exposes fields as functions:
- path
- interfaces
- nodes

### DIFF
--- a/gio/src/dbus_node_info.rs
+++ b/gio/src/dbus_node_info.rs
@@ -1,0 +1,44 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::ffi::CStr;
+
+use glib::translate::ToGlibPtr;
+
+use crate::{DBusInterfaceInfo, DBusNodeInfo};
+
+impl DBusNodeInfo {
+    pub fn path(&self) -> Option<&str> {
+        let c_obj = self.to_glib_none().0;
+        unsafe {
+            let path = (*c_obj).path;
+            if path.is_null() {
+                return None;
+            }
+            let c_str = CStr::from_ptr(path);
+            Some(c_str.to_str().unwrap())
+        }
+    }
+
+    pub fn interfaces(&self) -> &[DBusInterfaceInfo] {
+        let c_obj = self.to_glib_none().0;
+        unsafe {
+            let c_ii = (*c_obj).interfaces;
+            if c_ii.is_null() {
+                return &[];
+            }
+
+            glib::collections::PtrSlice::from_glib_borrow(c_ii)
+        }
+    }
+
+    pub fn nodes(&self) -> &[DBusNodeInfo] {
+        let c_obj = self.to_glib_none().0;
+        unsafe {
+            let c_ni = (*c_obj).nodes;
+            if c_ni.is_null() {
+                return &[];
+            }
+            glib::collections::PtrSlice::from_glib_borrow(c_ni)
+        }
+    }
+}
--- a/gio/src/lib.rs
+++ b/gio/src/lib.rs
@@ -34,6 +34,7 @@ pub use self::dbus_connection::{
 };
 mod dbus_message;
 mod dbus_method_invocation;
+mod dbus_node_info;
 #[cfg(any(feature = "v2_72", feature = "dox"))]
 #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_72")))]
 mod debug_controller_dbus;
--- a/glib/src/collections.rs
+++ b/glib/src/collections.rs
@@ -145,6 +145,18 @@ impl<
 {
     // rustdoc-stripper-ignore-next
     /// Borrows a static C array.
+    pub unsafe fn from_glib_borrow<'a>(ptr: *const <T as GlibPtrDefault>::GlibType) -> &'a [T] {
+        let mut len = 0;
+        if !ptr.is_null() {
+            while !(*ptr.add(len)).is_null() {
+                len += 1;
+            }
+        }
+        Self::from_glib_borrow_num(ptr, len)
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Borrows a static C array.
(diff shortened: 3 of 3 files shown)


gio/src/dbus_node_info.rs
```diff
@@ -0,0 +1,38 @@
...
+        let c_obj = self.to_glib_none().0;
+        let path = unsafe { (*c_obj).path };
+        if path.is_null() {
+            return "";
+        }
+        let c_str = unsafe { CStr::from_ptr(path) };
+        c_str.to_str().unwrap()
+    }
+
+    pub fn interfaces(self) -> Vec<DBusInterfaceInfo> {
```
bilelmoussaoui, line 20: Do you think you can also expose the fields of the other DBus*Info types?

abergmeier:
Not totally sure whether ownership semantics are correct.

gio/src/dbus_node_info.rs
abergmeier: I could but I currently have no use case for them. And I feel kind of bad adding too much code that is not backed up by tests.

gio/src/dbus_node_info.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use std::ffi::CStr;
+
+use glib::translate::ToGlibPtr;
+
+use crate::{DBusInterfaceInfo, DBusNodeInfo};
+
+impl DBusNodeInfo {
+    pub fn path(self) -> &'static str {
+        let c_obj = self.to_glib_none().0;
```
sdroege, line 11: I'd make this one big `unsafe` block in this function as it's all related really

gio/src/dbus_node_info.rs
```diff
@@ -0,0 +1,38 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::ffi::CStr;
+
+use glib::translate::ToGlibPtr;
+
+use crate::{DBusInterfaceInfo, DBusNodeInfo};
+
+impl DBusNodeInfo {
+    pub fn path(self) -> &'static str {
```
sdroege, line 10:
```suggestion
    pub fn path<'a>(&'a self) -> &'a str {
```

probably. Can this be `NULL`, if so it should probably be

```suggestion
    pub fn path<'a>(&'a self) -> Option<&'a str> {
```

gio/src/dbus_node_info.rs
sdroege, line 20:
```suggestion
    pub fn interfaces<'a>(&'a self) -> &'a [DBusInterfaceInfo] {
```

via https://gtk-rs.org/gtk-rs-core/git/docs/glib/collections/struct.PtrSlice.html#method.from_glib_borrow_num should also work. We should add a non-`_num` version of this.

gio/src/dbus_node_info.rs
```diff
@@ -0,0 +1,38 @@
...
+        let c_obj = self.to_glib_none().0;
+        let c_ii = unsafe { (*c_obj).interfaces };
+        if c_ii.is_null() {
+            return Vec::new();
+        }
+
+        unsafe { glib::translate::FromGlibPtrContainer::from_glib_none(c_ii) }
+    }
+
+    pub fn nodes(self) -> Vec<DBusNodeInfo> {
```
sdroege, line 30:
```suggestion
    pub fn nodes<'a>(&'a self) -> &'a [DBusNodeInfo] {
```

as above

gio/src/dbus_node_info.rs
abergmeier: Done

gio/src/dbus_node_info.rs
abergmeier: Done

gio/src/dbus_node_info.rs
abergmeier: Done

gio/src/dbus_node_info.rs
abergmeier: Done

gio/src/dbus_node_info.rs
abergmeier: Doing this fails the CI: https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes

sdroege: Oh yes you can omit all the `'a` here, sorry

abergmeier: Done

gio/src/dbus_node_info.rs
```diff
@@ -0,0 +1,44 @@
...
+            let path = (*c_obj).path;
+            if path.is_null() {
+                return None;
+            }
+            let c_str = CStr::from_ptr(path);
+            Some(c_str.to_str().unwrap())
+        }
+    }
+
+    pub fn interfaces(self) -> &'static [DBusInterfaceInfo] {
```
sdroege, line 22:
```suggestion
    pub fn interfaces(&self) -> &[DBusInterfaceInfo] {
```

gio/src/dbus_node_info.rs
```diff
@@ -0,0 +1,44 @@
...
+            let c_ii = (*c_obj).interfaces;
+            if c_ii.is_null() {
+                return &[];
+            }
+
+            glib::collections::PtrSlice::from_glib_borrow(c_ii)
+        }
+    }
+
+    pub fn nodes(self) -> &'static [DBusNodeInfo] {
```
sdroege, line 34:
```suggestion
    pub fn nodes(&self) -> &[DBusNodeInfo] {
```

gio/src/dbus_node_info.rs
abergmeier: Done

gio/src/dbus_node_info.rs
abergmeier: Done

sdroege (approved):

sdroege:
Thanks!

PR #782, opened 2022-10-08
glib/BindingBuilder: add concrete type based transform setters

fengalin:
Not sure about the naming though :)

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/382

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -2146,12 +2146,12 @@ pub trait ObjectExt: ObjectType {
     ///
     /// The binding can be unidirectional or bidirectional and optionally it is possible to
     /// transform the property values before they're passed to the other object.
-    fn bind_property<'a, O: ObjectType>(
+    fn bind_property<'a, 'f, 't, O: ObjectType>(
         &'a self,
         source_property: &'a str,
         target: &'a O,
         target_property: &'a str,
-    ) -> BindingBuilder<'a>;
+    ) -> BindingBuilder<'a, 'f, 't>;
 
     // rustdoc-stripper-ignore-next
     /// Returns the strong reference count of this object.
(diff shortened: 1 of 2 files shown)


glib/src/object.rs
```diff
@@ -3572,6 +3591,25 @@ impl<'a> BindingBuilder<'a> {
...
+    >(
+        self,
+        func: F,
+    ) -> Self {
+        Self {
+            transform_to: Some(Box::new(move |binding, from_value| {
+                let from_value = from_value.get().expect("Wrong value type");
+                func(binding, from_value).map(|r| r.to_value())
+            })),
+            ..self
```
GuillaumeGomez, line 3609: I didn't know it was a thing. :o

glib/src/object.rs
```diff
@@ -3500,23 +3500,23 @@ impl<T: ObjectType> From<WeakRef<T>> for SendWeakRef<T> {
...
+pub struct BindingBuilder<'a, 'f, 't> {
     source: &'a ObjectRef,
     source_property: &'a str,
     target: &'a ObjectRef,
     target_property: &'a str,
     flags: crate::BindingFlags,
-    transform_to: TransformFn,
-    transform_from: TransformFn,
+    transform_from: TransformFn<'f>,
+    transform_to: TransformFn<'t>,
```
sdroege, line 3516: Why did you change the order? :)

glib/src/object.rs
```diff
@@ -3560,6 +3560,25 @@ impl<'a> BindingBuilder<'a> {
         }
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Transform changed property values from the target object to the source object with the given closure.
+    pub fn transform_from_concrete<
```
sdroege, line 3565: I would call this one `transform_from()` and the other one `transform_from_with_values()` or so

glib/src/object.rs
fengalin: So the signature for the Builder uses `<'a, 'f, 't>` instead of `<'a, 't, 'f>`. :)

glib/src/object.rs
fengalin: All right!

glib/src/object.rs
sdroege: OK, so no deeper reason :) Thanks

fengalin:
Should be good now! :)

sdroege:

bilelmoussaoui (approved):

PR #784, opened 2022-10-08
pango: Manually implement GlyphItemIter and AttrIterator with lifetimes

### LINKED ISSUE #610: [BUG] pango::GlyphItemIter is unsound
**Bug description**

The struct holds a borrowed pointer to the `GlyphItem`. It needs to have a lifetime parameter scoped to the `GlyphItem`. Minimal crash reproducer:

```rust
fn main() {
    const LOREM: &str = "\
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor \
incididunt ut labore et dolore magna aliqua. Ut enim ad sapientiam perveniri \
potest, non paranda nobis solum ea, sed fruenda etiam sapientia est; sive hoc \
difficile est.";

    use pango::prelude::*;

    let mut iter = {
        let map = pangocairo::FontMap::default().unwrap();
        let pc = map.create_context().unwrap();
        let layout = pango::Layout::new(&pc);
        layout.set_width(256 * pango::SCALE);
        layout.set_height(256 * pango::SCALE);
        layout.set_text(LOREM);
        let mut layout_iter = layout.iter().unwrap();
        let item = layout_iter.run().unwrap();
        pango::GlyphItemIter::new_start(&item, LOREM).unwrap()
    };
    loop {
        println!(
            "{} {} {} {} {} {}",
            iter.start_glyph(), iter.start_index(), iter.start_char(),
            iter.end_glyph(), iter.end_index(), iter.end_char(),
        );
        if !iter.next_cluster() {
            break;
        }
    }
}
```

**Backtrace**

```
Stack trace of thread 2735378:
#0  0x00007fb484e15e0e pango_glyph_item_iter_next_cluster (libpango-1.0.so.0 + 0x28e0e)
#1  0x00005612ab9df66e n/a (pangotest + 0x2866e)
```

jf2048:
Also make them implement `IntoIterator`. The test is in `pangocairo` to get access to the font map, but is there a better way to do that so the test can be moved into `pango`?

Fixes #610


jf2048:
The doc embedding seems to have problems with lifetime `impl`s

sdroege:
> The test is in `pangocairo` to get access to the font map, but is there a better way to do that so the test can be moved into `pango`

Seems fine to me like this :)

pango/src/attr_iterator.rs
```diff
@@ -1,35 +1,220 @@
...
+#[cfg(any(feature = "v1_44", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_44")))]
+impl<'list> glib::StaticType for AttrIterator<'list> {
+    #[inline]
+    fn static_type() -> glib::Type {
+        unsafe { from_glib(ffi::pango_attr_iterator_get_type()) }
+    }
+}
+
+impl<'list> AttrIterator<'list> {
```
sdroege, line 50: I assume these methods are all just copy&paste of the autogenerated code plus adding the lifetime?

pango/src/attr_iterator.rs
```diff
@@ -1,35 +1,220 @@
...
+            from_glib_none(ffi::pango_attr_iterator_get(
+                mut_override(self.to_glib_none().0),
+                type_.into_glib(),
+            ))
+        }
+    }
+
+    #[doc(alias = "pango_attr_iterator_get_attrs")]
+    #[doc(alias = "get_attrs")]
+    pub fn attrs(&self) -> Vec<Attribute> {
```
sdroege, line 63: Not sure if it matters but you could use `glib::collections::SList` here

pango/src/glyph_item_iter.rs
```diff
@@ -1,63 +1,186 @@
...
+            from_glib(ffi::pango_glyph_item_iter_prev_cluster(
+                self.to_glib_none_mut().0,
+            ))
+        }
+    }
     #[doc(alias = "pango_glyph_item_iter_init_end")]
-    pub fn new_end(glyph_item: &GlyphItem, text: &str) -> Result<Self, glib::BoolError> {
+    pub fn new_end(
+        glyph_item: &'item GlyphItem,
+        text: impl AsRef<GStr> + 'text,
```
sdroege, line 43: Do we want to change this here to `AsRef<GStr>` already, or defer that to next release with the other related changes?

pango/src/glyph_item_iter.rs
```diff
@@ -1,63 +1,186 @@
...
                 Err(glib::bool_error!("Failed to create glyph item iter"))
             }
         }
     }
 
-    pub fn start_char(&self) -> i32 {
-        unsafe { (*self.to_glib_none().0).start_char }
+    #[inline]
+    pub fn glyph_item(&self) -> &'item GlyphItem {
+        unsafe { mem::transmute(&self.inner.glyph_item) }
```
sdroege, line 92:
```suggestion
        unsafe { &*(&self.inner.glyph_item as *const _ as *const GlyphItem) }
```

Is slightly safer (it checks for more things than transmute)

sdroege (commented):
Looks good to me otherwise

pango/src/glyph_item_iter.rs
jf2048: We can, but either way if you want to remove `AsRef<GStr>` then the `GlyphItemIter` Rust struct will have to be able to store a string. When we do `IntoGStr` then it will have to be generic on that

pango/src/attr_iterator.rs
jf2048: Yes, except `font` was rewritten because that one was just broken

pango/src/glyph_item_iter.rs
jf2048: Changed to store a `GString` for now

sdroege (approved):

PR #785, opened 2022-10-09
Pango odds and ends

jf2048:
The only thing missing now is the C bitfields


glib/src/auto/markup_parse_context.rs
```diff
@@ -0,0 +1,101 @@
...
+    match fn {
+        ref => |ptr| ffi::g_markup_parse_context_ref(ptr),
+        unref => |ptr| ffi::g_markup_parse_context_unref(ptr),
+        type_ => || ffi::g_markup_parse_context_get_type(),
+    }
+}
+
+impl MarkupParseContext {
+    //#[doc(alias = "g_markup_parse_context_new")]
+    //pub fn new(parser: /*Ignored*/&MarkupParser, flags: /*Ignored*/MarkupParseFlags, user_data: /*Unimplemented*/Option<Basic: Pointer>) -> MarkupParseContext {
```
sdroege, line 23: Is it useful without the constructor?

pango/src/glyph_string.rs
```diff
@@ -32,4 +32,63 @@ impl GlyphString {
             }
         }
     }
+
+    pub fn log_clusters(&self) -> &[i32] {
+        unsafe {
+            let ptr = (*self.to_glib_none().0).log_clusters as *const i32;
```
sdroege, line 38: https://gtk-rs.org/gtk-rs-core/git/docs/glib/collections/struct.Slice.html#method.from_glib_borrow_num if you feel like it :)

sdroege (commented):
Looks good to me otherwise

pango/src/script_iter.rs
```diff
@@ -0,0 +1,173 @@
...
+impl<'text> glib::StaticType for ScriptIter<'text> {
+    #[inline]
+    fn static_type() -> glib::Type {
+        unsafe { from_glib(ffi::pango_script_iter_get_type()) }
+    }
+}
+
+impl<'text> ScriptIter<'text> {
+    #[doc(alias = "pango_script_iter_new")]
+    pub fn new(text: impl AsRef<GStr> + 'text) -> Self {
```
sdroege, line 53: Do we want to change this here to `AsRef<GStr>` already, or defer that to next release with the other related changes?

Note: only for parameters this is potentially problematic

glib/src/auto/markup_parse_context.rs
jf2048: Yes, you can create one with `pango_markup_parser_new` which is why I included it here. I can try to bind `g_markup_parse_context_new` in another PR

pango/src/glyph_string.rs
jf2048: Fixed, and added `Slice::from_glib_borrow_num_mut` too

pango/src/script_iter.rs
jf2048: I think this one should stay `AsRef<GStr>` for now, with an owned `GString` it forces the `Iterator` implementation to copy the string every time.

With `IntoGStr` we can fix this and also optimize out those `GStr::from_ptr` calls.

sdroege: Works for me

sdroege (dismissed):

sdroege (approved):

PR #786, opened 2022-10-09
cairo: Use AsRef<T> for custom subclassing types

### LINKED ISSUE #189: Figure out how to handle custom "subclassing" traits
See https://github.com/gtk-rs/gtk4-rs/pull/500 for context

Currently for cairo:

You have a parent type `cairo::Surface` and other sub-types like `cairo::ImageSurface`. If a function takes a `cairo::Surface` and you have a `cairo::ImageSurface`, you pass it as such `some_func(&*image_surface)`  that's because there's a `impl Deref for cairo::WhateverSurface` with a `target = cairo::Surface`

for gtk4-rs, we use `AsRef` in those functions, we can then just call `some_func(&some_render_node)` , we should probably decide on something and use it across the crates. 

I would vote for `AsRef` as it's kind of similar to `IsA` in terms of "syntax"  and it's less magical than a Deref implementation :)

bilelmoussaoui:
Instead of requiring the user to go through the Deref implementation
Fixes #189

### DIFF
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -161,8 +161,8 @@ impl Context {
         status_to_result(status)
     }
 
-    pub fn new(target: &Surface) -> Result<Context, Error> {
-        let ctx = unsafe { Self::from_raw_full(ffi::cairo_create(target.to_raw_none())) };
+    pub fn new(target: impl AsRef<Surface>) -> Result<Context, Error> {
+        let ctx = unsafe { Self::from_raw_full(ffi::cairo_create(target.as_ref().to_raw_none())) };
         ctx.status().map(|_| ctx)
     }
 
@@ -223,7 +223,8 @@ impl Context {
     }
 
     #[doc(alias = "cairo_set_source")]
-    pub fn set_source(&self, source: &Pattern) -> Result<(), Error> {
+    pub fn set_source(&self, source: impl AsRef<Pattern>) -> Result<(), Error> {
+        let source = source.as_ref();
         source.status()?;
         unsafe {
             ffi::cairo_set_source(self.0.as_ptr(), source.to_raw_none());
@@ -238,7 +239,13 @@ impl Context {
     }
 
     #[doc(alias = "cairo_set_source_surface")]
-    pub fn set_source_surface(&self, surface: &Surface, x: f64, y: f64) -> Result<(), Error> {
+    pub fn set_source_surface(
+        &self,
+        surface: impl AsRef<Surface>,
+        x: f64,
+        y: f64,
+    ) -> Result<(), Error> {
+        let surface = surface.as_ref();
         surface.status()?;
         unsafe {
             ffi::cairo_set_source_surface(self.0.as_ptr(), surface.to_raw_none(), x, y);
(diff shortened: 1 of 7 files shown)


cairo/src/context.rs
```diff
@@ -474,17 +479,17 @@ impl Context {
     }
 
     #[doc(alias = "cairo_mask")]
-    pub fn mask(&self, pattern: &Pattern) -> Result<(), Error> {
-        pattern.status()?;
-        unsafe { ffi::cairo_mask(self.0.as_ptr(), pattern.to_raw_none()) };
+    pub fn mask(&self, pattern: impl AsRef<Pattern>) -> Result<(), Error> {
```
sdroege, line 482:
```suggestion
    pub fn mask(&self, pattern: &impl AsRef<Pattern>) -> Result<(), Error> {
```

etc or not? There's no reason to take these things by value :)

cairo/src/context.rs
```diff
@@ -223,10 +223,10 @@ impl Context {
     }
 
     #[doc(alias = "cairo_set_source")]
-    pub fn set_source(&self, source: &Pattern) -> Result<(), Error> {
-        source.status()?;
+    pub fn set_source(&self, source: impl AsRef<Pattern>) -> Result<(), Error> {
+        source.as_ref().status()?;
```
sdroege, line 227: Ideally you everywhere do something like `let source = source.as_ref()` in the beginning instead of calling it multiple times. Less line-noise

sdroege (commented):

sdroege:
> Instead of requiring the user to go through the Deref implementation

The user would just do `bla(&surface)` currently or not? Autoderef would figure out the correct type. The same would of course also still work now :)

Also good that you kept the `Deref` impls: otherwise you'd need to add a lot of extension traits which would be unnecessarily verbose in the single-inheritance case. Do you implement `Deref` also in gsk/gtk or only `AsRef`?

cairo/src/context.rs
bilelmoussaoui: I thought, Impl AsRef<T> would already take things by ref

cairo/src/context.rs
bilelmoussaoui: that is true, will change that

bilelmoussaoui:
> The user would just do bla(&surface) currently or not? Autoderef would figure out the correct type. The same would of course also still work now :)

Yes, and it is clearer in terms of API from the type system directly.

> Do you implement Deref also in gsk/gtk or only AsRef?

I think so yes

cairo/src/context.rs
sdroege: Indeed: https://doc.rust-lang.org/std/convert/trait.AsRef.html#impl-AsRef%3CU%3E-for-%26T

Does that work? You can now pass `&surface` and `surface`, I guess? I hope we have some code in the examples that does this :)

sdroege:
> Yes, and it is clearer in terms of API from the type system directly.

Also documentation-wise. It's clearer that you can pass any kind of surface, not just `Surface`.

cairo/src/context.rs
bilelmoussaoui: Just tried to change a test and it works with both `&surface` / `surface`

sdroege (approved):

PR #796, opened 2022-10-14
Add information about pangocairo::FontMap generation and fix some methods

GuillaumeGomez:


pangocairo/src/font_map.rs
```diff
@@ -1,38 +0,0 @@
...
-    }
-
-    #[allow(clippy::new_ret_no_self)]
-    #[doc(alias = "pango_cairo_font_map_new")]
-    pub fn new() -> Option<pango::FontMap> {
-        unsafe { from_glib_full(ffi::pango_cairo_font_map_new()) }
-    }
-
-    #[doc(alias = "pango_cairo_font_map_set_default")]
-    pub fn set_default(font_map: Option<Self>) {
```
sdroege, line 33: `set_default()` should probably stay an associated function. `font_map.set_default()` seems strange compared to `pangocairo::FontMap::set_default(&font_map)`. But document that in Gir.toml then

sdroege: But it should take a `font_map: Option<&Self>` instead of `Option<Self>`.

Also it should be an associated function because of the `Option` part.

pangocairo/src/auto/font_map.rs
```diff
@@ -23,22 +23,63 @@ impl FontMap {
     pub fn default() -> pango::FontMap {
         unsafe { from_glib_none(ffi::pango_cairo_font_map_get_default()) }
     }
+
+    #[doc(alias = "pango_cairo_font_map_new")]
+    pub fn new() -> pango::FontMap {
```
GuillaumeGomez, line 28: I didn't realize but this generation is completely wrong too.

GuillaumeGomez:
Added comments and fixed the argument.

sdroege (approved):

pangocairo/src/font_map.rs
```diff
@@ -30,7 +30,7 @@ impl FontMap {
     }
```
sdroege, line 30: ^------ this doesn't return an `Option`

sdroege: The new function is not nullable, I mean

GuillaumeGomez: Oh right!

sdroege (approved):

pangocairo/Gir.toml
```diff
@@ -106,15 +106,21 @@ name = "PangoCairo.FontMap"
 status = "generate"
 manual_traits = ["FontMapExtManual"]
     [[object.function]]
+    # Needed because cairo types don't implement `.into_glib()`.
```
bilelmoussaoui, line 109: I am a bit late but shouldn't it be possible to implement such traits for cairo types somehow?

Issue #789, opened 2022-10-11
Use trust_return_value_nullability after making sure it doesn't cause regressions


sdroege:
* [x] gdk-pixbuf ([done](https://github.com/gtk-rs/gtk-rs-core/pull/905))
* [x] gio ([done](https://github.com/gtk-rs/gtk-rs/pull/364))
* [x] glib ([done](https://github.com/gtk-rs/gtk-rs/commit/bf4198183a78a333166c4096b1843537040dd75c))
* [x] graphene
* [x] pango ([done](https://github.com/gtk-rs/gtk-rs-core/pull/797))
* [x] pangocairo ([done](https://github.com/gtk-rs/gtk-rs-core/pull/795))

CC @bilelmoussaoui @GuillaumeGomez

GuillaumeGomez:
I'll try to take a look in the next days.

GuillaumeGomez:
For pango, we need to wait for my MRs to be merged:

 * https://gitlab.gnome.org/GNOME/pango/-/merge_requests/648
 * https://gitlab.gnome.org/GNOME/pango/-/merge_requests/647
 * https://gitlab.gnome.org/GNOME/pango/-/merge_requests/646

And for new GIR files to be generated.

GuillaumeGomez:
And for `gdk-pixbuf` we need to wait for https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/146.

PR #797, opened 2022-10-16
pango: Trust upstream nullable annotations for return values

GuillaumeGomez:
Part of #789.


pango/Gir.toml
```diff
@@ -9,6 +9,7 @@ min_cfg_version = "1.40"
 use_gi_docgen = true
 single_version_file = true
 deprecate_by_min_version = true
+trust_return_value_nullability = true
```
sdroege, line 12: Also remove all the `nullable` overrides from this file then :)

GuillaumeGomez: Ok!

GuillaumeGomez:
Updated!

pango/src/auto/context.rs
```diff
@@ -30,7 +30,7 @@ glib::wrapper! {
 
 impl Context {
     #[doc(alias = "pango_context_new")]
-    pub fn new() -> Context {
+    pub fn new() -> Option<Context> {
```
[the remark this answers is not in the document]
GuillaumeGomez: https://gitlab.gnome.org/GNOME/pango/-/merge_requests/651

pango/src/auto/context.rs
```diff
@@ -83,8 +83,8 @@ impl Context {
 
     #[doc(alias = "pango_context_get_language")]
     #[doc(alias = "get_language")]
-    pub fn language(&self) -> Option<Language> {
-        unsafe { from_glib_full(ffi::pango_context_get_language(self.to_glib_none().0)) }
+    pub fn language(&self) -> Language {
```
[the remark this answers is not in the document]
GuillaumeGomez: I don't think it's possible: https://gitlab.gnome.org/GuillaumeGomez/pango/-/blob/main/pango/pango-context.c#L70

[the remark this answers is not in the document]
GuillaumeGomez: And when you set the language too: https://gitlab.gnome.org/GuillaumeGomez/pango/-/blob/main/pango/pango-context.c#L400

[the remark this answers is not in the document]
sdroege: That's why I said it looks like a bug :) Look at what the setter is actually returning: https://gitlab.gnome.org/GuillaumeGomez/pango/-/blob/main/pango/pango-context.c#L416

`set_language` is only set by `pango_context_set_language()` to the value passed in, and that can be `NULL`. It is also initialized to `NULL` on construction (implicitly).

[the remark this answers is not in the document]
GuillaumeGomez: Oh indeed.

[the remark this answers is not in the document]
sdroege: I would keep it as nullable here for now and report, separately, an issue against pango to maybe have this function fixed if that's indeed unintentionally strange.

sdroege:
For moving forward here, I would suggest you add overrides for the wrong nullable cases to `Gir.toml`. Then the API here is fixed at least and we can get rid of the overrides once we have the new .gir file

pango/src/auto/context.rs
```diff
@@ -84,7 +84,7 @@ impl Context {
     #[doc(alias = "pango_context_get_language")]
     #[doc(alias = "get_language")]
     pub fn language(&self) -> Option<Language> {
-        unsafe { from_glib_full(ffi::pango_context_get_language(self.to_glib_none().0)) }
+        unsafe { from_glib_none(ffi::pango_context_get_language(self.to_glib_none().0)) }
     }
 
     #[doc(alias = "pango_context_get_matrix")]
```
sdroege, line 90: Further below, `fn metrics()` should return no `Option` according to your pango MR

pango/Gir.toml
```diff
@@ -206,6 +203,14 @@ boxed_inline = true
 name = "Pango.Context"
 status = "generate"
     [[object.function]]
+    name = "new"
+        [object.function.return]
+        nullable = false
+    [[object.function]]
+    name = "get_language"
+        [object.function.return]
+        nullable = true
```
sdroege, line 212: Add links here to your MR that makes the overrides unnecessary

GuillaumeGomez: Good idea!

pango/src/auto/context.rs
GuillaumeGomez: Indeed, fixed.

sdroege:
regen check is unhappy

pango/Gir.toml
```diff
@@ -206,10 +203,26 @@ boxed_inline = true
...
 status = "generate"
     [[object.function]]
+    name = "new"
+        # Can be removed once https://gitlab.gnome.org/GNOME/pango/-/merge_requests/651
+        # has been merged.
+        [object.function.return]
+        nullable = false
+    [[object.function]]
+    name = "get_language"
+        # Can be removed once https://gitlab.gnome.org/GNOME/pango/-/merge_requests/651
```
sdroege, line 213: You don't change that one there. I think that needs a separate issue to first of all figure out how this function is supposed to work.

GuillaumeGomez: Done.

glib/src/auto/enums.rs
```diff
@@ -1346,6 +1346,10 @@ pub enum UnicodeScript {
     #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_72")))]
     #[doc(alias = "G_UNICODE_SCRIPT_MATH")]
     Math,
+    #[doc(alias = "G_UNICODE_SCRIPT_KAWI")]
+    Kawi,
+    #[doc(alias = "G_UNICODE_SCRIPT_NAG_MUNDARI")]
+    NagMundari,
```
sdroege, line 1352: These two needs to be marked as `version = 2.74` in the `Gir.toml`

sdroege (approved):

PR #799, opened 2022-10-18
implement FromValue on EnumValue

gdesmott:
not building:

error: lifetime may not live long enough
   --> glib/src/enums.rs:290:9
    |
285 | unsafe impl<'a> FromValue<'a> for &EnumValue {
    |             -- lifetime `'a` defined here
...
288 |     unsafe fn from_value(value: &'a Value) -> Self {
    |                                               ---- return type is &'1 EnumValue
289 |         let (_, v) = EnumValue::from_value(value).unwrap();
290 |         v
    |         ^ associated function was supposed to return data with lifetime `'1` but it is returning data with lifetime `'a`

### DIFF
--- a/glib/src/enums.rs
+++ b/glib/src/enums.rs
@@ -1,7 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::translate::*;
-use crate::value::Value;
+use crate::value::{FromValue, Value, ValueTypeChecker};
 use crate::Type;
 use std::ffi::CStr;
 use std::{cmp, fmt, ptr};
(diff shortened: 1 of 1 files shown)


glib/src/enums.rs
```diff
@@ -282,6 +282,40 @@ impl Ord for EnumValue {
     }
 }
 
+unsafe impl<'a> FromValue<'a> for &EnumValue {
```
sdroege, line 285:
```suggestion
unsafe impl<'a> FromValue<'a> for &'a EnumValue {
```

glib/src/enums.rs
```diff
@@ -282,6 +282,43 @@ impl Ord for EnumValue {
     }
 }
 
+unsafe impl<'a> FromValue<'a> for &'a EnumValue {
```
sdroege, line 285:
```suggestion
unsafe impl<'a, 'b> FromValue<'a> for &'b EnumValue {
```

sdroege (approved):

gdesmott:
Added a test.

glib/src/enums.rs
sdroege, line 285:
```suggestion
unsafe impl<'a, 'b> FromValue<'a> for &'b EnumValue {
```

glib/src/enums.rs
```diff
@@ -282,6 +282,43 @@ impl Ord for EnumValue {
...
+    type Checker = EnumTypeChecker;
+
+    unsafe fn from_value(value: &'a Value) -> Self {
+        let (_, v) = EnumValue::from_value(value).unwrap();
+        v
```
sdroege, line 290:
```suggestion
        // SAFETY: The enum class and its values live forever
        unsafe { std::mem::transmute(v) }
```

sdroege (approved):

PR #803, opened 2022-10-20
cairo: Fix rectangle getter

bilelmoussaoui:


sdroege (approved):
:facepalm:

sdroege:
This is a candidate for a clippy lint. If a getter/setter function is called *almost* like the field it is accessing then it's probably a mistake

PR #806, opened 2022-10-22
glib: Add `ObjectSubclass::obj()` as a shorter alias for `instance()`

sdroege:
For the lazy among us and it also mirrors `imp()` more closely, which is not called `implementation()` either.

----

CC @A6GibKm


sdroege:
Are there any opinions on this? :)

bilelmoussaoui (approved):

Issue #763, opened 2022-09-05
Bind `g_utf8` / `g_unichar` APIs


sdroege:
`std` doesn't have almost anything of that, and the crate implementing all that seems is using icu, which is huge and annoying to build. The utf8 APIs should probably take `&GStr` / `GString`, otherwise we end up creating lots of copies.

bilelmoussaoui:
Should this be closed now? cc @sdroege

RealKC:
When binding the `g_utf8` APIs I didn't include the ones that I thought were already offered by the rust stdlib through `str` (which `GStr` gets through its `Deref<Target = &str>` impl), IIRC that was one of them.

However, the `g_unichar` APIs are still missing (some of them are offered by the stdlib already, but not all), so this issue should stay open for the time being.

PR #813, opened 2022-10-23
glib: Bind more `g_utf8` APIs

RealKC:
I decided to make most of these associated functions on `GStr`, except for the utf8<->utf32 conversions.

Ref #763

### DIFF
--- a/glib/src/convert.rs
+++ b/glib/src/convert.rs
@@ -2,7 +2,7 @@
 
 use std::{io, os::raw::c_char, path::PathBuf, ptr};
 
-use crate::{translate::*, ConvertError, Error, GStr, GString, Slice};
+use crate::{translate::*, ConvertError, Error, GStr, GString, NormalizeMode, Slice};
 
 // rustdoc-stripper-ignore-next
 /// A wrapper for [`ConvertError`](crate::ConvertError) that can hold an offset into the input
@@ -318,6 +318,74 @@ pub fn locale_to_utf8(opsysstring: &[u8]) -> Result<(crate::GString, usize), Cvt
     }
 }
 
+#[doc(alias = "g_utf8_to_ucs4")]
+#[doc(alias = "g_utf8_to_ucs4_fast")]
+#[doc(alias = "utf8_to_ucs4")]
+pub fn utf8_to_utf32(str: impl AsRef<str>) -> Slice<char> {
+    unsafe {
+        let mut items_written = 0;
+
+        let str_as_utf32 = ffi::g_utf8_to_ucs4_fast(
+            str.as_ref().as_ptr().cast::<c_char>(),
+            str.as_ref().len() as _,
+            &mut items_written,
+        );
+
+        // NOTE: We assume that u32 and char have the same layout and trust that glib won't give us
+        //       invalid UTF-32 codepoints
+        Slice::from_glib_full_num(str_as_utf32.cast::<char>(), items_written as usize)
+    }
+}
+
+#[doc(alias = "g_ucs4_to_utf8")]
+#[doc(alias = "ucs4_to_utf8")]
+pub fn utf32_to_utf8(str: impl AsRef<[char]>) -> GString {
+    let mut items_read = 0;
+    let mut items_written = 0;
+    let mut error = ptr::null_mut();
+
+    unsafe {
+        let str_as_utf8 = ffi::g_ucs4_to_utf8(
+            str.as_ref().as_ptr().cast::<u32>(),
+            str.as_ref().len() as _,
+            &mut items_read,
+            &mut items_written,
+            &mut error,
+        );
+
+        assert!(
+            error.is_null(),
+            "Rust `char` should always be convertible to UTF-8"
+        );
+
+        GString::from_glib_full_num(str_as_utf8, items_written as usize)
+    }
+}
+
+#[doc(alias = "g_utf8_casefold")]
+#[doc(alias = "utf8_casefold")]
+pub fn casefold(str: impl AsRef<str>) -> GString {
+    unsafe {
+        let str = ffi::g_utf8_casefold(str.as_ref().as_ptr().cast(), str.as_ref().len() as isize);
+
+        from_glib_full(str)
+    }
+}
+
+#[doc(alias = "g_utf8_normalize")]
+#[doc(alias = "utf8_normalize")]
+pub fn normalize(str: impl AsRef<str>, mode: NormalizeMode) -> GString {
+    unsafe {
+        let str = ffi::g_utf8_normalize(
+            str.as_ref().as_ptr().cast(),
+            str.as_ref().len() as isize,
+            mode.into_glib(),
+        );
+
+        from_glib_full(str)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     #[test]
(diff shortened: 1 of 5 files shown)


glib/src/convert.rs
```diff
@@ -317,6 +317,64 @@ pub fn locale_to_utf8(opsysstring: &[u8]) -> Result<(crate::GString, usize), Cvt
     }
 }
 
+#[doc(alias = "g_utf8_to_ucs4")]
+pub fn utf8_to_utf32(str: &[u8]) -> Result<(Slice<u32>, usize), CvtError> {
```
RealKC, line 321: I initially wanted to make this return a `Slice<u32>` instead of `Slice<char>`, but I'm not sure whether the layouts of u32 and char are guaranteed to be the same, so that casting the returned pointer from `*mut u32` to `*mut char` would be sound, any thoughts?

glib/src/gstring.rs
```diff
@@ -102,6 +102,67 @@ impl GStr {
...
+        // SAFETY: GStr must be valid UTF-8
+        let str_as_utf32 =
+            unsafe { ffi::g_utf8_to_ucs4_fast(self.as_ptr(), self.len() as _, &mut items_written) };
+        unsafe { Slice::from_glib_full_num(str_as_utf32.cast(), items_written as usize) }
+    }
+}
+
+#[doc(alias = "GNormalizeMode")]
+pub enum NormalizeMode {
+    #[doc(alias = "G_NORMALIZE_DEFAULT")]
```
bilelmoussaoui, line 141: this should be auto generated, add it to the Gir.toml file & run the generator script

glib/src/gstring.rs
```diff
@@ -102,6 +102,67 @@ impl GStr {
...
+    pub fn collate(&self, other: &Self) -> Ordering {
+        unsafe { ffi::g_utf8_collate(self.as_ptr(), other.as_ptr()) }.cmp(&0)
+    }
+
+    #[doc(alias = "g_utf8_casefold")]
+    #[inline]
+    pub fn casefold(&self) -> GString {
+        let str = unsafe { ffi::g_utf8_casefold(self.as_ptr(), self.len() as isize) };
+
+        unsafe { from_glib_full(str) }
```
sdroege, line 117:
```suggestion
        unsafe {
            let str = ffi::g_utf8_casefold(self.as_ptr(), self.len() as isize);
            from_glib_full(str)
        }
```

The whole block is unsafe, not just each of those smaller blocks on their own

sdroege: (same elsewhere)

glib/src/convert.rs
sdroege: I would've expected this to be `&str -> Slice<char>` :)

glib/src/gstring.rs
```diff
@@ -102,6 +102,67 @@ impl GStr {
...
+
+    #[doc(alias = "g_utf8_collate")]
+    #[inline]
+    pub fn collate(&self, other: &Self) -> Ordering {
+        unsafe { ffi::g_utf8_collate(self.as_ptr(), other.as_ptr()) }.cmp(&0)
+    }
+
+    #[doc(alias = "g_utf8_casefold")]
+    #[inline]
+    pub fn casefold(&self) -> GString {
```
sdroege, line 114: Having these on `&GStr` requires unnecessarily the `\0` terminator to be added. That's not required for the functions here and they would as well work on a normal `&str` (that's the only difference between those two types).

RealKC: Oh yeah I didn't consider that, I guess taking `impl AsRef<str>` would be ideal then? Would these names be okay even if the functions were exposed at the module level?

sdroege: Yeah, maybe as part of the convert module?

RealKC: I decided to keep `collate` as an associated function as it takes null terminated strings, but I moved the rest to `convert`

glib/src/convert.rs
```diff
@@ -317,6 +317,65 @@ pub fn locale_to_utf8(opsysstring: &[u8]) -> Result<(crate::GString, usize), Cvt
     }
 }
 
+#[doc(alias = "g_utf8_to_ucs4")]
+#[doc(alias = "g_utf8_to_ucs4_fast")]
+pub fn utf8_to_utf32(str: &str) -> Slice<char> {
```
sdroege, line 322:
```suggestion
pub fn utf8_to_utf32(str: impl AsRef<str>) -> Slice<char> {
```

for consistency with the others maybe

glib/src/convert.rs
```diff
@@ -317,6 +317,65 @@ pub fn locale_to_utf8(opsysstring: &[u8]) -> Result<(crate::GString, usize), Cvt
...
+        );
+
+        // NOTE: We assume that u32 and char have the same layout and trust that glib won't give us
+        //       invalid UTF-32 codepoints
+        Slice::from_glib_full_num(str_as_utf32.cast::<char>(), items_written as usize)
+    }
+}
+
+#[doc(alias = "g_ucs4_to_utf8")]
+pub fn utf32_to_utf8(str: &[char]) -> GString {
```
sdroege, line 339:
```suggestion
pub fn utf32_to_utf8(str: AsRef<[char]>) -> GString {
```

glib/src/convert.rs
```diff
@@ -317,6 +317,65 @@ pub fn locale_to_utf8(opsysstring: &[u8]) -> Result<(crate::GString, usize), Cvt
...
+
+    unsafe {
+        let str_as_utf8 = ffi::g_ucs4_to_utf8(
+            str.as_ptr().cast::<u32>(),
+            str.len() as _,
+            &mut items_read,
+            &mut items_written,
+            &mut error,
+        );
+
```
sdroege, line 352: Maybe you want to `assert!(error.is_null())` here. Can this actually not fail?

glib/src/gstring.rs
```diff
@@ -102,6 +102,11 @@ impl GStr {
     pub fn as_c_str(&self) -> &CStr {
         unsafe { CStr::from_bytes_with_nul_unchecked(self.to_bytes_with_nul()) }
     }
+
+    #[doc(alias = "g_utf8_collate")]
+    pub fn collate(&self, other: impl AsRef<GStr>) -> Ordering {
+        unsafe { ffi::g_utf8_collate(self.as_ptr(), other.as_ref().as_ptr()) }.cmp(&0)
```
sdroege, line 108: Sad that there is no variant that takes a string length

glib/src/convert.rs
bilelmoussaoui, line 322: all the new manually written functions should be added to the Gir.toml and marked as manual, otherwise they won't be automatically documented

bilelmoussaoui: also if the function was renamed like in this case, it has to be configured in the Gir.toml as well for the docs links to work properly if there are any

RealKC: At least for the `utf8_` functions this isn't trivial, as there's a section in the `Gir.toml` ignoring all of them, and adding specific sections for individual ones didn't seem to generate docs for them in my attempts, but maybe I did it wrong?

For the `ucs4_to_utf8` function, we could just replace the current regex that ignores everything starting with `ucs4_` to one only ignoring `ucs4_to_ucs16`, but I still couldn't get it to generate docs just for `utf32_to_utf8` (renamed from `ucs4_to_utf8`) even with that.

Any tips?

glib/src/convert.rs
sdroege: Worth adding a comment and assertion IMHO :)

glib/src/convert.rs
bilelmoussaoui: If you can push your changes, I can try and see what is going wrong afterwards.

RealKC: I pushed my changes in 456cecdb6a34ae0947cf88cbde01af8c0a882f92

sdroege: @bilelmoussaoui What's the status here?

bilelmoussaoui: I wouldn't block the PR on that, I will try to find some time during the end of year holidays and improve the docs situation for glib/gio & other gtk-rs-core libraries.

PR #816, opened 2022-10-25
Fix new clippy lints

GuillaumeGomez:


GuillaumeGomez:
New warning is wild:

```
error: use of deprecated constant `libc::MSG_OOB`: recvmmsg call expects an unsigned type on musl/emscripten
  --> gio/sys/src/manual.rs:32:64
   |
32 |     pub const G_SOCKET_MSG_OOB: super::GSocketMsgFlags = libc::MSG_OOB;
   |                                                                ^^^^^^^
   |
   = note: `-D deprecated` implied by `-D warnings`
```

Any idea what's going on?

AaronErhardt:
@GuillaumeGomez looks like a bunch of glibc types were deprecated with no real alternative in place yet (so this has nothing to do with clippy actually). It seems like this got reverted already: https://github.com/rust-lang/libc/pull/2978

GuillaumeGomez:
Thanks for the information @AaronErhardt !

sdroege:
Yeah, let's not worry about those for now.

gio/src/lib.rs
```diff
@@ -7,6 +7,7 @@
 #![allow(clippy::missing_safety_doc)]
 #![allow(clippy::upper_case_acronyms)]
 #![allow(clippy::non_send_fields_in_send_ty)]
+#![allow(clippy::should_implement_trait)]
```
sdroege, line 10: This is generally a good lint. Can you only allow it for the one instance where it's wrong?

gio/src/list_model.rs
```diff
@@ -103,8 +103,8 @@ impl<'a, T: IsA<glib::Object>> Iterator for ListModelIter<'a, T> {
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let n = (self.reverse_pos - self.i) as usize;
-        (n as usize, Some(n))
+        let n: usize = (self.reverse_pos - self.i) as _;
+        (n as _, Some(n))
```
sdroege, line 107:
```suggestion
        (n, Some(n))
```

You already make it an `usize` before. This warning was actually correct

glib/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 #![allow(clippy::missing_safety_doc)]
 #![allow(clippy::wrong_self_convention)]
 #![allow(clippy::non_send_fields_in_send_ty)]
+#![allow(clippy::should_implement_trait)]
```
sdroege, line 8: also here

pangocairo/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+#![allow(clippy::should_implement_trait)]
```
sdroege, line 3: Also here

gio/src/lib.rs
GuillaumeGomez: That would require another gir update but I can do that.

sdroege: As discussed with @bilelmoussaoui, just put it on the `auto` module for now via `src/lib.rs`

pangocairo/src/lib.rs
GuillaumeGomez: Ah too late. I put the attribute on the methods directly.

gio/src/list_model.rs
sdroege: You didn't change anything here :)

GuillaumeGomez: I did one `git checkout --` too much.

GuillaumeGomez:
Updated!

sdroege (approved):

PR #817, opened 2022-10-25
image: Rebuild once every week

AaronErhardt:
Rebuild the gtk-rs-core image once every week. The current image is already 5 months old and I suspect this to be the reason why the relm4 docs pipeline [fails](https://github.com/Relm4/docs/actions/runs/3322872513/jobs/5492499363#step:3:331) with some weird permission errors while building libpanel (but I might be wrong). 

Anyway, there's no good reason to pull the image from Fedora rawhide and then let it age for several weeks, right?
I think weekly builds are reasonable and if desired, I will send a PR for the gtk-rs image as well.

Btw. is there any reason why all the shell commands are connected with `&& \` in one single docker instruction?

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,6 +1,9 @@
 name: github packages
 
 on:
+  # Rebuild the container once every week
+  schedule:
+  - cron: '0 0 * * 1'
   push:
     branches:
       - "master"


jf2048:
> Btw. is there any reason why all the shell commands are connected with `&& \` in one single docker instruction?

See this comment: https://github.com/gtk-rs/gtk-rs-core/pull/676#discussion_r868243599

jf2048:
The issue with `glib-compile-resources` seems to be because of outdated libseccomp, so it's likely rebuilding the image will fix it: https://github.com/moby/moby/issues/43595

There is another issue with the freetype subproject, I've tried to update that too to see if it fixes anything: https://gitlab.freedesktop.org/fontconfig/fontconfig/-/merge_requests/238

AaronErhardt:
Thanks @jf2048 for the additional information!

I hope this can be merged soon because the Relm4 CI will fail until the image rebuilds libpanel with its latest release.

jf2048:
@bilelmoussaoui Can you try to rerun the docker CI on this? It might be fixed now

jf2048:
Docker appears fixed :) But now broken by this: https://github.com/gtk-rs/gtk-rs-core/pull/816#issuecomment-1290594258

Dockerfile
```diff
@@ -29,5 +29,4 @@ RUN git clone https://gitlab.gnome.org/GNOME/glib.git --depth=1 && \
     (cd /graphene && \
         meson setup builddir --prefix=/usr --buildtype release -Dintrospection=disabled -Dtests=false -Dinstalled_tests=false && \
         meson install -C builddir) && \
-    rm -rf /glib /cairo /harfbuzz /pango /gdk-pixbuf /graphene
-
+    rm -rf /glib /cairo /harfbuzz /pango /gdk-pixbuf /graphene
```
bilelmoussaoui, line 32: The change on this file is unneeded, please revert it

bilelmoussaoui:
otherwise lgtm, we can land it before the the remaining of the CI is fixed.

Dockerfile
AaronErhardt: The idea was to trigger a rebuild of the image with this, but I guess that can be done manually too?

bilelmoussaoui (approved):

Issue #74, opened 2021-05-18
Not catching panics across FFI boundaries


SimonSapin:
https://github.com/gtk-rs/cairo/pull/257 added bindings for “user data” owned by cairo objects. Each user data entry has a destructor function pointer that cairo calls when the object is destroyed. This helps solve life time issues, since with reference counting it can be hard to predict when an object will be destroyed exactly.

The bindings are generic and accept user data of any type. They forward destructor function pointer calls to `Drop`, which potentially panic and unwind. However, until https://github.com/rust-lang/rust/issues/58794 is resolved, unwinding into C is undefined behavior.

A possible fix is using `std::panic::catch_unwind` in the destructor function. However even if we stash the panic payload somewhere at that point, there is no good place in the code to call `resume_unwind`. So the best we can do might be to abort the process.

piegamesde:
I don't understand one thing in here: the initial issue description is specifically about Cairo "user data". But doesn't the problem also apply in any signal callbacks? I've already panicked in my code more than once and it's never blown up badly, so was it just by luck?

sdroege:
Yes, it's about any callbacks that go through FFI.

piegamesde:
So if I'm getting a stack trace like

```
   2: glib::source::fnmut_callback_wrapper_local::{{closure}}
             at ~/.cache/cargo/registry/src/github.com-1ecc6299db9ec823/glib-0.14.0/src/source.rs:180:9
   3: glib::source::trampoline
             at ~/.cache/cargo/registry/src/github.com-1ecc6299db9ec823/glib-0.14.0/src/source.rs:90:5
   4: <unknown>
   5: g_main_context_dispatch
   6: <unknown>
   7: g_main_context_iteration
   8: g_application_run
   9: <O as gio::application::ApplicationExtManual>::run_with_args
             at ~/.cache/cargo/registry/src/github.com-1ecc6299db9ec823/gio-0.14.0/src/application.rs:30:13
```

then that's totally not guarantee to work in the general case?

-----

If I understand the linked threads above correctly, then this *ought* to work in a defined matter some time in the future. Nevertheless, I'd argue for running signal callbacks in a `catch_unwind`. This would allow at least a minimum of clean shutdown within Gtk. Because the current behavior straight up aborts the Gtk main loop, leaving any resources open.

How do other language bindings handle exceptions?

sdroege:
[...]

PR #818, opened 2022-10-31
gio: add spawn_blocking()

### LINKED ISSUE #792: Add a `spawn_blocking` method to run futures on a thread
It would be great to have a `spawn_blocking` akin to https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

I am trying to port some code that in C use g_task_run_in_thread and now that it is all rust, using gtask seems ugly.

The method could use gtask internally to go on the gio thread pool. Since glib does not depend on gio, it could be an extension trait that lives in gio.

jf2048:
Fixes #792

After looking at the equivalent tokio API I noticed we don't install unwind handlers, maybe that is something that should happen?

### DIFF
--- a/gio/src/task.rs
+++ b/gio/src/task.rs
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::{boxed::Box as Box_, mem::transmute, ptr};
+use std::{boxed::Box as Box_, future::Future, mem::transmute, panic, ptr};
 
 use glib::{
     prelude::*,
@@ -8,6 +8,8 @@ use glib::{
     translate::*,
     value::ValueType,
 };
+
+use futures_channel::oneshot;
 
 use crate::{AsyncResult, Cancellable};
 
@@ -379,6 +381,70 @@ impl<V: ValueType + Send> Task<V> {
 
 unsafe impl<V: ValueType + Send> Send for Task<V> {}
 unsafe impl<V: ValueType + Send> Sync for Task<V> {}
+
+// rustdoc-stripper-ignore-next
+/// A handle to a task running on the I/O thread pool.
+///
+/// Like [`std::thread::JoinHandle`] for a blocking I/O task rather than a thread. The return value
+/// from the task can be retrieved by awaiting on this handle. Dropping the handle "detaches" the
+/// task, allowing it to complete but discarding the return value.
+#[derive(Debug)]
+pub struct JoinHandle<T> {
+    rx: oneshot::Receiver<std::thread::Result<T>>,
+}
+
+impl<T> JoinHandle<T> {
+    #[inline]
+    fn new() -> (Self, oneshot::Sender<std::thread::Result<T>>) {
+        let (tx, rx) = oneshot::channel();
+        (Self { rx }, tx)
+    }
+}
+
+impl<T> Future for JoinHandle<T> {
+    type Output = std::thread::Result<T>;
+    #[inline]
+    fn poll(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        std::pin::Pin::new(&mut self.rx)
+            .poll(cx)
+            .map(|r| r.unwrap())
+    }
+}
+
+impl<T> futures_core::FusedFuture for JoinHandle<T> {
+    #[inline]
+    fn is_terminated(&self) -> bool {
+        self.rx.is_terminated()
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// Runs a blocking I/O task on the I/O thread pool.
+///
+/// Calls `func` on the internal Gio thread pool for blocking I/O operations. The thread pool is
+/// shared with other Gio async I/O operations, and may rate-limit the tasks it receives. Callers
+/// may want to avoid blocking indefinitely by making sure blocking calls eventually time out.
+///
+/// This function should not be used to spawn async tasks. Instead, use
+/// [`glib::MainContext::spawn`] or [`glib::MainContext::spawn_local`] to run a future.
+pub fn spawn_blocking<T, F>(func: F) -> JoinHandle<T>
+where
+    T: Send + 'static,
+    F: FnOnce() -> T + Send + 'static,
+{
+    // use Cancellable::NONE as source obj to fulfill `Send` requirement
+    let task = unsafe { Task::<bool>::new(Cancellable::NONE, Cancellable::NONE, |_, _| {}) };
+    let (join, tx) = JoinHandle::new();
+    task.run_in_thread(move |_, _: Option<&Cancellable>, _| {
+        let res = panic::catch_unwind(panic::AssertUnwindSafe(func));
+        let _ = tx.send(res);
+    });
+
+    join
+}
 
 #[cfg(test)]
 mod test {
--- a/glib/src/main_context_futures.rs
+++ b/glib/src/main_context_futures.rs
@@ -1,16 +1,16 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use std::{
-    mem,
-    pin::{self, Pin},
-    ptr,
+    any::Any, cell::Cell, marker::PhantomData, mem, num::NonZeroU32, panic, pin::Pin, ptr, thread,
 };
 
+use futures_channel::oneshot;
 use futures_core::{
     future::Future,
     task::{Context, Poll, RawWaker, RawWakerVTable, Waker},
 };
 use futures_task::{FutureObj, LocalFutureObj, LocalSpawn, Spawn, SpawnError};
+use futures_util::FutureExt;
 
 use crate::{
     thread_guard::ThreadGuard, translate::*, MainContext, MainLoop, Priority, Source, SourceId,
@@ -20,16 +20,21 @@ use crate::{
 // if the non-Send Future is polled/dropped from a different thread
 // than where this was created.
 enum FutureWrapper {
-    Send(FutureObj<'static, ()>),
-    NonSend(ThreadGuard<LocalFutureObj<'static, ()>>),
+    Send(FutureObj<'static, Box<dyn Any + Send + 'static>>),
+    NonSend(ThreadGuard<LocalFutureObj<'static, Box<dyn Any + 'static>>>),
 }
 
-impl Future for FutureWrapper {
-    type Output = ();
+unsafe impl Send for FutureWrapper {}
+unsafe impl Sync for FutureWrapper {}
 
-    fn poll(self: pin::Pin<&mut Self>, ctx: &mut Context) -> Poll<()> {
+impl Future for FutureWrapper {
+    type Output = Box<dyn Any + 'static>;
+
+    fn poll(self: Pin<&mut Self>, ctx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.get_mut() {
-            FutureWrapper::Send(fut) => Pin::new(fut).poll(ctx),
+            FutureWrapper::Send(fut) => {
+                Pin::new(fut).poll(ctx).map(|b| b as Box<dyn Any + 'static>)
+            }
             FutureWrapper::NonSend(fut) => Pin::new(fut.get_mut()).poll(ctx),
         }
     }
@@ -47,6 +52,7 @@ struct TaskSource {
     source: ffi::GSource,
     future: FutureWrapper,
     waker: Waker,
+    return_tx: Option<oneshot::Sender<thread::Result<Box<dyn Any + 'static>>>>,
 }
 
 #[repr(C)]
@@ -60,7 +66,7 @@ impl TaskSource {
         callback: ffi::GSourceFunc,
         _user_data: ffi::gpointer,
     ) -> ffi::gboolean {
-        let source = &mut *(source as *mut TaskSource);
+        let source = &mut *(source as *mut Self);
         assert!(callback.is_none());
 
         // Poll the TaskSource and ensure we're never called again if the
@@ -73,7 +79,7 @@ impl TaskSource {
     }
 
     unsafe extern "C" fn finalize(source: *mut ffi::GSource) {
-        let source = source as *mut TaskSource;
+        let source = source as *mut Self;
 
         // This will panic if the future was a local future and is dropped from a different thread
         // than where it was created so try to drop it from the main context if we're on another
@@ -89,8 +95,7 @@ impl TaskSource {
                 ptr::drop_in_place(&mut (*source).future);
             }
             FutureWrapper::NonSend(ref mut future) => {
-                let context =
-                    ffi::g_source_get_context(source as *mut TaskSource as *mut ffi::GSource);
+                let context = ffi::g_source_get_context(source as *mut Self as *mut ffi::GSource);
                 if !context.is_null() {
                     let future = ptr::read(future);
                     let context = MainContext::from_glib_none(context);
@@ -159,7 +164,11 @@ unsafe impl Sync for WakerSource {}
 impl TaskSource {
     #[allow(clippy::new_ret_no_self)]
     // checker-ignore-item
-    fn new(priority: Priority, future: FutureWrapper) -> Source {
+    fn new(
+        priority: Priority,
+        future: FutureWrapper,
+        return_tx: Option<oneshot::Sender<thread::Result<Box<dyn Any + 'static>>>>,
+    ) -> Source {
         unsafe {
             static TASK_SOURCE_FUNCS: ffi::GSourceFuncs = ffi::GSourceFuncs {
                 check: None,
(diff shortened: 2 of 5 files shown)


gio/src/task.rs
```diff
@@ -382,6 +384,64 @@ impl<V: ValueType + Send> Task<V> {
...
+    #[inline]
+    pub fn cancellable(&self) -> Option<&Cancellable> {
+        self.cancellable.as_ref()
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum JoinError {
+    #[error("task cancelled")]
+    Cancelled,
```
sdroege, line 408: As discussed on Matrix, it would be great if this would catch panics like tokio also does and returns them as another kind of error instead of aborting the process.

Similarly, the normal `spawn()` should probably do the same and also return such a join handle.

sdroege:
Looks good to me otherwise

gio/src/task.rs
```diff
@@ -382,6 +384,64 @@ impl<V: ValueType + Send> Task<V> {
...
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        std::pin::Pin::new(&mut self.rx)
+            .poll(cx)
+            .map(|r| r.map_err(|_| JoinError::Cancelled))
+    }
+}
+
+pub fn spawn_blocking<T, F, P>(cancellable: Option<&P>, func: F) -> JoinHandle<T>
```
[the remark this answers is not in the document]
sdroege: Maybe there can be two functions to make this a bit easier. `spawn_blocking()` without cancellable and `spawn_blocking_with_cancellable()` that would be this one.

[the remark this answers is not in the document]
pbor: If we keep them both, can we implement one in terms of the other so that we are sure they always behave the same?

With that said, `spawn_blocking(CancellableFuture::new(f, c))` seems as convenient as `spawn_blocking_with_cancellable(f, c)`...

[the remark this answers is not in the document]
pbor: Ignore my comments, I was clearly not thinking yesterday (in my defense it was holiday :) ).

The function takes a closure, not a future. I am ok with either the proposed version or with the two variants.

glib/src/thread_pool.rs
```diff
@@ -77,11 +78,12 @@ impl ThreadPool {
     pub fn push_future<T: Send + 'static, F: FnOnce() -> T + Send + 'static>(
```
sdroege, line 78: Hm, we could have `push()` above also return a `JoinHandle` kind of thing around a `std::sync::mpsc` channel, I guess

glib/src/main_context_futures.rs
```diff
@@ -230,19 +241,108 @@ impl TaskSource {
...
+        let id = source.attach(Some(ctx));
+        let id = Cell::new(Some(unsafe { NonZeroU32::new_unchecked(id.as_raw()) }));
+        Self { rx, id }
+    }
+    // rustdoc-stripper-ignore-next
+    /// Returns the internal source ID.
+    ///
+    /// Returns `None` if the handle was aborted.
+    #[inline]
+    pub unsafe fn as_raw_source_id(&self) -> Option<u32> {
```
sdroege, line 290: Would it be useful to also have a getter for it as `SourceId`?

glib/src/main_context_futures.rs
```diff
@@ -149,26 +155,30 @@ impl WakerSource {
...
-unsafe impl Sync for TaskSource {}
+unsafe impl<T> Send for TaskSource<T> {}
+unsafe impl<T> Sync for TaskSource<T> {}
 
 unsafe impl Send for WakerSource {}
 unsafe impl Sync for WakerSource {}
 
-impl TaskSource {
+impl<T: 'static> TaskSource<T> {
+    const TASK_SOURCE_FUNCS: ffi::GSourceFuncs = ffi::GSourceFuncs {
```
sdroege, line 165: This was `static` for a reason :) `g_source_new()` just takes a pointer and assumes that this stays valid forever.

glib/src/main_context_futures.rs
```diff
@@ -18,15 +24,15 @@ use crate::SourceId;
 // Wrapper around Send Futures and non-Send Futures that will panic
 // if the non-Send Future is polled/dropped from a different thread
 // than where this was created.
-enum FutureWrapper {
-    Send(FutureObj<'static, ()>),
-    NonSend(ThreadGuard<LocalFutureObj<'static, ()>>),
+enum FutureWrapper<T> {
+    Send(FutureObj<'static, T>),
+    NonSend(ThreadGuard<LocalFutureObj<'static, T>>),
 }
```
sdroege, line 30: This (for completeness) should probably get a `Send` impl now

glib/src/main_context_futures.rs
RealKC, line 290: Does this need to be unsafe? If so could it get a `# Safety` section?

gio/src/task.rs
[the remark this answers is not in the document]
jf2048: I removed the cancellable. If you need cancellation, just pass one into the closure or wrap the `JoinHandle` in an abortable future.

glib/src/main_context_futures.rs
jf2048: Not a getter, but there could be `into_source_id`. `SourceId` avoids implementing `Clone` and `Copy` so the source cannot accidentally be removed twice

sdroege: Yes

sdroege: Makes sense

jf2048: @sdroege What do you think? Can we remove the unsafe from both?

sdroege: Yes

glib/src/thread_pool.rs
jf2048: I think that would make it identical to `push_future`

glib/src/main_context_futures.rs
jf2048: I don't think it's needed, this is not a public type

glib/src/main_context_futures.rs
jf2048: It seems to be working

glib/src/thread_pool.rs
sdroege: It would return a blocking `JoinHandle` like `std::thread::spawn`

glib/src/main_context_futures.rs
sdroege: I think that's just coincidence. If you check the gobject code, `g_source_new()` assumes that the pointer to the functions stays valid forever.

glib/src/main_context_futures.rs
sdroege: Not needed but it would document the API a bit more

glib/src/main_context_futures.rs
jf2048: I made an example where it shows the compiler does do static monomorphizations of the vtable: https://rust.godbolt.org/z/6fxMT87zs

But this behavior does not seem to be specified anywhere.

sdroege: Yes, `const` are only guaranteed to be like a `#define` in C. The behaviour should be equivalent to copying the value into every place where it is used.

jf2048: Changed it to use `Box<dyn Any>`, that should still avoid allocating if return type is `()`

glib/src/thread_pool.rs
jf2048: Done, feeling a bit better about using that after seeing this land for 1.67: https://github.com/rust-lang/rust/pull/93563

glib/src/main_context_futures.rs
sdroege: Why the box instead of the plain value? This also seems to be for a different discussion thread but not sure which? :)

glib/src/main_context_futures.rs
```diff
@@ -230,19 +246,191 @@ impl TaskSource {
...
+            let id = unsafe { SourceId::from_glib(id.get()) };
+            id.remove();
+        }
+    }
+    // rustdoc-stripper-ignore-next
+    /// Safely converts the handle into a [`SourceId`].
+    ///
+    /// An be used to discard the return value while still retaining the ability to abort the
+    /// underlying task. Returns `Err(self)` if the handle was aborted already.
+    pub fn into_source_id(self) -> Result<SourceId, Self> {
```
sdroege, line 325: I wonder if access to the `&glib::Source` would also be useful. The source IDs are hard to use correctly in multi-threaded applications while the `Source` is refcounted and doesn't have such problems.

glib/src/main_context_futures.rs
```diff
@@ -230,19 +246,191 @@ impl TaskSource {
 
                 // This will panic if the future was a local future and is called from
                 // a different thread than where it was created.
-                Pin::new(&mut self.future).poll(&mut context)
+                if let Some(tx) = self.return_tx.take() {
```
sdroege, line 249: IIRC tokio for example still catches panics in these cases but just ignores them. For consistency we might want to do the same here

glib/src/main_context_futures.rs
jf2048: The `T` parameter cannot be used in a `static` variable so I had to remove it.

sdroege: Ah I see. The `T` would have to propagate up to `TaskSource<T>` and all the `GSourceFuncs`.

Seems like a good solution this way

glib/src/main_context_futures.rs
jf2048: The reason this is like this is so `MainContext::block_on` just propagates the panic up the stack.  Side effect is that `spawn_obj` and `spawn_local_obj` also do that so they kill the process on panic. Should they ignore panics? I am not sure the tokio behavior.

sdroege: Seems OK either way so let's just keep it like this I guess?

> The reason this is like this is so MainContext::block_on can pass None to just propagate the panic up the stack. 

As the C functions are not `extern "C-unwind"`, this will not actually happen in practice (or at least in the future)

sdroege: @jf2048 So what do you think? Let's get it in like this or do you want to change something? It seems fine to me either way for now

jf2048: This is just how it was before, IMO anything else to deal with the FFI problem should be part of #74

sdroege: Yeah I agree, I'm just asking because you seem to have explicitly implemented it the way it is because you thought that panics would be propagated out of `block_on()`

jf2048: Currently they are, even though it technically is undefined behavior

jf2048: I should clarify, it should be like this anyway even if the behavior is to abort on panic. Without this, a panic inside `block_on` is quietly ignored.

To really come up with a good fix for it, we would have to find some way to propagate the panic from any source func to the `g_main_loop_run` call. And of course in all other places where we pass callbacks to a C function.

sdroege: Good point. I don't like the quiet ignoring but OTOH that's what tokio does, and also what we would do here if people just ignore the join handle. Maybe if the channel for the join handle is closed, we should not catch the panic either but rethrow it?

> And of course in all other places where we pass callbacks to a C function.

I think this will only be possible case by case. In some cases we could simply mark the C callbacks as `C-unwind` instead of `C`, but it needs careful checking if this wouldn't introduce memory leaks or worse.

glib/src/main_context_futures.rs
```diff
@@ -230,19 +246,199 @@ impl TaskSource {
...
+    pub fn as_raw_source_id(&self) -> Option<u32> {
+        self.id.get().map(|i| i.get())
+    }
+    // rustdoc-stripper-ignore-next
+    /// Aborts the task associated with the handle.
+    #[inline]
+    pub fn abort(&self) {
+        if let Some(id) = self.id.take() {
+            let id = unsafe { SourceId::from_glib(id.get()) };
+            id.remove();
```
sdroege, line 319: This is not correct if a non-default main context is used (and also is not thread-safe). What you want here is `self.source.destroy()` instead

glib/src/main_context_futures.rs
sdroege: > ...  but rethrow it?

Also for the other cases btw, not just the futures case

gio/src/task.rs
```diff
@@ -382,6 +385,71 @@ impl<V: ValueType + Send> Task<V> {
...
 unsafe impl<V: ValueType + Send> Sync for Task<V> {}
 
+// rustdoc-stripper-ignore-next
+/// A handle to a task running on the I/O thread pool.
+///
+/// Like [`std::thread::JoinHandle`] for a blocking I/O task rather than a thread. The return value
+/// from the task can be retrieved by awaiting on this handle. Dropping the handle "detaches" the
+/// task, allowing it to complete but discarding the return value.
+#[derive(Debug)]
+#[repr(transparent)]
```
sdroege, line 395: Why?

glib/src/main_context_futures.rs
```diff
@@ -230,19 +246,199 @@ impl TaskSource {
...
+    #[inline]
+    fn is_terminated(&self) -> bool {
+        self.rx.is_terminated()
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// Task failure from awaiting a [`JoinHandle`].
+#[derive(Debug)]
+#[repr(transparent)]
```
sdroege, line 367: Why?

glib/src/thread_pool.rs
```diff
@@ -13,6 +14,31 @@ pub struct ThreadPool(ptr::NonNull<ffi::GThreadPool>);
...
 unsafe impl Sync for ThreadPool {}
 
+// rustdoc-stripper-ignore-next
+/// A handle to a thread running on a [`ThreadPool`].
+///
+/// Like [`std::thread::JoinHandle`] for a GLib thread. The return value from the task can be
+/// retrieved by calling [`ThreadHandle::join`]. Dropping the handle "detaches" the thread,
+/// allowing it to complete but discarding the return value.
+#[derive(Debug)]
+#[repr(transparent)]
```
sdroege, line 24: Why?

glib/src/main_context_futures.rs
sdroege: @jf2048 What do you think? :)

glib/src/main_context_futures.rs
jf2048: Changed it, but does that mean `Source` should be `!Send`?

jf2048: Sorry not `Source`, I meant `SourceId`?

gio/src/task.rs
jf2048: No reason, I just removed it

glib/src/main_context_futures.rs
sdroege: Sending it to other threads is only one of the many ways how you can make it fail. Ideally the API should just not exist at all :)

jf2048: Ok, let's save that for later https://github.com/gtk-rs/gtk-rs-core/issues/860

glib/src/main_context_futures.rs
jf2048: I suppose what we should do is always catch the panic and log it with `g_critical` if the channel is dropped from either end. There is no way to rethrow it in that case. It is not reliable to fall back on the default rust behavior right now since we are always in an FFI stack frame.

The other options are to make it into an abort, or just wait to see if abort becomes the default for `extern "C"`...

sdroege: `g_critical()` would be the tokio approach, right? If so, let's go with that

jf2048: Actually I don't think anything needs to be done, I was under the impression tokio set the hook but it does not. It just uses the default hook to print the message and ignores the panic: [current_thread.rs#L481-L483](https://github.com/tokio-rs/tokio/blob/39766220f4adc969ef7d026e04d752f5dfe55fb9/tokio/src/runtime/scheduler/current_thread.rs#L481-L483).

It also has an unstable feature that lets you configure the behavior: [https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.unhandled_panic](Builder::unhandled_panic). But it "may change or be removed in the future". Practically for users it seems nearly the same as just setting `panic=abort`.

jf2048: I made a couple changes, it should now match the default behavior of `tokio`:

- `MainContext::spawn` and `MainContext::spawn_local` will propagate the panic into the `JoinHandle` and ignore it if dropped
- `MainContext::block_on` will rethrow the panic
- `Spawn` and `LocalSpawn` will just ignore the panic. Based on the behavior in `futures_executor`, it seems with those the caller is expected to use [`FuturesExt::catch_unwind`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.catch_unwind) anyway if they want to change the behavior.

That should be good for now, if someone asks for some more ways to configure this we could add those in later

sdroege (approved):

PR #819, opened 2022-11-04
fix CI for 1.65/1.66

jf2048:


sdroege (approved):

glib/src/translate.rs
```diff
@@ -2663,6 +2663,7 @@ mod tests {
 
         impl TryFromGlib<libc::c_uint> for SpecialU32 {
             type Error = GlibNoneError;
+            #[allow(clippy::unnecessary_cast)]
```
sdroege, line 2666: Could also do `val as _` here but either is OK for me

sdroege:
Thanks!

PR #820, opened 2022-11-04
Move `g_cancellable_set_error_if_cancelled()` to manual

cgwalters:
Regression from this change:
https://github.com/gtk-rs/gtk-rs-core/commit/7c9b36c270a945012a2f5d07c8d0f11912d66d34#diff-8439ea6875c9b60b7d07a7b39f6dde1d31d38459f7c27173e56b4cecb0eadefaR142

Basically `g_cancellable_set_error_if_cancelled()` doesn't follow the GError conventions - it's the inverse, it returns `TRUE` if an error was set, and `FALSE` otherwise.

We need to move this to manual implementation.

I caught this while updating ostree-rs-ext to use gio 0.15, and our https://github.com/ostreedev/ostree-rs-ext/blob/main/lib/src/tokio_util.rs unit test started to panic.

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -398,6 +398,9 @@ manual_traits = ["CancellableExtManual"]
     name = "cancelled"
     # Racy, use 'connect' and 'disconnect' instead
     ignore = true
+    [[object.function]]
+    name = "set_error_if_cancelled"
+    manual = true
 
 [[object]]
 name = "Gio.CharsetConverter"
--- a/gio/src/auto/cancellable.rs
+++ b/gio/src/auto/cancellable.rs
@@ -5,7 +5,6 @@
 use glib::object::IsA;
 use glib::translate::*;
 use std::fmt;
-use std::ptr;
 
 glib::wrapper! {
     #[doc(alias = "GCancellable")]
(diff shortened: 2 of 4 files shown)


sdroege (approved):

sdroege:
@cgwalters How urgent is it for you to have this in a release? Should I do a bugfix release this weekend?

(We don't consider moving functions between auto/manual traits as a breaking change)

cgwalters:
Not urgent at all, we're still on 0.14, but bumping from 0.15 -> 0.16 is just somewhere on the todo list.  So fine to batch into whatever any usual backporting cycle is (monthly?).

sdroege:
Whenever I get to it and there's something interesting to backport, or someone needs it :)

lucab:
I'm sorry for having caused this regression, and I'd love to get this fixed in 0.15 too.
@sdroege may I send a backport PR and ask you to tag a `0.15.13` with that?

sdroege:
I'd consider 0.15 EOL at this point. We don't really have the resources to do bugfix releases for two stable branches.

lucab:
Ack, thanks for the notice!

PR #822, opened 2022-11-08
gio: socket/stream extras

jf2048:
Adds `DatagramBased`, all functions related to socket control messages, all `Stream::writev` methods. Unfortunately the method names in `DatagramBased` conflict with those in `Socket` and have to be fully qualified.


gio/Gir.toml
```diff
@@ -1187,6 +1225,20 @@ status = "generate"
...
         nullable = true
 
+[[object]]
+name = "Gio.SocketControlMessage"
+status = "generate"
+manual_traits = ["SocketControlMessageExtManual"]
+    [[object.function]]
+    name = "deserialize"
+    # writes from a void pointer
+    manual = true
```
sdroege, line 1235: Should probably just get `element-type` annotations in C?

gio/src/datagram_based.rs
```diff
@@ -0,0 +1,339 @@
...
+        let mut msgs = smallvec::SmallVec::<[_; 4]>::new();
+        let mut meta = vectors
+            .into_iter()
+            .map(|v| {
+                msgs.push(unsafe {
+                    std::mem::MaybeUninit::<ffi::GInputMessage>::zeroed().assume_init()
+                });
+                let vs = v
+                    .iter_mut()
+                    .map(|v| ffi::GInputVector {
```
sdroege, line 227: Maybe we can use `IoSlice` from `std` for this?

gio/src/datagram_based.rs
```diff
@@ -0,0 +1,339 @@
...
+                        buffer: v.as_mut_ptr() as *mut _,
+                        size: v.len(),
+                    })
+                    .collect::<Vec<_>>();
```
sdroege, line 231: This seems suboptimal. Every receive being a new heap allocation even if there's nothing to actually receive currently...

Maybe this can make use of smallvec or the API can be defined in a way that the caller already provides something that is ABI-compatible with the C type?

sdroege: Ah this already uses smallvec. The question about having the caller provide something ABI-compatible stays though :)

gio/src/datagram_based.rs
```diff
@@ -0,0 +1,339 @@
...
+            .iter()
+            .map(|m| {
+                let v = m
+                    .vectors
+                    .iter()
+                    .map(|v| ffi::GOutputVector {
+                        buffer: v.as_ptr() as *const _,
+                        size: v.len(),
+                    })
+                    .collect::<Vec<_>>();
```
sdroege, line 297: Same questions here

gio/src/output_stream.rs
```diff
@@ -65,6 +65,79 @@ pub trait OutputStreamExtManual: Sized + OutputStreamExt {
         >,
     >;
 
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    #[doc(alias = "g_output_stream_writev")]
+    fn writev(
+        &self,
+        vectors: &[&[u8]],
```
sdroege, line 73: This wouldn't allow to use a `Vec<Vec<u8>>` for example. Should be made a bit more generic at least

gio/src/output_stream.rs
```diff
@@ -65,6 +65,79 @@ pub trait OutputStreamExtManual: Sized + OutputStreamExt {
...
+        cancellable: Option<&impl IsA<Cancellable>>,
+    ) -> Result<usize, glib::Error>;
+
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    #[doc(alias = "g_output_stream_writev_async")]
+    fn writev_async<
+        B: AsRef<[u8]> + Send + 'static,
+        P: FnOnce(Result<(Vec<B>, usize), (Vec<B>, glib::Error)>) + 'static,
```
sdroege, line 82: Maybe can be made generic instead of using `Vec` here

gio/src/output_stream.rs
```diff
@@ -284,6 +357,304 @@ impl<O: IsA<OutputStream>> OutputStreamExtManual for O {
...
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    fn writev(
+        &self,
+        vectors: &[&[u8]],
+        cancellable: Option<&impl IsA<Cancellable>>,
+    ) -> Result<usize, glib::Error> {
+        let vectors = vectors
+            .iter()
+            .map(|v| ffi::GOutputVector {
```
sdroege, line 370: Same question as above about having the user pass in something ABI-compatible

gio/src/output_stream.rs
sdroege: Same applies to other functions here

gio/src/output_stream.rs
sdroege: Same applies to other functions here

gio/src/pollable_output_stream.rs
```diff
@@ -42,6 +44,15 @@ pub trait PollableOutputStreamExtManual {
         priority: glib::Priority,
     ) -> Pin<Box<dyn Stream<Item = ()> + 'static>>;
 
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    #[doc(alias = "g_pollable_output_stream_writev_nonblocking")]
+    fn writev_nonblocking(
+        &self,
+        vectors: &[&[u8]],
```
sdroege, line 52: Should be more generic

gio/src/pollable_output_stream.rs
```diff
@@ -144,6 +155,40 @@ impl<O: IsA<PollableOutputStream>> PollableOutputStreamExtManual for O {
...
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    fn writev_nonblocking(
+        &self,
+        vectors: &[&[u8]],
+        cancellable: Option<&impl IsA<Cancellable>>,
+    ) -> Result<(PollableReturn, usize), glib::Error> {
+        let vectors = vectors
+            .iter()
+            .map(|v| ffi::GOutputVector {
```
sdroege, line 168: Same question about providing something ABI-compatible

gio/src/socket.rs
```diff
@@ -61,6 +68,21 @@ impl AsRawSocket for Socket {
     }
 }
 
+#[doc(alias = "GInputMessage")]
+pub struct InputMessage {
```
sdroege, line 72: These two can probably be made `#[repr(transparent)]` wrappers around the C types

gio/src/socket.rs
```diff
@@ -74,6 +96,20 @@ pub trait SocketExtManual: Sized {
         buffer: B,
         cancellable: Option<&C>,
     ) -> Result<(usize, SocketAddress), glib::Error>;
+    #[doc(alias = "g_socket_receive_message")]
+    fn receive_message<C: IsA<Cancellable>>(
+        &self,
+        vectors: &mut [&mut [u8]],
```
sdroege, line 102: More generic maybe

sdroege: Also elsewhere here

gio/src/socket_control_message.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+pub trait SocketControlMessageExtManual: Sized {
+    #[doc(alias = "g_socket_control_message_serialize")]
+    fn serialize(&self, data: &mut [MaybeUninit<u8>]);
+}
+
+impl<O: IsA<SocketControlMessage>> SocketControlMessageExtManual for O {
+    #[doc(alias = "g_socket_control_message_serialize")]
+    fn serialize(&self, data: &mut [MaybeUninit<u8>]) {
```
sdroege, line 30: We don't do that anywhere else. Maybe better `&mut [u8]` for consistency for now? Dealing with `MaybeUninit` slices is not very convenient at least

gio/src/socket_msg_flags.rs
```diff
@@ -0,0 +1,77 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+use glib::value::{FromValue, ToValue};
+use glib::{StaticType, Type};
+use std::fmt;
+
+bitflags::bitflags! {
+    #[doc(alias = "GSocketMsgFlags")]
```
sdroege, line 9: Why can't this be autogenerated?

gio/src/unix_fd_message.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[cfg(unix)]
+use std::os::unix::io::{AsRawFd, RawFd};
+
+#[cfg(all(not(unix), feature = "dox"))]
+use socket::{AsRawFd, RawFd};
+
+pub trait UnixFDMessageExtManual: Sized {
+    #[doc(alias = "g_unix_fd_message_append")]
+    fn append_fd<T: AsRawFd>(&self, fd: T) -> Result<(), glib::Error>;
```
sdroege, line 16: `IntoRawFd` maybe? This needs to take ownership of the fds or otherwise somehow have a lifetime bound to the fds

gio/src/socket_msg_flags.rs
jf2048: Gir thinks all flags are `u32`, but this is a `c_int`

gio/src/datagram_based.rs
jf2048: It would be redundant. `IoSlice` is only for converting the `&[u8]` to the memory layout of `struct iovec`, but these functions already do that internally with a `GInputVector`

gio/src/unix_fd_message.rs
jf2048: It calls `dup` to make a copy of the fd: https://docs.gtk.org/gio/method.UnixFDMessage.append_fd.html

This is the same behavior as `UnixFDListExtManual::append` which also takes `AsRawFd`.

gio/Gir.toml
jf2048: Seems to have it already: [gsocketcontrolmessage.c#L157](https://gitlab.gnome.org/GNOME/glib/-/blob/96bc6d6e7cdc1cd24e54f1499092c0354991e391/gio/gsocketcontrolmessage.c#L157)

gio/src/socket_msg_flags.rs
sdroege: Ah I see. Sucks but makes sense

gio/src/unix_fd_message.rs
sdroege: Ok, makes sense

gio/src/datagram_based.rs
sdroege: But at the cost of additional allocations

gio/src/output_stream.rs
jf2048: Changed to use `OutputVector` directly. It is not more generic but it avoids allocations. The other option is to keep the allocation and make it `IntoIterator<Item = &dyn AsRef<[u8]>>`

gio/src/output_stream.rs
jf2048: How? It needs to be an owned type so it can be held by the closure/future.

gio/src/output_stream.rs
jf2048: I think this is all set the way it is, any other API will require allocations. This way the caller can choose whether to allocate or use the stack.

gio/Gir.toml
sdroege: Hm, can you create a gir issue?

gio/src/socket.rs
```diff
@@ -61,6 +69,185 @@ impl AsRawSocket for Socket {
...
+                bytes_sent: 0,
+                control_messages: control_messages.as_ptr() as *mut _,
+                num_control_messages: control_messages.len().try_into().unwrap(),
+            },
+            address: PhantomData,
+            vectors: PhantomData,
+            control_messages: PhantomData,
+        }
+    }
+    #[inline]
```
sdroege, line 245: Why do you put `#[inline]` on everything everywhere btw? :) Does it actually make any difference here?

Generally the LLVM heuristic for inlining is "yes", so in the majority of cases this has no effect at all.

gio/src/socket.rs
```diff
@@ -258,6 +556,114 @@ impl<O: IsA<Socket>> SocketExtManual for O {
         }
     }
 
+    fn send_message<P: IsA<SocketAddress>, C: IsA<Cancellable>>(
+        &self,
+        address: Option<&P>,
+        vectors: &[&[u8]],
```
sdroege, line 562: Why not have this take a `&[OutputVector]`?

gio/src/socket.rs
```diff
@@ -258,6 +556,114 @@ impl<O: IsA<Socket>> SocketExtManual for O {
...
+            }
+        }
+    }
+
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    fn send_message_with_timeout<P: IsA<SocketAddress>, C: IsA<Cancellable>>(
+        &self,
+        address: Option<&P>,
+        vectors: &[&[u8]],
```
sdroege, line 601: Why not have this take a `&[OutputVector]`?

gio/src/socket.rs
```diff
@@ -258,6 +556,114 @@ impl<O: IsA<Socket>> SocketExtManual for O {
...
+                Ok((from_glib(ret), bytes_written as usize))
+            } else {
+                Err(from_glib_full(error))
+            }
+        }
+    }
+
+    fn send_messages<C: IsA<Cancellable>>(
+        &self,
+        messages: &mut [OutputMessage],
```
sdroege, line 644: Why does this have to be mutable?

gio/src/socket_control_message.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::SocketControlMessage;
+use glib::object::IsA;
+use glib::translate::*;
+
+impl SocketControlMessage {
+    #[doc(alias = "g_socket_control_message_deserialize")]
+    pub fn deserialize(level: i32, type_: i32, data: &[u8]) -> SocketControlMessage {
+        let size = data.len() as _;
+        unsafe {
+            from_glib_full(ffi::g_socket_control_message_deserialize(
```
sdroege, line 13: This can't fail?

sdroege: I'm quite sure it can

gio/src/datagram_based.rs
```diff
@@ -0,0 +1,266 @@
...
+        messages: &mut [InputMessage],
+        flags: i32,
+        timeout: Option<Duration>,
+        cancellable: Option<&C>,
+    ) -> Result<usize, glib::Error>;
+
+    #[doc(alias = "g_datagram_based_send_messages")]
+    fn send_messages<C: IsA<Cancellable>>(
+        &self,
+        messages: &mut [OutputMessage],
```
sdroege, line 64: Why is this mutable?

gio/src/output_stream.rs
```diff
@@ -668,6 +668,30 @@ impl<T: IsA<OutputStream>> io::Write for OutputStreamWrite<T> {
...
+        to_std_io_result(result)
+    }
+
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    fn write_vectored(&mut self, bufs: &[io::IoSlice<'_>]) -> io::Result<usize> {
+        let vectors = bufs
+            .iter()
+            .map(|v| OutputVector::new(&**v))
+            .collect::<smallvec::SmallVec<[_; 2]>>();
```
sdroege, line 686: OOC, why do you use 2 here and in the other places?

gio/src/datagram_based.rs
jf2048: The `bytes_sent` field is modified by this function

sdroege: Ah that makes sense

gio/src/socket_control_message.rs
jf2048: It can, I filed a glib MR: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3099

gio/Gir.toml
jf2048: Not sure why it was screwing up before but it works now, removed the manual implementation

gio/src/socket.rs
jf2048: Missed this one, It takes `OutputVector` now :)

gio/src/output_stream.rs
jf2048: Random choice for a simple ring buffer usage, should I increase the number? Not sure what a good value would be.

gio/src/socket.rs
jf2048: No particular reason, I noticed some places we have it and some places we don't

gio/Gir.toml
jf2048: Oh now I remember, it detects the right type but does not cast the pointer correctly to `gpointer`

jf2048: https://github.com/gtk-rs/gir/pull/1407

gio/src/output_stream.rs
sdroege: I don't know. So far I always used this API from C with exactly 2 too :) Let's keep it at that then.
I was just asking because making it e.g. 4 or 16 is also not really going to cause problems, these things are small.

gio/Gir.toml
sdroege: Merged now, can you regenerate? I think then we're ready here? :)

gio/src/socket_control_message.rs
sdroege: You can work around that in the meantime via `nullable = true` in the `Gir.toml`. Please add a link to the MR in there so we can get rid of it in the near future :)

sdroege: Ah you did.

gio/src/socket.rs
```diff
@@ -61,6 +69,185 @@ impl AsRawSocket for Socket {
...
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct OutputVector<'v> {
+    _vector: ffi::GOutputVector,
+    buffer: PhantomData<&'v [u8]>,
+}
+
+impl<'v> OutputVector<'v> {
+    #[inline]
+    pub const fn new(buffer: &'v [u8]) -> Self {
```
sdroege, line 202: For completeness etc we could `impl Deref<Target=[u8]>` here and/or provide getters for the two fields. Also for the `InputVector` and input/output messages

jf2048: Added `Deref` impls and getters for `vectors`

sdroege (approved):

sdroege:
Thanks, this looks great :)

PR #826, opened 2022-11-17
Implement From<T> for Value, Variant

### LINKED ISSUE #788: Implement `Into<Value>`, `Into<Variant>`
Or more accurately, we should have `impl From<...> for Value` and `impl From<...> for Variant` in every place where there is `ToValue` or `ToVariant`. Then we can correctly support things like `g_value_take_*`, `g_variant_new_take_string`, etc.

Then `Closure`, `BindingBuilder::transform`, `Object::new`, etc should be changed to use that instead of `ToValue`

jf2048:
Avoids unnecessary copying and `ref`ing in some places. A few extra PRs will be needed to finish this up in gir, gtk4-rs, gstreamer-rs, but nothing too bad.

Fixes #788


cairo/src/enums.rs
```diff
@@ -42,6 +42,13 @@ macro_rules! gvalue_impl {
                 <Self as glib::StaticType>::static_type()
             }
         }
+
+        impl From<$name> for glib::Value {
```
sdroege, line 46: This also needs a gir change to implement the same there for enums/flags or not?

cairo/src/lib.rs
```diff
@@ -86,6 +86,21 @@ macro_rules! gvalue_impl {
...
+        impl From<$name> for glib::Value {
+            fn from(v: $name) -> Self {
+                unsafe {
+                    let mut value =
+                        glib::Value::from_type(<$name as glib::StaticType>::static_type());
+                    glib::gobject_ffi::g_value_take_boxed(
+                        value.to_glib_none_mut().0,
+                        glib::translate::ToGlibPtr::to_glib_none(&&v).0 as *mut _,
+                    );
+                    let _v = std::mem::ManuallyDrop::new(v);
```
sdroege, line 98: Why not `into_glib_ptr()`?

glib-macros/src/boxed_derive.rs
```diff
@@ -161,6 +161,20 @@ pub fn impl_boxed(input: &syn::DeriveInput) -> TokenStream {
             }
         }
 
+        impl ::std::convert::From<#name> for #crate_ident::Value {
+            fn from(v: #name) -> Self {
+                unsafe {
+                    let ptr: *mut #name = ::std::boxed::Box::into_raw(::std::boxed::Box::new(v));
+                    let mut value = #crate_ident::Value::from_type(<#name as #crate_ident::StaticType>::static_type());
+                    #crate_ident::gobject_ffi::g_value_take_boxed(
+                        #crate_ident::translate::ToGlibPtrMut::to_glib_none_mut(&mut value).0,
```
sdroege, line 170: This needs to be `into_glib_ptr()` or otherwise you'll have a double-free

glib-macros/src/shared_boxed_derive.rs
```diff
@@ -181,6 +181,20 @@ pub fn impl_shared_boxed(input: &syn::DeriveInput) -> proc_macro2::TokenStream {
             }
         }
 
+        impl ::std::convert::From<#name> for #crate_ident::Value {
+            fn from(v: #name) -> Self {
+                unsafe {
+                    let ptr = #refcounted_type_prefix::into_raw(v.0);
+                    let mut value = #crate_ident::Value::from_type(<#name as #crate_ident::StaticType>::static_type());
+                    #crate_ident::gobject_ffi::g_value_take_boxed(
+                        #crate_ident::translate::ToGlibPtrMut::to_glib_none_mut(&mut value).0,
```
sdroege, line 190: `into_glib_ptr()`

glib/src/object.rs
```diff
@@ -1170,6 +1170,21 @@ macro_rules! glib_object_wrapper {
...
+        #[doc(hidden)]
+        impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? ::std::convert::From<$name $(<$($generic),+>)?> for $crate::Value {
+            fn from(o: $name $(<$($generic),+>)?) -> Self {
+                unsafe {
+                    let mut value = $crate::Value::from_type(<$name $(<$($generic),+>)? as $crate::StaticType>::static_type());
+                    $crate::gobject_ffi::g_value_take_object(
+                        $crate::translate::ToGlibPtrMut::to_glib_none_mut(&mut value).0,
+                        $crate::translate::ToGlibPtr::<*mut $ffi_name>::to_glib_none(&o).0 as *mut _,
+                    );
+                    let _o = ::std::mem::ManuallyDrop::new(o);
```
sdroege, line 1182: `into_glib_ptr()`. And also in many other places, I guess :)

glib/src/object.rs
```diff
@@ -2284,15 +2299,15 @@ impl<T: ObjectType> ObjectExt for T {
     }
 
     #[track_caller]
-    fn set_property<V: ToValue>(&self, property_name: &str, value: V) {
+    fn set_property<V: Into<Value>>(&self, property_name: &str, value: V) {
```
sdroege, line 2302: This doesn't allow passing e.g. a `&some_object` anymore but will require passing `some_object.clone()`, correct?

cairo/src/enums.rs
jf2048: Yes: https://github.com/gtk-rs/gir/pull/1402

glib/src/object.rs
jf2048: That will still work, I didn't change this `impl From<&T> for Value`: [value.rs#L615-L620](https://github.com/gtk-rs/gtk-rs-core/blob/11222fe785210ef35667a81c14f86085bdf2d129/glib/src/value.rs#L615-L620)

sdroege: Ah so this doesn't actually change the API in a breaking way, and we don't have to change all the code everywhere? Just that we could if we wanted to have fewer copies

sdroege: Not sure this is the right direction for `set_property()` though. The value is effectively borrowed to the object, so the object implementation will have to do a copy anyway when it gets the value.

In C, `g_object_set()` will only borrow the value to the `GValue`, as an optimization.

jf2048: Yes, this does not require any major code changes :) Using `ToValue` will always create a copy so this change is a strict optimization to remove copies in cases where you pass an owned value. To fully avoid all copies here, we would have to add a lifetime parameter to `ToValue` and `Value` so it can hold borrowed values.

sdroege: Yeah I was thinking of some kind of `BorrowedValue<'a>`. All doable but so much to do! :)

cairo/src/lib.rs
jf2048: There was not any implementations of it, I wrote some

glib-macros/src/boxed_derive.rs
jf2048: Added an implementation of `IntoGlibPtr` to the macro

cairo/src/enums.rs
sdroege: Can you update this PR with that?

glib/src/variant_dict.rs
```diff
@@ -130,12 +130,12 @@ impl VariantDict {
     /// instance already, you should use the [`insert_value()`](#method.insert_value)
     /// method instead.
     #[doc(alias = "g_variant_dict_insert_value")]
-    pub fn insert<T: ToVariant>(&self, key: &str, value: &T) {
+    pub fn insert<T: Into<Variant>>(&self, key: &str, value: T) {
```
sdroege, line 133: This could also be `impl Into<Variant>`. Less noise

sdroege: Also this is the only place where this is used currently. Do you have specific code where you want to make use of this / where this makes things easier?

gio/src/settings.rs
```diff
@@ -170,7 +170,7 @@ impl<'a> BindingBuilder<'a> {
 pub trait SettingsExtManual {
     fn get<U: FromVariant>(&self, key: &str) -> U;
 
-    fn set<U: ToVariant>(&self, key: &str, value: &U) -> Result<(), BoolError>;
+    fn set<U: Into<Variant>>(&self, key: &str, value: U) -> Result<(), BoolError>;
```
sdroege, line 173: `impl Into<Variant>`

gio/src/settings.rs
```diff
@@ -194,8 +194,8 @@ impl<O: IsA<Settings>> SettingsExtManual for O {
         })
     }
 
-    fn set<U: ToVariant>(&self, key: &str, value: &U) -> Result<(), BoolError> {
-        self.set_value(key, &value.to_variant())
+    fn set<U: Into<Variant>>(&self, key: &str, value: U) -> Result<(), BoolError> {
```
sdroege, line 197: `impl Into<Variant>`

glib/src/gstring.rs
```diff
@@ -1076,6 +1088,18 @@ impl crate::value::ToValue for Vec<GString> {
     }
 }
 
+impl From<Vec<GString>> for Value {
+    fn from(mut v: Vec<GString>) -> Self {
+        unsafe {
+            let mut value = Value::for_value_type::<Vec<GString>>();
+            let ptr: *mut *mut c_char = ToGlibContainerFromSlice::to_glib_full_from_slice(&v);
+            v.set_len(0);
```
sdroege, line 1096: Why? In the line above you copy the whole strings anyway. Maybe you wanted `to_glib_container_from_slice()` above to steal the strings?

glib/src/value.rs
```diff
@@ -822,6 +836,13 @@ impl ToValue for String {
     }
 }
 
+impl From<String> for Value {
```
sdroege, line 839: Not also `From<&str>` and `From<&[&str]>`?

glib/src/object.rs
```diff
@@ -1608,13 +1608,13 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
 
     // rustdoc-stripper-ignore-next
     /// Set property `name` to the given value `value`.
-    pub fn property<T: ToValue + 'a>(self, name: &'a str, value: T) -> Self {
+    pub fn property<T: Into<Value> + 'a>(self, name: &'a str, value: T) -> Self {
```
sdroege, line 1611: `impl Into<Value>`

glib/src/object.rs
```diff
@@ -1692,7 +1692,7 @@ pub trait ObjectExt: ObjectType {
     /// If the property does not exist, if the type of the property is different than
     /// the provided value, or if the property is not writable.
     #[doc(alias = "g_object_set_property")]
-    fn set_property<V: ToValue>(&self, property_name: &str, value: V);
+    fn set_property<V: Into<Value>>(&self, property_name: &str, value: V);
```
sdroege, line 1695: `impl Into<Value>`

glib/src/object.rs
```diff
@@ -2298,15 +2298,15 @@ impl<T: ObjectType> ObjectExt for T {
     }
 
     #[track_caller]
-    fn set_property<V: ToValue>(&self, property_name: &str, value: V) {
+    fn set_property<V: Into<Value>>(&self, property_name: &str, value: V) {
```
sdroege, line 2301: `impl Into<Value>`

sdroege: ... and everywhere else where possible :)

sdroege:
Otherwise looks ready to me

glib/src/variant_dict.rs
jf2048: I have used `Settings::set`, but also I will want to use this in the class macro for gio actions and widget actions that have return values.

cairo/src/enums.rs
jf2048: Updated, but something else is breaking in `gio::Icon`: https://github.com/gtk-rs/gtk-rs-core/commit/245a088c38166343a88807e8b564a542724138b7#diff-56fdede3ae1018d2c5e30d97bd2f117b17c659689e9bee7818680e735625d932

```
error[E0283]: type annotations needed
  --> gio/src/auto/icon.rs:31:65
   |
31 |                 ToGlibPtr::<*mut ffi::GIcon>::to_glib_none(self.as_ref()).0
```

glib/src/gstring.rs
jf2048: No, we need to call `to_glib_full` on each of the strings to force them to a C string

glib/src/value.rs
jf2048: Those get handled by the blanket `From<&T>` impl

glib/src/gstring.rs
jf2048: Actually no, it should be `into_glib_ptr`, I guess those `GString` implementations should be calling that because they can avoid the copy

glib/src/variant_dict.rs
sdroege: Sounds good

glib/src/gstring.rs
sdroege: `into_glib_ptr()` on `Vec<GString>` would steal the strings themselves but store the array in newly allocated memory? Makes sense, yes. That's what I meant above and what you would effectively get by calling `to_glib_container_from_slice()` and truncating the `Vec` without dropping the values

cairo/src/enums.rs
sdroege: That should be fixed by https://github.com/gtk-rs/gtk-rs-core/pull/829#discussion_r1027121550

sdroege: Hmm, but without the gir-files update it stays in the `impl Icon` instead of going into the trait impl so it shouldn't have gotten the `as_ref()` added.

sdroege: I'll take a look at this after you solved the other discussion item here. I'll get that all fixed and merged then :)

glib/src/gstring.rs
sdroege: Looks like this is done now

cairo/src/enums.rs
sdroege: Ok, so my job now to get these two PRs in :) Thanks!

jf2048: I think this was my bad, I updated the gir-files and now it's working. This looks merge ready now

sdroege:
@jf2048 You send regen PR/MRs for gtk3/4 and gstreamer?

sdroege:
I guess now :) gstreamer-rs at least fails to compile now and needs some changes. Will take care of that then.

PR #835, opened 2022-11-27
gdk-pixbuf: Add subclassing support

### LINKED ISSUE #834: [FEATURE REQUEST] both gdk_pixbuf::{PixbufAnimation,PixbufAnimationIter} are missing a IsSubclassable trait implementation
To write a Gdk Pixbuf loader with animation support in Rust one needs to subclass GdkPixbufAnimation and GdkPixbufAnimationIter. Without the IsSubclassavble trait implementation this is not possible.

bilelmoussaoui:
Fixes #834

### DIFF
--- a/gdk-pixbuf/src/subclass/pixbuf_animation.rs
+++ b/gdk-pixbuf/src/subclass/pixbuf_animation.rs
@@ -1,0 +1,170 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+// rustdoc-stripper-ignore-next
+//! Traits intended for subclassing [`PixbufAnimation`](crate::PixbufAnimation).
+
+use std::mem::MaybeUninit;
+use std::time::Duration;
+
+use crate::{Pixbuf, PixbufAnimation, PixbufAnimationIter};
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+pub trait PixbufAnimationImpl: ObjectImpl {
+    fn is_static_image(&self) -> bool {
+        self.parent_is_static_image()
+    }
+
+    fn static_image(&self) -> Option<Pixbuf> {
+        self.parent_static_image()
+    }
+
+    fn size(&self) -> (i32, i32) {
+        self.parent_size()
+    }
+
+    fn iter(&self, start_time: Duration) -> PixbufAnimationIter {
+        self.parent_iter(start_time)
+    }
+}
+
+pub trait PixbufAnimationImplExt: ObjectSubclass {
+    fn parent_is_static_image(&self) -> bool;
+    fn parent_static_image(&self) -> Option<Pixbuf>;
+    fn parent_size(&self) -> (i32, i32);
+    fn parent_iter(&self, start_time: Duration) -> PixbufAnimationIter;
+}
+
+impl<T: PixbufAnimationImpl> PixbufAnimationImplExt for T {
+    fn parent_is_static_image(&self) -> bool {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GdkPixbufAnimationClass;
+            let f = (*parent_class)
+                .is_static_image
+                .expect("No parent class implementation for \"is_static_image\"");
+
+            from_glib(f(self
+                .obj()
+                .unsafe_cast_ref::<PixbufAnimation>()
+                .to_glib_none()
+                .0))
+        }
+    }
+
+    fn parent_static_image(&self) -> Option<Pixbuf> {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GdkPixbufAnimationClass;
+            let f = (*parent_class)
+                .get_static_image
+                .expect("No parent class implementation for \"get_static_image\"");
+
+            from_glib_none(f(self
+                .obj()
+                .unsafe_cast_ref::<PixbufAnimation>()
+                .to_glib_none()
+                .0))
+        }
+    }
+
+    fn parent_size(&self) -> (i32, i32) {
+        unsafe {
+            let data = T::type_data();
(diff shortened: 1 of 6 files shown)


gdk-pixbuf/src/subclass/pixbuf_animation_iter.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+unsafe extern "C" fn animation_iter_advance<T: PixbufAnimationIterImpl>(
+    ptr: *mut ffi::GdkPixbufAnimationIter,
+    time_ptr: *const glib::ffi::GTimeVal,
+) -> glib::ffi::gboolean {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    let time = if time_ptr.is_null() {
```
sdroege, line 168: Can this actually be `NULL`?

gdk-pixbuf/src/subclass/pixbuf_animation_iter.rs
```diff
@@ -0,0 +1,178 @@
...
+        None
+    } else {
+        Some(Duration::new(
+            (*time_ptr).tv_sec.try_into().unwrap(),
+            (*time_ptr).tv_usec.try_into().unwrap(),
```
sdroege, line 173: The `Duration` constructor takes nanoseconds, `tv_usec` sounds very much like microseconds

gdk-pixbuf/src/subclass/pixbuf_animation_iter.rs
bilelmoussaoui: Per the docs `As a shortcut, if start_time is NULL, the result of g_get_current_time() will be used automatically.`

gdk-pixbuf/src/subclass/pixbuf_animation_iter.rs
bilelmoussaoui: Oh good catch!

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+unsafe extern "C" fn animation_get_iter<T: PixbufAnimationImpl>(
+    ptr: *mut ffi::GdkPixbufAnimation,
+    start_time_ptr: *const glib::ffi::GTimeVal,
+) -> *mut ffi::GdkPixbufAnimationIter {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    let time = if start_time_ptr.is_null() {
```
sdroege, line 166: This can't possibly get `NULL` according to the C code

gdk-pixbuf/src/subclass/pixbuf_animation_iter.rs
sdroege: It can't according to the C code

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -0,0 +1,176 @@
...
+        None
+    } else {
+        Some(Duration::new(
+            (*start_time_ptr).tv_sec.try_into().unwrap(),
+            (*start_time_ptr).tv_usec.try_into().unwrap(),
```
sdroege, line 171: microseconds vs. nanoseconds here too

gdk-pixbuf/src/subclass/pixbuf_animation.rs
bilelmoussaoui: I only checked the docs which seems to mention the opposite, https://docs.gtk.org/gdk-pixbuf/vfunc.PixbufAnimation.get_iter.html

gdk-pixbuf/src/subclass/pixbuf_loader.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl<T: PixbufLoaderImpl> PixbufLoaderImplExt for T {
+    fn parent_size_prepared(&self, width: i32, height: i32) {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GdkPixbufLoaderClass;
+            let f = (*parent_class)
+                .size_prepared
+                .expect("No parent class implementation for \"size_prepared\"");
```
sdroege, line 43: All the vfuncs here are signal class handlers. So if they don't exist (which they don't by default), you would simply do nothing at all here and return quitely instead of panicking

gdk-pixbuf/src/subclass/pixbuf_animation_iter.rs
bilelmoussaoui: Got confused by the vfunc/func code docs...

gdk-pixbuf/src/subclass/pixbuf_animation.rs
sdroege: Yes it will automatically use the current time then.

The code is basically

```cpp
        if (start_time)
                val = *start_time;
        else
                g_get_current_time (&val);

        return GDK_PIXBUF_ANIMATION_GET_CLASS (animation)->get_iter (animation, &val);
```

You'll never actually get `NULL` inside the vfunc, but the caller of the public function can pass in `NULL` which is then replaced with the current time before calling the vfunc

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    fn static_image(&self) -> Option<Pixbuf> {
+        self.parent_static_image()
+    }
+
+    fn size(&self) -> (i32, i32) {
+        self.parent_size()
+    }
+
+    fn iter(&self, start_time: Option<Duration>) -> PixbufAnimationIter {
```
sdroege, line 27:
```suggestion
    fn iter(&self, start_time: Duration) -> PixbufAnimationIter {
```

etc

sdroege (approved):

PR #838, opened 2022-11-27
update windows CI to glib 2.74

jf2048:
I figured this should be updated along with https://github.com/gtk-rs/gtk3-rs/pull/793 before I forget. Also update the windows VM so we can get the latest version of python and gvsbuild, and update gvsbuild command as it now works like a real python package.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -10,11 +10,11 @@ jobs:
 
   gvsbuild:
     name: build GTK binaries with gvsbuild
-    runs-on: windows-2019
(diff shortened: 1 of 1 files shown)


sdroege (approved):

bilelmoussaoui (approved):

Hofer-Julian:
I haven't noticed that we use gvsbuild to build gtk nowadays.
I guess the book should use that then too.

sdroege:
That's for gtk3

Hofer-Julian:
I see

PR #843, opened 2022-12-01
Add CastNone trait

### LINKED ISSUE #842: [FEATURE REQUEST] downcast/upcast on `Option<Object>`
There are many cases where gtk returns an `Option<Widget>` which must be downcasted.
Doing `let label: gtk::Label = list_item.child().unwrap().downcast().unwrap();` is not fun (note: `child()` returns `Option<Widget>`).

Implementing downcast and upcast for `Option<Widget>` would remove an `unwrap`.
I think it makes sense. In OOP languages you can have `null` objects, and `null` objects can be downcasted to `null` subclassess, can't they? (Edit: in Java you can upcast a null, but not downcast it. In any case, if we return a Result when downcasting, everything is alright)

ranfdev:
Closes #842

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -291,6 +291,11 @@ impl<T: ObjectType> Cast for T {}
 // rustdoc-stripper-ignore-next
 /// Convenience trait mirroring `Cast`, implemented on `Option<Object>` types.
 ///
+/// # Warning
+/// Inveitably this trait will discard informations about a downcast failure:
+/// you don't know if the object was not of the expected type, or if it was `None`.
+/// If you need to handle the downcast error, use `Cast` over a `glib::Object`.
+///
 /// # Example
 /// ```ignore
 /// let widget: Option<Widget> = list_item.child();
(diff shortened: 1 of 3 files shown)


glib/src/object.rs
```diff
@@ -288,6 +288,55 @@ pub trait Cast: ObjectType {
 
 impl<T: ObjectType> Cast for T {}
 
+// rustdoc-stripper-ignore-next
+/// Convenience trait mirroring `Cast`, implemented on `Option<Object>` types.
+pub trait CastNone: Sized {
+    type Inner;
+    fn downcast<T: ObjectType>(self) -> Result<T, Self>
```
sdroege, line 295: This seems like strange API. How would you distinguish `None` from failing to downcast now?

Similar question for `downcast_ref()` below. It should probably all returned `Option<T>` with another wrapping around?

glib/src/object.rs
```diff
@@ -288,6 +288,55 @@ pub trait Cast: ObjectType {
 
 impl<T: ObjectType> Cast for T {}
 
+// rustdoc-stripper-ignore-next
+/// Convenience trait mirroring `Cast`, implemented on `Option<Object>` types.
+pub trait CastNone: Sized {
```
sdroege, line 293: You need to put it into the prelude or it won't easily be usable

glib/src/object.rs
```diff
@@ -288,6 +288,55 @@ pub trait Cast: ObjectType {
 
 impl<T: ObjectType> Cast for T {}
 
+// rustdoc-stripper-ignore-next
+/// Convenience trait mirroring `Cast`, implemented on `Option<Object>` types.
```
sdroege, line 292: Maybe add some more examples here to explain what this trait is good for

glib/src/object.rs
ranfdev: A failing downcast on `Some(t)` will return `Err(Some(t))`.

ranfdev: Indeed on `downcast_ref()` currently you lose the error information, you can't distinguish why `None` was returned.
But hey, if the user wants to handle every conversion step, he can skip using this trait and keep doing whatever he is doing now.

glib/src/object.rs
```diff
@@ -288,6 +288,75 @@ pub trait Cast: ObjectType {
...
+///
+/// # Example
+/// ```ignore
+/// let widget: Option<Widget> = list_item.child();
+///
+/// // Without using `CastNone`
+/// let label: gtk::Label = widget.unwrap().downcast().unwrap();
+///
+/// // Using `CastNone` we can avoid the first `unwrap()` call
+/// let label: gtk::Label = widget.downcast().unwrap();
```
bilelmoussaoui, line 302:
```suggestion
/// let label = widget.unwrap().downcast::<gtk::Label>().unwrap();
///
/// // Using `CastNone` we can avoid the first `unwrap()` call
/// let label = widget.downcast::<gtk::Label>().unwrap();
```

sdroege:
I don't like the API but if people think this is useful to have let's merge it. @bilelmoussaoui @GuillaumeGomez?

bilelmoussaoui:
> I don't like the API but if people think this is useful to have let's merge it. @bilelmoussaoui @GuillaumeGomez?

I think the use case for this API would be mostly GTK and libraries building on top of it as mentioned on the Matrix channel. I would suggest adding a warning to not go through this API unless downcasting/upcasting to the wrong type is a programmer error and the fallback shouldn't handled in that case.

glib/src/object.rs
```diff
@@ -288,6 +288,75 @@ pub trait Cast: ObjectType {
...
+///
+/// // Without using `CastNone`
+/// let label = widget.unwrap().downcast::<gtk::Label>().unwrap();
+///
+/// // Using `CastNone` we can avoid the first `unwrap()` call
+/// let label = widget.downcast::<gtk::Label>().unwrap();
+/// ````
+pub trait CastNone: Sized {
+    type Inner;
+    fn downcast<T: ObjectType>(self) -> Result<T, Self>
```
bilelmoussaoui, line 306: @pbor suggests to prefix all those functions in this trait with `then_`, that would make it extra clear that you are dealing with a cast from an optional type and not the ones from the Cast trait

sdroege: My suggestion was to suffix with `_opt()` but either seems fine to me as distinction, I somewhat prefer @pbor's

ranfdev: I feel adding a prefix/suffix would defeat the purpose of the trait. It's a conviene trait to write less things. If we make the name as long as `.unwrap().downcast()`, the trait doesn't make sense anymore.

sdroege: My main concern here is that by using the same names you

  * make it likely that people use it accidentally and then handle type mismatches / non-existence of values wrong
  * if the value does not implement the trait, you get a compiler error list both traits and all their possible implementors, which is a lot

bilelmoussaoui: > I feel adding a prefix/suffix would defeat the purpose of the trait. It's a conviene trait to write less things. If we make the name as long as `.unwrap().downcast()`, the trait doesn't make sense anymore.

Adding then_ as a prefix is waaay to less code than what we currently have and I completely agree with the concerns raised by Sebastian here.

ranfdev: I've chosen the prefix `and_`. Why?

Because `then` is used when the user should provide a function. See for example `Option::and_then` or `bool::then`.

The `and_` prefix is used when you the user wants to map an option to another option, and the process returns itself an option.
I think the concept maps well: 
```rust
// Option::and
(None::<usize>).and(Some(2u8)) -> None::<u8>
// CastNone::and_downcast
(None::<gtk::Widget>).and_downcast::<gtk::Button>() -> None::<gtk::Button>
```

sdroege: I like the naming. @bilelmoussaoui?

Could maybe also be implemented on `Result` now instead of just on `Option`?

ranfdev: Going from a Result to an Option only requires calling `result.ok()`, so I believe having a shortcut method on Result it's unnecessary. It would also require a new trait, because every method `and_downcast*` would need to return a `Result` instead of an `Option`, to keep consistency with the other `and_*` functions defined on `Result`.

sdroege (approved):

glib/src/object.rs
bilelmoussaoui: Looks good to me!

bilelmoussaoui (approved):

PR #846, opened 2022-12-04
pango: Backport Language changes

bilelmoussaoui:


pango/src/language.rs
```diff
@@ -119,29 +31,30 @@ impl Language {
...
-    pub fn sample_string(&self) -> GString {
-        unsafe { from_glib_none(ffi::pango_language_get_sample_string(self.to_glib_none().0)) }
-    }
-
     #[cfg(any(feature = "v1_48", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_48")))]
     #[doc(alias = "get_preferred")]
     #[doc(alias = "pango_language_get_preferred")]
-    pub fn preferred(&self) -> Vec<Language> {
+    pub fn preferred() -> Vec<Self> {
```
bilelmoussaoui, line 38: This was always wrong, should we keep the old function signature even in this case?

pango/Gir.toml
```diff
@@ -332,6 +331,29 @@ concurrency = "send+sync"
...
+    name = "from_string"
+    # This only returns NULL when passing NULL
+        [[object.function.parameter]]
+        name = "language"
+        nullable = false
+        [object.function.return]
+        nullable = false
+    [[object.function]]
+    name = "to_string"
+    manual = true # avoids renaming the function to not cause an API break
```
sdroege, line 355: I would have `to_str()` autogenerated and implement `to_string()` manually

bilelmoussaoui: I don't know why I did the opposite...

pango/src/language.rs
```diff
@@ -1,113 +1,25 @@
...
-
-#[doc(hidden)]
-impl FromGlibPtrFull<*const ffi::PangoLanguage> for Language {
-    unsafe fn from_glib_full(ptr: *const ffi::PangoLanguage) -> Self {
-        assert!(!ptr.is_null());
-        Self(ptr as *mut _)
-    }
-}
-
-impl Default for Language {
```
sdroege, line 64: This got lost

sdroege:
Seems OK to backport otherwise

pango/src/language.rs
bilelmoussaoui: It used to take &self when it wasn't needed...

sdroege: Ah... yes. Seems fine, it wouldn't have been usable for its purpose before.

pango/src/language.rs
sdroege: That seems like exactly the behaviour you'd want or not?

sdroege: And yes, someone might have used this :)

bilelmoussaoui: > That seems like exactly the behaviour you'd want or not?

I thought Default is only meant for cases where you would get a new instance instead of the same one?

sdroege: `Language` is immutable or not? So that's the same effectively :)

sdroege (approved):

PR #847, opened 2022-12-07
gio: Add a GioInfaliableFuture

bilelmoussaoui:
Needed for https://github.com/gtk-rs/gir/pull/1397

Note this is only needed for 0.16, for 0.17 we should break the GioFuture API to make use of `T` directly instead of `Result<T, E>`


bilelmoussaoui:
Closing in favour of https://github.com/gtk-rs/gtk-rs-core/pull/849

PR #851, opened 2022-12-09
Regenerate with latest gir/gir-files

bilelmoussaoui:


sdroege (approved):
Boring, not many changes here :)

I guess there are some things in manual code where a `to_glib_full()` could be converted in a similar way. Can you create an issue about checking that if you don't want to do that now?

bilelmoussaoui:
> Boring, not many changes here :)
> 
> I guess there are some things in manual code where a `to_glib_full()` could be converted in a similar way. Can you create an issue about checking that if you don't want to do that now?

Left a comment on the open issue instead, https://github.com/gtk-rs/gtk-rs-core/issues/50#issuecomment-1344160480

PR #852, opened 2022-12-09
glib: Use actual function name in `glib::BoolError` and include function name/source file/line number in structured logs

sdroege:
See individual commits


sdroege:
CC @jf2048

bilelmoussaoui (approved):

PR #856, opened 2022-12-10
glib: Implement `GStringBuilder` as `BoxedInline` to avoid a useless …

sdroege:
…additional heap allocation

CC @jf2048


GuillaumeGomez (dismissed):

GuillaumeGomez:
Thanks! We should maybe add memory checks for such changes to ensure by how much they change the memory usage...

sdroege:
> Thanks! We should maybe add memory checks for such changes to ensure by how much they change the memory usage...

What do you mean? The way how I'm currently looking at these for my use case (GStreamer debug logging) is to look at the generated assembly and check what can be avoided there :)

PR #865, opened 2022-12-17
Various `Stash` / `to_glib_none()` related optimizations and bugfixes

sdroege:
See individual commits.

CC @jf2048

### DIFF
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -1,5 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+#[cfg(feature = "use_glib")]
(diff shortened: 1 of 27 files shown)


sdroege:
@bilelmoussaoui You probably have to check for similar things in gtk4-rs

sdroege:
I also found this (changing to `PhantomData` if we can) to have actual effect in release builds as the stash is not always completely optimized away (without further inlining and LTO). It reduces required stack space for functions.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

PR #866, opened 2022-12-19
Inline various trivial functions

sdroege:
CC @jf2048 

Draft until
  * [x] https://github.com/gtk-rs/gtk-rs-core/pull/865 is finally merged
  * [x] Checked effect on generated assembly
  * [x] Checked effect on compile times
  * [x] Checked effect on code size


glib/src/subclass/types.rs
```diff
@@ -673,14 +697,17 @@ pub trait ObjectSubclassExt: ObjectSubclass {
...
         self.obj()
     }
 
+    #[inline]
     fn from_instance(obj: &Self::Type) -> &Self {
         Self::from_obj(obj)
     }
 
+    #[inline]
     fn obj(&self) -> crate::BorrowedObject<Self::Type> {
```
sdroege, line 711: @jf2048 Despite this, I still have this not being inlined in an `-O3` release build with LTO if it's used multiple times in a function. Hrm!

sdroege: Turning various assertions into `debug_assert!` like in the last commit is allowing to inline these functions in all the cases I found. I'll go through all our assertions to see which ones can safely be changed to debug assertions.

The reason why it isn't inlined with assertions in place is that the panic code (the message printing plus stack unwinding) is not very small and has to be inlined in every call side too.

sdroege: @GuillaumeGomez @bilelmoussaoui So we need some kind of decision here about how we handle assertions. Just randomly putting assertions everywhere is suboptimal for code size and optimizations :) The assertion printing and stack unwinding code is not small and has serious effects on both.

My suggestion would be:
  - Safe functions (i.e. that are callable from **safe** code) and call **unsafe** code, and require certain invariants to be complied to by the caller to operate safely will get proper `assert!` assertions that are also enabled in release builds, i.e. checks for valid lengths/indizes and things like that
  - Unsafe functions use debug `debug_assert!` assertions for invariants that the caller should comply to, i.e. that a pointer is not `NULL`, that it is of the correct type, that the refcount is still > 0 and things like that. This also implies that we assume C code behaves correctly (which we do anyway!) in release mode for the very few things we could potentially check.
  - Code uses debug `debug_assert!` assertions for internal consistency checks, i.e. situations that can only happen if our code is buggy (that's the case of the change in the top commit here)

This would be similar to how `std` treats assertion, but the second item is even stronger than what `std` does. E.g. `CStr::from_ptr()` does not check that the pointer is not `NULL` even in debug modes, and will simply segfault instead of panicking.

sdroege: Both @GuillaumeGomez and @bilelmoussaoui said on Matrix that we should go ahead with this plan.

sdroege:
@bilelmoussaoui Changes like in this PR in gtk4-rs are your job :)

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me but for such changes, having perf comparisons would be really nice. At some point we should really write benchmark to not blindly merge such changes.

sdroege:
Yeah, let's discuss at the next hackfest how we can somewhat automate such things. This was manually checked on various GStreamer plugins and some GTK application code with `cargo asm`. Performance was approximately the same but the generated assembly is considerably shorter and cleaner, and has fewer function calls.

Issue #845, opened 2022-12-02
[BUG] `ActionEntryBuilder.state()` has the wrong signature


pbor:
It should take a `&Variant` not a `str`

PR #870, opened 2022-12-27
ActionEntry: take proper types instead of strings

pbor:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/845

### DIFF
--- a/gio/src/action_entry.rs
+++ b/gio/src/action_entry.rs
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use glib::{prelude::*, Variant};
+use glib::{prelude::*, Variant, VariantTy, VariantType};
 
 use crate::{ActionMap, SimpleAction};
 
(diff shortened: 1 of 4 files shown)


gio/src/action_entry.rs
```diff
@@ -69,13 +69,13 @@ where
...
-    pub fn parameter_type(mut self, parameter_type: &str) -> Self {
-        self.0.parameter_type = Some(parameter_type.to_owned());
+    pub fn parameter_type(mut self, parameter_type: Option<&VariantTy>) -> Self {
+        self.0.parameter_type = parameter_type.map(|vt| vt.to_owned());
         self
     }
 
-    pub fn state(mut self, state: &str) -> Self {
-        self.0.state = Some(state.to_owned());
+    pub fn state(mut self, state: Variant) -> Self {
```
pbor, line 77: Here I decided to take the variant by value... this avoids a copy since I assume in the common case this will be some state created on the spot (eg `false.to_variant()`). However `SimpleAction::new_stateful` takes it by reference, let me know if you prefer to change it for consistency

sdroege: Variants are refcounted so this is not too important, but seems nicer like this indeed. Can you change the other one too?

sdroege (dismissed):

bilelmoussaoui (dismissed):

bilelmoussaoui (approved):

PR #871, opened 2022-12-28
Implement safe io

A6GibKm:
Needs reviewing

Stuff thats missing

- [ ] unix_open_pipe (marked as unsafe for the moment)
- [ ] spawn_async_with_pipes (marked as unsafe for the moment)
- [x] spawn_async_with_fds
- [x] g_subprocess_launcher_take_fd
- [x] g_subprocess_launcher_take_stderr_fd
- [x] g_subprocess_launcher_take_stderr_fd
- [x] g_subprocess_launcher_take_stdin_fd
- [x] g_subprocess_launcher_take_stdout_fd
- [x] into_raw_unix_fd
- [x] into_raw_unix_fd_local
- [x] g_unix_fd_add_full

### DIFF
--- a/gio/src/unix_fd_list.rs
+++ b/gio/src/unix_fd_list.rs
@@ -1,12 +1,12 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #[cfg(unix)]
-use std::os::unix::io::{AsFd, AsRawFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
 use std::{mem, ptr};
 
 use glib::{prelude::*, translate::*};
 #[cfg(all(not(unix), docsrs))]
-use socket::{AsFd, AsRawFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
+use socket::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
 
 use crate::{ffi, UnixFDList};
 
@@ -69,16 +69,79 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
     }
 
     #[doc(alias = "g_unix_fd_list_steal_fds")]
-    fn steal_fds(&self) -> Vec<RawFd> {
+    fn steal_fds(&self) -> FdArray {
         unsafe {
             let mut length = mem::MaybeUninit::uninit();
-            let ret = FromGlibContainer::from_glib_full_num(
-                ffi::g_unix_fd_list_steal_fds(self.as_ref().to_glib_none().0, length.as_mut_ptr()),
-                length.assume_init() as usize,
-            );
-            ret
+
+            let ptr =
+                ffi::g_unix_fd_list_steal_fds(self.as_ref().to_glib_none().0, length.as_mut_ptr());
+
+            FdArray {
+                ptr: ptr::NonNull::new(ptr).unwrap(),
+                len: length.assume_init() as usize,
+            }
         }
     }
 }
 
 impl<O: IsA<UnixFDList>> UnixFDListExtManual for O {}
+
+pub struct FdArray {
+    ptr: ptr::NonNull<libc::c_int>,
+    len: usize,
+}
+
+pub struct FdIterator {
+    ptr: ptr::NonNull<libc::c_int>,
+    len: usize,
+}
+
+impl Iterator for FdIterator {
+    type Item = OwnedFd;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.len > 0 {
+            let current = self.ptr.as_ptr();
+            if self.len > 1 {
+                let next = unsafe { self.ptr.as_ptr().add(1) };
+                self.ptr = ptr::NonNull::new(next).unwrap();
+            }
+            self.len -= 1;
+            Some(unsafe { OwnedFd::from_raw_fd(*current) })
+        } else {
+            None
+        }
+    }
+}
+
+impl Drop for FdArray {
+    fn drop(&mut self) {
+        while self.len > 0 {
+            unsafe {
+                let current = self.ptr.as_ptr();
+                libc::close(*current);
+            }
+            if self.len > 1 {
+                let next = unsafe { self.ptr.as_ptr().add(1) };
+                self.ptr = ptr::NonNull::new(next).unwrap();
+            }
+            self.len -= 1;
+        }
+    }
+}
+
+impl std::iter::IntoIterator for FdArray {
+    type Item = OwnedFd;
+    type IntoIter = FdIterator;
+
+    fn into_iter(mut self) -> Self::IntoIter {
+        let len = std::mem::take(&mut self.len);
+        FdIterator { len, ptr: self.ptr }
+    }
+}
+
+impl FdArray {
+    pub fn as_slice(&self) -> &[BorrowedFd<'_>] {
+        unsafe { std::slice::from_raw_parts(self.ptr.as_ptr() as *const BorrowedFd, self.len) }
+    }
+}
--- a/gio/src/unix_input_stream.rs
+++ b/gio/src/unix_input_stream.rs
@@ -13,24 +13,12 @@ impl UnixInputStream {
     // rustdoc-stripper-ignore-next
     /// Creates a new [`Self`] that takes ownership of the passed in fd.
     #[doc(alias = "g_unix_input_stream_new")]
-    pub fn take_fd(fd: OwnedFd) -> UnixInputStream {
+    pub fn from_fd(fd: OwnedFd) -> UnixInputStream {
         let fd = fd.into_raw_fd();
         let close_fd = true.into_glib();
         unsafe {
             InputStream::from_glib_full(ffi::g_unix_input_stream_new(fd, close_fd)).unsafe_cast()
         }
-    }
-
-    // rustdoc-stripper-ignore-next
-    /// Creates a new [`Self`] that does not take ownership of the passed in fd.
-    ///
-    /// # Safety
-    /// You may only close the fd if all references to this stream have been dropped.
-    #[doc(alias = "g_unix_input_stream_new")]
-    pub unsafe fn with_fd<T: AsRawFd>(fd: T) -> UnixInputStream {
-        let fd = fd.as_raw_fd();
-        let close_fd = false.into_glib();
-        InputStream::from_glib_full(ffi::g_unix_input_stream_new(fd, close_fd)).unsafe_cast()
     }
 }
 
--- a/gio/src/unix_output_stream.rs
+++ b/gio/src/unix_output_stream.rs
@@ -13,24 +13,12 @@ impl UnixOutputStream {
     // rustdoc-stripper-ignore-next
     /// Creates a new [`Self`] that takes ownership of the passed in fd.
     #[doc(alias = "g_unix_output_stream_new")]
-    pub fn take_fd(fd: OwnedFd) -> UnixOutputStream {
+    pub fn from_fd(fd: OwnedFd) -> UnixOutputStream {
         let fd = fd.into_raw_fd();
         let close_fd = true.into_glib();
         unsafe {
             OutputStream::from_glib_full(ffi::g_unix_output_stream_new(fd, close_fd)).unsafe_cast()
         }
-    }
-
-    // rustdoc-stripper-ignore-next
-    /// Creates a new [`Self`] that does not take ownership of the passed in fd.
-    ///
-    /// # Safety
-    /// You may only close the fd if all references to this stream have been dropped.
-    #[doc(alias = "g_unix_output_stream_new")]
-    pub unsafe fn with_fd<T: AsRawFd>(fd: T) -> UnixOutputStream {
-        let fd = fd.as_raw_fd();
-        let close_fd = false.into_glib();
-        OutputStream::from_glib_full(ffi::g_unix_output_stream_new(fd, close_fd)).unsafe_cast()
     }
 }
 
--- a/glib/src/functions.rs
+++ b/glib/src/functions.rs
@@ -241,7 +241,7 @@ pub fn compute_checksum_for_string(
 
 #[cfg(unix)]
 #[doc(alias = "g_unix_open_pipe")]
-pub fn unix_open_pipe(flags: i32) -> Result<(RawFd, RawFd), Error> {
+pub unsafe fn unix_open_pipe(flags: i32) -> Result<(RawFd, RawFd), Error> {
     unsafe {
         let mut fds = [0, 2];
         let mut error = ptr::null_mut();
--- a/glib/src/functions.rs
+++ b/glib/src/functions.rs
@@ -135,7 +135,7 @@ pub fn spawn_async_with_fds<P: AsRef<std::path::Path>>(
 #[cfg(not(windows))]
 #[cfg_attr(docsrs, doc(cfg(not(windows))))]
 #[doc(alias = "g_spawn_async_with_pipes")]
-pub fn spawn_async_with_pipes<
+pub unsafe fn spawn_async_with_pipes<
     P: AsRef<std::path::Path>,
     T: FromRawFd,
     U: FromRawFd,
--- a/glib/src/source.rs
+++ b/glib/src/source.rs
@@ -1,12 +1,10 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #[cfg(unix)]
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd};
 use std::{cell::RefCell, mem::transmute, num::NonZeroU32, time::Duration};
 
 use crate::ffi::{self, gboolean, gpointer};
-#[cfg(all(not(unix), docsrs))]
-use libc::c_int as RawFd;
 
 #[cfg(unix)]
 use crate::IOCondition;
@@ -154,33 +152,35 @@ fn into_raw_child_watch_local<F: FnMut(Pid, i32) + 'static>(func: F) -> gpointer
 #[cfg(unix)]
 #[cfg_attr(docsrs, doc(cfg(unix)))]
 unsafe extern "C" fn trampoline_unix_fd<
-    F: FnMut(RawFd, IOCondition) -> ControlFlow + Send + 'static,
+    F: FnMut(BorrowedFd, IOCondition) -> ControlFlow + Send + 'static,
 >(
-    fd: i32,
+    raw_fd: i32,
     condition: ffi::GIOCondition,
     func: gpointer,
 ) -> gboolean {
     let func: &RefCell<F> = &*(func as *const RefCell<F>);
+    let fd = BorrowedFd::borrow_raw(raw_fd);
     (*func.borrow_mut())(fd, from_glib(condition)).into_glib()
 }
 
 #[cfg(unix)]
 #[cfg_attr(docsrs, doc(cfg(unix)))]
 unsafe extern "C" fn trampoline_unix_fd_local<
-    F: FnMut(RawFd, IOCondition) -> ControlFlow + 'static,
+    F: FnMut(BorrowedFd, IOCondition) -> ControlFlow + 'static,
 >(
-    fd: i32,
+    raw_fd: i32,
     condition: ffi::GIOCondition,
     func: gpointer,
 ) -> gboolean {
     let func: &ThreadGuard<RefCell<F>> = &*(func as *const ThreadGuard<RefCell<F>>);
+    let fd = BorrowedFd::borrow_raw(raw_fd);
     (*func.get_ref().borrow_mut())(fd, from_glib(condition)).into_glib()
 }
 
 #[cfg(unix)]
 #[cfg_attr(docsrs, doc(cfg(unix)))]
 unsafe extern "C" fn destroy_closure_unix_fd<
-    F: FnMut(RawFd, IOCondition) -> ControlFlow + Send + 'static,
+    F: FnMut(BorrowedFd, IOCondition) -> ControlFlow + Send + 'static,
 >(
     ptr: gpointer,
 ) {
@@ -190,7 +190,7 @@ unsafe extern "C" fn destroy_closure_unix_fd<
 #[cfg(unix)]
 #[cfg_attr(docsrs, doc(cfg(unix)))]
 unsafe extern "C" fn destroy_closure_unix_fd_local<
-    F: FnMut(RawFd, IOCondition) -> ControlFlow + 'static,
+    F: FnMut(BorrowedFd, IOCondition) -> ControlFlow + 'static,
 >(
     ptr: gpointer,
 ) {
@@ -199,7 +199,7 @@ unsafe extern "C" fn destroy_closure_unix_fd_local<
 
 #[cfg(unix)]
 #[cfg_attr(docsrs, doc(cfg(unix)))]
-fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> ControlFlow + Send + 'static>(
+fn into_raw_unix_fd<F: FnMut(BorrowedFd, IOCondition) -> ControlFlow + Send + 'static>(
     func: F,
 ) -> gpointer {
     let func: Box<RefCell<F>> = Box::new(RefCell::new(func));
@@ -208,7 +208,7 @@ fn into_raw_unix_fd<F: FnMut(RawFd, IOCondition) -> ControlFlow + Send + 'static
 
 #[cfg(unix)]
 #[cfg_attr(docsrs, doc(cfg(unix)))]
-fn into_raw_unix_fd_local<F: FnMut(RawFd, IOCondition) -> ControlFlow + 'static>(
+fn into_raw_unix_fd_local<F: FnMut(BorrowedFd, IOCondition) -> ControlFlow + 'static>(
     func: F,
 ) -> gpointer {
     let func: Box<ThreadGuard<RefCell<F>>> = Box::new(ThreadGuard::new(RefCell::new(func)));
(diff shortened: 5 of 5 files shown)


gio/Cargo.toml
```diff
@@ -13,7 +13,7 @@ exclude = [
     "gir-files/*",
 ]
 edition = "2021"
-rust-version = "1.57"
```
sdroege, line 16: This already happened in https://github.com/gtk-rs/gtk-rs-core/commit/86eeafe6e0b1d4133df2c5d5f8c1fdd5edff7e24 or not?

A6GibKm: Yes, I was in an ancient local copy, tanks.

A6GibKm:
Added commit descriptions where there were api questions.

A6GibKm:
stuff that takes `impl AsFd` should take `&impl AsFd`.

gio/src/unix_fd_list.rs
```diff
@@ -54,37 +54,41 @@ impl<O: IsA<UnixFDList>> UnixFDListExtManual for O {
...
+    fn steal_fds(&self) -> Vec<OwnedFd> {
         unsafe {
             let mut length = mem::MaybeUninit::uninit();
-            let ret = FromGlibContainer::from_glib_full_num(
+            let ret: Vec<RawFd> = FromGlibContainer::from_glib_full_num(
                 ffi::g_unix_fd_list_steal_fds(self.as_ref().to_glib_none().0, length.as_mut_ptr()),
                 length.assume_init() as usize,
             );
-            ret
+            ret.into_iter().map(|fd| OwnedFd::from_raw_fd(fd)).collect()
```
sdroege, line 91: Why so complicated? Implement `glib::translate::TransparentType for OwnedFd` (in glib/src/translate.rs), and then you can do `glib::Slice::from_glib_full_num()` from the return value directly without all the vecs

glib/src/functions.rs
```diff
@@ -23,15 +23,15 @@ use crate::{Error, Pid, SpawnFlags};
...
     argv: &[&str],
     envp: &[&str],
     flags: SpawnFlags,
     child_setup: Option<Box_<dyn FnOnce() + 'static>>,
-    stdin_fd: T,
-    stdout_fd: U,
-    stderr_fd: V,
+    stdin_fd: &T,
+    stdout_fd: &U,
+    stderr_fd: &V,
```
sdroege, line 34: Everywhere you can use `T` instead of `&T` for `AsFd` (there's `impl<T: AsFd> AsFd for &T`), but also you can use `impl AsFd` instead of the generic parameter. Less line noise.

sdroege (commented):
Seems good to me otherwise

gio/src/unix_fd_list.rs
sdroege: https://gtk-rs.org/gtk-rs-core/git/docs/glib/translate/trait.TransparentType.html

A6GibKm: It requires Clone, which OwnedFd does not have.

sdroege: Ah I see, that's sad. I wonder why it doesn't have clone, that's just `dup()`.

A6GibKm:
Rebased. There were quite many conflicts so a review would be appreciated.

gio/src/unix_fd_list.rs
A6GibKm: well, according to rust, duplicating  a fd can fail, https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#94.

sdroege:
Is this ready for another review and potentially merging?

A6GibKm:
> Is this ready for another review and potentially merging?

Yes and maybe

gio/src/desktop_app_info.rs
```diff
@@ -48,21 +48,16 @@ pub trait DesktopAppInfoExtManual: IsA<DesktopAppInfo> {
...
         launch_context: Option<&P>,
         spawn_flags: glib::SpawnFlags,
         user_setup: Option<Box_<dyn FnOnce() + 'static>>,
         pid_callback: Option<&mut dyn (FnMut(&DesktopAppInfo, glib::Pid))>,
-        stdin_fd: &mut T,
-        stdout_fd: &mut U,
-        stderr_fd: &mut V,
+        stdin_fd: T,
+        stdout_fd: U,
+        stderr_fd: V,
```
sdroege, line 60: This should be `Option`s (and `None` maps to -1, see the docs)

sdroege: Also `T`, `U`, `V` should probably just become `Option<impl AsFd>` to reduce the syntax noise here

gio/src/subprocess_launcher.rs
```diff
@@ -30,30 +30,60 @@ impl SubprocessLauncher {
         }
     }
 
+    #[cfg(any(unix, docsrs))]
+    #[cfg_attr(docsrs, doc(cfg(unix)))]
+    #[doc(alias = "g_subprocess_launcher_take_fd")]
+    pub fn take_owned_fd(&self, source_fd: OwnedFd, target_fd: OwnedFd) {
```
sdroege, line 36: This etc should also be `Option`, see above

gio/src/unix_fd_list.rs
```diff
@@ -25,12 +25,12 @@ impl UnixFDList {
 
 pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
     #[doc(alias = "g_unix_fd_list_append")]
-    fn append<T: AsRawFd>(&self, fd: T) -> Result<i32, glib::Error> {
+    fn append<T: AsFd>(&self, fd: T) -> Result<i32, glib::Error> {
```
sdroege, line 28:
```suggestion
    fn append(&self, fd: impl AsFd) -> Result<i32, glib::Error> {
```

etc

gio/src/unix_fd_list.rs
```diff
@@ -42,39 +42,43 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
     }
 
     #[doc(alias = "g_unix_fd_list_get")]
-    fn get(&self, index_: i32) -> Result<RawFd, glib::Error> {
+    fn get(&self, index_: i32) -> Result<OwnedFd, glib::Error> {
```
sdroege, line 45: The returned fd is not owned, is it? It's just borrowed

glib/src/source.rs
```diff
@@ -805,9 +808,10 @@ where
 /// This function panics if called from a different thread than the one that
 /// owns the main context.
 #[doc(alias = "g_unix_fd_add_full")]
-pub fn unix_fd_add_local<F>(fd: RawFd, condition: IOCondition, func: F) -> SourceId
+pub fn unix_fd_add_local<D, F>(fd: D, condition: IOCondition, func: F) -> SourceId
 where
-    F: FnMut(RawFd, IOCondition) -> Continue + 'static,
+    D: AsFd,
```
sdroege, line 813: `impl AsFd`, etc

sdroege:
Same comments also apply in other places in this PR

bilelmoussaoui:
Any chance we could get this for the next release?

glib/src/source.rs
A6GibKm: Did you mean to use `x: impl AsFd` rather than `x: D` (with `where D: AsFd`)?

sdroege: Yes

gio/src/unix_fd_list.rs
```diff
@@ -42,39 +42,43 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
...
         }
     }
+
     #[doc(alias = "g_unix_fd_list_steal_fds")]
-    fn steal_fds(&self) -> Vec<RawFd> {
+    fn steal_fds(&self) -> Vec<OwnedFd> {
         unsafe {
             let mut length = mem::MaybeUninit::uninit();
-            let ret = FromGlibContainer::from_glib_full_num(
+            let ret: Vec<RawFd> = FromGlibContainer::from_glib_full_num(
```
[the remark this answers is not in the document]
A6GibKm: Done, but the return type is still `Vec<BorrowedFd<'_>>`, which is kinda odd.

gio/src/subprocess_launcher.rs
A6GibKm: Done, but I am not sure if the change I made should have been done in the way I did (rather than baking the option into `fn unsafe take_fd` itself)

gio/src/unix_fd_list.rs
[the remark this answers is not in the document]
sdroege: It says `steal` here so they're probably not borrowed but owned?

Also why don't you use `glib::Slice` here to avoid the copy to a `Vec`?

gio/src/subprocess_launcher.rs
sdroege: Yes I think that would be better

gio/src/unix_fd_list.rs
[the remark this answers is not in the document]
A6GibKm: I wasn't able to do the map RawFd->BorrowedFd without ownership issues at the time.

A6GibKm:
rebased. I don't quite remember the state of this. It should be fine, iirc.

gio/src/subprocess_launcher.rs
sdroege: This still applies. `None` should map to `-1`

gio/src/unix_fd_list.rs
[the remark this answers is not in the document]
sdroege: This can actually just return `&[BorrowedFd]`. You can use `std::slice::from_raw_parts()` plus a cast. `BorrowedFd` is guaranteed to have the same layout as an actual fd.

gio/src/unix_fd_list.rs
```diff
@@ -47,39 +47,43 @@ pub trait UnixFDListExtManual: sealed::Sealed + IsA<UnixFDList> + Sized {
...
             } else {
                 Err(from_glib_full(error))
             }
         }
     }
 
     #[doc(alias = "g_unix_fd_list_peek_fds")]
-
-    fn peek_fds(&self) -> Vec<RawFd> {
+    fn peek_fds(&self) -> Vec<BorrowedFd<'_>> {
```
sdroege, line 65: This can actually just return `&[BorrowedFd]`. You can use `std::slice::from_raw_parts()` plus a cast. `BorrowedFd` is guaranteed to have the same layout as an actual fd.

gio/src/unix_fd_list.rs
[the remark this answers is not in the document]
sdroege: `OwnedFd` has the same layout as an actual fd so you could just cast things.

I think the correct thing to do here (because `glib::Slice` requires all the traits for the contained type) is to define a new `struct FdArray(...)` that stores a `glib::Slice<i32>` and can `Deref` into `BorrowedFd` and `IntoIterator<Item = OwnedFd>`, and on `Drop` frees things accordingly.

gio/src/desktop_app_info.rs
```diff
@@ -54,21 +54,16 @@ mod sealed {
 pub trait DesktopAppInfoExtManual: sealed::Sealed + IsA<DesktopAppInfo> {
     #[cfg_attr(docsrs, doc(cfg(all(feature = "v2_58", unix))))]
     #[doc(alias = "g_desktop_app_info_launch_uris_as_manager_with_fds")]
-    fn launch_uris_as_manager_with_fds<
-        P: IsA<AppLaunchContext>,
-        T: AsRawFd,
-        U: AsRawFd,
-        V: AsRawFd,
-    >(
+    fn launch_uris_as_manager_with_fds<P: IsA<AppLaunchContext>>(
```
sdroege, line 57: It would be good to keep an unsafe version of this that works via raw fds.

gio/src/subprocess_launcher.rs
```diff
@@ -31,28 +31,60 @@ impl SubprocessLauncher {
...
+    pub unsafe fn take_stdin_fd(&self, fd: impl IntoRawFd) {
         unsafe {
             ffi::g_subprocess_launcher_take_stdin_fd(self.to_glib_none().0, fd.into_raw_fd());
         }
     }
 
     #[cfg(unix)]
     #[cfg_attr(docsrs, doc(cfg(unix)))]
+    #[doc(alias = "g_subprocess_launcher_take_stdin_fd")]
+    pub fn take_stdin_owned_fd(&self, fd: OwnedFd) {
```
sdroege, line 71: Also these functions all should take an `Option`, `None` is -1

glib/src/functions.rs
```diff
@@ -24,15 +24,15 @@ use crate::{Error, Pid, SpawnFlags};
 #[cfg_attr(docsrs, doc(cfg(all(feature = "v2_58", not(windows)))))]
 #[allow(clippy::too_many_arguments)]
 #[doc(alias = "g_spawn_async_with_fds")]
-pub fn spawn_async_with_fds<P: AsRef<std::path::Path>, T: AsRawFd, U: AsRawFd, V: AsRawFd>(
+pub fn spawn_async_with_fds<P: AsRef<std::path::Path>>(
```
sdroege, line 27: Should keep an unsafe version around too

glib/src/log.rs
```diff
@@ -972,16 +972,20 @@ pub fn log_variant(log_domain: Option<&str>, log_level: LogLevel, fields: &crate
 #[cfg_attr(docsrs, doc(cfg(unix)))]
 #[doc(alias = "g_log_writer_supports_color")]
 #[inline]
-pub fn log_writer_supports_color<T: AsRawFd>(output_fd: T) -> bool {
-    unsafe { from_glib(ffi::g_log_writer_supports_color(output_fd.as_raw_fd())) }
+pub fn log_writer_supports_color(output_fd: impl AsFd) -> bool {
```
sdroege, line 975: Should keep unsafe versions of this and below around

glib/src/source.rs
```diff
@@ -872,14 +872,14 @@ where
 /// The default main loop almost always is the main loop of the main thread.
 /// Thus, the closure is called on the main thread.
 #[doc(alias = "g_unix_fd_add_full")]
-pub fn unix_fd_add<F>(fd: RawFd, condition: IOCondition, func: F) -> SourceId
+pub fn unix_fd_add<F>(fd: impl AsFd, condition: IOCondition, func: F) -> SourceId
```
sdroege, line 875: unsafe versions of these too

sdroege: (callback still uses `BorrowedFd`)

glib/src/log.rs
A6GibKm: I don't think so, you can unsafely use the safe versions with BorrowedFd::borrow_raw or similar.

sdroege: Hm good point. But then we should probably remove all the others too?

glib/src/source.rs
A6GibKm: I don't know how we can do better regarding the callback to be honest. Cant have 

```
F: FnMut(impl AsFd, IOCondition) -> ControlFlow + Send + 'static,
```

sdroege: No I meant for the callback it's fine like it is

A6GibKm:
I removed the extra functions, since yeah in all cases you can unsafely use them with BorrowedFd::borrow_raw and OwnedFd::form_raw_fd.

gio/src/socket.rs
```diff
@@ -19,14 +19,16 @@ impl Socket {
     #[cfg(unix)]
     #[cfg_attr(docsrs, doc(cfg(unix)))]
     #[allow(clippy::missing_safety_doc)]
-    pub unsafe fn from_fd(fd: impl IntoRawFd) -> Result<Socket, glib::Error> {
+    pub fn from_fd(fd: OwnedFd) -> Result<Socket, glib::Error> {
```
sdroege, line 22:
```suggestion
    #[doc(alias = "g_socket_new_from_fd")]
    pub fn from_fd(fd: OwnedFd) -> Result<Socket, glib::Error> {
```

gio/src/socket.rs
```diff
@@ -19,14 +19,16 @@ impl Socket {
...
-        if error.is_null() {
-            Ok(from_glib_full(ret))
-        } else {
-            Err(from_glib_full(error))
+        unsafe {
+            let ret = ffi::g_socket_new_from_fd(fd, &mut error);
+            if error.is_null() {
+                Ok(from_glib_full(ret))
+            } else {
+                Err(from_glib_full(error))
```
sdroege, line 30: > On success, the returned GSocket takes ownership of fd. On failure, the caller must close fd themselves.

So you need to `close()` the fd here

gio/src/subprocess_launcher.rs
```diff
@@ -1,7 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #[cfg(any(unix, all(docsrs, unix)))]
-use std::os::unix::io::IntoRawFd;
+use std::os::unix::io::{IntoRawFd, OwnedFd};
```
sdroege, line 4: Below: "unused import: `ffi`" on Windows so maybe just move that part into a `#[cfg(unix)]` while we're at it

gio/src/subprocess_launcher.rs
```diff
@@ -19,40 +19,41 @@ impl SubprocessLauncher {
     #[cfg(unix)]
     #[cfg_attr(docsrs, doc(cfg(unix)))]
     #[doc(alias = "g_subprocess_launcher_take_fd")]
-    pub fn take_fd(&self, source_fd: impl IntoRawFd, target_fd: impl IntoRawFd) {
+    pub fn take_fd(&self, source_fd: Option<OwnedFd>, target_fd: Option<OwnedFd>) {
```
sdroege, line 22: I don't think this function allows `-1`. Only the other 3 (stdin/err/out) do

sdroege, line 22: Only `source_fd` is taken ownership of, `target_fd` is just borrowed> AFAIU

> This function takes ownership of the source_fd; it will be closed in the parent when self is freed.

gio/src/unix_fd_list.rs
```diff
@@ -47,41 +47,107 @@ pub trait UnixFDListExtManual: sealed::Sealed + IsA<UnixFDList> + Sized {
...
+        }
+    }
+}
+
+impl std::iter::IntoIterator for FdArray {
+    type Item = OwnedFd;
+    type IntoIter = FdIterator;
+
+    fn into_iter(mut self) -> Self::IntoIter {
+        // Can I .take() the ptr and set it to null here?
```
sdroege, line 143: No, but but you should set the `len` in the `FdArray` to `0` so you don't drop the fds twice

gio/src/unix_fd_list.rs
```diff
@@ -47,41 +47,107 @@ pub trait UnixFDListExtManual: sealed::Sealed + IsA<UnixFDList> + Sized {
...
+        let len = std::mem::take(&mut self.len);
+        FdIterator { len, ptr: self.ptr }
+    }
+}
+
+impl FdArray {
+    pub fn as_slice(&self) -> &[BorrowedFd<'_>] {
```
sdroege, line 150: As you didn't implement `Deref<Target = [OwnedFd]>`, it would be nice if you implemented a non-destructive `iter()` too. Personally, I would just implement `Deref` :)

gio/src/unix_fd_list.rs
```diff
@@ -47,41 +47,107 @@ pub trait UnixFDListExtManual: sealed::Sealed + IsA<UnixFDList> + Sized {
...
+
+impl Iterator for FdIterator {
+    type Item = OwnedFd;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.len > 0 {
+            let current = self.ptr.as_ptr();
+            if self.len > 1 {
+                let next = unsafe { self.ptr.as_ptr().add(1) };
+                self.ptr = ptr::NonNull::new(next).unwrap();
```
sdroege, line 112: This doesn't work. You'll have to keep track of the original pointer so it can be freed on `Drop` (of the `FdIterator`)

gio/src/unix_fd_list.rs
```diff
@@ -47,41 +47,107 @@ pub trait UnixFDListExtManual: sealed::Sealed + IsA<UnixFDList> + Sized {
...
+impl Drop for FdArray {
+    fn drop(&mut self) {
+        while self.len > 0 {
+            unsafe {
+                let current = self.ptr.as_ptr();
+                libc::close(*current);
+            }
+            if self.len > 1 {
+                let next = unsafe { self.ptr.as_ptr().add(1) };
+                self.ptr = ptr::NonNull::new(next).unwrap();
```
sdroege, line 131: This doesn't work, you'll also have to free the original pointer

gio/src/unix_fd_list.rs
```diff
@@ -47,41 +47,107 @@ pub trait UnixFDListExtManual: sealed::Sealed + IsA<UnixFDList> + Sized {
...
+    ptr: ptr::NonNull<libc::c_int>,
+    len: usize,
+}
+
+pub struct FdIterator {
+    ptr: ptr::NonNull<libc::c_int>,
+    len: usize,
+}
+
+impl Iterator for FdIterator {
```
sdroege, line 104: It would be nice to also implement `size_hint()` and `DoubleEndedIterator` and `ExactSizeIterator` and `FusedIterator` here

gio/src/unix_input_stream.rs
```diff
@@ -1,27 +1,24 @@
...
     // rustdoc-stripper-ignore-next
     /// Creates a new [`Self`] that takes ownership of the passed in fd.
-    ///
-    /// # Safety
-    /// You must not close the fd unless you've previously called [`UnixInputStreamExtManual::set_close_fd`]
-    /// with `true` on this stream. At which point you may only do so when all references to this
-    /// stream have been dropped.
     #[doc(alias = "g_unix_input_stream_new")]
-    pub unsafe fn take_fd(fd: impl IntoRawFd) -> UnixInputStream {
+    pub fn take_fd(fd: OwnedFd) -> UnixInputStream {
```
sdroege, line 16: This seems misnamed. Should be `from_fd()` or even just `new()`?

gio/src/unix_input_stream.rs
```diff
@@ -1,27 +1,24 @@
...
         let fd = fd.into_raw_fd();
         let close_fd = true.into_glib();
-        InputStream::from_glib_full(ffi::g_unix_input_stream_new(fd, close_fd)).unsafe_cast()
+        unsafe {
+            InputStream::from_glib_full(ffi::g_unix_input_stream_new(fd, close_fd)).unsafe_cast()
+        }
```
sdroege, line 21: `with_fd()` below is unnecessary now AFAICS

gio/src/unix_output_stream.rs
```diff
@@ -1,27 +1,24 @@
...
     // rustdoc-stripper-ignore-next
     /// Creates a new [`Self`] that takes ownership of the passed in fd.
-    ///
-    /// # Safety
-    /// You must not close the fd unless you've previously called [`UnixOutputStreamExtManual::set_close_fd`]
-    /// on this stream. At which point you may only do so when all references to this stream have
-    /// been dropped.
     #[doc(alias = "g_unix_output_stream_new")]
-    pub unsafe fn take_fd(fd: impl IntoRawFd) -> UnixOutputStream {
+    pub fn take_fd(fd: OwnedFd) -> UnixOutputStream {
```
sdroege, line 16: Same two comments as for the UNIX input stream

glib/src/functions.rs
```diff
@@ -24,15 +24,15 @@ use crate::{Error, Pid, SpawnFlags};
...
     argv: &[&str],
     envp: &[&str],
     flags: SpawnFlags,
     child_setup: Option<Box_<dyn FnOnce() + 'static>>,
-    stdin_fd: T,
-    stdout_fd: U,
-    stderr_fd: V,
+    stdin_fd: Option<impl AsFd>,
+    stdout_fd: Option<impl AsFd>,
+    stderr_fd: Option<impl AsFd>,
```
sdroege, line 35: `Option<impl ...>` is usually a bad idea. Just try calling this function with a `None` :)

I think here and in the other APIs using an `Option<impl ...>` currently you'll have to add a builder pattern instead

glib/src/functions.rs
```diff
@@ -238,7 +241,7 @@ pub fn compute_checksum_for_string(
 
 #[cfg(unix)]
 #[doc(alias = "g_unix_open_pipe")]
-pub fn unix_open_pipe(flags: i32) -> Result<(RawFd, RawFd), Error> {
+pub unsafe fn unix_open_pipe(flags: i32) -> Result<(RawFd, RawFd), Error> {
```
sdroege, line 244: Should probably return `OwnedFd`?

glib/src/functions.rs
```diff
@@ -132,7 +135,7 @@ pub fn spawn_async_with_fds<P: AsRef<std::path::Path>, T: AsRawFd, U: AsRawFd, V
 #[cfg(not(windows))]
 #[cfg_attr(docsrs, doc(cfg(not(windows))))]
 #[doc(alias = "g_spawn_async_with_pipes")]
-pub fn spawn_async_with_pipes<
+pub unsafe fn spawn_async_with_pipes<
```
sdroege, line 138: Why not make this use the new traits?

gio/src/unix_input_stream.rs
A6GibKm: Renamed to from_fd

glib/src/functions.rs
A6GibKm: Is that compatible with the `flags` argument that allows setting O_CLOEXEC/FD_CLOEXEC?

sdroege: I think so

nagisa: As a prior art `rustix::fd::open` also returns `OwnedFd` and allows passing in the CLOEXEC flag.

A6GibKm:
Perhaps we could split this into smaller chunks.

sdroege:
Go for it, or maybe @nagisa wants to take this over to bring it over the finish line?

A6GibKm:
> Go for it, or maybe @nagisa wants to take this over to bring it over the finish line?

See https://github.com/gtk-rs/gtk-rs-core/pull/1687.

bilelmoussaoui:
@A6GibKm would you mind rebasing this one? i would like to switch to 2024 edition before the next breaking release

sdroege:
How does this relate to 2024 edition? But regardless, it would be nice to finish this :)

A6GibKm:
I think all the important bits were added in https://github.com/gtk-rs/gtk-rs-core/pull/1687 and https://github.com/gtk-rs/gtk-rs-core/pull/1696. I will check and close if this is the case.

gio/src/unix_fd_list.rs
sdroege: In gstreamer-rs I have [this macro](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/bf982f422156f1424fc894efa77de964757fb7fd/gstreamer/src/utils.rs#L82) for defining iterators like this because they happen quite often. Might be helpful here

A6GibKm:
I am closing in favor of #1745. If someone wants to pick up [wip: unix_fd_list: Use custom iterator for steal_fds](https://github.com/gtk-rs/gtk-rs-core/pull/871/commits/696197e0a78e4e18a0a059f43c846fc59ca98ae4), feel free to do so.

sdroege:
Can you create an issue for tracking the missing parts?

PR #875, opened 2022-12-31
Improve `glib::collections` API and make it more versatile

sdroege:
CC @jf2048


glib/src/collections/strv.rs
```diff
@@ -0,0 +1,1520 @@
...
+use std::{ffi::c_char, fmt, marker::PhantomData, mem, ptr};
+
+use crate::{translate::*, GStr, GString};
+
+// rustdoc-stripper-ignore-next
+/// `NULL`-terminated UTF-8 string as stored in [`StrV`].
+///
+/// Unlike [`&GStr`] this does not have its length stored.
+#[repr(transparent)]
+pub struct StrVItem(ptr::NonNull<c_char>);
```
sdroege, line 12: @bilelmoussaoui @GuillaumeGomez @jf2048 Any suggestions for a better name here? `StrPtr` maybe?

GuillaumeGomez (approved):

glib/src/collections/strv.rs
GuillaumeGomez: Let's go with `StrPtr` then.

sdroege: > @pbor: the downside if that if we make it a simple name people will reach out to it even when they should use GStr/GString

So keeping it as `StrVItem` might actually not be a bad idea.

PR #876, opened 2022-12-31
gio: simplify async initable

pbor:
Use CancellableFuture to clean up the code a little

### DIFF
--- a/gio/src/subclass/async_initable.rs
+++ b/gio/src/subclass/async_initable.rs
@@ -4,7 +4,9 @@ use std::{future::Future, pin::Pin, ptr};
 
 use glib::{prelude::*, subclass::prelude::*, thread_guard::ThreadGuard, translate::*, Error};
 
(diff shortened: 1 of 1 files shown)


gio/src/subclass/async_initable.rs
```diff
@@ -100,12 +102,12 @@ unsafe extern "C" fn async_initable_init_async<T: AsyncInitableImpl>(
 ) {
     let instance = &*(initable as *mut T::Instance);
     let imp = instance.imp();
-    let cancellable = from_glib_borrow::<_, Option<Cancellable>>(cancellable);
+    let cancellable = Option::<Cancellable>::from_glib_none(cancellable);
```
sdroege, line 105: Why?

pbor: Because I need to pass it by value to the cancellable future

sdroege (approved):

PR #878, opened 2023-01-02
glib: Optimize various from/to `Vec` FFI translation functions

sdroege:
Use memcpys if possible instead of manual loops, and directly write into
the `Vec` memory and set the length afterwards to avoid having the
reallocate logic at every step.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #879, opened 2023-01-03
glib: Fix `glib::wrapper!` for `BoxedInline` with generic parameters

sdroege:


sdroege:
Needed for https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/1180

glib/src/boxed_inline.rs
```diff
@@ -29,7 +29,7 @@ macro_rules! glib_boxed_inline_wrapper {
         impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? std::marker::Copy for $name $(<$($generic),+>)? {}
 
         $crate::glib_boxed_inline_wrapper!(
-            @generic_impl [$($attr)*] $name $(<$($generic $(: $bound $(+ $bound2)*)?),+>)?, $ffi_name,
+            @generic_impl $name $(<$($generic $(: $bound $(+ $bound2)*)?),+>)?, $ffi_name,
```
GuillaumeGomez, line 32: Removing this means you won't be able to add documentation anymore for these types. Is it expected?

sdroege: Yes, because it's completely unused in that macro call. The documentation is already added in another place.

We unnecessarily passed the attributes into this macro call, i.e. they are not actually used in there because they're already used up elsewhere.

GuillaumeGomez: No `repr(c)` or `repr(transparent)` either?

sdroege: Same answer: that's already handled before (and in this case it is always `#[repr(transparent)]` anyway)

sdroege: Look in line 12/13

GuillaumeGomez: Perfect then!

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #880, opened 2023-01-04
application: Return ExitCode

A6GibKm:
Allows writing

```rust
    fn main() -> ExitCode {
        let app = gio::Application::new(None, Default::default());
        // ...
        app.run()
    }
```

in Rust 1.61 or newer. 

Note that the [termination](https://doc.rust-lang.org/std/process/trait.Termination.html) trait is supported for any Result<T: Termination, E: Debug>, we might want to return that instead, but it is kinda redundant at gtk already returns errors as a number.

### DIFF
--- a/gio/src/application.rs
+++ b/gio/src/application.rs
@@ -6,16 +6,16 @@ use glib::{
     prelude::*,
     signal::{connect_raw, SignalHandlerId},
     translate::*,
-    GString,
+    ExitCode, GString,
 };
 
 use crate::{Application, File};
 
 pub trait ApplicationExtManual {
     #[doc(alias = "g_application_run")]
-    fn run(&self) -> i32;
+    fn run(&self) -> ExitCode;
     #[doc(alias = "g_application_run")]
-    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> i32;
+    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> ExitCode;
     fn connect_open<F: Fn(&Self, &[File], &str) + 'static>(&self, f: F) -> SignalHandlerId;
 
     #[doc(alias = "g_application_hold")]
@@ -26,16 +26,17 @@ pub trait ApplicationExtManual {
 }
 
 impl<O: IsA<Application>> ApplicationExtManual for O {
-    fn run(&self) -> i32 {
+    fn run(&self) -> ExitCode {
         self.run_with_args(&std::env::args().collect::<Vec<_>>())
     }
 
-    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> i32 {
+    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> ExitCode {
         let argv: Vec<&str> = args.iter().map(|a| a.as_ref()).collect();
         let argc = argv.len() as i32;
-        unsafe {
+        let exit_code = unsafe {
             ffi::g_application_run(self.as_ref().to_glib_none().0, argc, argv.to_glib_none().0)
-        }
+        };
+        ExitCode::from(exit_code)
     }
 
     fn connect_open<F: Fn(&Self, &[File], &str) + 'static>(&self, f: F) -> SignalHandlerId {
--- a/gio/src/subclass/application.rs
+++ b/gio/src/subclass/application.rs
@@ -2,7 +2,7 @@
 
 use std::{ffi::OsString, fmt, ops::Deref, ptr};
 
-use glib::{subclass::prelude::*, translate::*, Cast, VariantDict};
+use glib::{subclass::prelude::*, translate::*, Cast, ExitCode, VariantDict};
 use libc::{c_char, c_int, c_void};
 
 use crate::Application;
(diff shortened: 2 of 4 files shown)


A6GibKm:
mmm since it does not provide hash, partialeq or eq the test can only be removed...

gio/src/application.rs
```diff
@@ -26,16 +26,17 @@ pub trait ApplicationExtManual {
...
-    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> i32 {
+    fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> ExitCode {
         let argv: Vec<&str> = args.iter().map(|a| a.as_ref()).collect();
         let argc = argv.len() as i32;
-        unsafe {
+        let exit_code = unsafe {
             ffi::g_application_run(self.as_ref().to_glib_none().0, argc, argv.to_glib_none().0)
-        }
+        };
+        ExitCode::from(exit_code as u8)
```
sdroege, line 39: Isn't it common to return `-1`? That would not work anymore now.

Please check what `g_application_run()` can potentially return and if we would run into problems here.

sdroege:
> mmm since it does not provide hash, partialeq or eq the test can only be removed...

Yes, remove that assertion

gio/src/application.rs
sdroege: Yes but is that going to behave correctly? I guess so, as exit code on POSIX at least are always u8 anyway IIRC.

Where does this return value come from exactly, and can an application override what `run()` is returning somehow?

A6GibKm: > Where does this return value come from exactly, and can an application override what run() is returning somehow?

There are 
```
    fn command_line(&self, command_line: &crate::ApplicationCommandLine) -> i32 {
        self.parent_command_line(command_line)
    }

    fn local_command_line(&self, arguments: &mut ArgumentList) -> Option<i32> {
        self.parent_local_command_line(arguments)
    }
```
in ApplicationImpl, but the return type cannot be ExitCode as you cannot cast it back to u8/i32 and use that in the C trampoline.

> Yes but is that going to behave correctly? I guess so, as exit code on POSIX at least are always u8 anyway IIRC.

I have no clue, but ExitCode intentionally has only From<u8> for compatibility reasons.

sdroege: I think we can't really use `ExitCode` here then as that would break applications that return some specific `i32` there

sdroege: Or what do you think @A6GibKm ?

A6GibKm: I am not sure, losing info when doing i32 -> u8 is not ideal but tolerable and makes things portable... on the other hand not being able to read the u8... that might completely ruin someone's code, after app.run() might not be the actual exit code of an app, maybe https://doc.rust-lang.org/std/process/struct.ExitStatus.html its more appropiate? If not we can always do a wrapper around the i32 that implements the termination trait.

sdroege: I think implementing `Termination` around some wrapper type that allows getting back the original `i32` would make most sense.

sdroege:
@A6GibKm What should we do about this?

gio/src/application.rs
A6GibKm: Added that

glib/src/exit_code.rs
```diff
@@ -0,0 +1,22 @@
+use std::process::Termination;
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct ExitCode(i32);
```
sdroege, line 4: Maybe make the field public too, and `PartialOrd` / `Ord`, and maybe success/failure constants like in `std`?

gio/src/subclass/application.rs
```diff
@@ -468,7 +468,5 @@ mod tests {
             .build();
 
         app.set_inactivity_timeout(10000);
-
-        assert_eq!(app.run_with_args(&["--local"]), EXIT_STATUS);
```
sdroege, line 472: This can be done again now or not?

But you probably want to switch `local_command_line()`, `command_line()` and `handle_local_options()` to also use the new type here

sdroege (commented):
Seems good to me otherwise, thanks for updating :)

gio/src/subclass/application.rs
A6GibKm: Done, thanks I forgot about it since it wasn't possible without this wrapper.

sdroege: You missed `handle_local_options()`

glib/src/exit_code.rs
sdroege: I think this is the only one missing?

glib/src/exit_code.rs
```diff
@@ -0,0 +1,26 @@
+use std::process::Termination;
```
sdroege, line 1:
```
==> Checking license headers from `../glib/src`
xx> Missing header in `../glib/src/exit_code.rs`
```

Need to add the first two lines from the other files here too

glib/src/exit_code.rs
A6GibKm: > Maybe make the field public too

since we have From in both directions I am not supper happy about this, but we could have `pub fn value(&self) ->i32`.

A6GibKm: I mean the source    at rust `pub const FAILURE: ExitCode = ExitCode(imp::ExitCode::FAILURE); ` is super obscure and I couldn't find where it comes from.

sdroege: Ok, just a `pub fn value(&self) -> i32` then :)

sdroege: And `FAILURE = 1`

sdroege (approved):

PR #890, opened 2023-01-13
glib: implement ToGlibPtr<*mut _> for boxed types

### LINKED ISSUE #889: [BUG] boxed types do not implement the `ToGlibPtr` for `*const` pointers
**Bug description**

It looks like boxed types do not implement the `ToGlibPtr` to convert from the wrapped object to the `*const` pointer.

Seems that [only the `*mut`](https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/boxed.rs#L78) implementation is present.
Is there any specific reason for that?
Thanks!

mbiggio:
Fixes #889

### DIFF
--- a/glib/src/value_array.rs
+++ b/glib/src/value_array.rs
@@ -114,8 +114,8 @@ impl ops::Deref for ValueArray {
 
         unsafe {
             slice::from_raw_parts(
-                (*self.to_glib_none().0).values as *const Value,
-                (*self.to_glib_none().0).n_values as usize,
+                (*ToGlibPtr::<*const _>::to_glib_none(self).0).values as *const Value,
+                (*ToGlibPtr::<*const _>::to_glib_none(self).0).n_values as usize,
             )
         }
     }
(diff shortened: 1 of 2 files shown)


bilelmoussaoui (approved):

glib/src/value_array.rs
```diff
@@ -114,8 +114,8 @@ impl ops::Deref for ValueArray {
 
         unsafe {
             slice::from_raw_parts(
-                (*self.to_glib_none().0).values as *const Value,
-                (*self.to_glib_none().0).n_values as usize,
+                (*ToGlibPtr::<*const _>::to_glib_none(self).0).values as *const Value,
```
sdroege, line 117: Might be nicer to just call `self.as_ptr()` here

sdroege (commented):

sdroege:
Closing in favour of https://github.com/gtk-rs/gtk-rs-core/pull/895

Issue #805, opened 2022-10-21
RFC: re-opening the `Into<Option<_>>` debate :)


fengalin:
## Another `Into<Option<_>>` experiment

In `gstreamer-rs` a [fix for nullability inconsistency] had an API changed induced an update to all the call sites for a function from`q.set_uri(uri)` to `q.set_uri(Some(uri))`. This revived the debate whether `Into<Option<_>>` should be used in argument position when the argument is optional.

This debate was [previously concluded]: "Into<Option<_>> considered harmful". The main arguments leading to this conclusion are:

- In some cases, this requires annotations, which can result messy, particularly   when dealing with closures.
- Users can be confused and not figure out that the API can also accept `None`.

We wanted to figure out whether the problems encountered at the time should still be considered harmefull and whether `Into<Option<_>>` should be reintroduce, at least in some cases, as we ended up doing for `set_uri`.

### TLDR

The TLDR is the above arguments are still valid:

- Type annotations are still necessary in some cases when we want to pass `None`. We probably don't want to use this with closures or complex function arguments.
- When users find code such as `q.set_uri(uri)`, they will probably not guess they can use `q.set_uri(None)` too.

Nevertheless, code looks nicer IMO when at least some of these arguments use this feature. I conducted an [experiment on `gstreamer-rs`] changing most candidates in manual code and updated `gst-plugins-rs` to use the resulting code base. See the difference in [this `gst-plugins-rs` commit].

In this issue, I'd like to show the result of a quick experimentation on `gtk-rs-core/glib` to illustrate the pros & cons. The changes and some illustrative test cases are available in [this `gtk-rs-core/glib` commit].

### The easy cases

There are easy cases for which there's no issue apart from the users not immediately figuring out that the arguments can be an `Option`.

#### `Copy + Clone` types

The `Copy + Clone` types handling is straightforward with limited impact on the function signature and body.

See the [`Channel::new`] implementation and the above test cases.

#### Reference to concrete types

When the argument's inner type is a reference to a concrete type, the implementation is quite straightforward too, with the addition of a lifetime.

See the [`DateTime::from_iso8601`] implementation and the above test cases.

#### Reference to a type by one of its trait implementation

This is where it starts to get tricky.

##### Strings

One very common case for this is when the argument is a string, like in `set_uri`. Currently, most functions use an `Option<&str>` for these kind of arguments. This allows using:

```rust
    q.my_function(Some("a `str` literal"));
[...]

PR #891, opened 2023-01-13
glib: Implement `IntoOptionalGStr` for plain string types

sdroege:
This way strings can be passed to such functions without explicit `Option` wrapping, see https://github.com/gtk-rs/gtk-rs-core/issues/805.

----

CC @jf2048 @fengalin 

Not sure if we want this change already for this release

### DIFF
--- a/glib/src/convert.rs
+++ b/glib/src/convert.rs
@@ -71,7 +71,7 @@ pub fn convert_with_fallback(
     str_: &[u8],
     to_codeset: impl IntoGStr,
     from_codeset: impl IntoGStr,
-    fallback: Option<impl IntoGStr>,
+    fallback: impl IntoOptionalGStr,
 ) -> Result<(Slice<u8>, usize), CvtError> {
     assert!(str_.len() <= isize::MAX as usize);
     let mut bytes_read = 0;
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):

jf2048:
TBH not a fan of this. I think `Into<Option<String>>` is getting too far away from idiomatic Rust

sdroege:
@jf2048 You probably want to comment in #805 then. This PR comes from the discussions there :)

fengalin:
My position has shifted two days after my initial try on #805. See [this comment](https://github.com/gtk-rs/gtk-rs-core/issues/805#issuecomment-1288079608). Sorry if that lead to confusions, I made it clear in the initial comment now.

The short version of my most recent position is:

- If the `None` case is the default, then the function with the same name as in the `sys` crate should NOT use any argument for the optional value. Another function with an explicit name would accept a non-`Option` argument to pass the value when needed.
- If the `Some` case is the default, then the function with the same name as in the `sys` crate should accept a non-`Option` argument to pass the value. Another function with an explicit name could allow selecting the `None` case.
- In both cases, another function such as `unset_*`, `discard_*` or `reset_*` would allow switching from `Some` value to `None` when necessary.

PR #895, opened 2023-01-14
glib: implement ToGlibPtr<*mut _> for boxed types

pbor:


pbor:
Same as https://github.com/gtk-rs/gtk-rs-core/pull/890 but with the requested changes

sdroege:
Now "only" need to fix the usage everywhere :)

pbor:
> Now "only" need to fix the usage everywhere :)

Fixed (I hope) but please double check that using `as_ptr()` instead of `to_glib_none().0` is legit

sdroege:
> Fixed (I hope) but please double check that using `as_ptr()` instead of `to_glib_none().0` is legit

They're doing the same thing, yes.

sdroege (approved):

PR #896, opened 2023-01-14
pango: use the new List api to simplify reorder_items

pbor:
As discussed in chat, we could also change the signature to take `&[Item]` and avoid the cloning, but I am not familiar enough with the use of this API to understand if that would be problematic for the caller

### DIFF
--- a/pango/src/functions.rs
+++ b/pango/src/functions.rs
@@ -1,7 +1,5 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::ptr;
-
 use glib::translate::*;
 
 #[cfg(any(feature = "v1_44", feature = "dox"))]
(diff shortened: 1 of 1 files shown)


pango/src/functions.rs
```diff
@@ -10,18 +8,9 @@ use crate::{Analysis, GlyphString, Item};
 
 #[doc(alias = "pango_reorder_items")]
 pub fn reorder_items(logical_items: &[&Item]) -> Vec<Item> {
```
sdroege, line 10:
```suggestion
pub fn reorder_items(logical_items: impl IntoIterator<Item = Item>) -> glib::List<Item> {
```

maybe? Or even

```suggestion
pub fn reorder_items(logical_items: glib::List<Item>) -> glib::List<Item> {
```

pbor: I am ok with either... as I mentioned I never used this API, so I am not sure how the caller gets the list, so I am not sure if taking a glib::List will be annoying

sdroege: Me neither but that's what you'd have in C so it's probably not wrong

pango/src/functions.rs
```diff
@@ -1,27 +1,17 @@
...
-            .map(|v| v.to_glib_none())
-            .collect();
-        let mut list: *mut glib::ffi::GList = ptr::null_mut();
-        for stash in &stash_vec {
-            list = glib::ffi::g_list_prepend(list, Ptr::to(stash.0));
-        }
-
-        FromGlibPtrContainer::from_glib_full(ffi::pango_reorder_items(list))
+        FromGlibPtrContainer::from_glib_full(ffi::pango_reorder_items(
+            logical_items.as_ptr() as *mut _
```
sdroege, line 13: Oh I misunderstood the API. I thought it takes ownership of the parameter and returns a modified list.
You can make it take a reference then.

sdroege (approved):

PR #898, opened 2023-01-15
gio: use StrV for the simple proxy resolver API

pbor:


sdroege (dismissed):

gio/src/tls_connection.rs
```diff
@@ -20,6 +20,11 @@ pub trait TlsConnectionExtManual {
         &self,
         type_: TlsChannelBindingType,
     ) -> Result<glib::ByteArray, glib::Error>;
+
+    #[cfg(any(feature = "v2_60", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_60")))]
+    #[doc(alias = "g_tls_connection_set_advertised_protocols")]
+    fn set_advertised_protocols(&self, protocols: IntoStrV);
```
sdroege, line 27:
```suggestion
    fn set_advertised_protocols(&self, protocols: impl IntoStrV);
```

etc

gio/Gir.toml
```diff
@@ -1364,6 +1364,10 @@ manual_traits = ["TlsConnectionExtManual"]
     name = "get_channel_binding_data"
     # Gir confuses the mutability of the data
     manual = true
+    [[object.function]]
```
bilelmoussaoui, line 1367: does it make sense to manually implement this function just for this? ideally someday we will fix gir to automatically generate proper code here..

gio/src/tls_connection.rs
pbor: oops. I do not have 2.60 locally so I did not see those. Fixed

gio/Gir.toml
pbor: I would rather start changing the api for the users and as soon as gir is updated we can drop the manual impls

sdroege (approved):

PR #899, opened 2023-01-15
settings: implement strv setter and getter manually

pbor:


gio/src/file_info.rs
```diff
@@ -47,4 +47,28 @@ impl FileInfo {
             );
         }
     }
+
+    #[doc(alias = "g_file_info_get_attribute_stringv")]
+    #[doc(alias = "get_attribute_stringv")]
+    pub fn attribute_stringv(&self, attribute: &str) -> StrV {
```
bilelmoussaoui, line 53: Why not use IntoGStr here ?

pbor: I did not do it because I do not think we want to manually impl all attributes getters and setters, and the auto one for now take str

sdroege: But now it's inconsistent :) That's going to be confusing.

Also once the code generator supports these things, `StrV` will also be covered.

sdroege: Oh I misunderstood. Yes I agree, the `attribute` should stay a `&str`. I thought you changed it to `impl IntoGStr` and *that* would then be inconsistent.

gio/src/file_info.rs
```diff
@@ -47,4 +47,28 @@ impl FileInfo {
...
+        unsafe {
+            FromGlibPtrContainer::from_glib_none(ffi::g_file_info_get_attribute_stringv(
```
sdroege, line 55: `FileInfo` is immutable, so this could also just be borrowed instead of creating a copy or not? (of course the same for everything else here)

I plan to add a configuration for making that possible with the new code generator, but as you say it will take some time.

pbor: you are right, even the C api is `[transfer none]`

sdroege: `(transfer none)` in the C API is often a trap. Pointers that are only valid until *something* happens. It can only be borrowed directly if the underlying object is immutable (or it is a boxed type that is not refcounted).

sdroege: This is clearly such a case :) There's also a setter for this, and if you call the setter your pointer becomes invalid. As the info is refcounted and setting values uses interior mutability, this is unsound unfortunately and we always need to copy.

pbor: ok, rolled back the change

sdroege (approved):

PR #901, opened 2023-01-18
glib: Add `MainContext::spawn_from_within()` for spawning non-`Send` …

sdroege:
…futures from another thread

This requires a `Send` closure to construct the non-`Send` future and is for example useful if spawning a future that creates non-`Send` GObjects on the main context from another thread.

----

CC @elmarco @pbor

### DIFF
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -199,7 +199,7 @@ pub use self::bridged_logging::{rust_log_handler, GlibLogger, GlibLoggerDomain, 
 pub mod subclass;
 
 mod main_context_futures;
-pub use main_context_futures::{JoinError, JoinHandle};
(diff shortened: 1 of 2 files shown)


elmarco:
Sorry, a bit too much going on for me here with the JoinHandle stuff! But it looks like the API I was looking for (I just wrapped spawn_local with invoke)

You could perhaps update the doc for spawn_local(), and cross-ref the methods documentations.

sdroege:
> Sorry, a bit too much going on for me here with the JoinHandle stuff! But it looks like the API I was looking for (I just wrapped spawn_local with invoke)

It's just all a bit complicated to have some kind of `JoinHandle` for this new set of functions. You probably don't use that, so it's equivalent to yours if you simply drop the return value :)

> You could perhaps update the doc for spawn_local(), and cross-ref the methods documentations.

Good idea, will do!

PR #902, opened 2023-01-18
glib: Add helper functions to spawn a non-`Send` future on the curren…

sdroege:
…t thread-default `MainContext`

----

CC @elmarco @pbor @A6GibKm


glib/src/lib.rs
```diff
@@ -199,7 +199,7 @@ pub use self::bridged_logging::{rust_log_handler, GlibLogger, GlibLoggerDomain,
 pub mod subclass;
 
 mod main_context_futures;
-pub use main_context_futures::{JoinError, JoinHandle};
+pub use main_context_futures::{spawn, spawn_with_priority, JoinError, JoinHandle};
```
sdroege, line 202: The naming of the functions is a bit suboptimal as there's also this set of functions: https://gtk-rs.org/gtk-rs-core/git/docs/glib/fn.spawn_async.html

Any suggestions?

GuillaumeGomez: Or directly make these two functions methods of `MainContext` and add doc aliases on them.

sdroege: There are already methods with that name on `MainContext`. The goal here is to have a function that gets the correct main context instance and then calls the method on that.

sdroege: `spawn_on_thread_default_main_context()` could work of course, but that's almost as long as the implementation of the function :)

sdroege: Maybe could move this into a `glib::futures` module, and also potentially move some other things in there. Not sure.

sdroege:
This is apparently not too useful because `g_application_run()` and `g_main_loop_run()` are not pushing their main context as the thread default one.

sdroege:
Maybe something like this. It would take the thread-default main context if any, otherwise the global default. And then make sure that the current thread actually owns it.

sdroege:
I guess there is not really any interest in having such a function? Let's close this then?

BrainBlasted:
Huh? This would definitely be helpful

sdroege:
I got no feedback at all so I assumed nobody actually needed it :) As it's apparently useful to you, do you have suggestions about the name and where to put it? Where/how would you expect it?

A6GibKm:
> Huh? This would definitely be helpful. Apps currently use a macro for this.

Note that they do ``MainContext::default()` rather than `MainContext::thread_default()`

sdroege:
> Note that they do `MainContext::default()` rather than `MainContext::thread_default()`, that has me a bit confused, if they are equivalent then sure this would be rather helpful.

For an application that didn't configure a custom thread default main context this would end up the same.

sdroege:
Conclusion seems to be to make a new `glib::futures` module and add this as `glib::futures::spawn()`, and I guess move various other function in there too.

BrainBlasted:
Has anyone started working in that direction?

sdroege:
I'm not aware of anybody having started such work

glib/src/main_context_futures.rs
```diff
@@ -552,6 +552,36 @@ impl Spawn for MainContext {
...
+    spawn_with_priority(crate::PRIORITY_DEFAULT, f)
+}
+
+// rustdoc-stripper-ignore-next
+/// Spawns an infallible, non-`Send` on the current thread-default [`MainContext`] with a
+/// non-default priority.
+///
+/// Panics if there is no thread-default main context on this thread and if this thread
+/// does not own the global default main context.
+pub fn spawn_with_priority<R: 'static, F: Future<Output = R> + 'static>(
```
bilelmoussaoui, line 570: shouldn't this be named spawn_local_with_priority?

PR #903, opened 2023-01-19
action_entry: Add activate_async helper

BrainBlasted:
This is a helper function in a similar vein to
gtk4-rs's `install_action_async ()` for widgets.
It allows applications to install actions
that use async/await without having to manually
create a glib::MainContext in the callback.


gio/src/action_entry.rs
```diff
@@ -87,6 +88,25 @@ where
         self
     }
 
+    pub fn activate_async<Fut, F>(mut self, callback: F) -> Self
+    where
+        F: Fn(&O, &SimpleAction, Option<&Variant>) -> Fut + 'static + Clone,
+        Fut: Future<Output = ()>,
+    {
+        let future_cb = move |map: &O, action: &SimpleAction, variant: Option<&Variant>| {
+            let ctx = glib::MainContext::default();
```
sdroege, line 97: This is potentially problematic. Do we always want to run things on the global default main context here, or could this also be on the thread-default one? See also https://github.com/gtk-rs/gtk-rs-core/pull/902

gio/src/action_entry.rs
```diff
@@ -87,6 +88,25 @@ where
...
+            let variant = variant.map(ToOwned::to_owned);
+            ctx.spawn_local(
```
sdroege, line 99: This will panic if the function is called from not the main thread

sdroege:
This should probably just call the helper function from https://github.com/gtk-rs/gtk-rs-core/pull/902 once merged.

gio/src/action_entry.rs
```diff
@@ -87,6 +88,30 @@ where
         self
     }
 
+    pub fn activate_async<Fut, F>(mut self, callback: F) -> Self
+    where
+        F: Fn(&O, &SimpleAction, Option<&Variant>) -> Fut + 'static + Clone,
```
sdroege, line 93: Why require `Clone` here? I think it would be better if you would instead use an `Rc` instead of a `Box` for storing the callback. Then you get cloning for free and don't make it the user's problem.

bilelmoussaoui: any chance we resolve this? it would be cool to merge this already without waiting endlessly for #902.

PR #908, opened 2023-01-21
glib: Deprecate paramspec `new()` functions in favour of the builder

sdroege:


sdroege:

sdroege:
Considering that nothing in gtk4-rs, gtk-rs-core or gstreamer-rs / gst-plugins-rs is using the normal constructors anymore and they're inconvenient to use, let's go ahead with this to get some noise out of the way.

Issue #831, opened 2022-11-22
`ParamSpecEnumBuilder` should take `Default::default()` into consideration


pbor:
**Description**

`ParamSpecEnumBuilder` today does not expose a `.default_value` setter. When asking on the chat, I think that today it is just `0` by default.

We should:
1. document the behavior
2. Since enums can implement the `Default` trait, we should honor it

sdroege:
> When asking on the chat, I think that today it is just `0` by default.

It requires passing the default value, always.

I think we should have

```rust
pub fn builder()<T: StaticType + FromGlib<i32> + IntoGlib<GlibType = i32> + Default>(
    name: &str,
) -> ParamSpecEnumBuilder<'_, T>;
pub fn builder()_with_default<T: StaticType + FromGlib<i32> + IntoGlib<GlibType = i32>>(
    name: &str,
    default_value: T,
) -> ParamSpecEnumBuilder<'_, T>
```

What do you think?

bilelmoussaoui:
I would leave the current constructor as is and add a `with_default` variant that would require `T: Default`

sdroege:
@bilelmoussaoui I also don't want to change the API, but a `with_default()` sounds like it as a default parameter :) Do you have another name suggestion?

pbor:
I do not think the API break is a problem as long it is something that the compiler can catch. The important thing is that we do not keep the same signature and change the behavior.

sdroege:
Well that's not possible here anyway (to keep the same signature but change the behaviour) :)

PR #909, opened 2023-01-21
glib: Implement enum paramspec builder variant that builds the defaul…

sdroege:
…t value automatically

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/831

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -1000,11 +1000,17 @@ impl ParamSpecEnum {
         }
     }
 
-    pub fn builder<T: StaticType + FromGlib<i32> + IntoGlib<GlibType = i32>>(
+    pub fn builder_with_default<T: StaticType + FromGlib<i32> + IntoGlib<GlibType = i32>>(
         name: &str,
         default_value: T,
     ) -> ParamSpecEnumBuilder<T> {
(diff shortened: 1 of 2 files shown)


sdroege:
@bilelmoussaoui @pbor What do you think?

pbor:
It still seems backward to me: if I have a enum with `Default` I just want to call `.build()` if I do not have a default, I can specify with it `with_default`.

What about adding a `with_default()` method that returns a new type `ParamSpecEnumBuilderWithDefault` which has a `build()` method and instead require `Default` for the `build()` method on `ParamSpecEnumBuilder`

sdroege:
How is that different from this API other than `build_with_default()` vs. `with_default()` as the function name?

pbor:
> How is that different from this API other than `build_with_default()` vs. `with_default()` as the function name?

Unless I misunderstood the PR, in your patch you use `build_with_default()` for enums that implement `Default`. In my proposal you need to use `with_default()` for enums that do *not* impl Default or if you want to force a different default value

sdroege:
Ah I see, so you want to have it the other way around. I think that makes more sense actually, yes.

sdroege:
Done that now.

pbor (approved):

PR #910, opened 2023-01-21
glib: Add new object constructor for constructing an object with default property values

sdroege:
Nicer than writing `Object::new(&[])`.

Also optimize the object builder a bit.

### DIFF
--- a/examples/gio_task/file_size/mod.rs
+++ b/examples/gio_task/file_size/mod.rs
@@ -13,7 +13,7 @@ unsafe impl Sync for FileSize {}
 
 impl FileSize {
     pub fn new() -> Self {
-        glib::Object::new(&[])
+        glib::Object::new_default()
     }
 
     pub fn retrieved_size(&self) -> Option<i64> {
--- a/gio/src/async_initable.rs
+++ b/gio/src/async_initable.rs
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::{boxed::Box as Box_, pin::Pin};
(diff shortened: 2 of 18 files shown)


pbor:
Is this really that common that we need to optimize for it? Usually this is wrapped into a `new()` method for the concrete object.

Maybe we should `impl Default` and have it call `Object::new(T::static_type(), &[])`?

`new_default_with_type` seems even less common and if one is writing some kind of factory object they can stick to passing in `&[]`

sdroege:
> Is this really that common that we need to optimize for it? Usually this is wrapped into a `new()` method for the concrete object.

All the `Default` impls are calling `new(&[])` currently, and I know that I'm having to write this quite often for subclasses. So at least for me it's going to save some annoying typing :)

> Maybe we should `impl Default` and have it call `Object::new(T::static_type(), &[])`?

This would be called from the `impl Default`, yes.

sdroege:
Also note that I'm working on deprecating `::new()` and `::with_type()` and `::with_values()` as those all unnecessarily copy the property values twice.

pbor:
Ah ok then, if the redundancy is removed the other way round then ok

pbor:
Initable and AsyncInitable will need the same updates for consistency

sdroege:
> Initable and AsyncInitable will need the same updates for consistency

Yeah they should also get builders. That API became quite bitrotten already...

sdroege:
As discussed, names are as good as it can be for this release and for the release afterwards (after having deprecated `new()` for this release), we can switch over from `new_default()` to `new()`. There's also an issue so we don't forget.

I'll finish the (async) initable changes and then merge this.

PR #914, opened 2023-01-21
Regenerate with latest gir and mark suboptimal `Object` constructors as deprecated

sdroege:


sdroege:
Actually don't merge this yet. I have another improvement for the builders pending.

gio/src/auto/data_output_stream.rs
```diff
@@ -36,65 +36,61 @@ impl DataOutputStream {
...
 
-    pub fn byte_order(mut self, byte_order: DataStreamByteOrder) -> Self {
-        self.byte_order = Some(byte_order);
-        self
+    pub fn base_stream(self, base_stream: &impl IsA<OutputStream>) -> Self {
+        Self {
+            builder: self
+                .builder
+                .property("base-stream", base_stream.clone().upcast()),
+        }
```
bilelmoussaoui, line 78: why not something like 
```rust 
pub fn base_stream(mut self, base_stream: &impl IsA<OutputStream>) -> Self {
    self.builder.property("base-stream", base_stream.clone().upcast());
    self
}
```

sdroege: Because that's not possible. `property()` consumes the builder and returns a new builder.

PR #915, opened 2023-01-21
gio: bind GFileDescriptorBased

jf2048:
This is a breaking change, the `fd` method conflicts with the implementations on `UnixInputStreamExtManual` and `UnixOutputStreamExtManual`

cc @BrainBlasted


gio/src/file_descriptor_based.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+pub trait FileDescriptorBasedExtManual: 'static {
+    #[doc(alias = "g_file_descriptor_based_get_fd")]
+    #[doc(alias = "get_fd")]
+    fn fd<T: FromRawFd>(&self) -> T;
+}
+
+impl<O: IsA<FileDescriptorBased>> FileDescriptorBasedExtManual for O {
+    fn fd<T: FromRawFd>(&self) -> T {
```
sdroege, line 37: > This is a breaking change, the `fd` method conflicts with the implementations on `UnixInputStreamExtManual` and `UnixOutputStreamExtManual`

Why don't you remove them then? :)

sdroege (commented):
Looks good to me otherwise

gio/src/file_descriptor_based.rs
jf2048: They have been removed

sdroege (approved):

PR #917, opened 2023-01-21
build-tools: Allow passing multiple source dirs to `compile_resources`

RealKC:
You can pass the command line tool multiple source dirs, so we should allow it too.

---

If this gets merged, I'll send corresponding PRs to the gtk4-rs/gtk3-rs repositories too

### DIFF
--- a/glib-build-tools/src/lib.rs
+++ b/glib-build-tools/src/lib.rs
@@ -10,18 +10,22 @@ use std::{env, path::Path, process::Command};
 ///
 /// ```no_run
 /// glib_build_tools::compile_resources(
-///     "resources",
+///     &["resources"],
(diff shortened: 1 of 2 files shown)


bilelmoussaoui (dismissed):

glib-build-tools/src/lib.rs
```diff
@@ -15,13 +15,17 @@ use std::{env, path::Path, process::Command};
 ///     "compiled.gresource",
```
sdroege, line 15: The docs here need to be updated too :)

RealKC: Fixed!

sdroege (approved):

PR #920, opened 2023-01-23
gio: use GStr for the manual extension point implenentation

pbor:


gio/src/io_extension_point.rs
```diff
@@ -80,16 +80,16 @@ impl IOExtensionPoint {
     // rustdoc-stripper-ignore-next
     /// Create a new builder for an extension point.
     #[doc(alias = "g_io_extension_point_register")]
-    pub fn builder(name: &str) -> IOExtensionPointBuilder {
-        IOExtensionPointBuilder::new(name)
+    pub fn builder(name: impl IntoGStr) -> IOExtensionPointBuilder {
+        name.run_with_gstr(|name| IOExtensionPointBuilder::new(name.to_owned()))
```
pbor, line 84: this makes a copy with `to_owned` but I assume a copy was happening anyway before when calling `to_glib_none` for `&str`... not that it matters in any way, just to confirm my understanding

sdroege: This defeats the purpose of the trait. Just do `impl Into<glib::GString>` here :)

sdroege (approved):

PR #925, opened 2023-01-27
ci: Only deny clippy warnings with stable clippy

sdroege:


graphene/src/lib.rs
```diff
@@ -19,7 +19,7 @@ macro_rules! skip_assert_initialized {
 #[allow(clippy::transmute_ptr_to_ref)]
 #[allow(clippy::type_complexity)]
 #[allow(clippy::unreadable_literal)]
-#[allow(clippy::derive_hash_xor_eq)]
```
pbor, line 22: shall we keep the old one with a comment like `// remove once we bump MSRV to 1.67` ?

sdroege: Problem is that right now it either fails the stable or beta clippy CI job. I think we should make the beta job warn-only.

PR #926, opened 2023-01-28
Use `PtrSlice<GStrPtr>` for `KeyFile` methods

pbor:


glib/src/key_file.rs
```diff
@@ -167,7 +203,7 @@ impl KeyFile {
 
     #[doc(alias = "g_key_file_get_string_list")]
     #[doc(alias = "get_string_list")]
-    pub fn string_list(&self, group_name: &str, key: &str) -> Result<Vec<GString>, Error> {
+    pub fn string_list(&self, group_name: &str, key: &str) -> Result<StrV, Error> {
```
pbor, line 206: Should it return `(StrV, usize)` to get access to the length instead? There is an inconsistency here even before my patch with `keys()` and `groups()`: either we return StrV and size in all of these or we return Vec and omit size

pbor: I went with the first option since I believe the codegen will generate that once it learns about strv

sdroege: I think you can generate a `glib::Slice<StrVItem>` or `glib::PtrSlice<StrVItem>` in this case?

pbor: With either of those (and the difference is not very clear btw) I get

`the trait `translate::TransparentPtrType` is not implemented for `StrVItem``

sdroege: You could implement that though.

The difference is
  - `TransparentType`: A reference of the Rust type `&T` is equivalent to the native C pointer `*const ffi::T`
  - `TransparentPtrType`: The Rust type itself `T` is equivalent to the native C pointer `*const ffi::T`, or in other words `&T` is equivalent to `*const *const ffi::T`

The first is true for `glib::Value`, boxed inline, and such types, the second is true for GObjects, shared, boxed types.

As `StrVItem` is basically a `*const c_char` this means that it is a `TransparentPtrType`.

glib/src/key_file.rs
```diff
@@ -85,6 +85,39 @@ impl KeyFile {
         }
     }
 
+    #[doc(alias = "g_key_file_get_groups")]
+    #[doc(alias = "get_groups")]
+    pub fn groups(&self) -> PtrSlice<GStrPtr> {
+        unsafe {
+            let mut length = mem::MaybeUninit::uninit();
+            let ret = ffi::g_key_file_get_groups(self.to_glib_none().0, length.as_mut_ptr());
+            FromGlibContainer::from_glib_container_num(ret, length.assume_init() as _)
```
sdroege, line 94: These are all copied of course

glib/src/key_file.rs
```diff
@@ -85,6 +85,39 @@ impl KeyFile {
...
+            let mut length = mem::MaybeUninit::uninit();
+            let mut error = ptr::null_mut();
+            let ret = ffi::g_key_file_get_keys(
+                self.to_glib_none().0,
+                group_name.to_glib_none().0,
+                length.as_mut_ptr(),
+                &mut error,
+            );
+            if error.is_null() {
+                Ok(FromGlibContainer::from_glib_full_num(
```
sdroege, line 111: These not, which is nicer

sdroege (commented):
Is it guaranteed for both that there's a `NULL` item at the end? If not you want to use `Slice` instead of `PtrSlice`. That's the main difference between the two

pbor:
Oops, I missed that one was not `full`. As far as I understand these function all return a newly allocated, null terminated array, so I think `PtrSlice` is correct

sdroege:
CI is not going to pass for a while until GNOME infrastructure is back up. But looks all good

sdroege (approved):

PR #928, opened 2023-01-30
gdk-pixbuf: check if either width/height is null before assignment in animation_get_size()

aruiz:


gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -141,8 +141,12 @@ unsafe extern "C" fn animation_get_size<T: PixbufAnimationImpl>(
     let imp = instance.imp();
 
     let (width, height) = imp.size();
```
sdroege, line 143: Maybe return directly without calling `size()` if both are `NULL`?

Looks good to me otherwise

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -137,12 +137,20 @@ unsafe extern "C" fn animation_get_size<T: PixbufAnimationImpl>(
     width_ptr: *mut libc::c_int,
     height_ptr: *mut libc::c_int,
 ) {
+    if width_ptr == null_mut() && height_ptr == null_mut() {
```
sdroege, line 140:
```suggestion
    if width_ptr.is_null() && height_ptr.is_null() {
```

also below

sdroege: You missed the "also below" :)

aruiz: Oopse.

aruiz: Fixed in the last force push

sdroege (approved):

PR #936, opened 2023-01-31
Add TransparentPtr marker trait to List, SList and StrVItem

pbor:


glib/src/collections/strv.rs
```diff
@@ -11,6 +11,21 @@ use crate::{translate::*, GStr, GString};
...
+
+#[doc(hidden)]
+impl GlibPtrDefault for StrVItem {
+    type GlibType = *mut c_char;
+}
+
+impl Clone for StrVItem {
+    #[inline]
+    fn clone(&self) -> StrVItem {
+        unsafe { StrVItem(ptr::NonNull::new_unchecked(ffi::g_strdup(self.0.as_ptr()))) }
```
sdroege, line 25: You also need `Drop` then

sdroege: Might make sense to rename this and give it a constructor too. By having `Clone` we make it proper API :)

sdroege: `GStrPtr`?

pbor: Ok... let me drop it from this PR and open a separate one

sdroege (approved):

Issue #932, opened 2023-01-30
properties: Can't use the macro for a custom glib::Boxed type property


melix99:
This is a minimal reproducer of what I'm trying to do: [lib.zip](https://github.com/gtk-rs/gtk-rs-core/files/10538714/lib.zip).

The compilation fails with:

```
error[E0277]: the trait bound `RefCell<Option<StrBoxed>>: Property` is not satisfied
  --> src/lib.rs:16:21
   |
16 |     #[derive(Debug, Properties, Default)]
   |                     ^^^^^^^^^^ the trait `Property` is not implemented for `RefCell<Option<StrBoxed>>`
   |
   = help: the trait `Property` is implemented for `RefCell<T>`
   = note: this error originates in the derive macro `Properties` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
```

Am I just doing something wrong?

bilelmoussaoui:
We are probably missing a HasParamSpec trait impl, could you try adding a manual implementation to your type and see if it helps?

melix99:
@bilelmoussaoui That gives me:

```
error[E0119]: conflicting implementations of trait `glib::HasParamSpec` for type `StrBoxed`
 --> src/lib.rs:4:17
  |
4 | #[derive(Clone, glib::Boxed)]
  |                 ^^^^^^^^^^^ conflicting implementation for `StrBoxed`
...
8 | impl glib::HasParamSpec for StrBoxed {
  | ------------------------------------ first implementation here
  |
  = note: this error originates in the derive macro `glib::Boxed` (in Nightly builds, run with -Z macro-backtrace for more info)
```

sdroege:
I think what we're missing is the `Option<T>` part here, not `StrBoxed`. That would need a generic impl in `glib`

bilelmoussaoui:
> I think what we're missing is the `Option<T>` part here, not `StrBoxed`. That would need a generic impl in `glib`

Isn't it expected that `ParamSpecBoxed` doesn't accept nullable values? as in, I would expect the user to have `StrBoxed(Option<Str>)`  instead

sdroege:
Most boxed/object properties are nullable, and you can derive boxed/shared as nullable too.

ranfdev:
`HasParamSpec` is implemented on every type `Option<T>` where `T: glib::value::ToValueOptional`.

[...]

PR #940, opened 2023-01-31
Add doc for ConstructCell, improve doc Boxed, Enum

ranfdev:
Related to the issue #932.

I'm not sure if types created with `glib::Enum` can be `nullable`...

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -530,7 +530,8 @@ pub fn error_domain_derive(input: TokenStream) -> TokenStream {
 }
 
 /// Derive macro for defining a [`BoxedType`]`::type_` function and
-/// the [`glib::Value`] traits.
+/// the [`glib::Value`] traits. Eventually, the type can be set as `nullable` to
+/// get an implemention of `glib::value::ToValueOptional`.
 ///
 /// # Example
 ///
@@ -539,7 +540,7 @@ pub fn error_domain_derive(input: TokenStream) -> TokenStream {
 /// use glib::subclass::prelude::*;
 ///
 /// #[derive(Clone, Debug, PartialEq, Eq, glib::Boxed)]
-/// #[boxed_type(name = "MyBoxed")]
+/// #[boxed_type(name = "MyBoxed", nullable)]
 /// struct MyBoxed(String);
 /// ```
 ///


sdroege:
> I'm not sure if types created with `glib::Enum` can be `nullable`...

They can't, neither can flags :)

glib-macros/src/lib.rs
```diff
@@ -445,7 +446,7 @@ pub fn closure_local(item: TokenStream) -> TokenStream {
 /// use glib::subclass::prelude::*;
 ///
 /// #[derive(Debug, Copy, Clone, PartialEq, Eq, glib::Enum)]
-/// #[enum_type(name = "MyEnum")]
+/// #[enum_type(name = "MyEnum", nullable)]
```
sdroege, line 449: I'd expect this to fail. It should :)

glib/src/construct_cell.rs
```diff
@@ -12,8 +12,15 @@ use std::sync::RwLock;
 macro_rules! define_construct {
     (@common $ident:ident, $container:ident, $inner:ty) => {
         // rustdoc-stripper-ignore-next
-        /// Wrapper around a container type to be used for custom glib properties which get set during
-        /// the construct phase. This is especially useful when used alongside the Properties macro.
+        /// Experimental.
+        /// This type is mostly useful when used alongside the `glib::Properties` macro, when you want to create a property
+        ///  of type `Option<T>` and
+        /// - `T` doesn't implement `glib::value::ToValueOptional`.
```
sdroege, line 18: I wouldn't support that. If `T` doesn't implement that then `T` doesn't support being `NULL`

glib-macros/src/lib.rs
ranfdev: Except it doesn't fail. The doc check failed for other network related reasons.

sdroege: Yeah it doesn't check if there are random other, unknown attributes apparently.

sdroege: https://github.com/gtk-rs/gtk-rs-core/issues/943

sdroege: So just remove the wrong attribute here for now :)

sdroege (approved):

PR #947, opened 2023-02-03
glib: Implement Regex

davidmhewitt:
I'm interested in generating bindings for https://github.com/elementary/granite which exposes GLib.Regex as part of its public API, so it would be good to have these in the `glib` crate to depend on.


glib/src/regex.rs
```diff
@@ -0,0 +1,172 @@
...
+        start_position: i32,
+        match_options: RegexMatchFlags,
+        max_tokens: i32,
+    ) -> Result<Vec<crate::GString>, crate::Error> {
+        let string_len = string.len() as _;
+        unsafe {
+            let mut error = ptr::null_mut();
+            let ret = ffi::g_regex_split_full(
+                self.to_glib_none().0,
+                string.to_glib_none().0,
```
sdroege, line 134: You probably want to implement all these like

```rust
string.as_ptr() as *const _,
string.len() as _,
```

otherwise the `to_glib_none()` will always make a copy just to add the `NUL`-terminator, which is not even needed here because you explicitly provide the length anyway.

glib/src/auto/regex.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    #[doc(alias = "g_regex_match")]
+    #[doc(alias = "match")]
+    pub fn match_(&self, string: &str, match_options: RegexMatchFlags) -> Option<MatchInfo> {
+        unsafe {
+            let mut match_info = ptr::null_mut();
+            let ret = from_glib(ffi::g_regex_match(
+                self.to_glib_none().0,
+                string.to_glib_none().0,
```
sdroege, line 97: And this and others you probably want to implement in terms of `match_full()` or so manually so that the strings don't have to be copied.

glib/src/auto/regex.rs
```diff
@@ -0,0 +1,207 @@
...
+    #[doc(alias = "g_regex_get_pattern")]
+    #[doc(alias = "get_pattern")]
+    pub fn pattern(&self) -> crate::GString {
+        unsafe { from_glib_none(ffi::g_regex_get_pattern(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "g_regex_get_string_number")]
+    #[doc(alias = "get_string_number")]
+    pub fn string_number(&self, name: &str) -> i32 {
+        unsafe { ffi::g_regex_get_string_number(self.to_glib_none().0, name.to_glib_none().0) }
```
sdroege, line 87: And here and elsewhere you probably want to use `impl IntoGStr` for the very same reason :)

glib/src/auto/match_info.rs
```diff
@@ -0,0 +1,141 @@
...
+        unsafe { from_glib(ffi::g_match_info_is_partial_match(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "g_match_info_matches")]
+    pub fn matches(&self) -> bool {
+        unsafe { from_glib(ffi::g_match_info_matches(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "g_match_info_next")]
+    pub fn next(&self) -> Result<(), crate::Error> {
```
sdroege, line 129: Should there maybe be an `Iterator` impl that works over this?

glib/src/auto/match_info.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    #[doc(alias = "g_match_info_fetch_all")]
+    pub fn fetch_all(&self) -> Vec<crate::GString> {
+        unsafe {
+            FromGlibPtrContainer::from_glib_full(ffi::g_match_info_fetch_all(self.to_glib_none().0))
+        }
+    }
+
+    #[doc(alias = "g_match_info_fetch_named")]
+    pub fn fetch_named(&self, name: &str) -> Option<crate::GString> {
```
sdroege, line 53: Also here `impl IntoGStr` I guess (also elsewhere in this file)

sdroege (commented):

jf2048:
Not a blocker, but it would be good if eventually we could have a documentation note here saying this should not really be used by Rust programs except for compatibility reasons. No one should ever be writing new code with this unless necessary, e.g. calling those libgranite APIs that need it. There are much better native Rust implementations, even the pcre2 crate is better than using this.

glib/src/auto/regex.rs
davidmhewitt: Is there a way to do this and still have the bindings automatically generated, maybe some way of manually overriding the type in the `Gir.toml`?

Otherwise, I'd be happy to manually implement these.

glib/src/auto/match_info.rs
davidmhewitt: Looking at this in a bit more detail, the `g_match_info_next` method doesn't return a new `MatchInfo`, it modifies the existing one in place, so it may be quite difficult to implement the `Iterator` trait over the top of this? I assume we'd need some kind of copy mechanics on the underlying struct?

glib/src/auto/regex.rs
sdroege: In a couple of weeks once I finished my rewrite of the code generator that would automatically happen. until then it would have to be manual

glib/src/auto/match_info.rs
sdroege: Let's not worry about that then for now :)

bilelmoussaoui:
Would you mind rebasing this one?

glib/Gir.toml
```diff
@@ -714,6 +716,65 @@ status = "generate"
     name = "get_user_data"
     ignore = true # unsafe pointer
 
+[[object]]
+name = "GLib.MatchInfo"
+status = "generate"
+    [[object.function]]
```
bilelmoussaoui, line 722: It would be nice if possible to document the reason why we have to manually implement those functions; so we can potentially re-visit them in the future if improvements are made to the generator.

glib/src/auto/regex.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    #[doc(alias = "g_regex_get_pattern")]
+    #[doc(alias = "get_pattern")]
+    pub fn pattern(&self) -> crate::GString {
+        unsafe { from_glib_none(ffi::g_regex_get_pattern(self.to_glib_none().0)) }
+    }
+
+    //#[doc(alias = "g_regex_replace_eval")]
+    //pub fn replace_eval(&self, string: &[&str], start_position: i32, match_options: RegexMatchFlags, eval: /*Unimplemented*/Fn(&MatchInfo, /*Ignored*/String) -> bool, user_data: /*Unimplemented*/Option<Basic: Pointer>) -> Result<crate::GString, crate::Error> {
```
bilelmoussaoui, line 84: Would you be interested in adding bindings for this one? seems to be the only missing part for complete API coverage. Not a blocker though.

bilelmoussaoui (commented):
otherwise lgtm

glib/src/auto/match_info.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+impl MatchInfo {
+    #[doc(alias = "g_match_info_fetch")]
+    pub fn fetch(&self, match_num: i32) -> Option<crate::GString> {
+        unsafe { from_glib_full(ffi::g_match_info_fetch(self.to_glib_none().0, match_num)) }
+    }
+
+    #[doc(alias = "g_match_info_fetch_all")]
+    pub fn fetch_all(&self) -> Vec<crate::GString> {
```
sdroege, line 25:
```suggestion
    pub fn fetch_all(&self) -> glib::PtrSlice<crate::GString> {
```

Would avoid one unnecessary copy (into the `Vec`) here

glib/src/regex.rs
```diff
@@ -0,0 +1,272 @@
...
+        unsafe {
+            from_glib_full(ffi::g_regex_escape_nul(
+                string.as_ptr() as *const _,
+                string.len() as _,
+            ))
+        }
+    }
+
+    #[doc(alias = "g_regex_escape_string")]
+    pub fn escape_string(string: &str) -> crate::GString {
```
sdroege, line 30: Why not `impl IntoGStr` here too?

sdroege: And in some other places below where `&str` is used

glib/src/regex.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    #[doc(alias = "g_regex_split_full")]
+    pub fn split_full(
+        &self,
+        string: &str,
+        start_position: i32,
+        match_options: RegexMatchFlags,
+        max_tokens: i32,
+    ) -> Result<Vec<crate::GString>, crate::Error> {
```
sdroege, line 210:
```suggestion
    ) -> Result<glib::PtrSlice<crate::GString>, crate::Error> {
```

sdroege: (and elsewhere in this file)

sdroege:
Looks good to me apart from those trivial things. Once that's all covered, can you squash the commits together?

glib/src/lib.rs
```diff
@@ -174,6 +174,10 @@ mod property;
 pub use self::property::*;
 mod quark;
 pub use self::quark::Quark;
+pub mod match_info;
+pub use self::match_info::*;
```
bilelmoussaoui, line 178:
```suggestion
```

glib/src/lib.rs
```diff
@@ -174,6 +174,10 @@ mod property;
...
+pub mod regex;
+pub use self::regex::*;
```
bilelmoussaoui, line 180:
```suggestion
```
the module doesn't provide any new types itself

davidmhewitt:
Thank you both for the feedback, I'll work on addressing the last few comments over the next few hours.

But first, a question:

It looks like I initially manually implemented a lot of these methods so we could use `impl IntoGStr`  in the arguments instead of `&str`. The methods generated by `gir` are otherwise identical.

Is it still the case that it would be preferred to have manually written bindings just to get `impl IntoGStr` here, or should I revert most of these methods back to the automatically generated versions that use `&str`?

sdroege:
Keep them like this until gir supports generating them like this. That's what the comment in `Gir.toml` should say why they're not autogenerated.

glib/src/auto/regex.rs
davidmhewitt: I've just attempted to implement this and the binding is very fiddly due to the callback function needing to append to an existing `GString` in place, so I'm going to skip the implementation of this function for now.

davidmhewitt:
I have resolved all feedback apart from the non-blocking comment about the missing implementation of `g_regex_replace_eval` and squashed my commits.

Thanks again for your reviews, let me know if there's anything else I can do here.

bilelmoussaoui (approved):

bilelmoussaoui:
Looks good enough, we can always iterate later if needed. Thank you for your contribution

sdroege:
This has some soundness issues, see https://github.com/gtk-rs/gtk-rs-core/issues/1223

PR #950, opened 2023-02-04
Rename GStrPtr to GStringPtr

pbor:
Since this is an owned type this is more consistent.

### DIFF
--- a/glib/src/collections/strv.rs
+++ b/glib/src/collections/strv.rs
@@ -2,7 +2,7 @@
 
 use std::{ffi::c_char, fmt, marker::PhantomData, mem, ptr};
 
-use crate::{translate::*, GStr, GStrPtr, GString};
+use crate::{translate::*, GStr, GString, GStringPtr};
 
 // rustdoc-stripper-ignore-next
 /// Minimum size of the `StrV` allocation.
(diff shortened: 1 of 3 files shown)


jf2048:
If you want there could also be a `to_str` that just calls `to_gstr().as_str()`

pbor:
> If you want there could also be a `to_str` that just calls `to_gstr().as_str()`

yeah, I opted for the minimal public API, but I can add it if that is preferred

jf2048:
It's up to you, either way is fine :)

glib/src/collections/strv.rs
```diff
@@ -201,12 +201,12 @@ impl IntoIter {
     // rustdoc-stripper-ignore-next
     /// Returns the remaining items as slice.
     #[inline]
-    pub fn as_slice(&self) -> &[GStrPtr] {
+    pub fn as_slice(&self) -> &[GStringPtr] {
```
jf2048, line 204:
```suggestion
    pub const fn as_slice(&self) -> &[GStringPtr] {
```

glib/src/collections/strv.rs
```diff
@@ -675,14 +678,14 @@ impl StrV {
     }
 
     // rustdoc-stripper-ignore-next
-    /// Borrows this slice as a `&[GStrPtr]`.
+    /// Borrows this slice as a `&[GStringPtr]`.
     #[inline]
-    pub fn as_slice(&self) -> &[GStrPtr] {
+    pub fn as_slice(&self) -> &[GStringPtr] {
```
jf2048, line 683:
```suggestion
    pub const fn as_slice(&self) -> &[GStringPtr] {
```

jf2048 (changes requested):

sdroege:
> yeah, I opted for the minimal public API, but I can add it if that is preferred

Well, doesn't cost much so why not add it? :)

glib/src/collections/strv.rs
```diff
@@ -443,11 +446,11 @@ impl StrV {
         if len == 0 {
             StrV::default()
         } else {
-            // Need to fully copy the array here.
-            let slice = Self::from_glib_borrow_num(ptr, len);
-            let mut s = Self::new();
-            s.extend_from_slice(slice);
-            s
+            StrV {
+                ptr: ptr::NonNull::new_unchecked(ffi::g_strdupv(ptr as *mut _)),
```
sdroege, line 450: The reason for the previous implementation was that we know the length while `g_strdupv()` first needs to determine the length before copying

sdroege (commented):
Looks good otherwise, thanks :)

glib/src/collections/strv.rs
pbor: Yes, but `extend_from_slice` requires `AsRef` so it cannot be used anymore, I guess I can re-implement it

sdroege: Ah I see. Well, worst case start with `g_memdup()` :)

pbor: I need to copy all the strings as well, so memdup does not sound the right thing... I simply allocated and then copied all the strings

sdroege (approved):

sdroege:
Also fixed gtk4-rs for the rename.

Issue #951, opened 2023-02-04
[FEATURE REQUEST] impl FromIterator for ListStore


ranfdev:
We already have `IntoIterator`, so why not add `FromIterator`?

sdroege:
Yes why not? :)

PR #954, opened 2023-02-05
gio: implement FromIterator for ListStore

pbor:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/951

### DIFF
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -107,6 +107,16 @@ impl ListStore {
     }
 }
 
+impl<P: IsA<glib::Object>> std::iter::FromIterator<P> for ListStore {
+    fn from_iter<I: IntoIterator<Item = P>>(iter: I) -> Self {
+        let store = Self::new(P::static_type());
+        for item in iter.into_iter() {
+            store.append(&item)
+        }
+        store
+    }
+}
+
 impl<'a> std::iter::IntoIterator for &'a ListStore {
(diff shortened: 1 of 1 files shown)


pbor:
Mind you, I have no idea what I am doing :)

gio/src/list_store.rs
```diff
@@ -107,6 +107,16 @@ impl ListStore {
     }
 }
 
+impl<P: IsA<glib::Object>> std::iter::FromIterator<P> for ListStore {
```
sdroege, line 110: Could also implement it in addition for `&P` but looks good otherwise!

sdroege: Apparently not possible, causes conflicting impls

pbor: That does not seem to work: 

```
110 | impl<P: IsA<glib::Object>> std::iter::FromIterator<P> for ListStore {
    | ------------------------------------------------------------------- first implementation here
...
120 | impl<P: IsA<glib::Object>> std::iter::FromIterator<&P> for ListStore {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `auto::list_store::ListStore`
    |
    = note: downstream crates may implement trait `glib::IsA<glib::Object>` for type `&_`
```

sdroege (approved):

PR #955, opened 2023-02-05
Use strcmp for GStringPtr comparisons

jf2048:
Benchmarks show this to be a bit faster because `to_str` causes the string to be iterated twice.

Also add `GStr::from_ptr_checked` to ease with doing a utf8 check from a C pointer.

### DIFF
--- a/glib/src/gstring.rs
+++ b/glib/src/gstring.rs
@@ -684,12 +684,37 @@ impl GStringPtr {
     pub fn to_str(&self) -> &str {
         self.to_gstr().as_str()
     }
+
+    // rustdoc-stripper-ignore-next
+    /// Returns the string's C pointer.
+    #[inline]
+    pub const fn as_ptr(&self) -> *const c_char {
+        self.0.as_ptr()
+    }
+
+    // rustdoc-stripper-ignore-next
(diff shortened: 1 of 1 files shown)


jf2048:
cc @pbor

pbor:
Thanks for cleaning up after me :)

(looks good to me)

glib/src/gstring.rs
```diff
@@ -684,6 +697,28 @@ impl GStringPtr {
...
+    ///
+    /// # Safety
+    ///
+    /// `a` and `b` must be non-null pointers to nul-terminated C strings.
+    #[inline]
+    unsafe fn strcmp(a: *const c_char, b: *const c_char) -> Ordering {
+        match libc::strcmp(a, b) {
+            0 => Ordering::Equal,
+            x if x < 0 => Ordering::Less,
+            _ => Ordering::Greater,
```
sdroege, line 719: https://github.com/gtk-rs/gtk-rs-core/blob/0d38c9a72540336665663bea73243a30f2592e68/glib/src/translate.rs#L1424-L1429 btw :) I'm surprised clippy didn't complain here that you should use `cmp()`

sdroege (approved):

Issue #957, opened 2023-02-05
`properties` macro should special case `default = `


pbor:
Right now we are passing down `foo = bar` attributes as `builder.foo(bar)`, however it would be more ergonomic to have `default = 42` instead of `default_value = 42`. The builder use `default_value` as a function name just because default conflicts with `Default::default`

PR #958, opened 2023-02-05
Support `default = ...` in the properties macro

pbor:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/957

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -114,6 +114,11 @@ impl Parse for PropAttr {
                 "set" => PropAttr::Set(Some(input.parse()?)),
                 "type" => PropAttr::Type(input.parse()?),
                 "member" => PropAttr::Member(input.parse()?),
+                // Special case "default = ..." and map it to .default_value(...)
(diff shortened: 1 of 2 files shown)


pbor:
This does not work yet because the test does not pass, but I am confused on the reason why: the macro generates the corrected thing:

```
                    <<RefCell<
                        String,
                    > as glib::Property>::Value as glib::HasParamSpec>::param_spec_builder()(
                            "with-default",
                        )
                        .readwrite()
                        .default_value("hello")
                        .build(),
```

However getting the property still returns `""`... is there a bug somewhere else that ignores the default value?

pbor:
As confusing as it is, I think that was expected, since the prop default is just what is set in the glib type system, but it is not automatically used when initializing an instance. I changed the unit test.

cc @ranfdev

sdroege:
CC @ranfdev

ranfdev:
Looks good to me

sdroege (approved):

sdroege (approved):

PR #963, opened 2023-02-06
Greatly improve `clone!` proc macro error output

### LINKED ISSUE #956: [BUG] `clone!` loses closure arguments span information, leading to poor diagnostics
**Bug description**

If you provide a closure to the `clone!` macro that takes the wrong number of arguments, you'll get a diagnostic like this:

```
error[E0593]: closure is expected to take 4 arguments, but it takes 0 arguments
   --> src/places_sidebar.rs:301:25
    |
301 |               drop_target.connect_drop(clone!(@weak item => move || {
    |  _________________________^^^^^^^^^^^^_-
    | |                         |
    | |                         expected closure that takes 4 arguments
302 | |             }));
    | |______________- takes 0 arguments
    |
help: consider changing the closure to take and ignore the expected arguments
    |
301 |             drop_target.connect_drop(|_, _, _, _|);
    | 
```

This is because rustc is unable to determine the span of the closure arguments, so it falls back to the span of the entire closure. I believe we should be able to thread the correct span back to rustc if we emit the expanded closure arguments with the same span as the original closure. That would lead to a diagnostic like this:

```
error[E0593]: closure is expected to take 4 arguments, but it takes 0 arguments
   --> src/places_sidebar.rs:301:25
    |
301 |               drop_target.connect_drop(clone!(@weak item => move || {
    |                           ^^^^^^^^^^^^                           -- takes 0 arguments
    |                           |
    |                           expected closure that takes 4 arguments
302 |               }));
    |
    |
help: consider changing the closure to take and ignore the expected arguments
    |
301 |             drop_target.connect_drop(clone!(@weak item => move |_, _, _, _| {;
    | 
```

**Backtrace**

N/A

GuillaumeGomez:
Fixes #956.

What is it this PR doing? Instead of using `panic!` whenever we encounter an invalid input, it generates a `compile_error!` call with the span of the token which created the issue to trick the compiler into given the location we want.

Unfortunately, the spans are limited with stable rustc so I can't use `Span::join` to make it look better in some cases, but I think it remains a great improvement nonetheless.

### DIFF
--- a/glib-macros/src/clone.rs
+++ b/glib-macros/src/clone.rs
@@ -3,11 +3,95 @@
 use std::iter::Peekable;
 
 use proc_macro::{
-    token_stream::IntoIter as ProcIter, Delimiter, Group, Ident, Punct, Spacing, Span, TokenStream,
-    TokenTree,
+    token_stream::IntoIter as ProcIter, Delimiter, Group, Ident, Literal, Punct, Spacing, Span,
+    TokenStream, TokenTree,
 };
 
 use crate::utils::crate_ident_new;
+
+struct PeekableProcIter {
+    inner: Peekable<ProcIter>,
+    current_span: Option<Span>,
+    next_span: Option<Span>,
+}
+
+impl<'a> From<&'a Group> for PeekableProcIter {
+    fn from(f: &'a Group) -> Self {
+        let current_span = Some(f.span());
+        let mut inner = f.stream().into_iter().peekable();
+        let next_span = inner.peek().map(|n| n.span());
+        Self {
+            inner,
+            current_span,
+            next_span,
+        }
+    }
+}
+
+impl From<TokenStream> for PeekableProcIter {
+    fn from(f: TokenStream) -> Self {
+        let mut inner = f.into_iter().peekable();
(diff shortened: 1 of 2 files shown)


glib-macros/src/clone.rs
```diff
@@ -3,12 +3,100 @@
...
+            stream.extend(tokens);
+            let mut t = TokenTree::Group(Group::new(Delimiter::Brace, stream));
+            t.set_span(span);
+
+            let mut stream = TokenStream::new();
+            stream.extend(vec![t]);
+            stream
+        } else {
+            // for some reason there is no span yet...
+            panic!("{message}");
```
jf2048, line 95: Could it fall back to `Span::call_site` instead of panicking? That should be equivalent but without the weird "proc macro panicked" messages.

jf2048 (changes requested):

sdroege (commented):
Looks good to me otherwise

glib-macros/src/clone.rs
GuillaumeGomez: Ah sure.

GuillaumeGomez:
Updated!

sdroege (approved):

PR #968, opened 2023-02-07
ThreadGuard: Implement ToValue, Fromvalue and Default traits

alatiera:
This will also allow ThreadGuard to be used with the new property macro.

### DIFF
--- a/glib-macros/tests/properties.rs
+++ b/glib-macros/tests/properties.rs
@@ -87,6 +87,7 @@ mod foo {
     }
 
     pub mod imp {
+        use glib::thread_guard::ThreadGuard;
         use glib::{ParamSpec, Value};
         use std::rc::Rc;
 
@@ -145,6 +146,8 @@ mod foo {
             cell: Cell<u8>,
             #[property(get = Self::overridden, override_class = Base)]
             overridden: PhantomData<u32>,
+            #[property(get, set)]
+            thread_guard: ThreadGuard<Mutex<String>>,
         }
 
         impl ObjectImpl for Foo {
@@ -194,6 +197,12 @@ fn props() {
     // Read bar
     let bar: String = myfoo.property("bar");
     assert_eq!(bar, "".to_string());
+
+    // Set the thread guard
+    myfoo.set_property("thread-guard", "foobar".to_value());
+    // And grab directly the string from the guard after
+    let bar: String = myfoo.property("thread-guard");
+    assert_eq!(bar, "foobar".to_string());
 
     // Set bar
     myfoo.set_property("bar", "epic".to_value());
--- a/glib/src/property.rs
+++ b/glib/src/property.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::RwLock;
 
+use crate::thread_guard::ThreadGuard;
 use crate::HasParamSpec;
 
 // rustdoc-stripper-ignore-next
@@ -35,6 +36,9 @@ impl<T: Property> Property for Mutex<T> {
     type Value = T::Value;
 }
 impl<T: Property> Property for RwLock<T> {
+    type Value = T::Value;
+}
+impl<T: Property> Property for ThreadGuard<T> {
     type Value = T::Value;
 }
 impl<T: Property> Property for once_cell::sync::OnceCell<T> {
@@ -128,6 +132,19 @@ impl<T> PropertySetNested for RwLock<T> {
     type SetNestedValue = T;
     fn set_nested<F: FnOnce(&mut Self::SetNestedValue)>(&self, f: F) {
         f(&mut self.write().unwrap());
+    }
+}
+
+impl<T: PropertyGet> PropertyGet for ThreadGuard<T> {
+    type Value = T::Value;
+    fn get<R, F: Fn(&Self::Value) -> R>(&self, f: F) -> R {
+        self.get_ref().get(f)
+    }
+}
+impl<T: PropertySetNested> PropertySetNested for ThreadGuard<T> {
+    type SetNestedValue = T::SetNestedValue;
+    fn set_nested<F: FnOnce(&mut Self::SetNestedValue)>(&self, f: F) {
+        self.get_ref().set_nested(f)
     }
 }
 
--- a/glib/src/thread_guard.rs
+++ b/glib/src/thread_guard.rs
@@ -4,6 +4,7 @@ use std::{
     mem, ptr,
     sync::atomic::{AtomicUsize, Ordering},
 };
+
 fn next_thread_id() -> usize {
     static COUNTER: AtomicUsize = AtomicUsize::new(0);
     COUNTER.fetch_add(1, Ordering::SeqCst)
@@ -108,6 +109,12 @@ impl<T> Drop for ThreadGuard<T> {
             self.thread_id == thread_id(),
             "Value dropped on a different thread than where it was created"
         );
+    }
+}
+
+impl<T: Default> Default for ThreadGuard<T> {
+    fn default() -> Self {
+        Self::new(T::default())
     }
 }
 


glib/src/thread_guard.rs
```diff
@@ -111,4 +112,40 @@ impl<T> Drop for ThreadGuard<T> {
...
+    }
+}
+
+unsafe impl<'a, T, C, E> crate::value::FromValue<'a> for ThreadGuard<T>
+where
+    T: crate::value::FromValue<'a, Checker = C> + crate::StaticType,
+    C: crate::value::ValueTypeChecker<Error = crate::value::ValueTypeMismatchOrNoneError<E>>,
+    E: std::error::Error + Send + Sized + 'static,
+{
+    type Checker = crate::value::ValueTypeOrNoneChecker<T, C, E>;
```
alatiera, line 139: No idea what checker to use

glib/src/thread_guard.rs
```diff
@@ -111,4 +112,40 @@ impl<T> Drop for ThreadGuard<T> {
...
+
+    unsafe fn from_value(value: &'a crate::Value) -> Self {
+        match T::Checker::check(value) {
+            Err(crate::value::ValueTypeMismatchOrNoneError::UnexpectedNone) => panic!(),
```
alatiera, line 143: No idea if there's any point in checking the value since we can only panic really

glib/src/thread_guard.rs
```diff
@@ -111,4 +112,40 @@ impl<T> Drop for ThreadGuard<T> {
...
+    fn to_value(&self) -> crate::Value {
+        T::to_value(self.get_ref())
+    }
+
+    fn value_type(&self) -> crate::Type {
+        T::static_type()
+    }
+}
+
+unsafe impl<'a, T, C, E> crate::value::FromValue<'a> for ThreadGuard<T>
```
alatiera, line 133: should probably import all the `crate::value::` types

glib/src/thread_guard.rs
sdroege: `T::Checker`

glib/src/param_spec.rs
```diff
@@ -2065,6 +2065,17 @@ pub trait HasParamSpec {
     fn param_spec_builder() -> Self::BuilderFn;
 }
 
+impl<T: crate::value::ToValueOptional + HasParamSpec> HasParamSpec
```
alatiera, line 2068: I think this will be obsolete by https://github.com/gtk-rs/gtk-rs-core/pull/967 ?

sdroege:
CC @ranfdev

glib/src/thread_guard.rs
sdroege: You need to check that the type matches and handle it accordingly. Just forward the checker result

glib/src/param_spec.rs
sdroege:
```suggestion
impl<T: HasParamSpec> HasParamSpec
```

glib/src/thread_guard.rs
sdroege: :shrug:

glib/src/thread_guard.rs
```diff
@@ -111,4 +115,40 @@ impl<T> Drop for ThreadGuard<T> {
...
 
+impl<T: Default> Default for ThreadGuard<T> {
+    fn default() -> Self {
+        Self::new(T::default())
+    }
+}
+
 unsafe impl<T> Send for ThreadGuard<T> {}
+
+impl<T: ToValue + StaticType> ToValue for ThreadGuard<T> {
```
sdroege, line 126: You also want `impl<T: From<Value>> From<ThreadGuard<T>> for Value` or so

glib/src/thread_guard.rs
alatiera: We are returning Self, so I think we can only unwrap it

glib/src/thread_guard.rs
```diff
@@ -111,4 +115,40 @@ impl<T> Drop for ThreadGuard<T> {
...
+unsafe impl<'a, T, C, E> FromValue<'a> for ThreadGuard<T>
+where
+    T: FromValue<'a, Checker = C> + StaticType,
+    C: ValueTypeChecker<Error = ValueTypeMismatchOrNoneError<E>>,
+    E: std::error::Error + Send + Sized + 'static,
+{
+    type Checker = T::Checker;
+
+    unsafe fn from_value(value: &'a Value) -> Self {
+        match T::Checker::check(value) {
```
sdroege, line 145: You don't have to call it at all here. Just get the value

glib/src/thread_guard.rs
alatiera: done, hopefully right on the first try

sdroege: No

glib/src/thread_guard.rs
```diff
@@ -111,4 +115,39 @@ impl<T> Drop for ThreadGuard<T> {
...
+    }
+
+    fn value_type(&self) -> Type {
+        T::static_type()
+    }
+}
+
+impl<T: From<Value> + ToValue> From<ThreadGuard<T>> for Value {
+    fn from(guard: ThreadGuard<T>) -> Self {
+        Self::from(guard.get_ref())
```
sdroege, line 138:
```suggestion
        Self::from(guard.into_inner())
```

glib/src/thread_guard.rs
```diff
@@ -111,4 +115,39 @@ impl<T> Drop for ThreadGuard<T> {
...
+    fn to_value(&self) -> Value {
+        T::to_value(self.get_ref())
+    }
+
+    fn value_type(&self) -> Type {
+        T::static_type()
+    }
+}
+
+impl<T: From<Value> + ToValue> From<ThreadGuard<T>> for Value {
```
sdroege, line 136:
```suggestion
impl<T: From<Value>> From<ThreadGuard<T>> for Value {
```

glib/src/thread_guard.rs
alatiera: The compiler says that it wants something like this, but its too late at night for me to understand how we are going from `T: From<Value> where Value: From<T>`

```
diff --git a/glib/src/thread_guard.rs b/glib/src/thread_guard.rs
index 1dac4d84686..333789cd724 100644
--- a/glib/src/thread_guard.rs
+++ b/glib/src/thread_guard.rs
@@ -133,9 +133,12 @@ impl<T: ToValue + StaticType> ToValue for ThreadGuard<T> {
     }
 }
 
-impl<T: From<Value> + ToValue> From<ThreadGuard<T>> for Value {
+impl<T: From<Value>> From<ThreadGuard<T>> for Value
+where
+    Value: From<T>,
+{
     fn from(guard: ThreadGuard<T>) -> Self {
-        Self::from(guard.get_ref())
+        Self::from(guard.into_inner())
     }
 }
 ```

File: glib/src/param_spec.rs [hunk 9]
```diff
@@ -2065,6 +2065,15 @@ pub trait HasParamSpec {
     fn param_spec_builder() -> Self::BuilderFn;
 }
 
+impl<T: HasParamSpec> HasParamSpec for crate::thread_guard::ThreadGuard<T> {
```
ranfdev, line 2068: Actually, unless the type has a "direct" mapping to a ParamSpec, the type shouldn't implement `HasParamSpec` but the `Property` trait.

`HasParamSpec` creates a direct mapping to a `ParamSpec` and adds some metadata to the type system (eg: the associated `ParamSpecXBuilder` for the specific type).

 `Property` is made for "wrapper" types, like `Cell` or `Mutex` and handles recursive types.
 Thanks to the `Property` trait we can support types like `Rc<RefCell<String>>` or even (I think)
 `Rc<Arc<ThreadGuard<String>>>`

glib/src/thread_guard.rs
```diff
@@ -111,4 +115,42 @@ impl<T> Drop for ThreadGuard<T> {
...
+    fn default() -> Self {
+        Self::new(T::default())
+    }
+}
+
 unsafe impl<T> Send for ThreadGuard<T> {}
+
+impl<T: ToValue + StaticType> ToValue for ThreadGuard<T> {
+    fn to_value(&self) -> Value {
+        T::to_value(self.get_ref())
```
alatiera, line 128: This is basically taking `ThreadGuard<T>` and converting it to `Value<T>` effectively getting rid of the guard. This can be problematic and shouldn't be done implicitly.

glib/src/property.rs
```diff
@@ -37,6 +38,9 @@ impl<T: Property> Property for Mutex<T> {
 impl<T: Property> Property for RwLock<T> {
     type Value = T::Value;
 }
+impl<T: Property> Property for ThreadGuard<T> {
+    type Value = T::Value;
```
alatiera, line 42: Value should probably be Self here, probably a Boxed object?

glib/src/thread_guard.rs
ranfdev: That's bad, from what I understand this will _always_ dump the threadguard.

So the returned ThreadGuard we get when calling `obj.property("thread-guard")` is actually a  guard created on the fly, not the one we put in.

glib/src/property.rs
ranfdev: No, because `Property::Value` is used to retrieve the innermost type, the one with `HasParamSpec` trait.

We can't put `type Value = Self` because `Self` doesn't implement `HasParamSpec`

glib/src/param_spec.rs
sdroege: If we can go without all the value trait impls on the `ThreadGuard` that would indeed be preferable. Can `Property` work with types that don't offer interior mutability though? I guess so because that's how `Rc` works?

sdroege: Yeah that part works but it only allows

```rust
            #[property(name = "thread-guard", get, set)]
            thread_guard: ThreadGuard<Mutex<String>>,
```

It does not allow the actually sensible one

```rust
            #[property(name = "thread-guard", get, set)]
            thread_guard: Mutex<ThreadGuard<String>>,
```

```
error[E0277]: the trait bound `ThreadGuard<std::string::String>: glib::ToValue` is not satisfied
error[E0277]: the trait bound `ThreadGuard<std::string::String>: FromValue<'_>` is not satisfied
```

or

```rust
            #[property(name = "thread-guard", get, set)]
            thread_guard: Mutex<ThreadGuard<Arc<String>>>,
```

```
error[E0277]: the trait bound `ThreadGuard<Arc<std::string::String>>: glib::ToValue` is not satisfied
error[E0277]: the trait bound `ThreadGuard<Arc<std::string::String>>: FromValue<'_>` is not satisfied
```

or

```rust
            #[property(name = "thread-guard", get, set)]
            thread_guard: Mutex<ThreadGuard<Option<std::sync::Arc<String>>>>,
```

```
error[E0277]: the trait bound `std::sync::Mutex<ThreadGuard<Option<Arc<std::string::String>>>>: Property` is not satisfied
```

The last one is actually approximately what @alatiera wants to use: `Mutex<Option<ThreadGuard<gdk::Paintable>>>`.

 I think the same problem also exists in other type nestings with `Option` and `Rc`.

sdroege: FTR

```diff
diff --git a/glib/src/property.rs b/glib/src/property.rs
index 2183b9b4c8..75783c1deb 100644
--- a/glib/src/property.rs
+++ b/glib/src/property.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::RwLock;
 
+use crate::thread_guard::ThreadGuard;
 use crate::HasParamSpec;
 
 // rustdoc-stripper-ignore-next
@@ -50,6 +51,9 @@ impl<T: Property> Property for Rc<T> {
 impl<T: Property> Property for Arc<T> {
     type Value = T::Value;
 }
+impl<T: Property> Property for ThreadGuard<T> {
+    type Value = T::Value;
+}
 
 // rustdoc-stripper-ignore-next
 /// A container type implementing this trait can be read by the default getter generated by the `Props` macro.
@@ -196,6 +200,19 @@ impl<T: PropertySetNested> PropertySetNested for Arc<T> {
     }
 }
 
+impl<T: PropertyGet> PropertyGet for ThreadGuard<T> {
+    type Value = T::Value;
+    fn get<R, F: Fn(&Self::Value) -> R>(&self, f: F) -> R {
+        self.get_ref().get(f)
+    }
+}
+impl<T: PropertySetNested> PropertySetNested for ThreadGuard<T> {
+    type SetNestedValue = T::SetNestedValue;
+    fn set_nested<F: FnOnce(&mut Self::SetNestedValue)>(&self, f: F) {
+        self.get_ref().set_nested(f)
+    }
+}
+
 macro_rules! impl_atomic {
     ($atomic:ty, $valuety:ty) => {
         impl Property for $atomic {
```

ranfdev: That thought doesn't return a `ThreadGuard` from the getter. It returns the type contained in the `ThreadGuard`, doesn't it?

Yesterday we wanted to return the entire `ThreadGuard`:

`let t: ThreadGuard<_> = obj.prop("thread-guard").get()`.

sdroege: Yes, I don't think that's sensible. You can't store a `ThreadGuard` in a `Value`, so that wrapping will necessarily go away. Nothing can know about it at the `Value` level.

I think the properties macro should consider the `ThreadGuard` just as a storage container.

glib-macros/tests/properties.rs
```diff
@@ -145,6 +146,8 @@ mod foo {
             cell: Cell<u8>,
             #[property(get = Self::overridden, override_class = Base)]
             overridden: PhantomData<u32>,
+            #[property(name = "thread-guard", get, set)]
```
ranfdev, line 149: No need to manually set the name. The macro generates it for you. The underscore _ becomes a dash -.

sdroege:
I thought we decided that this is not actually useful and correct

ranfdev:
I don't know, I thought we would still merge what's possible.

This PR still works, even though the use cases it covers are limited.

ranfdev:
It is correct, just it doesn't cover every use case we have thought of.

alatiera:
I am fine either way, though not sure that would want get prop to not return a threadguard

sdroege:
Let's close this then in favour of https://github.com/gtk-rs/gtk-rs-core/pull/984

Issue #976, opened 2023-02-09
[FEATURE REQUEST] Add more implementations of `IntoStrV`, `IntoOptionalGStr` etc


andy128k:
Currently `IntoStrV` is not implemented for `[&str; N]` and a cast to a slice is needed.

`IntoOptionalGStr` could also be implemented for string types, similar to how `impl Into<Option<&str>>` accepts string literals.

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/977

PR #977, opened 2023-02-10
glib: Add various `IntoStrV` impls

sdroege:
And don't copy any strings in the ones for containers of already `NUL`-terminated strings.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/976

### DIFF
--- a/glib/src/collections/strv.rs
+++ b/glib/src/collections/strv.rs
@@ -1054,31 +1054,121 @@ const MAX_STACK_ALLOCATION: usize = 16;
 impl IntoStrV for Vec<GString> {
     #[inline]
     fn run_with_strv<R, F: FnOnce(&[*mut c_char]) -> R>(self, f: F) -> R {
-        let required_len = (self.len() + 1) * mem::size_of::<*mut c_char>()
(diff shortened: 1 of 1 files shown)


glib/src/collections/strv.rs
```diff
@@ -1151,7 +1241,8 @@ impl<const N: usize> IntoStrV for [GString; N] {
                 let mut strs = ptrs.add(self.len() + 1) as *mut c_char;
 
                 for (i, item) in self.iter().enumerate() {
-                    ptr::copy_nonoverlapping(item.as_ptr(), strs, item.len() + 1);
+                    ptr::copy_nonoverlapping(item.as_ptr() as *const _, strs, item.len());
+                    *strs.add(item.len()) = 0;
                     *ptrs.add(i) = strs;
                     strs = strs.add(item.len() + 1);
```
andy128k, line 1247: Should we care about alignment?

sdroege: char arrays have an alignment of 1 so that's fine here. But that's exactly why the pointers come first and afterwards the characters :)

andy128k (approved):

PR #979, opened 2023-02-10
Implement Default trait for vectors

yuraiz:
I made a gtk widget that uses graphene::Vec3 internally, 
https://github.com/yuraiz/telegrand/blob/animated-background/src/session/content/gradient_bg.rs
and because vectors doesn't implement Default I can't derive Default for widget implementation 

also it would be useful to have gdk::RGBA to vector conversions but that's compicated because those libraries are independent

### DIFF
--- a/graphene/src/matrix.rs
+++ b/graphene/src/matrix.rs
@@ -213,6 +213,12 @@ impl fmt::Debug for Matrix {
     }
 }
 
+impl Default for Matrix {
+    fn default() -> Self {
+        Self::new_identity()
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::Matrix;
--- a/graphene/src/box_.rs
+++ b/graphene/src/box_.rs
@@ -79,6 +79,12 @@ impl Box {
     }
 }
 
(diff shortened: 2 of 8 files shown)


sdroege (dismissed):
Probably makes sense to also implement that for the `Point`s the same way, and for the `Matrix` as the identity matrix.

Do you want to add those?

yuraiz:
> Probably makes sense to also implement that for the `Point`s the same way, and for the `Matrix` as the identity matrix.
> 
> Do you want to add those?

Yes

yuraiz:
I found that vectors have zero() fn, it's better to use it

graphene/src/matrix.rs
```diff
@@ -213,6 +213,12 @@ impl fmt::Debug for Matrix {
     }
 }
 
+impl Default for Matrix {
+    fn default() -> Self {
+        Self::from_float(Default::default())
```
sdroege, line 218:
```suggestion
        Self::new_identity()
```

sdroege (approved):

PR #984, opened 2023-02-11
Props type support

### LINKED ISSUE #983: [FEATURE REQUEST] Newtype support for `#[property]`
I have a newtype over f64:
```rust
pub enum ScaleRequest {
    FitToAllocation,
    Set(f64),
}
```
It maps one-to-one to f64 (0. maps to the first item, others map to the second item). I'd like to use it with the properties macro. For GObject it should appear as an f64 property.

Currently this seems to require implementing `HasParamSpec`, `ToValue`, `From<ScaleRequest> for Value` (for some reason?) and `FromValue`, the last of which requires unsafe code:
```rust
pub enum ScaleRequest {
    FitToAllocation,
    Set(f64),
}

impl From<f64> for ScaleRequest {
    fn from(value: f64) -> Self {
        if value == 0. {
            ScaleRequest::FitToAllocation
        } else {
            ScaleRequest::Set(value.clamp(0., 10.))
        }
    }
}

impl glib::HasParamSpec for ScaleRequest {
    type ParamSpec = glib::ParamSpecDouble;
    type SetValue = Self;
    type BuilderFn = fn(&str) -> glib::ParamSpecDoubleBuilder;

    fn param_spec_builder() -> Self::BuilderFn {
        Self::ParamSpec::builder
    }
}

impl From<ScaleRequest> for glib::Value {
    fn from(value: ScaleRequest) -> Self {
        value.to_value()
    }
}

impl glib::ToValue for ScaleRequest {
    fn to_value(&self) -> glib::Value {
        match *self {
            ScaleRequest::FitToAllocation => 0.,
            ScaleRequest::Set(scale) => scale,
        }
        .to_value()
    }

    fn value_type(&self) -> glib::Type {
        f64::static_type()
    }
}

unsafe impl<'a> glib::value::FromValue<'a> for ScaleRequest {
    type Checker = glib::value::GenericValueTypeChecker<f64>;

    unsafe fn from_value(value: &'a glib::Value) -> Self {
        f64::from_value(value).into()
    }
}
```

Would be great if this could be representable without unsafe code with some kind of pair of `From<f64> for T` + `From<T> for f64`.

It's also worth noting that I could instead implement `Property` with `Value = f64` and custom get and set returning `f64`. However, that would result in my Rust type having getter and setter working with `f64` instead of the real type, which is not as nice.

ranfdev:
fixes #983 
enables support for a correct handling of `ThreadGuard`, therefore supersedes #968.

In general, this kinda enables two-level nesting: this makes supporting newtypes trivial (conceptually, they are a nested type). 

The change is backwards compatible

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -379,13 +379,24 @@ fn expand_property_fn(props: &[PropDesc]) -> TokenStream2 {
                 ),
                 (None, MaybeCustomFn::Default) => quote!(
                     DerivedPropertiesEnum::#enum_ident =>
-                        #crate_ident::PropertyGet::get(&self.#field_ident, |v| ::std::convert::From::from(v))
+                        #crate_ident::PropertyGet::get(
+                            &self.#field_ident,
+                            |v| #crate_ident::PropertyGet::get(
+                                v,
+                                |v| ::std::convert::From::from(v)
+                            )
+                        )
 
                 ),
                 (Some(member), MaybeCustomFn::Default) => quote!(
                     DerivedPropertiesEnum::#enum_ident =>
-                        #crate_ident::PropertyGet::get(&self.#field_ident, |v| ::std::convert::From::from(&v.#member))
-
+                        #crate_ident::PropertyGet::get(
+                            &self.#field_ident,
+                            |v| #crate_ident::PropertyGet::get(
+                                &v.#member,
+                                |v| ::std::convert::From::from(v)
+                            )
+                        )
                 ),
             };
             quote_spanned!(span=> #body)
@@ -415,6 +426,7 @@ fn expand_set_property_fn(props: &[PropDesc]) -> TokenStream2 {
             field_ident,
             member,
             set,
+            ty,
             ..
         } = p;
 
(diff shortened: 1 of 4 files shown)


ranfdev:
Nesting multiple writable types is still not supported.
So, `Rc<Mutex<Rc<u32>>>` is now supported, but `Mutex<Mutex<Rc<u32>>>` isn't supported at all (as before).

ranfdev:
I guess there's still one problem: the setter on the wrapper type accepts the inner type, that is, `i32`, not `MyInt(i32)`. 

CC: @YaLTeR

YaLTeR:
Yeah, it would be good if the generated public-facing methods had newtype in their signatures.

glib-macros/tests/properties.rs
```diff
@@ -101,10 +120,16 @@ mod foo {
...
+            #[property(get, set)]
+            custom_type: RefCell<MyInt>,
             // The following property doesn't store any data. The value of the property is calculated
             // when the value is accessed.
             #[property(get = Self::hello_world)]
             _buzz: PhantomData<String>,
+            #[property(get, set)]
+            thread_guard_wrapped: Mutex<ThreadGuard<u32>>,
+            #[property(get, set)]
+            thread_guard_wrapping: ThreadGuard<Mutex<u32>>,
```
sdroege, line 132: Would this also already handle a `Mutex<Option<ThreadGuard<Paintable>>>`? Asking for the `Option` there. It's a write-only property so I think that potentially works because there's `From<T> for Option<T>` but I think it doesn't work in practice because you'd need to call `from()` twice?

glib-macros/src/properties.rs
```diff
@@ -426,22 +438,28 @@ fn expand_set_property_fn(props: &[PropDesc]) -> TokenStream2 {
...
                     DerivedPropertiesEnum::#enum_ident => {
+                        let value: <#ty as #crate_ident::Property>::Value =
+                            #crate_ident::Value::get(value)#expect;
                         #crate_ident::PropertySetNested::set_nested(
                             &self.#field_ident,
-                            move |v| v.#member = #crate_ident::Value::get(value)#expect
+                            move |v| v.#member = ::std::convert::From::from(value)
                         );
                     }
                 ),
```
sdroege, line 465: Adding a discussion item for this somewhere so we don't forget

> I guess there's still one problem: the setter on the wrapper type accepts the inner type, that is, i32, not MyInt(i32).

bilelmoussaoui:
Would you mind rebasing this one?

ranfdev:
I can rebase, but I don't think this is ready to be merged. There are still some unresolved things, as pointed out by the unresolved comments

PR #985, opened 2023-02-11
Make `WeakRef` and `SendWeakRef` useable with the `Properties` derive macro

### LINKED ISSUE #980: [FEATURE REQUEST] Make `glib::WeakRef` useable with the `Properties` derive macro
Currently it's missing some trait impls

RealKC:
There's still some things to do, but I'm not sure how to do them:

1. How should I test these?
2. I'm not sure how to implement `PropertySetNested` for them

Closes #980

### DIFF
--- a/glib/src/property.rs
+++ b/glib/src/property.rs
@@ -10,6 +10,9 @@ use std::sync::Mutex;
 use std::sync::RwLock;
 
 use crate::HasParamSpec;
+use crate::IsA;
+use crate::Object;
+use crate::WeakRef;
 
 // rustdoc-stripper-ignore-next
 /// A type that can be used as a property. It covers every type which have an associated `ParamSpec`
@@ -49,6 +52,9 @@ impl<T: Property> Property for Rc<T> {
 }
 impl<T: Property> Property for Arc<T> {
     type Value = T::Value;
+}
+impl<T: IsA<Object> + HasParamSpec> Property for WeakRef<T> {
+    type Value = Option<T>;
 }
 
 // rustdoc-stripper-ignore-next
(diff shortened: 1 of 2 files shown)


RealKC:
> I'm not sure how to implement PropertySetNested for them

Thinking about it, this may not actually be needed?

ranfdev:
> How should I test these?

You could try setting something in the property and retrieving the value back. But If the property compiles, the code should just work anyway.

> I'm not sure how to implement PropertySetNested for them

The trait `PropertySetNested` is used by the macro to set an inner field of a struct. These `WeakRef` types take a `T: IsA<Object>`, they can't take a simple struct, neither they can wrap a type taking a simple struct. 

In brief, it's ok for them to not implement `PropertySetNested`

glib/src/property.rs
```diff
@@ -50,6 +54,12 @@ impl<T: Property> Property for Rc<T> {
 impl<T: Property> Property for Arc<T> {
     type Value = T::Value;
 }
+impl<T: IsA<Object> + Property> Property for WeakRef<T> {
+    type Value = T::Value;
+}
+impl<T: IsA<Object> + Property> Property for SendWeakRef<T> {
+    type Value = T::Value;
```
ranfdev, line 61: `T` is a gobject, so `Property::Value` is `T` itself.  You should be able to just put `Value = T` in there.
Then, the bound can just be `T: IsA<Object>`

RealKC:
> You could try setting something in the property and retrieving the value back. But If the property compiles, the code should just work anyway.

Yeah I wasn't sure of an object that would've been easily weak-ref-able while also being easy to compare

glib/src/property.rs
RealKC: It doesn't seem like I can get it to just `T: IsA<Object>`, but I could get it to `T: IsA<Object> + HasParamSpec`, should I do that?

ranfdev: Yes, that's correct, `HasParamSpec` is needed

ranfdev (dismissed):

glib/src/property.rs
```diff
@@ -169,6 +179,41 @@ impl<T> PropertySet for once_cell::unsync::OnceCell<T> {
     }
 }
 
+impl<T: IsA<Object>> PropertyGet for WeakRef<T> {
+    type Value = T;
+
+    fn get<R, F: Fn(&Self::Value) -> R>(&self, f: F) -> R {
+        match self.upgrade() {
+            Some(value) => f(&value),
+            None => panic!("failed to upgrade weak reference"),
```
sdroege, line 188: Is that really the behaviour we want? I would've expected it to simply return `None`/`NULL` in the property instead of panicking.

RealKC: Changed it to use `Option<T>`

sdroege (approved):
@ranfdev What do you think?

ranfdev:
lgtm

PR #987, opened 2023-02-11
Add FromValueDelegate

ranfdev:
Makes implementing `FromValue` safe and trivial

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -370,20 +370,6 @@ pub trait ToValue {
 }
 
 // rustdoc-stripper-ignore-next
-/// Blanket implementation for all references.
-impl<T: ToValue + StaticType> ToValue for &T {
-    #[inline]
-    fn to_value(&self) -> Value {
-        T::to_value(*self)
-    }
-
-    #[inline]
-    fn value_type(&self) -> Type {
-        T::static_type()
-    }
-}
-
-// rustdoc-stripper-ignore-next
 /// Trait to convert an `Option` to a `Value` for optional types.
 pub trait ToValueOptional {
     // rustdoc-stripper-ignore-next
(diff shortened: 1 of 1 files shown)


glib/src/value.rs
```diff
@@ -297,6 +297,26 @@ mod private {
...
+// rustdoc-stripper-ignore-next
+/// Given a type implementing `From<i32>`, the type can easily and safely
+/// implement `FromValue` by doing
+/// ```
+/// impl FromValueDelegate for MyInt {
+///   type Delegate = i32;
+/// }
+/// ```
+pub trait FromValueDelegate<'a>: From<Self::Delegate> {
+    type Delegate: FromValue<'a>;
```
sdroege, line 309: Wouldn't be nicer to also automatically get the `ToValue` `impl` here at the same time? `ToValue` is safe to implement but if we require `From` in both directions I think having a single `ValueDelegate` trait like this would be nicer.

glib/src/value.rs
```diff
@@ -369,20 +369,6 @@ pub trait ToValue {
     fn value_type(&self) -> Type;
 }
 
-// rustdoc-stripper-ignore-next
-/// Blanket implementation for all references.
-impl<T: ToValue + StaticType> ToValue for &T {
```
ranfdev, line 374: I needed to remove this to auto-implement `ToValue` for `T: ValueDelegate`. Else they were in conflict.

The tests don't complain so...  🤷🏻

ranfdev:
Uh oh, we don't have GATs in rust 1.64, only from 1.65. 😢 

That ValueDelegate trait implementation will require a `T: Clone` bound then

sdroege:
> Uh oh, we don't have GATs in rust 1.64, only from 1.65. cry

For the next release we can update to 1.65 but that won't do for backporting :)

RealKC:
I don't think this can be backported anyway? It removes a trait impl and I'm not sure the new one covers the cases the old trait impl did

glib/src/value.rs
sdroege: I think this makes it impossible to pass `&Object` to e.g. `Object::set_property()` and only `Object` would work

ranfdev:
So, in brief, implementing this with traits is  impossible. To keep the old ToValue implementation for `&T`, Rust would need support for specialization.

Fortunately, we have macros!
Instead of implementing ToValue/FromValue with a blanket implementation, we could have a `#[derive(ValueDelegate)]`
 to achieve the same thing.

And that would work for every rust version. 
The result is equivalent.  The macro would just generate a `ToValue/FromValue` implementation calling `From<DelegatedType>` and `Into<DelegatedType>`

sdroege:
I think that makes sense. It's even less user code than implementing a trait manually ;)

pbor:
https://github.com/gtk-rs/gtk-rs-core/issues/983 has the macro approach

PR #998, opened 2023-02-15
RFC: cairo-rs: add Context API to enforce stack correctness

epilys:
(This is a pattern I implement with a trait in my applications.)

This pull requests proposes a simple API to make sure Context stacks are correct on compile-time. The following is an example usage:

```rust
pub fn draw_stuff(area: &gtk::DrawingArea, mut cr: ContextRef, flag: bool) {
    cr.set_source_rgba(1.0, 1.0, 1.0, 1.0);
    let mut cr1 = cr.push(); // mut is necessary in order to create new stacks from cr1.

    if flag {
        let cr2 = cr1.push();
        // Use cr2 as a regular cairo::Context reference, since it derefs to it automatically.
        let mut m = cairo::Matrix::new();
        m.translate(234.0, 324.0);
        cr2.transform(matrix);

        cr2.rectangle(
            0.0,
            0.0,
            1.0,
            1.0,
        );
        cr2.fill().unwrap();
    } // cr2 drops here

    cr1.move_to(0.0, 0.5);
    cr2.set_source_rgba(1.0, 0.0, 0.0, 1.0);
    cr1.line_to(0.5, 15.0);
    cr1.stroke().unwrap();
    drop(cr1);
    cr.arc(
        0.0,
        0.0,
        5.0,
        0.0,
        2.0 * std::f64::consts::PI,
    );
    cr.fill().unwrap();
    // cr drops, all stacks have been restored.
}
``` 

Commit message:

`Context` has the `Context::save` and `Context::restore` methods to save the current Context state, do changes to the new state, and be able to restore the original state. If a save() is not followed by a restore() the result is visual bugs (at least in my experience).

In C there's no way to prevent a user from forgetting a restore().

In Rust you can nest stack levels naturally using the borrowing rules. Since a mutable reference of a variable prevents the variable to be used while the reference is still alive, we can ensure the user cannot create a new stack level and forget about the previous one. The restore() is performed inside the Drop implementation. That means that if it compiles, the stacking is correct, and since this is on the type level it is a zero-cost abstraction.

Questions:
- should `Context` be available to the user without `unsafe`? As long as it is, the user can create as many `ContextRef`s as they wish and bypass this API.
- If restore().unwrap() in the Drop impl panics, the process aborts. Is that acceptable?

### DIFF
--- a/cairo/src/context.rs
+++ b/cairo/src/context.rs
@@ -944,11 +944,81 @@ impl Context {
             ffi::cairo_tag_end(self.0.as_ptr(), tag_name.as_ptr())
         }
     }
+
+    pub fn push<'a, 'b: 'a>(self: &'a mut &'b Self) -> ContextRef<'a, 'b> {
+        self.save().unwrap();
+        ContextRef(self)
+    }
 }
 
 impl fmt::Display for Context {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "Context")
+    }
+}
+
+/// `Context` wrapper to enforce valid stacking with type-level semantics.
+///
+/// ```compile_fail
+/// # use cairo::ContextRef;
+/// # fn func(mut cr: ContextRef) {
+///     let mut new_context: ContextRef = cr.push();
+///     let new_context2 = cr.push(); // `new_context` is still alive, so this fails to compile.
+/// # }
+/// ```
+///
+/// ```no_run
+/// # use cairo::ContextRef;
+/// # fn func(mut cr: ContextRef, x: f64, y: f64) {
+///     let mut new_context = cr.push();
+///
+///     {
+///         // `let` creates its own scope, but by using {}
+///         // we don't have to manually call `drop(new_context2)`
+///         // in order to be able to use `new_context` again.
+///         let new_context2 = new_context.push();
+///         new_context2.line_to(x, y);
+///     }
+///
+///     new_context.line_to(y, x);
+/// # }
+/// ```
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct ContextRef<'a, 'b: 'a>(&'a mut &'b Context);
+
+impl std::ops::Deref for ContextRef<'_, '_> {
+    type Target = Context;
+    fn deref(&self) -> &Self::Target {
+        self.0
+    }
+}
+
+impl Drop for ContextRef<'_, '_> {
+    fn drop(&mut self) {
+        self.0.restore().unwrap();
+    }
+}
+
+impl<'c, 'a: 'c, 'b: 'a> ContextRef<'a, 'b> {
+    pub fn save(&self, _do_not_call: ()) {
+        // Override save method so that self.save().unwrap() fails?
+    }
+
+    pub fn restore(&self, _do_not_call: ()) {
+        // Override restore method so that self.save().unwrap() fails?
+    }
+
+    pub fn push(self: &'c mut ContextRef<'a, 'b>) -> ContextRef<'c, 'b> {
+        self.0.save().unwrap();
+        ContextRef(self.0)
+    }
+}
+
+impl<'a, 'b> From<&'a mut &'b Context> for ContextRef<'a, 'b> {
+    fn from(cr: &'a mut &'b Context) -> Self {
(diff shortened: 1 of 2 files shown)


federicomenaquintero:
The following is a bit of anecdotal evidence, not really a review of your code, but it may help you decide what to do with it. I hope it's useful :smile: 

Librsvg [used to have a similar abstraction](https://gitlab.gnome.org/GNOME/librsvg/-/issues/759) in the days when cairo-rs would panic if a Context entered an error state.  We found this inconvenient, so changed the cairo-rs API to return Result in more places and to not panic.

One of your questions is this:

> - If restore().unwrap() in the Drop impl panics, the process aborts. Is that acceptable?

I'd say it depends.  For example, librsvg may deal with untrusted SVGs and it should not panic, so it needs to be able to catch those errors from Cairo when cairo-rs offers them as Results.  For an application which only does trusted drawing and the worst that could happen to it could be to OOM inside Cairo, maybe panicking is fine.

Librsvg replaced the restore-on-Drop abstraction with this:

```rust
pub fn with_saved_cr<O, F>(cr: &cairo::Context, f: F) -> Result<O, RenderingError>
where
    F: FnOnce() -> Result<O, RenderingError>,
{
    cr.save()?;
    match f() {
        Ok(o) => {
            cr.restore()?;
            Ok(o)
        }

        Err(e) => Err(e),
    }
}
```

Ignore the `O` and the conversion to librsvg's own `RenderingError`.  The point here is that errors are caught and reported.

I think librsvg has had [a single bug where it forgot to restore()](https://gitlab.gnome.org/GNOME/librsvg/-/issues/746); it was because something like `cr.save(); something_else()?; cr.restore();` was returning in the `?`.  I added a `SavedCr` abstraction like yours after that, but ended up changing it for that `with_saved_cr()` function.

**Cairo annoyances:**

**gstate doesn't save everything:** Not everything that you can mutate in a `cr` is saved/restored!  For example, the current path is not saved, but the current clipping path and the current transform are.  The idea is that you can do this:

```rust
   cr.new_path();
   cr.move_to(...);
   cr.line_to(...); // start a current path...

   cr.save();

   cr.transform(...); // change the CTM to be more convenint or something
   cr.curve_to(); // frob the current path some more...

   cr.restore(); // restore the CTM, keep the path in its current state

   cr.line_to(...); // from the curve's endpoint to a point in the original CTM
   cr.stroke();
```

I think cairo-rs's API would get very complicated if we tried to model this accurately.  But who knows?  Take a dive into `cairo-gstate-private.h` and the surrounding code :smiley: 

**One cannot restore() out of an error state:** A `cr` can go into an error state for various reasons, like OOM or setting a non-invertible transform.  It can also be totally non-obvious: if Cairo doesn't like the dimensions of the temporary surface that it uses while rendering patterns (gradients, etc.), it will return an internal error, which the high-level code misinterprets as OOM (see e.g. [this bug](https://gitlab.gnome.org/GNOME/librsvg/-/issues/683)).

The problem is easy to reproduce:

```rust
  cr.save();
  cr.transform(matrix_with_all_zeros);
  cr.restore() // properly returns Err(cairo::Error::InvalidMatrix)

  // Now you can't use the cr anymore because it is in an error state.
```

After a `cr` gets into an error state, there is no way out of it - the cr is useless and just "eats" commands without doing anything.

Librsvg interprets an error from `cr.restore()` as "there's nothing left to do; exit early and return an error".

Addendum: this behavior is **TERRIBLE** for things like EOG, or GTK3 in general, which generally pass the `cr` from their `GtkWidget.draw()` into librsvg.  If they render an SVG that puts the `cr` in an error state, then the rest of GTK can't draw anything and the whole window gets messed up.  See [this EOG bug](https://gitlab.gnome.org/GNOME/eog/-/issues/259) and the horrible example in [this MR](https://gitlab.gnome.org/GNOME/eog/-/merge_requests/123#note_1390167).

**Fallible Drop:** Or, dropping but being aware that something may fail - I haven't followed [the discourse around this](https://internals.rust-lang.org/t/pre-pre-rfc-nodrop-marker-trait/15682).

epilys:
Perhaps all of the error state problems would be handled if the `inner` Context kept state itself. Whenever an error happens, it is stored in the `inner` Context and subsequent ContextRef methods always fail with the same error.

PR #1000, opened 2023-02-15
Add ValueDelegate macro

### LINKED ISSUE #983: [FEATURE REQUEST] Newtype support for `#[property]`

ranfdev:
closes #983 

It works pretty well I'd say.

Should I add support for generics?


ranfdev:
Added support to make the newtype `nullable`, implementing `ToValueOptional`.
I won't add support for generics, unless requested.
Done

pbor:
The code itself looks good to me, though given that this is a new API etc, before merging I would prefer to get a thumbs-up from someone who had approval permission for more than a couple of days :)

In the mean time let me bring up some bikeshedding: which alternatives have we considered to the `ValueDelegate` name? I do not find it super clear... how about `ValueRepresentation` or `ValueRepr` ?

pbor:
Also, if `nullable` is specified, should we accept/use `TryFrom` and set `None` in case of error?

ranfdev:
> naming
`ValueRepresentation` is a bit too long. `ValueRepr` would be ok, but in any case I think `ValueDelegate` is clearer. The macro doesn't provide a _full_ representation. The representation is _delegated_ to the chosen type. GLib can't distinguish `MyInt(i32)` and an `i32`. So `MyInt(i32)` doesn't have a full representation. That would require registering another glib type.

> should we accept/use `TryFrom` and set `None` in case of error?

No, because that would silently ignore the error. We would need a specific `Checker` inside the `ToValue` trait to handle the conversion and provide a useful error. I agree this can be useful.

pbor (dismissed):
As mentioned, this looks good to me, but let's wait for other reviews as well before adding new API

ranfdev:
@sdroege we would like a thumbs up before merging a new API :)

sdroege (dismissed):

sdroege:
Seems good to me, please merge after solving the conflicts :)

sdroege:
So let's get this in now? @pbor ?

ranfdev:
I wanted to merge it, I just needed a review to do so

sdroege (approved):

Issue #1009, opened 2023-02-17
[BUG] Leak in `Variant::to_variant`


mbiggio:
**Bug description**
This small reproducer is leaking the variant when `true` is converted to a `Variant`:

```
use gio::prelude::*;
use gio::{SimpleAction, SimpleActionGroup};

fn main() {
    let actions = SimpleActionGroup::new();
    let my_action = SimpleAction::new(
        "my-action",
        Some(&bool::static_variant_type()),
    );  
    actions.add_action(&my_action);

    actions.lookup_action("my-action")
        .unwrap_or_else(|| panic!("missing action"))
        .downcast::<gio::SimpleAction>()
        .unwrap()
        .set_state(true.to_variant());
}
```

This can be reproduced only with gio 0.17, not with 0.16

**Leak**

```
==3355619== 89 (48 direct, 41 indirect) bytes in 1 blocks are definitely lost in loss record 254 of 349 
==3355619==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==3355619==    by 0x4AFDE98: g_malloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==3355619==    by 0x4B16485: g_slice_alloc (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==3355619==    by 0x4B376BA: ??? (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==3355619==    by 0x4B3784B: g_variant_new_from_bytes (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==3355619==    by 0x4B30627: ??? (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==3355619==    by 0x4B306B2: g_variant_new_boolean (in /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0.6400.6)
==3355619==    by 0x113C32: <bool as glib::variant::ToVariant>::to_variant (variant.rs:1135)
==3355619==    by 0x1116D1: action_test_case::main (main.rs:16)
==3355619==    by 0x1122FA: core::ops::function::FnOnce::call_once (function.rs:251)
==3355619==    by 0x11227D: std::sys_common::backtrace::__rust_begin_short_backtrace (backtrace.rs:121)
[...]

PR #1010, opened 2023-02-17
SimpleAction: take ownership of value without leaking it

mbiggio:
The rust API takes ownership of value, but the wrapped C API does not


mbiggio:
Solves #1009 for release 0.17

sdroege:
Can you also send a PR that reverts the commit for master?

gio/src/simple_action.rs
```diff
@@ -0,0 +1,35 @@
+use crate::SimpleAction;
```
sdroege, line 1: Misses the two comment lines from all the other files here

sdroege (approved):

mbiggio:
> Can you also send a PR that reverts the commit for master?

Sure. We prefer to revert in the end or should we keep the API change?
@pbor

sdroege:
> Sure. We prefer to revert in the end or should we keep the API change?

Revert, the API change makes no sense

PR #1012, opened 2023-02-17
gio: Add subclassing support to AppLaunchContext & FileMonitor

bilelmoussaoui:


bilelmoussaoui:
Don't merge yet, I will be adding other stuff to this PR like GIcon/GLoadableIcon.

gio/src/subclass/icon.rs
```diff
@@ -0,0 +1,174 @@
...
+pub trait IconImpl: ObjectImpl + Hash + PartialEq {
+    fn serialize(&self) -> Option<glib::Variant> {
+        self.parent_serialize()
+    }
+
+    fn to_tokens(&self) -> Option<(Vec<String>, i32)> {
+        self.parent_to_tokens()
+    }
+
+    fn from_tokens(tokens: &[String], version: i32) -> Result<Icon, Error> {
```
sdroege, line 21: This and above is a `StrV` if you want to avoid all the copying

gio/src/subclass/loadable_icon.rs
```diff
@@ -0,0 +1,263 @@
...
+    ) -> Pin<Box<dyn Future<Output = Result<(InputStream, Option<GString>), glib::Error>> + 'static>>
+    {
+        self.parent_load_future(size)
+    }
+}
+
+pub trait LoadableIconImplExt: ObjectSubclass {
+    fn parent_load(
+        &self,
+        size: i32,
```
sdroege, line 31: Should this be an `IconSize`?

gio/src/subclass/file_monitor.rs
```diff
@@ -0,0 +1,85 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{subclass::prelude::*, File, FileMonitor, FileMonitorEvent};
+use glib::{prelude::*, translate::*};
+
+pub trait FileMonitorImpl: ObjectImpl {
```
sdroege, line 6: Is this actually useful?

gio/src/subclass/app_launch_context.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+unsafe extern "C" fn app_launch_context_launch_failed<T: AppLaunchContextImpl>(
+    ptr: *mut ffi::GAppLaunchContext,
+    startup_id: *const c_char,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    imp.launch_failed(&GString::from_glib_borrow(startup_id))
```
sdroege, line 166: Should maybe be a `&GStr`?

gio/src/subclass/loadable_icon.rs
bilelmoussaoui: It cannot be, IconSize is a gtk only enum

gio/src/subclass/file_monitor.rs
bilelmoussaoui: Today it is not for sure. My intention is to cover up whatever APIs we can manually to have a better understanding of what is needed for an automatically generated of the subclassing code in the future

sdroege: I mean, is this actually intended to be subclasses? The API exposed here looks quite incomplete for implementing a new file monitor

PR #1023, opened 2023-02-21
Add nullable attribute on properties macro

### LINKED ISSUE #990: [BUG] properties: setter (on wrapper type) for a `RefCell<Option<Obj>>` takes `Borrow<Obj>` instead of `Borrow<Option<Obj>>`
Using `Borrow` in the setter was a trick to let `String` properties have a setter taking `Borrow<str>` and  `Object` properties having a setter taking any `Borrow<$Subclass>`.

But once the type is wrapped inside an `Option`, everything breaks. `Borrow` isn't recursive.
```
328 |         myfoo.set_object(Some(glib::BoxedAnyObject::new("")));
    |               ---------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Borrow<Option<Object>>` is not implemented for `Option<BoxedAnyObject>`
    |               |
    |               required by a bound introduced by this call
```

I currently don't know how to solve this.

ranfdev:
Fixes #990

I know, if a `ParamSpecBuilder` has a method `nullable`, this attribute will now shadow it. But I don't know any ParamSpec with a method named `nullable`. And even if that exists, the builder pattern can always be used: `builder().nullable()`.

This change is backwards compatible (expect if someone is using a builder with a `nullable` method I guess...).

Any other solution doesn't seem to be more backward compatible than this.

Also, the `nullable` attribute is used already by other macros, such as `glib::Boxed`, so this behavior seems consistent to me. 

If this isn't enough, we will introduce something better, introducing breaking changes, for the next milestone.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -872,7 +872,8 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 /// `PropertySet` and `PropertySetNested` if possible.
 ///
 /// The type `Option<T>` is supported as a property only if `Option<T>` implements `ToValueOptional`.
-/// Optional types also require the `nullable` attribute.
+/// Optional types also require the `nullable` attribute: without it, the generated setter on the wrapper type
+/// will take `T` instead of `Option<T>`, preventing the user from ever calling the setter with a `None` value.
 /// If your type doesn't support `PropertySet`, you can't use the generated setter, but you can
 /// always define a custom one.
 ///
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -872,6 +872,7 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 /// `PropertySet` and `PropertySetNested` if possible.
 ///
 /// The type `Option<T>` is supported as a property only if `Option<T>` implements `ToValueOptional`.
+/// Optional types also require the `nullable` attribute.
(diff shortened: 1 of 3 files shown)


jf2048:
This is a good fix for now. But still it is incorrect to be using `Borrow` here

sdroege:
> This change is backwards compatible (expect if someone is using a builder with a `nullable` method I guess...).

I wouldn't worry about that. All known pspecs are defined by GObject itself, GTK and GStreamer.

glib-macros/src/lib.rs
```diff
@@ -872,6 +872,7 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 /// `PropertySet` and `PropertySetNested` if possible.
 ///
 /// The type `Option<T>` is supported as a property only if `Option<T>` implements `ToValueOptional`.
+/// Optional types also require the `nullable` attribute.
```
sdroege, line 875: This probably needs a bit more explanation in the docs. It also works without but explaining what the difference between with/without is would be useful here.

sdroege:
> This is a good fix for now. But still it is incorrect to be using Borrow here

We can't fix this for 0.17 so this is fine for now. Switching to a different trait has to happen on master, consistently for everything.

sdroege (approved):
Seems fine to me otherwise

ranfdev:
Feel free to merge

sdroege (approved):

PR #1024, opened 2023-02-22
glib: Switch Priority to associated constants

bilelmoussaoui:
It makes more sense compared a struct + various constants

### DIFF
--- a/glib/src/source.rs
+++ b/glib/src/source.rs
@@ -827,8 +827,27 @@ where
 
 // rustdoc-stripper-ignore-next
 /// The priority of sources
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
 pub struct Priority(i32);
+
+impl Priority {
+    #[doc(alias = "G_PRIORITY_HIGH")]
+    pub const HIGH: Self = Self(ffi::G_PRIORITY_HIGH);
+    #[doc(alias = "G_PRIORITY_DEFAULT")]
+    pub const DEFAULT: Self = Self(ffi::G_PRIORITY_DEFAULT);
+    #[doc(alias = "G_PRIORITY_HIGH_IDLE")]
+    pub const HIGH_IDLE: Self = Self(ffi::G_PRIORITY_HIGH_IDLE);
+    #[doc(alias = "G_PRIORITY_DEFAULT_IDLE")]
+    pub const DEFAULT_IDLE: Self = Self(ffi::G_PRIORITY_DEFAULT_IDLE);
+    #[doc(alias = "G_PRIORITY_LOW")]
+    pub const LOW: Self = Self(ffi::G_PRIORITY_LOW);
+}
+
+impl Default for Priority {
+    fn default() -> Self {
+        Self::DEFAULT
+    }
+}
 
 #[doc(hidden)]
 impl IntoGlib for Priority {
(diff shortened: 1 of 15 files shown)


glib/src/source.rs
```diff
@@ -827,44 +827,54 @@ where
 
 // rustdoc-stripper-ignore-next
 /// The priority of sources
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub struct Priority(i32);
+#[derive(Default, Clone, Copy, Debug, Eq, PartialEq)]
```
sdroege, line 830: You probably also want to implement `PartialOrd` / `Ord` based on the integer values

glib/src/source.rs
```diff
@@ -827,44 +827,54 @@ where
...
+    #[doc(alias = "G_PRIORITY_DEFAULT")]
+    #[default]
+    Default,
+    #[doc(alias = "G_PRIORITY_HIGH_IDLE")]
+    HighIdle,
+    #[doc(alias = "G_PRIORITY_DEFAULT_IDLE")]
+    DefaultIdle,
+    #[doc(alias = "G_PRIORITY_LOW")]
+    Low,
+    Custom(i32),
```
sdroege, line 843: Not sure if that matters here but this increases the size of the type from 32 bits to 40 bits, which means that it e.g. will require two registers on 32 bit platforms and usually will require 64 bits of storage (padding, etc).

You might instead want to keep the previous definition and add associated constants instead? Then you can still have `Priority::DEFAULT`, `Priority::HIGH`, etc.

bilelmoussaoui: Interesting. Switched to using associated consts instead

sdroege: Ideally Rust would allow defining enums in such a way that there are a few fixed values and all others are mapped to a generic one, but that's not possible right now except for things inside std.

glib/src/source.rs
```diff
@@ -830,6 +830,37 @@ where
...
+    pub const DEFAULT: Self = Self(ffi::G_PRIORITY_DEFAULT);
+    #[doc(alias = "G_PRIORITY_HIGH_IDLE")]
+    pub const HIGH_IDLE: Self = Self(ffi::G_PRIORITY_HIGH_IDLE);
+    #[doc(alias = "G_PRIORITY_DEFAULT_IDLE")]
+    pub const DEFAULT_IDLE: Self = Self(ffi::G_PRIORITY_DEFAULT_IDLE);
+    #[doc(alias = "G_PRIORITY_LOW")]
+    pub const LOW: Self = Self(ffi::G_PRIORITY_LOW);
+}
+
+impl PartialOrd for Priority {
```
sdroege, line 846: Now you can also just derive it

glib/src/source.rs
```diff
@@ -827,9 +827,28 @@ where
...
+    #[doc(alias = "G_PRIORITY_HIGH")]
+    pub const HIGH: Self = Self(ffi::G_PRIORITY_HIGH);
+    #[doc(alias = "G_PRIORITY_DEFAULT")]
+    pub const DEFAULT: Self = Self(ffi::G_PRIORITY_DEFAULT);
+    #[doc(alias = "G_PRIORITY_HIGH_IDLE")]
+    pub const HIGH_IDLE: Self = Self(ffi::G_PRIORITY_HIGH_IDLE);
+    #[doc(alias = "G_PRIORITY_DEFAULT_IDLE")]
+    pub const DEFAULT_IDLE: Self = Self(ffi::G_PRIORITY_DEFAULT_IDLE);
+    #[doc(alias = "G_PRIORITY_LOW")]
+    pub const LOW: Self = Self(ffi::G_PRIORITY_LOW);
```
sdroege, line 843: Should probably also get a `From<i32>` as any value is valid here

sdroege (dismissed):
Otherwise seems good to me

sdroege (approved):

sdroege:
Merging this probably requires some changes in the other repos.

bilelmoussaoui:
> Merging this probably requires some changes in the other repos.

Nothing to be updated on gtk4-rs, it is only used on the book which uses stable releases

Issue #1021, opened 2023-02-21
[FEATURE REQUEST] Add manual `connect_notify_local` method to glib::SignalGroup


melix99:
Currently to listen to notify signals of the properties of the target object of a `glib::SignalGroup`, we are forced to use the regular `connect_local`, which is a bit annoying. It would be cool to have a `connect_notify_local` to make things simpler to use.

melix99:
More info here: https://github.com/melix99/telegrand/pull/462#discussion_r1108776250

sdroege:
Do you want to provide a PR for this?

PR #1027, opened 2023-02-22
glib: Add connect_notify* methods to SignalGroup

melix99:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1021.

Didn't have a chance to test it locally because I couldn't use cargo patch in my project :/

### DIFF
--- a/glib/src/gobject/signal_group.rs
+++ b/glib/src/gobject/signal_group.rs
@@ -42,6 +42,50 @@ impl SignalGroup {
         self.connect_closure(signal_name, after, RustClosure::new_local(callback));
     }
 
+    #[inline]
+    pub fn connect_notify<F>(&self, name: Option<&str>, callback: F)
+    where
+        F: Fn(&crate::Object, &crate::ParamSpec) + Send + Sync + 'static,
+    {
+        let signal_name = if let Some(name) = name {
+            format!("notify::{name}")
+        } else {
+            "notify".into()
+        };
+
+        let closure = crate::RustClosure::new(move |values| {
+            let obj = values[0].get().unwrap();
+            let pspec = values[1].get().unwrap();
+            callback(obj, pspec);
+
+            None
+        });
+
+        self.connect_closure(&signal_name, false, closure);
+    }
+
+    #[inline]
+    pub fn connect_notify_local<F>(&self, name: Option<&str>, callback: F)
+    where
+        F: Fn(&crate::Object, &crate::ParamSpec) + 'static,
+    {
+        let signal_name = if let Some(name) = name {
+            format!("notify::{name}")
+        } else {
+            "notify".into()
+        };
+
+        let closure = crate::RustClosure::new_local(move |values| {
+            let obj = values[0].get().unwrap();
+            let pspec = values[1].get().unwrap();
+            callback(obj, pspec);
+
+            None
+        });
+
+        self.connect_closure(&signal_name, false, closure);
+    }
+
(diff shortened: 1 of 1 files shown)


glib/src/gobject/signal_group.rs
```diff
@@ -42,6 +42,44 @@ impl SignalGroup {
         self.connect_closure(signal_name, after, RustClosure::new_local(callback));
     }
 
+    #[inline]
+    pub fn connect_notify<F>(&self, name: Option<&str>, callback: F)
+    where
+        F: Fn() + Send + Sync + 'static,
```
sdroege, line 48: You probably want to pass the object and pspec into the closure

glib/src/gobject/signal_group.rs
```diff
@@ -42,6 +42,44 @@ impl SignalGroup {
         self.connect_closure(signal_name, after, RustClosure::new_local(callback));
     }
 
+    #[inline]
+    pub fn connect_notify<F>(&self, name: Option<&str>, callback: F)
```
sdroege, line 46: Why is this needed? It should be possibly to directly call http://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/object/trait.ObjectExt.html#tymethod.connect_notify on the `SignalGroup` as it's a `glib::Object`

melix99: Because of the nature of `SignalGroup`.

`SignalGroup` is an object that makes it easy to listen to multiple signals from the `target` object. This means that the `SignalGroup::connect` will listen for signals of the `target` object, instead of the `SignalGroup` itself. This PR adds an helper to connect to notify signals of properties of the `target` object, which is different than `ObjectExt::connect_notify`, because the latter only listens to notify signals of the `SignalGroup` object, which is not what we want.

glib/src/gobject/signal_group.rs
melix99: I'm not sure how to get the pspec of the `target` object though.

bilelmoussaoui: Using https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/object/trait.ObjectExt.html#tymethod.find_property?

sdroege: It's passed as argument to the callback

sdroege: Together with the object itself

glib/src/gobject/signal_group.rs
```diff
@@ -42,6 +42,50 @@ impl SignalGroup {
...
+        F: Fn(&Self, &crate::ParamSpec) + Send + Sync + 'static,
+    {
+        let signal_name = if let Some(name) = name {
+            format!("notify::{name}")
+        } else {
+            "notify".into()
+        };
+
+        let closure = crate::RustClosure::new(move |values| {
+            let obj = values[0].get().unwrap();
```
melix99, line 57: Ah, it returns the target object instead of the SignalGroup. Not sure what to do here. I think we can either return a glib::Object of the target object or clone the SignalGroup and return that instead. I'm inclined for the former.

melix99: I'll push that for now.

sdroege: Yes that's what I would expect. The other signals should work the same way

sdroege (approved):

PR #1031, opened 2023-02-25
pangocairo-sys: fix package.description in Cargo.toml

decathorpe:
It appears that this was overwritten by accident at some point.


sdroege (approved):
Oops. Thanks!

Issue #935, opened 2023-01-31
RFC: group futures related API in a `futures` mod


pbor:
We are running into namespacing issues where good function names clash, eg `spawn` in glib refers to spawning processes but would be handy to spawn futures.

The idea floated on the chat channel is to group all futures related API in a `futures` module.

To do that we need to collect a list of the functions we would move and how they would be renamed.

sdroege:
See also https://github.com/gtk-rs/gtk-rs-core/pull/902

pbor:
See also https://github.com/gtk-rs/gtk-rs-core/pull/528

PR #1033, opened 2023-02-26
Draft: Futures module

ranfdev:
Related issue: https://github.com/gtk-rs/gtk-rs-core/issues/935

I think there are currently too many timeout functions, with too many combinations of scheduling priority and scheduling precision (second, millisecond).

Instead of just moving every timeout function to a `futures` module, I've decided to redesign the API taking advantage of the Builder pattern and the [IntoFuture](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) trait.

API example:
```rust
c.block_on(async {
    sleep(Duration::from_millis(10)).await;
    sleep(Duration::from_secs(1))
        .priority(crate::PRIORITY_LOW)
        .precision(SchedulingPrecision::Seconds)
        .await;
});
```

I'm still looking if it's possible to redesign  `spawn` and `idle` with the same mindset. I don't particularly like having `spawn`, `spawn_local`, `spawn_local_with_priority`, `spawn_with_priority`, `spawn_within`, `spawn_within_with_priority`, ...


sdroege:
> I'm still looking if it's possible to redesign spawn and idle with the same mindset. I don't particularly like having spawn, spawn_local, spawn_local_with_priority, spawn_with_priority, spawn_within, spawn_within_with_priority, ...

That sounds like a nice idea. You probably won't get around distinguishing `spawn()` and `spawn_local()` though because of the different trait bounds.

glib/src/futures/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    type IntoFuture = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        use SchedulingPrecision::*;
+        match self.precision {
+            Millisecond => crate::timeout_future_with_priority(self.priority, self.duration),
+            Second => crate::timeout_future_seconds_with_priority(
+                self.priority,
+                self.duration.as_secs_f32().round() as u32,
```
sdroege, line 35: Why not `as_secs()`?

glib/src/futures/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+}
+
+impl Sleep {
+    pub fn priority(mut self, priority: Priority) -> Self {
+        self.priority = priority;
+        self
+    }
+    pub fn precision(mut self, precision: SchedulingPrecision) -> Self {
```
sdroege, line 46: Maybe the precision should be defined based on whether the `Duration` has fractional seconds or not? Or is that too implicit?

Otherwise maybe having a `sleep(Duration)` and `sleep_seconds(u32)` would be simpler API.

glib/src/futures/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+
+pub fn sleep(duration: Duration) -> Sleep {
+    Sleep {
+        priority: crate::PRIORITY_DEFAULT,
+        duration,
+        precision: SchedulingPrecision::Millisecond,
+    }
+}
+
+pub struct Timeout<F: Future> {
```
sdroege, line 60: This all needs docs

glib/src/futures/mod.rs
ranfdev: We can't make it implicit based on the choosen `Duration`, because the user could want to actually wait `2000` milliseconds precisely, and not `2` seconds roughly.

What about having the methods `precision_seconds`, `precision_milliseconds`, so that we can remove the public enum `SchedulingPrecision`?

sdroege: Good point. Yes that would also work

ranfdev: Thinking about it... In gtk adding a new enum for a specific function seems to be pretty normal. I think it fits adding an enum here as well

PR #1034, opened 2023-02-27
glib: Only optimize `IntoGStr` for `String` when capacity allows

jf2048:
And add a bench too, for me it shows this is faster for the small string case.

### DIFF
--- a/glib/benches/gstring.rs
+++ b/glib/benches/gstring.rs
@@ -1,0 +1,46 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use glib::IntoGStr;
+
+pub fn str_into_gstr(c: &mut Criterion) {
+    c.bench_function("str as IntoGStr", |b| {
+        b.iter(|| {
+            "abcdefg".run_with_gstr(|g| {
+                black_box(g);
+            })
+        })
+    });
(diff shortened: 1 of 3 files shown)


glib/benches/gstring.rs
```diff
@@ -0,0 +1,25 @@
...
+                black_box(g);
+            })
+        })
+    });
+}
+
+pub fn string_into_gstr(c: &mut Criterion) {
+    c.bench_function("String as IntoGStr", |b| {
+        b.iter(|| {
+            String::from("abcdefg").run_with_gstr(|g| {
```
sdroege, line 17: How does the `From` impl here work? Will it reserve exactly the length of the string or the next power of two?

Also I guess for benchmarking it would make sense to actually look at the different cases and compare the behaviour.

sdroege (commented):
The actual change looks good to me, thanks!

glib/benches/gstring.rs
jf2048: Right now it makes it with `len == capacity` but that behaviour seems unspecified. So I added a `shrink_to_fit` and an assert to make sure that holds.

sdroege (approved):

PR #1035, opened 2023-02-27
gio: add FileEnumerator::into_stream

jf2048:
cc @ebassi


gio/src/file_enumerator.rs
```diff
@@ -16,3 +17,124 @@ impl Iterator for FileEnumerator {
...
+        f.debug_struct("FileEnumeratorStream")
+            .field("enumerator", &self.enumerator)
+            .field("num_files", &self.num_files)
+            .field("priority", &self.priority)
+            .finish()
+    }
+}
+
+impl futures_core::Stream for FileEnumeratorStream {
+    type Item = Result<Vec<FileInfo>, glib::Error>;
```
sdroege, line 60: Why not a `glib::List<FileInfo>`?

ranfdev (commented):
Maybe this could be implemented with https://docs.rs/futures-util/latest/futures_util/stream/struct.IntoStream.html?

But then we can't specify the needed arguments, like the number of files... Feel free to ignore this comment.

gio/src/file_enumerator.rs
ebassi: tbh, I'd leave `glib::List` to API that interface directly with C. This is a Rust convenience, so I'd recommend using native/idiomatic Rust type wherever possible.

sdroege: `glib::List` works as an `Iterator` so that for the majority of uses it should be equivalent except for not having to copy over everything into a `Vec`.

jf2048: This is just passing the vec from the underlying `next_files_future`, it is already converted by the time the stream gets it. It should probably be changed in gir to use `List`, and then it can be changed here

jf2048:
That is only for a single element stream unfortunately, no first-class support for it like `IntoFuture`

gio/src/file_enumerator.rs
jf2048: See https://github.com/gtk-rs/gir/pull/1447

PR #1038, opened 2023-03-02
Fixed unit tests under macOS and possibly other *nix flavors

xanathar:
Fixes unit tests under macOS and possibly some other *nix flavors :

- Optional use of `SOCK_CLOEXEC` on platforms that do not support the flag
- Fixed `Path` constructor usage in macOS tests

About the `SOCK_CLOEXEC` option: I did a deep dive into libc codebase to find what are the supported platforms and I think I came with an exhaustive list (that includes enscriptem... I'm skeptical about that one, but libc defines the flag, so I did put it in). That said, I don't think there's a real need for the unit test to use `SOCK_CLOEXEC` (but I'm not the author of the test, so who knows), so another option is to just remove the flag for all platforms to reduce the clutter in the code.

### DIFF
--- a/gio/src/socket.rs
+++ b/gio/src/socket.rs
@@ -925,12 +925,7 @@ mod tests {
 
         let mut fds = [0 as libc::c_int; 2];
         let (out_sock, in_sock) = unsafe {
-            let ret = libc::socketpair(
-                libc::AF_UNIX,
-                libc::SOCK_STREAM | libc::SOCK_CLOEXEC,
-                0,
-                fds.as_mut_ptr(),
-            );
+            let ret = libc::socketpair(libc::AF_UNIX, libc::SOCK_STREAM, 0, fds.as_mut_ptr());
             if ret != 0 {
                 panic!("{}", io::Error::last_os_error());
             }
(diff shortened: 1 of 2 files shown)


gio/src/socket.rs
```diff
@@ -923,11 +923,48 @@ mod tests {
         use super::Socket;
         use crate::{prelude::*, Cancellable, UnixFDMessage};
 
+        #[cfg(any(
+            target_env = "newlib",
+            target_os = "android",
+            target_os = "dragonfly",
+            target_os = "emscripten",
+            target_os = "freebsd",
+            target_os = "fuchsia",
```
xanathar, line 932: There is the possibility that `fuchsia` actually does not meet the parent `cfg(unix)` condition - I have to check.

sdroege:
> About the `SOCK_CLOEXEC` option

I think removing that flag is OK here, and it keeps the code simple. It probably was just added for good measure. @jf2048?

jf2048:
You can just get rid of it, it's unnecessary. No need for this big cfg block. Probably I copied it from some Linux example somewhere.

pbor (approved):

xanathar:
I removed the `SOCK_CLOEXEC` stuff, should be "ready" if there aren't other issues.

PR #1039, opened 2023-03-02
Added subclassing support for gio::SocketControlMessage

xanathar:
Adds the possibility to subclass [`gio::SocketControlMessage`](https://gtk-rs.org/gtk-rs-core/git/docs/gio/struct.SocketControlMessage.html) to be subclassed to implement new messages.

### DIFF
--- a/gio/src/subclass/socket_control_message.rs
+++ b/gio/src/subclass/socket_control_message.rs
@@ -1,0 +1,262 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{subclass::prelude::*, translate::*, Cast};
+
+use crate::SocketControlMessage;
+
+pub trait SocketControlMessageImpl: ObjectImpl + SocketControlMessageImplExt {
+    fn level(&self) -> i32 {
+        self.parent_level()
+    }
+
+    fn msg_type(&self) -> i32 {
+        self.parent_msg_type()
+    }
+
+    fn size(&self) -> usize {
+        self.parent_size()
+    }
+
+    fn serialize(&self, data: &mut [u8]) {
+        self.parent_serialize(data);
+    }
+
+    fn deserialize(level: i32, type_: i32, data: &[u8]) -> Option<SocketControlMessage> {
+        Self::parent_deserialize(level, type_, data)
+    }
+}
+
+pub trait SocketControlMessageImplExt: ObjectSubclass {
+    fn parent_level(&self) -> i32;
+
+    fn parent_msg_type(&self) -> i32;
+
+    fn parent_size(&self) -> usize;
+
+    fn parent_serialize(&self, data: &mut [u8]);
+
+    fn parent_deserialize(level: i32, type_: i32, data: &[u8]) -> Option<SocketControlMessage>;
+}
+
+impl<T: SocketControlMessageImpl> SocketControlMessageImplExt for T {
+    fn parent_level(&self) -> i32 {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GSocketControlMessageClass;
+            let f = (*parent_class)
+                .get_level
+                .expect("No parent class implementation for \"level\"");
+
+            f(self
+                .obj()
+                .unsafe_cast_ref::<SocketControlMessage>()
+                .to_glib_none()
+                .0)
+        }
+    }
+
+    fn parent_msg_type(&self) -> i32 {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GSocketControlMessageClass;
+            let f = (*parent_class)
+                .get_type
+                .expect("No parent class implementation for \"msg_type\"");
+
+            f(self
+                .obj()
+                .unsafe_cast_ref::<SocketControlMessage>()
+                .to_glib_none()
+                .0)
+        }
+    }
+
+    fn parent_size(&self) -> usize {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GSocketControlMessageClass;
+            let f = (*parent_class)
+                .get_size
+                .expect("No parent class implementation for \"size\"");
+
+            f(self
+                .obj()
+                .unsafe_cast_ref::<SocketControlMessage>()
+                .to_glib_none()
+                .0)
+        }
+    }
+
+    fn parent_serialize(&self, data: &mut [u8]) {
+        unsafe {
+            let type_data = T::type_data();
+            let parent_class =
+                type_data.as_ref().parent_class() as *mut ffi::GSocketControlMessageClass;
+            let f = (*parent_class)
+                .serialize
+                .expect("No parent class implementation for \"serialize\"");
+
+            f(
+                self.obj()
+                    .unsafe_cast_ref::<SocketControlMessage>()
+                    .to_glib_none()
+                    .0,
+                data.as_mut_ptr() as _,
+            )
+        }
+    }
+
+    fn parent_deserialize(level: i32, type_: i32, data: &[u8]) -> Option<SocketControlMessage> {
+        unsafe {
+            let type_data = T::type_data();
+            let parent_class =
+                type_data.as_ref().parent_class() as *mut ffi::GSocketControlMessageClass;
+
+            (*parent_class).deserialize.map(|f| {
+                let message_ptr = f(level, type_, data.len(), data.as_ptr() as _);
+                from_glib_full(message_ptr)
+            })
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


pbor (dismissed):
This looks good to me. But in full disclosure I discussed this with Marco offline.

xanathar:
Fixed a clippy warning that was breaking the CI.

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
...
+        #[glib::object_subclass]
+        impl ObjectSubclass for TestSocketControlMessage {
+            const NAME: &'static str = "TestSocketControlMessage";
+            const ABSTRACT: bool = false;
+            type Type = super::TestSocketControlMessage;
+            type ParentType = SocketControlMessage;
+
+            fn new() -> Self {
+                Self
+            }
```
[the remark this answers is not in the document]
xanathar: I've already removed the constructor entirely in the last commit because it's not needed.

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::prelude::*;
+
+    mod imp {
+        use super::*;
+
+        pub struct TestSocketControlMessage;
```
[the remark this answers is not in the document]
xanathar: I've already removed the constructor entirely in the last commit because it's not needed, so there's no need for the derive(Default) anymore I think.

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
...
+            fn msg_type(&self) -> i32 {
+                2
+            }
+
+            fn size(&self) -> usize {
+                std::mem::size_of::<u64>()
+            }
+        }
+
+        impl SocketControlMessageExtManual for TestSocketControlMessage {
```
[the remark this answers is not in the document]
xanathar: Will test without this impl early tomorrow.

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{subclass::prelude::*, translate::*, Cast};
+
+use crate::SocketControlMessage;
+
+pub trait SocketControlMessageImpl: ObjectImpl + SocketControlMessageImplExt + Send {
```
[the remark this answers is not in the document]
sdroege: If it's `Send` then it also needs to be `Sync`.

But why remove the bound on the `Ext` trait?

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        pub struct TestSocketControlMessage;
+
+        impl TestSocketControlMessage {}
+        impl ObjectImpl for TestSocketControlMessage {}
+
+        #[glib::object_subclass]
+        impl ObjectSubclass for TestSocketControlMessage {
+            const NAME: &'static str = "TestSocketControlMessage";
+            const ABSTRACT: bool = false;
```
sdroege, line 133: That's already the default and you can omit it

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
...
+        self.parent_level()
+    }
+
+    fn msg_type(&self) -> i32 {
+        self.parent_msg_type()
+    }
+
+    fn size(&self) -> usize {
+        self.parent_size()
+    }
```
sdroege, line 18: Without implementing `serialize` and `deserialize` this is all rather useless or not?

gio/src/subclass/socket_control_message.rs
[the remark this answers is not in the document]
sdroege: Yes it is not needed but you need to put `serialize` / `deserialize` into the `Impl` trait instead

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    #[test]
+    fn test_socket_control_message_subclassing() {
+        let stream = glib::Object::new::<TestSocketControlMessage>();
+
+        assert_eq!(stream.level(), 1);
+        assert_eq!(stream.msg_type(), 2);
+        assert_eq!(stream.size(), std::mem::size_of::<u64>());
+
+        let through_abstract: &dyn SocketControlMessageExt = &stream;
```
sdroege, line 176: This part doesn't make much sense :)

gio/src/subclass/socket_control_message.rs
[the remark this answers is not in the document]
sdroege: Yes but you won't be able to call the `parent_XXX()` functions from the default impls if you don't have it as trait bound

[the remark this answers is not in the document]
sdroege: Then the compiler figures it out itself somehow apparently. Having the circle closed here seems like a good idea nonetheless.

gio/src/subclass/socket_control_message.rs
xanathar: For some cases `serialize` is enough (e.g. to pass a `SO_TXTIME` to a socket) but I agree that a more complete implementation makes more sense.

gio/src/subclass/socket_control_message.rs
xanathar: Removed (you're right - my misunderstanding on how the dispatch worked there).

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    fn parent_deserialize(level: i32, type_: i32, data: &[u8]) -> Option<SocketControlMessage> {
+        unsafe {
+            let type_data = T::type_data();
+            let parent_class =
+                type_data.as_ref().parent_class() as *mut ffi::GSocketControlMessageClass;
+            let f = (*parent_class)
+                .deserialize
+                .expect("No parent class implementation for \"deserialize\"");
```
sdroege, line 116: Should this be optional and return `None` if not implemented?

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,266 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{subclass::prelude::*, translate::*, Cast};
+
+use crate::SocketControlMessage;
+
+pub trait SocketControlMessageImpl: ObjectImpl + Send {
```
sdroege, line 7:
```suggestion
pub trait SocketControlMessageImpl: ObjectImpl + SocketControlMessageImplExt + Send + Sync {
```

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,266 @@
...
+            fn msg_type(&self) -> i32 {
+                std::i32::MAX
+            }
+
+            fn size(&self) -> usize {
+                size_of::<u64>()
+            }
+
+            fn serialize(&self, data: &mut [u8]) {
+                data.clone_from_slice(&self.0.get().to_ne_bytes());
```
sdroege, line 224:
```suggestion
                data.copy_from_slice(&self.0.get().to_ne_bytes());
```

gio/src/subclass/socket_control_message.rs
```diff
@@ -0,0 +1,266 @@
...
+            }
+
+            fn deserialize(level: i32, type_: i32, data: &[u8]) -> Option<SocketControlMessage> {
+                if level == std::i32::MAX && type_ == std::i32::MAX {
+                    let obj = glib::Object::new::<super::TestSocketControlMessage>();
+                    obj.imp()
+                        .0
+                        .set(u64::from_ne_bytes(data.try_into().unwrap()));
```
sdroege, line 232:
```suggestion
                        .set(u64::from_ne_bytes(data.try_into().ok()?));
```

gio/src/subclass/socket_control_message.rs
xanathar: Ok for the `Sync` (I mean, I need to think about it, but I trust your judgement that is needed :) )... but on `SocketControlMessageImplExt` this is the opposite of what Bilal suggested yesterday here: https://github.com/gtk-rs/gtk-rs-core/pull/1039#discussion_r1123602116

sdroege: * For `Sync`, either you need `Send` and `Sync`, or you need neither :) Only `Send` is not possible with `GObject`s
* For the `SocketControlMessageImplExt` see my comments and also all (?) the other `Impl` traits in gtk-rs-core

xanathar: As agreed offline, removing `Send` and `Sync` entirely.

sdroege (approved):

PR #1044, opened 2023-03-06
glib: impl StaticType, FromValue, ToValue, HasParamSpec for Box<str>

SeaDve:
depends on https://github.com/gtk-rs/gtk-rs-core/pull/1043 (Of which commits are also included here)

### DIFF
--- a/glib-macros/tests/value_delegate_derive.rs
+++ b/glib-macros/tests/value_delegate_derive.rs
@@ -6,6 +6,10 @@ fn higher_level_types() {
     pub struct MyVec(Vec<String>);
 
     #[derive(Debug, glib::ValueDelegate)]
+    #[value_delegate(nullable)]
+    pub struct MyString(Box<str>);
+
+    #[derive(Debug, glib::ValueDelegate)]
     #[value_delegate(from = Option<String>)]
     struct MyVecManualFrom(Vec<String>);
 
@@ -23,7 +27,6 @@ fn higher_level_types() {
     let vec = vec!["foo".to_string(), "bar".to_string()];
     let vec_value = vec.to_value();
     let my_vec_value = MyVec(vec.clone()).to_value();
-
     assert_eq!(MyVec::static_type(), Vec::<String>::static_type());
     assert_eq!(
         vec_value.get::<Vec<String>>().unwrap(),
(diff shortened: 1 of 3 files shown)


glib-macros/tests/value_delegate_derive.rs
```diff
@@ -1,9 +1,10 @@
-use glib::{value::FromValue, HasParamSpec, StaticType, ToValue};
+use glib::{value::FromValue, StaticType, ToValue};
 
 #[test]
 fn higher_level_types() {
     #[derive(Debug, glib::ValueDelegate)]
-    pub struct MyVec(Vec<String>);
```
sdroege: Would be useful to have both tested or not?

sdroege (commented):
Looks good to me otherwise

glib-macros/tests/value_delegate_derive.rs
SeaDve: Both nullable and non, or both MyString and MyVec?

sdroege: `MyString` and `MyVec`

ranfdev: Yeah, let's keep both. They are good test cases for the ValueDelegate macro, which only had tests in the form of the example documentation.

glib/src/value.rs
```diff
@@ -904,6 +904,49 @@ impl ToValueOptional for String {
...
+    }
+}
+
+impl ToValue for Box<str> {
+    fn to_value(&self) -> Value {
+        <&str>::to_value(&self.as_ref())
+    }
+
+    fn value_type(&self) -> Type {
+        String::static_type()
```
ranfdev, line 933: Why not self.value_type()?

SeaDve: I simply copied it from `String` implementation of `value_type`, and it seems that the implementation of `str` of `value_type` also calls into `String::static_type()`.

glib-macros/tests/value_delegate_derive.rs
SeaDve: Added both `MyVec` and `MyString`

sdroege (approved):

sdroege (approved):

PR #1045, opened 2023-03-06
glib: Make WeakRef cheaply cloneable

heftig:
By using an `Arc` instead of a `Box`, we can make `WeakRef::clone` a cheap operation, compared to taking a global write lock for setting a `GWeakRef`. The extra space used for the refcounts should be a fine trade-off.


heftig:
Using `weakref.set(&obj)` will now affect not just `weakref` but all of its clones, which is probably surprising.

glib/src/object.rs
```diff
@@ -3461,27 +3468,27 @@ impl<T: ObjectType> Default for WeakRef<T> {
     }
 }
 
-unsafe impl<T: ObjectType + Sync + Sync> Sync for WeakRef<T> {}
```
sdroege, line 3464: Oops. Can you make a separate PR for this?

Issue #1047, opened 2023-03-07
[FEATURE REQUEST] FromValue::from_value should return a Result


ranfdev:
I want to add support for `TryFrom` in the ValueDelegate macro

Eg, I want to do this:
```rust
#[derive(ValueDelegate)]
#[value_delegate(try_from = u32)]
struct MyU16(u16);
```

Clearly, the conversion from a u32 to a u16 could fail. 
`FromValue::from_value` should return an error, else I can't return the error from the call of `TryFrom::try_from`.

Currently `FromValue` has an associated `Checker` which checks the type in the glib type system and throws an error when the conversion is not possible. A Checker can't be made to handle `TryFrom` though, because that would require calling `try_from` inside `FromValue::check` and `FromValue::from_value`, which is a waste of resources.

PR #1048, opened 2023-03-07
Try from delegate

ranfdev:
Currently if the conversion fails, the code will just panic. This isn't any different from `str::from_value`, which panics on invalid utf-8.
related #1047


sdroege:
I think this should come with some refactoring of `FromValue` so that you can a) return the `TryFrom::Error` here, and b) don't need to call `TryFrom` twice for checking and extraction.

PR #1055, opened 2023-03-14
glib-macros: generate "From<Ident> for Value" on ValueDelegate

SeaDve:
This allows `Ident` (not only `&Ident`) to be allowed on functions that take in `Into<Value>`.


sdroege (dismissed):

sdroege (approved):

SeaDve:
I added some more test cases on `nullable`. I'm also sorry for staling your review again

bilelmoussaoui (approved):

PR #1065, opened 2023-03-22
glib-macros: enable default features of syn

decathorpe:
The code in this crate uses features of "syn" which are only present
when the default features are enabled. This previously worked "by
accident" because other crates in the dependency tree pulled in the
"default" feature of syn, but these crates have started moving to
syn v2, so this is no longer the case.

### DIFF
--- a/glib-macros/Cargo.toml
+++ b/glib-macros/Cargo.toml
@@ -17,7 +17,7 @@ heck = "0.4"
 proc-macro-error = "1.0"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", features = ["full"], default-features = false }
+syn = { version = "1.0", features = ["full"] }
 proc-macro-crate = "1.0"
 
 [lib]


bilelmoussaoui (approved):

ralismark:
Can this be backported to previous version branches? I've got a project that is still on version 0.15 of the gtk-rs libraries because we're still using the v0_22 APIs.

decathorpe:
(I can confirm, this issue also affects the 0.16 and 0.15 branches of glib-macros, gtk3-macros, and the 0.5 and 0.4 branches of gtk4-macros. I already backported the fix to the Fedora Linux packages for these older crate versions, but having it upstream would be great for developers who don't build on Fedora 😉 )

decathorpe:
Thanks!

Side note: This also happens in projects that don't depend on syn only indirectly. The projects that depend on old versions of gtk-rs are starting to be affected just because crates like serde, anyhow, thiserror, etc. have started to move to syn v2, not because the projects themselves moved to syn v2.

PR #1066, opened 2023-03-24
glib: Replace Continue with ControlFlow

### LINKED ISSUE #1063: Replace glib::Continue with ControlFlow
Sadly, we can't make use of `std::ops::ControlFlow` as it requires a type param for the `Break` variant. Although, we can replace `glib::Continue` with `glib::ControlFlow` that would follow the semantics of `std::ops::ControlFlow`

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
pub enum ControlFlow {
    Continue,
    Break,
}

impl From<std::ops::ControlFlow<()>> for ControlFlow {
    fn from(c: std::ops::ControlFlow<()>) -> Self {
        match c {
            std::ops::ControlFlow::Break(_) => Self::Break,
            std::ops::ControlFlow::Continue(_) => Self::Continue,
        }
    }
}

impl From<ControlFlow> for std::ops::ControlFlow<()> {
    fn from(c: ControlFlow) -> Self {
        match c {
            ControlFlow::Break => Self::Break(()),
            ControlFlow::Continue => Self::Continue(()),
        }
    }
}

impl From<bool> for ControlFlow {
    fn from(c: bool) -> Self {
        if c {
            Self::Continue
        } else  {
            Self::Break
        }
    }
}

impl From<ControlFlow> for bool {
    fn from(c: ControlFlow) -> Self {
        match c {
            ControlFlow::Break => false,
            ControlFlow::Continue => true,
        }
    }
}

#[doc(hidden)]
impl IntoGlib for ControlFlow {
    type GlibType = gboolean;

    #[inline]
    fn into_glib(self) -> gboolean {
        bool::from(self).into_glib()
    }
}
```

Note that changes on https://github.com/gtk-rs/gir are needed to make the code generator uses `glib::ControlFlow` instead of `glib::Continue` is needed as well.

bilelmoussaoui:
Fixes #1063

### DIFF
--- a/gio/src/datagram_based.rs
+++ b/gio/src/datagram_based.rs
@@ -23,12 +23,12 @@ pub trait DatagramBasedExtManual: sealed::Sealed + IsA<DatagramBased> + Sized {
         func: F,
     ) -> glib::Source
     where
-        F: FnMut(&Self, glib::IOCondition) -> glib::Continue + 'static,
+        F: FnMut(&Self, glib::IOCondition) -> glib::ControlFlow + 'static,
         C: IsA<Cancellable>,
     {
         unsafe extern "C" fn trampoline<
             O: IsA<DatagramBased>,
-            F: FnMut(&O, glib::IOCondition) -> glib::Continue + 'static,
+            F: FnMut(&O, glib::IOCondition) -> glib::ControlFlow + 'static,
         >(
             datagram_based: *mut ffi::GDatagramBased,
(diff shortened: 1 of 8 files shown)


sdroege (approved):
Looks good to me. Can you update all the repos (incl gtk3-rs and gstreamer-rs / gst-plugins-rs) and then we'll merge it?

sdroege (approved):

PR #1068, opened 2023-04-01
glib-macros: Specify quoted types explicitly

felinira:
Having a custom eg. result type in scope will fail compilation, as the types don't match. Specify quoted types explicitly to not rely on the prelude.

### DIFF
--- a/glib-macros/src/error_domain_derive.rs
+++ b/glib-macros/src/error_domain_derive.rs
@@ -35,7 +35,7 @@ pub fn impl_error_domain(input: &syn::DeriveInput) -> TokenStream {
 
                 static QUARK: #crate_ident::once_cell::sync::Lazy<#crate_ident::Quark> =
                     #crate_ident::once_cell::sync::Lazy::new(|| unsafe {
-                        from_glib(#crate_ident::ffi::g_quark_from_static_string(concat!(#domain_name, "\0") as *const str as *const _))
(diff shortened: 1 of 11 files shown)


sdroege:
These are the only two places where this is needed in the macros?

sdroege:
That would be great, thanks. I would be surprised if these were the only places :)

To be even more correct, I think you'd do `::std::...` instead of `std::...`

bilelmoussaoui (dismissed):

glib-macros/src/error_domain_derive.rs
```diff
@@ -48,7 +48,7 @@ pub fn impl_error_domain(input: &syn::DeriveInput) -> TokenStream {
             #[inline]
             fn from(value: i32) -> ::core::option::Option<Self>
             where
-                Self: Sized
+                Self: std::marker::Sized
```
sdroege, line 51:
```suggestion
                Self: ::std::marker::Sized
```

maybe?

sdroege (commented):
Looks good to me otherwise, thanks :)

sdroege (approved):

sdroege:
Thanks a lot :) I'll do a new bugfix release with this on Monday or Tuesday.

sdroege:
Done, this is in 0.17.7 now

PR #1070, opened 2023-04-05
RFC: glib: Add optional support for serialization and deserialization with serde

rmnscnce:
If merged, some GLib types will gain serialization and deserialization support through serde. This features is gated as `serde`.

Supported types for both deserialization and serialization:
- `glib::Bytes`
- `glib::GString`

Some types are only serializable:
- `glib::ByteArray`
- `glib::GStr`
- `glib::GStrV`

Collection types are also supported as long as the type parameters implement the necessary traits:
- `glib::PtrSlice<T: TransparentPtrType + _>`
- `glib::Slice<T: TransparentType + _>`
- `glib::List<T: TransparentPtrType + _>`
- `glib::SList<T: TransparentPtrType + _>`


sdroege:
I like this and we have something similar in place in the GStreamer bindings already. You probably want to mention this new feature in the README.md though.

Also do you see a possibility of moving all that repetitive code into a macro or two? :)

GuillaumeGomez:
Looks like a good idea.

glib/Cargo.toml
```diff
@@ -59,6 +60,7 @@ log_macros = ["log"]
 dox = ["ffi/dox", "gobject_ffi/dox", "log_macros"]
 compiletests = []
 gio = ["gio_ffi"]
+serialize = ["dep:serde"]
```
sdroege, line 63: Why `serialize` instead of `serde`?

glib/src/serde.rs
```diff
@@ -0,0 +1,248 @@
...
+
+mod bytes {
+    use crate::Bytes;
+
+    serialize_impl!(Bytes, Bytes(b) => b);
+
+    deserialize_impl!(
+        Bytes,
+        Bytes(seq) => {
+            let mut byte_vec = vec![];
```
sdroege, line 147: You could use the size hint here to reserve initial capacity

glib/src/serde.rs
```diff
@@ -0,0 +1,248 @@
...
+
+            while let Some(_size @ 1..) = seq.size_hint() {
```
sdroege, line 149: Why the outer loop?

glib/Cargo.toml
rmnscnce: I wanted to use `serde` but seems like the tests break on it.

https://github.com/gtk-rs/gtk-rs-core/actions/runs/4616156154/jobs/8160815190#step:19:414
https://github.com/gtk-rs/gtk-rs-core/tree/5e2d7622885fd64cfbf7a727273d78368cdc139e

glib/src/serde.rs
rmnscnce: Would this otherwise be acceptable?

```rs
let mut byte_vec = match seq.size_hint() {
	Some(size) => Vec::with_capacity(size),
	None => vec![],
};
```

glib/Cargo.toml
sdroege: Hrm that looks like a bug in `trybuild2`. @GuillaumeGomez you're maintaining that, can you take a look? :)

glib/src/serde.rs
sdroege: I think so. I would do the same as the deserialize impl for `Vec` is doing inside serde.

rmnscnce: Serde initializes `Vec<_>` with `Vec::with_capacity(std::cmp::min(size_hint.unwrap_or(0), 4096))`. Should this be done as well to `Slice<_>` and `PtrSlice<_>`? They can be instantiated with an initial capacity as well.

glib/Cargo.toml
GuillaumeGomez: Sure.

GuillaumeGomez: I published a new version of `trybuild2` with the commits of the upstream `trybuild`.

glib/src/lib.rs
```diff
@@ -213,6 +213,9 @@ pub use self::thread_pool::{ThreadHandle, ThreadPool};
 
 pub mod thread_guard;
 
+#[cfg(feature = "serialize")]
```
bilelmoussaoui, line 216:
```suggestion
#[cfg(feature = "serde")]
```

glib/src/serde.rs
```diff
@@ -0,0 +1,401 @@
...
+
+                fn expecting(&self, formatter: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+                    formatter.write_str("an array of bytes")
+                }
+
+                fn visit_str<E>(self, _: &str) -> Result<Self::Value, E>
+                where
+                    E: ::serde::de::Error,
+                {
+                    unimplemented!()
```
sdroege, line 93: Why is this not implemented?

glib/src/serde.rs
```diff
@@ -0,0 +1,401 @@
...
+                        formatter: &mut ::std::fmt::Formatter,
+                    ) -> ::std::fmt::Result {
+                        formatter.write_str("a valid UTF-8 string")
+                    }
+
+                    fn visit_str<E>(self, _: &str) -> Result<Self::Value, E>
+                    where
+                        E: ::serde::de::Error,
+                    {
+                        unimplemented!()
```
sdroege, line 196: Also this

sdroege (commented):
Can you also add something to the CI so that it builds with and without the `serde` feature so we can make sure things continue to work?

.github/workflows/CI.yml
```diff
@@ -82,9 +82,15 @@ jobs:
       - name: build
         run: cargo build --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "${{ matrix.conf.features }}"
         if: matrix.rust != 'nightly'
+      - name: build-serde
```
bilelmoussaoui, line 85: i would instead add the serde feature to https://github.com/gtk-rs/gtk-rs-core/pull/1070/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R26, same for the windows ci file

.github/workflows/CI.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           - { name: "cairo", features: "png,pdf,svg,ps,use_glib,v1_18,freetype,script,xcb,xlib,win32-surface", nightly: "--features 'png,pdf,svg,ps,use_glib,v1_18,freetype,script,xcb,xlib,win32-surface'", test_sys: true }
           - { name: "gdk-pixbuf", features: "v2_42", nightly: "--all-features", test_sys: true }
           - { name: "gio", features: "v2_74", nightly: "--all-features", test_sys: true }
-          - { name: "glib", features: "v2_74", nightly: "--all-features", test_sys: true }
+          - { name: "glib", features: "v2_74,serde", nightly: "--all-features", test_sys: true }
```
sdroege, line 26:
```suggestion
          - { name: "glib", features: "v2_74", nightly: "--all-features", test_sys: true }
          - { name: "glib serde", features: "v2_74,serde", nightly: "--all-features", test_sys: true }
```

or so maybe so we test both configurations?

sdroege (dismissed):
Looks good to me otherwise

PR #1071, opened 2023-04-06
glib: Allow using `Path` / `PathBuf` in `glib::Value`s

sdroege:
There's no separate GType for that unfortunately and it's up to the caller to select whether this is an actual `String` or a `Path`. If getting this wrong it will fail the same way it would now already fail for strings.


bilelmoussaoui (approved):
Thanks!

PR #1081, opened 2023-04-13
glib-macros: Strip out r# prefix from property names inside the GObject

rmnscnce:
If merged, the `r#` prefix will be removed from names of the property, allowing the expected names to be used.

This also adds a section to the `glib::Properties` docs elaborating the use of keywords as property names with the macro.

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -11,6 +11,7 @@ use syn::parenthesized;
 use syn::parse::Parse;
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
+use syn::LitStr;
 use syn::Token;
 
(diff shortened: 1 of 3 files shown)


sdroege:
Can you get rid of the merge commit?

glib-macros/src/properties.rs
```diff
@@ -556,10 +569,14 @@ fn expand_wrapper_connect_prop_notify(props: &[PropDesc]) -> TokenStream2 {
     let crate_ident = crate_ident_new();
     let connection_fns = props.iter().map(|p| {
         let name = &p.name;
+        let stripped_name = syn::LitStr::new(
```
sdroege, line 572: Maybe this can go into a function

sdroege (approved):

PR #1083, opened 2023-04-17
glib: Enable various smallvec features

sdroege:
This all require a newer rustc than what smallvec depends on, but much older than glib's minimum supported Rust version.

Enabling these features enables smallvec to use less space, work with any array size and provide some more const functions.

----

Another reason why conservative MSRV are the plague.

### DIFF
--- a/glib/Cargo.toml
+++ b/glib/Cargo.toml
@@ -31,7 +31,7 @@ ffi = { package = "glib-sys", path = "sys" }
 gobject_ffi = { package = "gobject-sys", path = "gobject-sys" }
 glib-macros = { path = "../glib-macros" }
 rs-log = { package = "log", version = "0.4", optional = true }
-smallvec = "1.0"
+smallvec = { version = "1.0", features = ["union", "const_generics", "const_new"] }
 thiserror = "1"
 gio_ffi = { package = "gio-sys", path = "../gio/sys", optional = true }
 memchr = "2.5.0"


GuillaumeGomez (approved):

GuillaumeGomez:
> Another reason why conservative MSRV are the plague.

Not sure what it has to do with the rest though.

sdroege:
> Not sure what it has to do with the rest though.

Especially the union feature makes a quite big difference and should be enabled by default. That can't be done though if you want to support rustc from 3 years ago. And your software depending indirectly on smallvec will have no way to enable it explictly unless you depend directly on smallvec (even if you don't use it), assuming you even notice that there is this configuration somewhere in one of your dependencies.

GuillaumeGomez:
Yes but it doesn't concern us so I'm really not sure to understand your point.

sdroege:
In this case we're directly depending on it, so we "only" need to be aware that this configuration exists in our direct dependency. Which nobody did for 2 years, or did you? :)

GuillaumeGomez:
Ok I just got it: you said it because you realized that these features that improved this crate greatly were disabled by default because of MSRV. Now it makes sense.

sdroege:
Sorry, I should've been more clear :) I had no breakfast yet!

GuillaumeGomez:
Go eat young man!

PR #1089, opened 2023-05-03
properties: Allow to omit set for construct_only properties

### LINKED ISSUE #1088: property macro: Support construct_only
I have a bunch of

```rust
#[property(get, set, builder().construct_only())]
```

in my code. I don't really need the setter since it will always fail and the `builder().` doesn't look to too pretty. Would it be possible to have something like the following?

```rust
#[property(get, construct_only)]
```

felinira:
This allows to omit `set` for `construct_only` properties, as it is confusing from a semantic perspective, because we don't actually generate a public setter. It was required before, but didn't actually throw an error but panic at runtime.

Updates the docs about how to use `construct_only` and chooses a better example for the `builder()` attribute.

Closes #1088

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -851,7 +851,7 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 
 /// This macro enables you to derive object properties in a quick way.
 ///
-/// # Supported #[property] attributes
+/// # Supported `#[property]` attributes
 /// | Attribute | Description | Default | Example |
 /// | --- | --- | --- | --- |
 /// | `name = "literal"` | The name of the property | field ident where `_` (leading and trailing `_` are trimmed) is replaced into `-` | `#[property(name = "prop-name")]` |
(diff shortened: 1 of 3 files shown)


ranfdev:
It's still a bit confusing when setting a custom setter, but It's ok. When I'm on the PC I'll propose some small code changes.

glib-macros/tests/properties.rs
```diff
@@ -160,6 +160,10 @@ mod foo {
             weak_ref_prop: glib::WeakRef<glib::Object>,
             #[property(get, set)]
             send_weak_ref_prop: glib::SendWeakRef<glib::Object>,
+            #[property(get, default_value = 0, construct_only)]
```
bilelmoussaoui, line 163: How would a default value work with construct only ?

pbor: default actually *only* works for construct and construct_only

glib-macros/src/lib.rs
```diff
@@ -862,7 +862,8 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 /// | `override_interface = expr` | The type of interface of which to override the property from | | `#[property(override_interface = SomeInterface)]` |
 /// | `nullable` | Whether to use `Option<T>` in the wrapper's generated setter |  | `#[property(nullable)]` |
 /// | `member = ident` | Field of the nested type where property is retrieved and set | | `#[property(member = author)]` |
-/// | `builder(<required-params>)[.ident]*` | Used to input required params or add optional Param Spec builder fields | | `#[property(builder(SomeEnum::default()))]`, `#[builder().default_value(1).construct_only()]`, etc.  |
+/// | `construct_only` | Specify that the property is construct only. This will not generate a public setter and only allow the property to be set during object construction. The use of a custom setter is supported. | | `#[property(get, construct_only)]` or `#[property(get, set = set_prop, construct_only)]` |
```
ranfdev, line 865:
```suggestion
/// | `construct_only` | Specify that the property is construct only. This will not generate a public setter and only allow the property to be set during object construction. The use of a custom internal setter is supported. | | `#[property(get, construct_only)]` or `#[property(get, set = set_prop, construct_only)]` |
```

glib-macros/src/properties.rs
```diff
@@ -258,6 +259,17 @@ impl PropDesc {
             builder_fields,
         } = attrs;
 
+        let is_construct_only = if builder_fields.iter().any(|(k, _)| *k == "construct_only") {
```
ranfdev, line 262: I would define is_construct_only without the `if`, and move the if below:
```
let is_construct_only = builder_fields.iter().any(|(k, _)| *k == "construct_only");
// Insert a default setter automatically
if is_construct_only && set.is_none() {
  set = Some(MaybeCustomFn::Default);
}
```

ranfdev (changes requested):

ranfdev (approved):

bilelmoussaoui (approved):

PR #1091, opened 2023-05-04
Regenerate with latest gir

bilelmoussaoui:


gio/src/dbus_proxy.rs
```diff
@@ -6,25 +6,9 @@ use glib::{prelude::*, signal::connect_raw, translate::*, SignalHandlerId};
 
 use crate::DBusProxy;
 
-pub trait DBusProxyExtManual: 'static {
+pub trait DBusProxyExtManual: IsA<DBusProxy> + 'static {
     #[cfg(feature = "v2_72")]
     #[doc(alias = "g-signal")]
```
bilelmoussaoui, line 11: this was very badly handled by the automatic script, needs fixing

gdk-pixbuf/src/auto/pixbuf.rs
```diff
@@ -660,7 +660,7 @@ impl Pixbuf {
         unsafe { FromGlibPtrContainer::from_glib_container(ffi::gdk_pixbuf_get_formats()) }
     }
 
-    #[cfg(any(feature = "v2_40"))]
```
GuillaumeGomez, line 663: That should be a clippy lint.

gdk-pixbuf/src/auto/pixbuf_animation.rs
```diff
@@ -137,28 +137,15 @@ impl PixbufAnimation {
     }
 }
 
-pub trait PixbufAnimationExt: 'static {
```
sdroege, line 140: Before merging this we should make sure that it's impossible for external code to implement these traits on anything. Otherwise we need to mark the trait unsafe or implement one of the many sealing mechanisms.

bilelmoussaoui: I have opened https://github.com/gtk-rs/gtk-rs-core/issues/1093 to keep track of this task

jf2048: I think that is already happening as it requires `IsA<T>` and that is unsafe?

bilelmoussaoui: Could we merge this now?

Issue #1090, opened 2023-05-04
docs no longer display the required features


bilelmoussaoui:
See https://gtk-rs.org/gtk4-rs/git/docs/gtk4/subclass/symbolic_paintable/index.html and https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/subclass/symbolic_paintable/index.html

cc @AaronErhardt

AaronErhardt:
I'll look into it.

bilelmoussaoui:
thanks for the quick fix @AaronErhardt

PR #1092, opened 2023-05-04
Fix required features not shown in docs

AaronErhardt:
Fix for https://github.com/gtk-rs/gtk-rs-core/issues/1090 on the gtk-rs-core side

It seems like you need both RUSTFLAGS and RUSTDOCFLAGS because `build.rs` doesn't care about the doc flags and cargo only passes RUSTDOCFLAGS to rustdoc...


bilelmoussaoui:
Tested with gtk4-rs, it works as expected! thanks a lot

PR #1095, opened 2023-05-08
glib: prelude cleanup

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gir/pull/1469

### DIFF
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -21,20 +21,14 @@ pub use self::{
     closure::{Closure, RustClosure},
     enums::{EnumClass, EnumValue, FlagsBuilder, FlagsClass, FlagsValue, UserDirectory},
     error::{BoolError, Error},
-    object::{
-        BorrowedObject, Cast, CastNone, Class, InitiallyUnowned, Interface, IsA, Object,
-        ObjectClassExt, ObjectExt, ObjectType, SendWeakRef, WeakRef,
-    },
+    object::{BorrowedObject, Class, InitiallyUnowned, Interface, Object, SendWeakRef, WeakRef},
     signal::{
         signal_handler_block, signal_handler_disconnect, signal_handler_unblock,
         signal_stop_emission_by_name, Propagation, SignalHandlerId,
     },
-    types::{ILong, Pointer, StaticType, StaticTypeExt, Type, ULong},
-    value::{BoxedValue, SendValue, ToSendValue, ToValue, Value},
-    variant::{
-        FixedSizeVariantArray, FixedSizeVariantType, FromVariant, StaticVariantType, ToVariant,
-        Variant,
-    },
+    types::{ILong, Pointer, Type, ULong},
+    value::{BoxedValue, SendValue, Value},
+    variant::{FixedSizeVariantArray, Variant},
     variant_dict::VariantDict,
     variant_iter::{VariantIter, VariantStrIter},
     variant_type::{VariantTy, VariantTyIterator, VariantType},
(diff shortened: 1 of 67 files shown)


glib/src/prelude.rs
```diff
@@ -4,6 +4,16 @@
...
+    error::ErrorDomain,
+    gstring::{IntoGStr, IntoOptionalGStr},
+    object::{Cast, CastNone, IsA, ObjectExt, ObjectType},
+    param_spec::{HasParamSpec, ParamSpecBuilderExt, ParamSpecType},
+    types::{StaticType, StaticTypeExt},
+    value::{
+        ToSendValue, ToValue, ToValueOptional, ValueType, ValueTypeChecker, ValueTypeOptional,
+    },
+    variant::{FixedSizeVariantType, FromVariant, StaticVariantType, ToVariant},
+    Continue,
```
bilelmoussaoui, line 18: Continue should definitely be removed from here. But I would do it once #1063 I work on

glib/src/prelude.rs
```diff
@@ -4,6 +4,16 @@
...
-    ParamSpecType, StaticType, StaticTypeExt, StaticVariantType, ToSendValue, ToValue, ToVariant,
+    closure::{IntoClosureReturnValue, TryFromClosureReturnValue},
+    collections::{ptr_slice::IntoPtrSlice, strv::IntoStrV},
+    error::ErrorDomain,
+    gstring::{IntoGStr, IntoOptionalGStr},
+    object::{Cast, CastNone, IsA, ObjectExt, ObjectType},
+    param_spec::{HasParamSpec, ParamSpecBuilderExt, ParamSpecType},
+    types::{StaticType, StaticTypeExt},
+    value::{
+        ToSendValue, ToValue, ToValueOptional, ValueType, ValueTypeChecker, ValueTypeOptional,
```
sdroege, line 15: `ToValueOption` / `ValueTypeOptional` should be unnecessary here. Probably also others, I'd have to check each in detail.

Generally, the idea is that the prelude only contains the traits of which users would call methods directly (so they can call them without importing the trait explicitly). Exception for this are the FFI traits (so `IntoStrV` etc need some thinking).

Other traits that are only used in trait bounds (exception `IsA`) are not part of the prelude.

bilelmoussaoui: > Generally, the idea is that the prelude only contains the traits of which users would call methods directly (so they can call them without importing the trait explicitly). Exception for this are the FFI traits (so IntoStrV etc need some thinking).

I am sure i have to make use of  `ToValue` somewhere. Probably before the enum/flags paramspec builder taking a default value param.

For IntoStrV and other ffi traits, should we re-export them in translate then? That would make things much cleaner then exporting everything in the main crate lib.rs and ending up with a very very long list of types/traits/functions

sdroege: > I am sure i have to make use of `ToValue` somewhere.

`ToValue` yes but not `ToValueOptional`. `ToValue` should stay in the prelude

> For IntoStrV and other ffi traits, should we re-export them in translate then

Yes that's what I was thinking

bilelmoussaoui: Cleaned things a little bit, should be good for a review now

glib/src/lib.rs
```diff
@@ -104,9 +98,7 @@ mod exit_code;
 pub use exit_code::ExitCode;
 
 pub mod collections;
-pub use collections::{
-    ptr_slice::IntoPtrSlice, strv::IntoStrV, List, PtrSlice, SList, Slice, StrV,
```
sdroege, line 108: You're not adding these two traits to the prelude. Any reason why?

Are there other traits (outside `glib::translate`) that are not in the prelude?

sdroege (approved):
Seems fine to me otherwise

glib/src/lib.rs
bilelmoussaoui: Because they are specific to ffi usage so I re-exported them for translate instead.

There are various traits like FromValue that comes to mind. We are over exporting things in the main scope, I think we should reduce that going forward and just have people import things when they need them

PR #1096, opened 2023-05-09
Update syn: v2.0

### LINKED ISSUE #1061: glib-macros: Port to syn 2.0

ranfdev:
closes #1061 

**Problem**
syn 2.0 removed the type `NestedMeta`, that was extensively used by gtk-rs-core to parse macro attributes, without providing a direct replacement type.
Instead, the suggested way to parse an attribute is now to use [`Attribute::parse_nested_meta`](https://docs.rs/syn/latest/syn/struct.Attribute.html#method.parse_nested_meta), which iterates over the arguments of an attribute and forces you to parse them. `parse_nested_meta` tracks the spans of the arguments being parsed, so in the case an error is returned, it provides good error messages.

Most of the glib-macros depended on utility functions defined in `glib-macros::utils`. Those functions often tried to retrieve a single argument inside an attribute, without caring about the other arguments. This is not possible when using `Attribute::parse_nested_meta`, because it requires to handle _every possible argument_ in a single place. We can't just parse the argument `name = "value"` and skip  `nick = "nick"`, because `parse_nested_meta` will consider `nick` as an unexpected value and will throw an error.

**Proposed solution**
Rewriting the parsing of attributes taking advantage of `parse_nested_meta`, which is now the suggested way to parse attribute arguments (also see https://docs.rs/syn/latest/syn/meta/struct.ParseNestedMeta.html). A new set of utilities have been implemented in `glib-macro::utils` to ease the parsing process.

**To do**
implement some tests for the new parsing utility functions, to polish their interface and ensure correctness.

**Observations**
I would consider using the crate https://github.com/jf2048/deluxe (by @jf2048, #970) to handle attribute parsing.

**Conclusions**
- Improvements from syn 1.0:
  - We are up to date!
  - the errors from the macro should now be improved, because `parse_nested_meta` keeps track of argument spans and ensures every argument is parsed correctly.
- Drawbacks: even though the tests are passing, rewriting the parsing system can easily introduce some bugs. I fear I've missed some corner cases.

### DIFF
--- a/glib-macros/src/enum_derive.rs
+++ b/glib-macros/src/enum_derive.rs
@@ -6,9 +6,7 @@ use proc_macro_error::abort_call_site;
 use quote::{quote, quote_spanned};
 use syn::{punctuated::Punctuated, spanned::Spanned, token::Comma, Data, Ident, Variant};
 
-use crate::utils::{
-    crate_ident_new, gen_enum_from_glib, parse_item_attributes, parse_name, ItemAttribute,
-};
+use crate::utils::{crate_ident_new, gen_enum_from_glib, parse_nested_meta_items, NestedMetaItem};
 
 // Generate glib::gobject_ffi::GEnumValue structs mapping the enum such as:
 //     glib::gobject_ffi::GEnumValue {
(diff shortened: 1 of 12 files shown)


glib-macros/src/enum_derive.rs
```diff
@@ -2,13 +2,11 @@
 
 use heck::{ToKebabCase, ToUpperCamelCase};
 use proc_macro2::TokenStream;
-use proc_macro_error::abort_call_site;
+use proc_macro_error::{abort, abort_call_site};
```
sdroege, line 5:
```
error: unused import: `abort`
 --> glib-macros/src/enum_derive.rs:5:24
  |
5 | use proc_macro_error::{abort, abort_call_site};
  |                        ^^^^^
  |
  = note: `-D unused-imports` implied by `-D warnings`
```

sdroege (approved):

glib-macros/src/enum_derive.rs
```diff
@@ -29,21 +27,14 @@ fn gen_enum_values(
...
-            match attr {
-                ItemAttribute::Name(n) => value_name = n,
-                ItemAttribute::Nick(n) => value_nick = n,
-            }
-        );
+        let mut name_attr = NestedMetaItem::<syn::LitStr>::new("name").value_required();
+        let mut nick = NestedMetaItem::<syn::LitStr>::new("nick").value_required();
+
+        // the attribute `enum_value` is optional, we don't need to check if we found it
+        let _ = parse_nested_meta_items(&v.attrs, "enum_value", &mut [&mut name_attr, &mut nick]);
```
ranfdev, line 34: Even if the attribute is optional, I have to handle the case when the attribute has been found but the parsing of it has failed.

I need a match statement here.

ranfdev (commented):
Ah, I've noticed I still need to change some little things, please don't merge yet.

sdroege (approved):

PR #1099, opened 2023-05-12
Update rust-version = 1.65

### LINKED ISSUE #1097: [FEATURE REQUEST] Update rust version
In the matrix chat we talked about doing it, to keep the same min version as gstreamer-rs (rust 1.66, see here https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/main/gstreamer-app/Cargo.toml)

ranfdev:
closes #1097 
So that we can use GATs (which could be useful for the Properties macro) and `let-else` statements.

Rust versions between `1.65..=1.69` don't introduce significant changes to justify an upgrade to an higher version.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.64.0"
+          - "1.65.0"
         conf:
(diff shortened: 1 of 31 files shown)


sdroege (dismissed):

ranfdev:
Ah, I have to update the CI testing bed as well

PR #1104, opened 2023-05-26
Bump the gir-format-check dev-dependency to 0.1.2

schopin-pro:
In the commit 272bba5f2f159f5c15e1f0fea6e491038f427cf6 the removal of the to_string() call is only valid if the Errors struct implements Display, which is available in gir-format-check 0.1.2 on.

Fixes: 272bba5f2f15 ("Fix various 1.58 (beta) clippy warnings or silence them")

We've been hitting this issue when packaging the various crates for Debian and Ubuntu.

### DIFF
--- a/gdk-pixbuf/Cargo.toml
+++ b/gdk-pixbuf/Cargo.toml
@@ -36,4 +36,4 @@ gio = { path = "../gio" }
 once_cell = "1"
 
 [dev-dependencies]
-gir-format-check = "^0.1"
+gir-format-check = "^0.1.2"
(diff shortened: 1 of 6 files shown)


bilelmoussaoui:
I don't think this PR makes much sense.

sdroege:
The version given in `Cargo.toml` is the lowest version that is known/expected to work. If you use `-Zdirect-minimal-versions` you're explicitly asking for that to be used.

For Linux distros, increasing this version without any reason would even potentially complicate things because now they first have to update to a newer version although the old version would work just fine. Which then unnecessarily connects updates with each other and complicates the whole procedure.

Similarly, for "enterprise settings" this only complicates things. As you say yourself, dependencies might be whitelisted explicitly and might need some manual procedure to get updated to a newer version.

----

What problem are you exactly trying to solve here?

PR #1105, opened 2023-05-26
graphene: fix alignment of graphene_simd4f_t on armhf and s390x

schopin-pro:
On systems with the armhf ABI as used by Debian and Ubuntu (amdv7 + VFP, but without NEON), graphene falls back to the scalar definition of graphene_simd4f_t, which has the same alignment as its base type, float.

On s390x, the underlying type is a generic float
__attribute__((vector_size(16)). Surprinsingly, the s390x ABI does not mandate alignment bigger than 8 for the vector types.

Without these patches, the layout tests fail on those architectures. The armhf case is notable because there are a fair number of Raspberry Pis out there that use that ABI, either because their SoC isn't 64-bit ready or just because they have low memory.

Ideally, we could specifiy this in the Gir.toml file, but it doesn't seem like the tool is flexible enough to express this kind of nuances,

### DIFF
--- a/graphene/sys/src/lib.rs
+++ b/graphene/sys/src/lib.rs
@@ -227,7 +227,9 @@ impl ::std::fmt::Debug for graphene_rect_t {
     }
 }
 
-#[repr(align(16))]
+#[cfg_attr(target_arch = "arm", repr(align(4)))]
+#[cfg_attr(target_arch = "s390x", repr(align(8)))]
+#[cfg_attr(not(any(target_arch = "arm", target_arch = "s390x")), repr(align(16)))]
 #[derive(Copy, Clone)]
 #[repr(C)]
 pub struct graphene_simd4f_t {


sdroege:
Based on https://github.com/gtk-rs/gtk-rs-core/issues/1107 there is also a problem on armel and i386 in addition.

The main problem here is that `src/lib.rs` is autogenerated, so you can't just change it like this. It would have to be marked as `manual = true` in the `Gir.toml` and then we could implement it manually in `src/manual.rs` with whatever complicated alignment configuration is correct here.

See e.g. `gio/sys` or `glib/sys` as an example for this. Can you update it accordingly?

sdroege:
From that issue it sounds like `graphene_simd4f_t` is only wrong for i386, and `graphene_simd4x4f_t` is wrong for everything else. But here you modify only `graphene_simd4f_t` for s390x which is probably unnecessary?

schopin-pro:
Sorry, this fell off my radar entirely. I'll try to get back into it
this week.

bilelmoussaoui:
@schopin-pro any updates on this one?

schopin-pro:
Still on my todolist, still haven't found the time to work on it. Feel free
to close out the PR though, since the final implementation will likely be
different anyway :)

bilelmoussaoui:
Closing then as it been more than a year without any updates. There is an issue tracking it already

PR #1112, opened 2023-05-31
Make FileInfo bytestring methods use CStrings

jf2048:
This is more correct for the optional attributes that store a path, i.e. `trash::orig-path`, `thumbnail::path`, etc.

For all current cases this should actually be a `Path`, but that should be handled upstream by a couple new methods: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3461

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -822,6 +822,14 @@ status = "generate"
     name = "set_attribute_stringv"
     # use strv
     manual = true
+    [[object.function]]
+    name = "get_attribute_byte_string"
+    # use CString
+    manual = true
+    [[object.function]]
+    name = "set_attribute_byte_string"
+    # use CStr
+    manual = true
 
 [[object]]
 name = "Gio.FilterOutputStream"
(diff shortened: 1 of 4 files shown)


gio/Gir.toml
```diff
@@ -822,6 +822,14 @@ status = "generate"
...
     # use strv
     manual = true
+    [[object.function]]
+    name = "get_attribute_byte_string"
+    # use CString
+    manual = true
+    [[object.function]]
+    name = "set_attribute_byte_string"
+    # use CStr
+    manual = true
```
sdroege, line 832: Hm we can't do `string_type = "c_string"`... that only handles all the other options. I wonder how that got forgotten :)

sdroege:
Shouldn't this be `OString` instead? Why `CString`?

jf2048:
See the comments from upstream, it seems "byte string" suggests this is not actually always a file path, even though in every case it currently is?

sdroege:
Should it be just `Vec<u8>` / `&[u8]` then maybe? C strings are always a bit strange to deal with in Rust...

PR #1113, opened 2023-06-02
Add a std_once_cell feature 

A6GibKm:
That adds std::cell::OnceCell and std::sync::OnceLock support for the properties macro.

### DIFF
--- a/glib/Cargo.toml
+++ b/glib/Cargo.toml
@@ -58,6 +58,7 @@ log = ["rs-log"]
 log_macros = ["log"]
 compiletests = []
 gio = ["gio_ffi"]
+std_once_cell = []
 
 [package.metadata.docs.rs]
 all-features = true
--- a/glib/src/property.rs
+++ b/glib/src/property.rs
@@ -47,6 +47,14 @@ impl<T: Property> Property for once_cell::sync::OnceCell<T> {
 impl<T: Property> Property for once_cell::unsync::OnceCell<T> {
     type Value = T::Value;
 }
+#[cfg(feature = "std_once_cell")]
+impl<T: Property> Property for std::cell::OnceCell<T> {
+    type Value = T::Value;
+}
+#[cfg(feature = "std_once_cell")]
+impl<T: Property> Property for std::sync::OnceLock<T> {
+    type Value = T::Value;
+}
 // Handle smart pointers trasparently
 impl<T: Property> Property for Rc<T> {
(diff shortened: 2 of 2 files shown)


ranfdev:
We lose the CI testing on the minimum supported version, rust 1.65. Are we sure we are ok with that?

A6GibKm:
Why is the last test failing?

jf2048:
The docs regen thing is unrelated to this

.github/workflows/CI.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.65.0"
+          - "1.70.0"
```
sdroege, line 21: > We lose the CI testing on the minimum supported version, rust 1.65. Are we sure we are ok with that?

No, this needs to be optional and the CI needs to be updated accordingly.

A6GibKm: >  CI needs to be updated accordingly.

What do toy mean by this?

A6GibKm: I removed in in the meantime from ci.

sdroege: You need to disable this specific feature for the 1.65 CI run. It would be good to enable it for the others though.

A6GibKm: Done.

.github/workflows/CI.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           - { name: "cairo", features: "png,pdf,svg,ps,use_glib,v1_18,freetype,script,xcb,xlib,win32-surface", nightly: "--features 'png,pdf,svg,ps,use_glib,v1_18,freetype,script,xcb,xlib,win32-surface'", test_sys: true }
           - { name: "gdk-pixbuf", features: "v2_42", nightly: "--all-features", test_sys: true }
           - { name: "gio", features: "v2_74", nightly: "--all-features", test_sys: true }
-          - { name: "glib", features: "v2_74", nightly: "--all-features", test_sys: true }
+          - { name: "glib", features: "v2_74", nightly: "--features=v2_76,log,log_macros,compiletests", test_sys: true }
```
sdroege, line 26: Well for nightly it's obviously fine to enable that feature. The way how it was before seemed good to me :) Sorry

I only didn't merge this yet because there was some CI failure and I had no time to look into that yet

PR #1114, opened 2023-06-03
strv: Implement From for constant GStr slices

A6GibKm:
This was implemented for GString and &str but not for &GStr.

### DIFF
--- a/glib/src/collections/strv.rs
+++ b/glib/src/collections/strv.rs
@@ -358,6 +358,21 @@ impl<'a, const N: usize> From<[&'a str; N]> for StrV {
     }
 }
 
+impl<'a, const N: usize> From<[&'a GStr; N]> for StrV {
+    #[inline]
+    fn from(value: [&'a GStr; N]) -> Self {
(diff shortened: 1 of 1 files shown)


A6GibKm:
Please let me know if the test is redundant.

A6GibKm:
For reference, this is the code that wasn't working
```rust
    glib::StrV::from([
        gio::FILE_ATTRIBUTE_STANDARD_NAME,
        gio::FILE_ATTRIBUTE_TIME_CREATED,
        gio::FILE_ATTRIBUTE_TIME_CREATED_USEC,
        gio::FILE_ATTRIBUTE_TIME_MODIFIED,
        gio::FILE_ATTRIBUTE_TIME_MODIFIED_USEC,
    ])
    .join(Some(","))
```

jf2048:
This PR looks good but, for the use case of file attributes I think we should probably just make a macro at some point to match the C string concatting behavior

sdroege (approved):

PR #1115, opened 2023-06-05
glib: Do not use ptr::offset/offset_from for private/impl offset

heftig:
Doing this violates Rust's [strict provenance rules][1]. Unfortunately, the API for doing this properly ([`ptr::expose_addr`][2]) isn't ready yet. However, we can work around this on stable Rust by casting through `isize` for offset calculations.

[1]: https://doc.rust-lang.org/std/ptr/index.html
[2]: https://doc.rust-lang.org/std/primitive.pointer.html#method.expose_addr

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -88,16 +88,57 @@ pub unsafe trait InstanceStructExt: InstanceStruct {
     fn class(&self) -> &<Self::Type as ObjectSubclass>::Class;
 }
 
+// rustdoc-stripper-ignore-next
+/// Offset `ptr` by `offset` *bytes* and cast the result to `*const U`.
+///
+/// The result must be a correctly aligned pointer to a valid value of type `U`.
+///
+/// # Panics:
+///
+/// This function panics if debug assertions are enabled if adding `offset` causes `ptr` to
+/// overflow or if the resulting pointer is not correctly aligned.
+#[inline]
+fn offset_ptr_by_bytes<T, U>(ptr: *const T, offset: isize) -> *const U {
+    // FIXME: Use `ptr::expose_addr()` once stable
+    let ptr = ptr as usize;
+    let ptr = if offset < 0 {
+        ptr - (-offset) as usize
+    } else {
+        ptr + offset as usize
+    };
+    debug_assert_eq!(ptr & (mem::align_of::<U>() - 1), 0);
+    ptr as *const U
+}
+
+// rustdoc-stripper-ignore-next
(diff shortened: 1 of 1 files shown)


heftig:
Should fix [gstreamer-rs#470](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/470).

glib/src/subclass/types.rs
```diff
@@ -95,7 +95,7 @@ unsafe impl<T: InstanceStruct> InstanceStructExt for T {
             let data = Self::Type::type_data();
             let private_offset = data.as_ref().impl_offset();
             let ptr: *const u8 = self as *const _ as *const u8;
-            let imp_ptr = ptr.offset(private_offset);
+            let imp_ptr = ((ptr as isize) + private_offset) as *const u8;
             let imp = imp_ptr as *const Self::Type;
```
sdroege, line 99: I think this is still wrong. There should be no `*const u8` anywhere here, and wherever it is used you should use `isize` or `usize`.

sdroege:
Should also make a 0.16 bugfix release with this.

sdroege:
Also causes https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/issues/373 . Quite a horrible bug :)

glib/src/subclass/types.rs
```diff
@@ -95,7 +95,7 @@ unsafe impl<T: InstanceStruct> InstanceStructExt for T {
             let data = Self::Type::type_data();
             let private_offset = data.as_ref().impl_offset();
             let ptr: *const u8 = self as *const _ as *const u8;
-            let imp_ptr = ptr.offset(private_offset);
+            let imp_ptr = ((ptr as isize) + private_offset) as *const u8;
```
GuillaumeGomez, line 98: Could you add comments explaining why we do it like and also a note saying that once `expose_addr` is stable, we should use it instead please?

sdroege: Already done locally

glib/src/subclass/types.rs
```diff
@@ -88,16 +88,57 @@ pub unsafe trait InstanceStructExt: InstanceStruct {
     fn class(&self) -> &<Self::Type as ObjectSubclass>::Class;
 }
 
+// rustdoc-stripper-ignore-next
+/// Offset `ptr` by `offset` *bytes* and cast the result to `*const U`.
```
GuillaumeGomez, line 92: That's much better, thanks a lot!

PR #1122, opened 2023-07-05
Implement FromGlibPtrBorrow on boxed and shared-boxed types

mbiggio:


glib-macros/src/boxed_derive.rs
```diff
@@ -197,6 +197,22 @@ pub fn impl_boxed(input: &syn::DeriveInput) -> TokenStream {
...
+                #crate_ident::translate::FromGlibPtrBorrow::from_glib_borrow(ptr as *mut _)
+            }
+        }
+
+        impl #crate_ident::translate::FromGlibPtrBorrow<*mut #name> for #name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *mut #name) -> #crate_ident::translate::Borrowed<Self> {
+                debug_assert!(!ptr.is_null());
+
+                #crate_ident::translate::Borrowed::new(*::std::boxed::Box::from_raw(ptr))
```
sdroege, line 212: This is stealing the memory :) The `Borrowed` is not going to free it later so it's technically fine, but going via a cast and dereference or so would be better

mbiggio: Yeah, I was indeed relying on the `ManuallyDrop` inside the `Borrowed` to deal with that.
The problem is that I'm not 100% sure we can cast the raw pointer into the `Box` in an easy way.

The `as` keyword will work only for primitive types, maybe we could use `mem::transmute` but it's not going to be pretty either.

And actually, looking at the `from_raw` implementation of `Box`, I guess we would be missing the allocator part of the `Box` in the conversion:
https://doc.rust-lang.org/src/alloc/boxed.rs.html#997-999

The same thing for the shared type actually, which is also stealing the ptr with the `from_raw` implementation of the `RefCounted` trait.

Any specific idea on how to fix that?

mbiggio: Also dereferencing directly the raw pointer without going through the `Box` is not going to work:

```
#crate_ident::translate::Borrowed::new(*ptr)
```

```
error[E0507]: cannot move out of `*ptr` which is behind a raw pointer
```

sdroege: Hrm I see, we need a full `T` here. I think this is the only way then... I don't like it but OK :)
Can you add a comment why the `from_raw()` is not breaking everything here?

mbiggio: sure, just added that :)

sdroege: Wait, you're deconstructing the box here with the dereference. That's actually not valid like this. You would free the heap memory of the box and move out the `T` from there onto the stack.

mbiggio: Yes you're right, my apologies.

I will need to find another way, even if there's nothing obvious that comes to my mind.
One possibility maybe would be wrapping the box in a `ManuallyDrop` before dereferencing it, I can maybe try that out

sdroege: `ptr::read()` from the `*const T` should work as long as it's wrapped in the `ManuallyDrop` / `Borrowed` right after

glib-macros/src/shared_boxed_derive.rs
```diff
@@ -218,6 +218,26 @@ pub fn impl_shared_boxed(input: &syn::DeriveInput) -> proc_macro2::TokenStream {
...
         #impl_from_value
 
+        impl #crate_ident::translate::FromGlibPtrBorrow<*const #refcounted_type_prefix::InnerType> for #name {
+            #[inline]
+            unsafe fn from_glib_borrow(ptr: *const #refcounted_type_prefix::InnerType) -> #crate_ident::translate::Borrowed<Self> {
+                debug_assert!(!ptr.is_null());
+
+                // from_raw is taking ownership of the raw pointer here, but wrapping its result
+                // in Borrowed::new ensures that it won't be deallocated when it will go out of
+                // scope, so the pointer will still be valid afterwards
```
sdroege, line 228: I'd do it consistently with the other one (using `ptr::read`), even if it's not strictly necessary here

sdroege (commented):

glib-macros/src/shared_boxed_derive.rs
mbiggio: I think the `shared_boxed` case is a little different from the `boxed` one.
Our shared type is going to be something like:

```
MyShared(Arc<InnerType>)
```
or

```
MyShared(Rc<InnerType>)
```

so we actually need to get an `Arc` or an `Rc` from the raw pointer to the `InnerType` in order to get the `MyShared` object (in contrast to the `boxed` case in which we really needed just the `InnerType`). 

Another possibility would be wrapping the `InnerType` returned by `ptr::read` in a new `Arc` or `Rc`, but this sounds a bit strange to me, since the raw pointer actually comes from a different `Arc` or `Rc`, with its own already existing refcount

sdroege: Ah indeed. Hrm OK then! I don't like all this but I also don't see a better way :)

PR #1130, opened 2023-07-14
Fix clippy lints

GuillaumeGomez:


glib/src/lib.rs
```diff
@@ -131,8 +131,10 @@ pub mod closure;
 mod convert;
 pub use self::convert::*;
 mod enums;
-mod functions;
-pub use self::functions::*;
+// We need this to prevent shadowing `auto::functions` reexport from above.
+#[path = "functions.rs"]
+mod inner_functions;
+pub use self::inner_functions::*;
```
bilelmoussaoui, line 137: I would rather make the change to do `pub(crate) auto_functions;` in the auto generated code and we re-export it content from a functions module.

GuillaumeGomez: If I understand what you said correctly, it would still be the same issue: two `functions` at the top level.

GuillaumeGomez:
Is it normal for the doc check to fail like this?

glib/src/boxed.rs
```diff
@@ -59,6 +59,7 @@ macro_rules! glib_boxed_wrapper {
 
         impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? std::clone::Clone for $name $(<$($generic),+>)? {
             #[inline]
+            #[allow(unknown_lints, clippy::incorrect_clone_impl_on_copy_type)]
```
sdroege, line 62: Should we add `unknown_lints` just to the `lib.rs`?

glib/src/boxed_inline.rs
```diff
@@ -50,6 +51,7 @@ macro_rules! glib_boxed_inline_wrapper {
 
         impl $(<$($generic $(: $bound $(+ $bound2)*)?),+>)? std::clone::Clone for $name $(<$($generic),+>)? {
             #[inline]
+            #[allow(unknown_lints, clippy::incorrect_clone_impl_on_copy_type)]
```
sdroege, line 54: Why do we get this clippy lint at all? This should never be a `Copy` type. `BoxedInline` might, but all the others not.

Something seems fishy here :)

glib/src/lib.rs
sdroege: Or we could rename the manual one to `manual_functions.rs`? However, what is even the clippy warning here and why don't we get the same for all the other cases where we have a module in `auto` and at the top-level with the same name?

glib/src/main_context_channel.rs
```diff
@@ -66,6 +66,7 @@ impl<T> Clone for Channel<T> {
 }
 
 impl<T> Channel<T> {
+    #[allow(unknown_lints, clippy::arc_with_non_send_sync)]
```
sdroege, line 69: This should get a clippy issue. There's nothing wrong here and it's too strict. We use `T` instead of `T: Send + Sync` here to allow both cases, and there's no other way to allow both cases really.

clippy should probably only warn about this for concrete types.

glib/src/main_context_futures.rs
```diff
@@ -95,7 +95,8 @@ impl TaskSource {
                 ptr::drop_in_place(&mut (*source).future);
             }
             FutureWrapper::NonSend(ref mut future) => {
-                let context = ffi::g_source_get_context(source as *mut Self as *mut ffi::GSource);
+                let source: *mut Self = source as *mut _;
```
sdroege, line 98:
```suggestion
                let source = source as *mut Self;
```

glib/src/main_context_futures.rs
```diff
@@ -95,7 +95,8 @@ impl TaskSource {
...
+                let context = ffi::g_source_get_context(source as *mut ffi::GSource);
```
sdroege, line 99: What exactly is the problem here?

glib/src/param_spec.rs
```diff
@@ -46,9 +46,10 @@ unsafe impl<'a> crate::value::FromValue<'a> for ParamSpec {
     type Checker = crate::value::GenericValueTypeOrNoneChecker<Self>;
 
     unsafe fn from_value(value: &'a crate::Value) -> Self {
-        let ptr = gobject_ffi::g_value_dup_param(value.to_glib_none().0);
+        let ptr: *mut gobject_ffi::GParamSpec =
+            gobject_ffi::g_value_dup_param(value.to_glib_none().0) as *mut _;
```
sdroege, line 50: What is the problem here? Generally we want to put the type on the cast instead of having a wildcard cast and a type annotation.

glib/src/param_spec.rs
```diff
@@ -533,7 +534,8 @@ macro_rules! define_param_spec_default {
                 unsafe {
                     let ptr =
                         $crate::translate::ToGlibPtr::<*const $ffi_type>::to_glib_none(self).0;
-                    $from_glib((*ptr).default_value)
+                    let call = $from_glib;
+                    call((*ptr).default_value)
```
sdroege, line 538: What's the problem here?

sdroege (commented):
These changes look mostly suspicious and like clippy bugs to me.

Maybe we should *finally* switch to only using stable to run clippy on the CI.

glib/src/boxed_inline.rs
bilelmoussaoui: It should be only needed for BoxedInline right.

glib/src/lib.rs
bilelmoussaoui: > However, what is even the clippy warning here and why don't we get the same for all the other cases where we have a module in auto and at the top-level with the same name?

Because we do `pub mod functions` in both, which I believe is wrong. We should use a `pub (crate)` and re-export these auto functions from the manual functions, similar how we handle prelude & builders.

bilelmoussaoui:
> Is it normal for the doc check to fail like this?

It has been like this for a few weeks now, the rustdoc stripper does nonsense sometimes, restarting it should work

glib/src/lib.rs
sdroege: Yes

glib/src/boxed_inline.rs
bilelmoussaoui: Didn't investigate further but see https://github.com/gtk-rs/gtk4-rs/actions/runs/5557902405/jobs/10152309339?pr=1423

sdroege: That seems valid to ignore here, yes.

glib/src/lib.rs
GuillaumeGomez: `pub(crate)` won't solve the problem that rustc doesn't know what `functions` is actually. Hence why the rename is needed in the first place. I'll rename the file.

bilelmoussaoui: > `pub(crate)` won't solve the problem that rustc doesn't know what `functions` is actually. Hence why the rename is needed in the first place. I'll rename the file.

Yes, I rather rename the auto file, as that is easier to do then renaming all the other files

sdroege (approved):
Seems good to me. @bilelmoussaoui ?

GuillaumeGomez:
I can't reproduce the CI error locally. Yeay...

sdroege:
Are you running locally with 1.71? Happens for me, here's a diff to fix it

```diff
diff --git a/glib/tests/subclass_compiletest/03-object-no-auto-send-sync.stderr b/glib/tests/subclass_compiletest/03-object-no-auto-send-sync.stderr
index b4edfbe22b2..4fc7bdc3967 100644
--- a/glib/tests/subclass_compiletest/03-object-no-auto-send-sync.stderr
+++ b/glib/tests/subclass_compiletest/03-object-no-auto-send-sync.stderr
@@ -8,7 +8,11 @@ error[E0277]: `*mut c_void` cannot be sent between threads safely
   |
   = help: the trait `Send` is not implemented for `*mut c_void`
   = note: required for `TypedObjectRef<*mut c_void, ()>` to implement `Send`
-  = note: required because it appears within the type `Object`
+note: required because it appears within the type `Object`
+ --> src/object.rs
+  |
+  |     pub Object, *mut std::os::raw::c_void, (), gobject_ffi::GObject, @ffi_class gobject_ffi::GObjectClass, @type_ gobject_ffi::g_object_g...
+  |         ^^^^^^
 note: required by a bound in `main::check`
  --> tests/subclass_compiletest/03-object-no-auto-send-sync.rs:2:17
   |
@@ -25,7 +29,11 @@ error[E0277]: `*mut c_void` cannot be shared between threads safely
   |
   = help: the trait `Sync` is not implemented for `*mut c_void`
   = note: required for `TypedObjectRef<*mut c_void, ()>` to implement `Send`
-  = note: required because it appears within the type `Object`
+note: required because it appears within the type `Object`
+ --> src/object.rs
+  |
+  |     pub Object, *mut std::os::raw::c_void, (), gobject_ffi::GObject, @ffi_class gobject_ffi::GObjectClass, @type_ gobject_ffi::g_object_g...
+  |         ^^^^^^
 note: required by a bound in `main::check`
  --> tests/subclass_compiletest/03-object-no-auto-send-sync.rs:2:17
   |
```

GuillaumeGomez:
I use `1.71` indeed. Something must be wrong with my local rustup toolchain setup. Thanks for the patch, I applied it.

sdroege (approved):

PR #1131, opened 2023-07-17
Use `--generate-link-to-definition` option when generating documentation

GuillaumeGomez:


glib/patch
```diff
@@ -0,0 +1,30 @@
+diff --git a/glib/tests/subclass_compiletest/03-object-no-auto-send-sync.stderr b/glib/tests/subclass_compiletest/03-object-no-auto-send-sync.stderr
```
sdroege, line 1: What's this? :P

sdroege (dismissed):
Looks good to me otherwise

glib/patch
GuillaumeGomez: Oh no :o

GuillaumeGomez:
Removed the patch file. ^^'

PR #1133, opened 2023-07-20
Add retain method to ListStore

ranfdev:
Motivation:
People often endup doing
```rust
for item in list_store.iter()
  if item.unwrap().name() == 'Giorgio' {
    list_store.remove(&item)
  }
}
```
But that `unwrap` is going to panic, because we added checks to prevent that an iterator over a list is mutated during iteration.

`retain` is just a nice alternative to provide an easier to use way to remove items while iterating the list.
```rust
list_store.retain(|item| item.name() != 'Giorgio')
```

It will also remove consecutive items with a single `splice` call, so If the user removes every item, `retain` doesn't spam `items_changed` signals

### DIFF
--- a/gio/src/list_store.rs
+++ b/gio/src/list_store.rs
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::cmp::Ordering;
+use std::{cell::Cell, cmp::Ordering, rc::Rc};
 
 use glib::{prelude::*, translate::*, Object};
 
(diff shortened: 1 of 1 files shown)


gio/src/list_store.rs
```diff
@@ -76,6 +76,45 @@ impl ListStore {
...
+    // rustdoc-stripper-ignore-next
+    /// Retains only the elements specified by the predicate.
+    /// This method operates in place, visiting each element exactly once in the original order,
+    /// and preserves the order of the retained elements.
+    /// Because the elements are visited exactly once in the original order,
+    /// external state may be used to decide which elements to keep.
+    ///
+    /// # Panics
+    /// Panics if the predicate closure mutates the list by removing or adding items.
+    pub fn retain(&self, f: impl Fn(&glib::Object) -> bool) {
```
sdroege, line 88:
```suggestion
    pub fn retain(&self, f: impl FnMut(&glib::Object) -> bool) {
```

because why not :)

gio/src/list_store.rs
```diff
@@ -76,6 +76,45 @@ impl ListStore {
...
+        let mut consec_removed = 0;
+        let mut i = 0;
+        const ADDITIONS: &[glib::Object] = &[]; // To satisfy the type checker
+
+        let changed = Rc::new(Cell::new(false));
+        let changed_clone = changed.clone();
+        let signal_id = self.connect_items_changed(move |_list, _, _, _| changed_clone.set(true));
```
sdroege, line 95: Ideally you'd like to store that in something that would disconnect on `Drop`

gio/src/list_store.rs
```diff
@@ -76,6 +76,45 @@ impl ListStore {
...
+            }
+            if !keep {
+                consec_removed += 1;
+            } else if consec_removed > 0 {
+                self.splice(i - consec_removed, consec_removed, ADDITIONS);
+                changed.set(false);
+                i -= consec_removed;
+                consec_removed = 0;
+            }
+            i += 1;
```
sdroege, line 110: I hope the test covers the case of `consec_removed > 0` and afterwards `i += 1` and then checks that the next iteration ends up at the correct item?

That part is a bit hard to follow but otherwise this seems all correct :)

ranfdev: Eh, it should be covered. If it weren't, `retain` would remove or keep an unwanted item, but the test case checks that.

sdroege: I guess we should add coverage reports to the CI. That was very useful for these kind of things in gstreamer :) If it's covered, all good.

sdroege (approved):

Issue #1102, opened 2023-05-24
[BUG] `glib::closure` can cause clippy to complain


MathieuDuponchelle:
```
error: unneeded unit expression
    --> net/webrtc/src/webrtcsink/imp.rs:2186:37
     |
2186 | / ...                   glib::clone!(@strong session_id_str, @weak webrtcbin, @weak element => @default-return (), move |sess, pspec| {
2187 | | ...                       // Run the Loss-based control algorithm on new peer TWCC feedbacks
2188 | | ...                       element.imp().process_loss_stats(&element, &session_id_str, &sess.property::<gst::Structure>(pspec.name()));
2189 | | ...                   })
     | |________________________^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_unit
     = note: `-D clippy::unused-unit` implied by `-D warnings`
     = note: this error originates in the macro `glib::clone` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Perhaps `@default-return` should be supported with no following token?

SeaDve:
I think you can omit `@default-return (), ` here

PR #1138, opened 2023-07-22
Don't generate unit tuple in clone macro as default-return value

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1102.


bilelmoussaoui:
looks good to me, but i still believe the actual solution is to not use `@default-return` at all. Still good to merge though

bilelmoussaoui (approved):

PR #1144, opened 2023-07-25
glib: Re-introduce an event propagation specific type

bilelmoussaoui:
As the semantics of ControlFlow don't match 1:1 with the event propagations See https://github.com/gtk-rs/gtk4-rs/issues/1435

Would still need a change on gir's side as well as a regen + yanking the releases & redoing them..............................


glib/src/signal.rs
```diff
@@ -146,3 +146,81 @@ pub fn signal_has_handler_pending<T: ObjectType>(
         ))
     }
 }
+
```
awused, line 149: Yeah, I think this is a lot clearer.

glib/src/signal.rs
```diff
@@ -146,3 +146,81 @@ pub fn signal_has_handler_pending<T: ObjectType>(
...
+// rustdoc-stripper-ignore-next
+/// Whether an event was handled or not.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Propagation {
+    // Event was handled, don't propagate
+    Stop,
+    // Event was not handled, propagate
+    Proceed,
```
bilelmoussaoui, line 157: Maybe Propagate / Inhibit as variants names? no idea, english is hard, naming is harder

awused: Personally I think these are fine, `Propagation::Stop` and `Propagation::Proceed` seem very clear and unambiguous. I don't find `Propagation::Propagate` or `Propagation::Inhibit` to be any clearer.

glib/src/signal.rs
```diff
@@ -146,3 +146,81 @@ pub fn signal_has_handler_pending<T: ObjectType>(
...
+        } else {
+            Self::Stop
+        }
+    }
+}
+
+impl From<Propagation> for bool {
+    fn from(c: Propagation) -> Self {
+        match c {
+            Propagation::Stop => false,
```
andy128k, line 187: Does this correspond to `Inhibit(true)`?  Then values are flipped.

glib/src/signal.rs
bilelmoussaoui: I kept them as is for now & added some docs. If someone has better naming alternatives, we could always do that in future releases.

glib/src/signal.rs
bilelmoussaoui: Thanks! Fixed

PR #1145, opened 2023-07-25
Add object_subclass example

ranfdev:
I'd like having a copy/pastable file to use when I need a simple gobject to store some data. I think many people would find this useful.

### DIFF
--- a/examples/object_subclass/author.rs
+++ b/examples/object_subclass/author.rs
@@ -1,0 +1,49 @@
+// You can copy/paste this file every time you need a simple GObject
+// to hold some data
+
+use glib::once_cell::sync::Lazy;
+use glib::prelude::*;
+use glib::subclass::prelude::*;
+use glib::subclass::Signal;
+use glib::Properties;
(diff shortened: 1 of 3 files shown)


examples/object_subclass/author.rs
```diff
@@ -0,0 +1,46 @@
...
+        type Type = super::Author;
+    }
+}
+
+glib::wrapper! {
+    pub struct Author(ObjectSubclass<imp::Author>);
+}
+impl Author {
+    pub fn new() -> Self {
+        glib::Object::builder().build()
```
bilelmoussaoui, line 44:
```suggestion
        glib::Object::new()
```

sdroege: Or maybe the constructor should have parameters for the properties?

bilelmoussaoui:
@ranfdev any chance you could handle the review comment so we can land it? thanks

bilelmoussaoui (approved):

Issue #44, opened 2021-05-14
gio::Application::get_default() is unsafe


sdroege:
It allows retrieving the application from any thread, not just the main thread (or more general the thread where the application was created). `gio::Application` is not thread-safe, so this is a problem.

jf2048:
It seems this can never be completely safe because `g_application_set_default` is not thread safe, and that gets called internally by `GApplication` implementations. I guess what that means is, a `GApplication` can only be referenced in the main thread, but we don't have any way of knowing what the main thread really is, getting the current thread id is just a guess.

It could be made a little more robust by checking if we can acquire the main context before storing the `qdata`, but that still isn't completely safe in all cases because it only checks for a running application...

sdroege:
Yeah the problem with making `get_default()` unsafe is that it's a very useful API to have in GTK applications and requiring those to use unsafe code is not ideal (`set_default()` is fine IMHO, who needs that anyway...)

jf2048:
That will not fully make it sound but will probably correct most of the issues people have. The wrapper macro will need some modifications here to do the thread checking so we may need to see https://github.com/gtk-rs/gtk-rs-core/issues/916 for that

sdroege:
I think the easiest would be to implement `default()` manually for now and implement the same assertion that is also used in e.g. `glib::MainContext::spawn_local()`. That should cover the easiest mistakes at least.

More correct would be to forbid construction of a `gio::Application` from another thread but that requires `glib::wrapper!` macro changes as @jf2048 says, and not very easy ones.

Does someone want to do the easy part for now?

jf2048:
> I think the easiest would be to implement `default()` manually for now and implement the same assertion that is also used in e.g. `glib::MainContext::spawn_local()`. That should cover the easiest mistakes at least.

How can we do that, since `Application` does not reference a `MainContext`? I think any solution will need to modify the wrapper and do the `qdata` solution described above.

sdroege:
Can you actually use an `Application` on a non-default main context?

PR #1146, opened 2023-07-25
glib: Bind `g_unichar` APIs

### LINKED ISSUE #763: Bind `g_utf8` / `g_unichar` APIs

RealKC:
I'm not sure if documentation for these methods can be automatically generated given these are manually defined as methods on an extension trait.

I *think* I got all of the ones not offered by the stdlib

Closes #763 

---

Some notes:
* names were made "rustier", e.g.:` g_unichar_istitle` became `is_titlecase`
* I did not include APIs that were already availabe through Rust's char

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,5 +1,5 @@
 [files]
-extend-exclude = ["auto", "sys", "gobject-sys", "*.svg"]
+extend-exclude = ["auto", "sys", "gobject-sys", "*.svg", "glib/Gir.toml"]
 
 [default.extend-words]
 # Ignore false-positives
--- a/glib/src/auto/enums.rs
+++ b/glib/src/auto/enums.rs
@@ -1054,6 +1054,261 @@ impl FromGlib<ffi::GTimeType> for TimeType {
 
 #[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
 #[non_exhaustive]
+#[doc(alias = "GUnicodeBreakType")]
+pub enum UnicodeBreakType {
+    #[doc(alias = "G_UNICODE_BREAK_MANDATORY")]
+    Mandatory,
+    #[doc(alias = "G_UNICODE_BREAK_CARRIAGE_RETURN")]
+    CarriageReturn,
+    #[doc(alias = "G_UNICODE_BREAK_LINE_FEED")]
+    LineFeed,
+    #[doc(alias = "G_UNICODE_BREAK_COMBINING_MARK")]
+    CombiningMark,
+    #[doc(alias = "G_UNICODE_BREAK_SURROGATE")]
+    Surrogate,
+    #[doc(alias = "G_UNICODE_BREAK_ZERO_WIDTH_SPACE")]
+    ZeroWidthSpace,
+    #[doc(alias = "G_UNICODE_BREAK_INSEPARABLE")]
+    Inseparable,
+    #[doc(alias = "G_UNICODE_BREAK_NON_BREAKING_GLUE")]
+    NonBreakingGlue,
+    #[doc(alias = "G_UNICODE_BREAK_CONTINGENT")]
+    Contingent,
+    #[doc(alias = "G_UNICODE_BREAK_SPACE")]
+    Space,
+    #[doc(alias = "G_UNICODE_BREAK_AFTER")]
+    After,
+    #[doc(alias = "G_UNICODE_BREAK_BEFORE")]
+    Before,
+    #[doc(alias = "G_UNICODE_BREAK_BEFORE_AND_AFTER")]
+    BeforeAndAfter,
+    #[doc(alias = "G_UNICODE_BREAK_HYPHEN")]
+    Hyphen,
+    #[doc(alias = "G_UNICODE_BREAK_NON_STARTER")]
+    NonStarter,
+    #[doc(alias = "G_UNICODE_BREAK_OPEN_PUNCTUATION")]
+    OpenPunctuation,
+    #[doc(alias = "G_UNICODE_BREAK_CLOSE_PUNCTUATION")]
+    ClosePunctuation,
+    #[doc(alias = "G_UNICODE_BREAK_QUOTATION")]
+    Quotation,
+    #[doc(alias = "G_UNICODE_BREAK_EXCLAMATION")]
+    Exclamation,
+    #[doc(alias = "G_UNICODE_BREAK_IDEOGRAPHIC")]
+    Ideographic,
+    #[doc(alias = "G_UNICODE_BREAK_NUMERIC")]
+    Numeric,
+    #[doc(alias = "G_UNICODE_BREAK_INFIX_SEPARATOR")]
+    InfixSeparator,
+    #[doc(alias = "G_UNICODE_BREAK_SYMBOL")]
+    Symbol,
+    #[doc(alias = "G_UNICODE_BREAK_ALPHABETIC")]
+    Alphabetic,
+    #[doc(alias = "G_UNICODE_BREAK_PREFIX")]
+    Prefix,
+    #[doc(alias = "G_UNICODE_BREAK_POSTFIX")]
+    Postfix,
+    #[doc(alias = "G_UNICODE_BREAK_COMPLEX_CONTEXT")]
+    ComplexContext,
+    #[doc(alias = "G_UNICODE_BREAK_AMBIGUOUS")]
+    Ambiguous,
+    #[doc(alias = "G_UNICODE_BREAK_UNKNOWN")]
+    Unknown,
+    #[doc(alias = "G_UNICODE_BREAK_NEXT_LINE")]
+    NextLine,
+    #[doc(alias = "G_UNICODE_BREAK_WORD_JOINER")]
+    WordJoiner,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_L_JAMO")]
+    HangulLJamo,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_V_JAMO")]
+    HangulVJamo,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_T_JAMO")]
+    HangulTJamo,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_LV_SYLLABLE")]
+    HangulLvSyllable,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_LVT_SYLLABLE")]
+    HangulLvtSyllable,
+    #[doc(alias = "G_UNICODE_BREAK_CLOSE_PARENTHESIS")]
+    CloseParenthesis,
+    #[doc(alias = "G_UNICODE_BREAK_CONDITIONAL_JAPANESE_STARTER")]
+    ConditionalJapaneseStarter,
+    #[doc(alias = "G_UNICODE_BREAK_HEBREW_LETTER")]
+    HebrewLetter,
+    #[doc(alias = "G_UNICODE_BREAK_REGIONAL_INDICATOR")]
+    RegionalIndicator,
+    #[doc(alias = "G_UNICODE_BREAK_EMOJI_BASE")]
+    EmojiBase,
+    #[doc(alias = "G_UNICODE_BREAK_EMOJI_MODIFIER")]
+    EmojiModifier,
+    #[doc(alias = "G_UNICODE_BREAK_ZERO_WIDTH_JOINER")]
+    ZeroWidthJoiner,
+    #[doc(hidden)]
+    __Unknown(i32),
+}
+
+impl fmt::Display for UnicodeBreakType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "UnicodeBreakType::{}",
+            match *self {
+                Self::Mandatory => "Mandatory",
+                Self::CarriageReturn => "CarriageReturn",
+                Self::LineFeed => "LineFeed",
+                Self::CombiningMark => "CombiningMark",
+                Self::Surrogate => "Surrogate",
+                Self::ZeroWidthSpace => "ZeroWidthSpace",
+                Self::Inseparable => "Inseparable",
+                Self::NonBreakingGlue => "NonBreakingGlue",
+                Self::Contingent => "Contingent",
+                Self::Space => "Space",
+                Self::After => "After",
+                Self::Before => "Before",
+                Self::BeforeAndAfter => "BeforeAndAfter",
+                Self::Hyphen => "Hyphen",
+                Self::NonStarter => "NonStarter",
+                Self::OpenPunctuation => "OpenPunctuation",
+                Self::ClosePunctuation => "ClosePunctuation",
+                Self::Quotation => "Quotation",
+                Self::Exclamation => "Exclamation",
+                Self::Ideographic => "Ideographic",
+                Self::Numeric => "Numeric",
+                Self::InfixSeparator => "InfixSeparator",
+                Self::Symbol => "Symbol",
+                Self::Alphabetic => "Alphabetic",
+                Self::Prefix => "Prefix",
+                Self::Postfix => "Postfix",
+                Self::ComplexContext => "ComplexContext",
+                Self::Ambiguous => "Ambiguous",
+                Self::Unknown => "Unknown",
+                Self::NextLine => "NextLine",
+                Self::WordJoiner => "WordJoiner",
+                Self::HangulLJamo => "HangulLJamo",
+                Self::HangulVJamo => "HangulVJamo",
+                Self::HangulTJamo => "HangulTJamo",
+                Self::HangulLvSyllable => "HangulLvSyllable",
+                Self::HangulLvtSyllable => "HangulLvtSyllable",
+                Self::CloseParenthesis => "CloseParenthesis",
+                Self::ConditionalJapaneseStarter => "ConditionalJapaneseStarter",
+                Self::HebrewLetter => "HebrewLetter",
+                Self::RegionalIndicator => "RegionalIndicator",
+                Self::EmojiBase => "EmojiBase",
+                Self::EmojiModifier => "EmojiModifier",
+                Self::ZeroWidthJoiner => "ZeroWidthJoiner",
+                _ => "Unknown",
+            }
+        )
+    }
+}
+
+#[doc(hidden)]
+impl IntoGlib for UnicodeBreakType {
+    type GlibType = ffi::GUnicodeBreakType;
+
+    fn into_glib(self) -> ffi::GUnicodeBreakType {
+        match self {
+            Self::Mandatory => ffi::G_UNICODE_BREAK_MANDATORY,
+            Self::CarriageReturn => ffi::G_UNICODE_BREAK_CARRIAGE_RETURN,
+            Self::LineFeed => ffi::G_UNICODE_BREAK_LINE_FEED,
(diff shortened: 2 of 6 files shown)


glib/src/unichar.rs
```diff
@@ -0,0 +1,237 @@
...
+        match self {
+            Compatibility => true,
+            Canonical => false,
+        }
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// The result of a single step of the Unicode canonical decomposition algorithm
+pub enum DecompositionResult {
```
RealKC, line 36: I'm not too happy with the names, but I don't have better ideas

glib/src/unichar.rs
```diff
@@ -0,0 +1,237 @@
...
+                self as _,
+                // NOTE: This assumes that glib will only ever write valid unicode codepoints to the
+                //       buffer.
+                decomposition_kind.into_bool().into_glib(),
+                buffer.as_mut_ptr().cast(),
+                buffer_len,
+            )
+        };
+        debug_assert!(decomposition_length <= buffer_len);
+        unsafe { buffer.set_len(decomposition_length) };
```
RealKC, line 201: This feels kinda sketch but I think it's fine?

glib/src/auto/enums.rs
```diff
@@ -1052,6 +1052,261 @@ impl FromGlib<ffi::GTimeType> for TimeType {
...
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_V_JAMO")]
+    HangulVJamo,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_T_JAMO")]
+    HangulTJamo,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_LV_SYLLABLE")]
+    HangulLvSyllable,
+    #[doc(alias = "G_UNICODE_BREAK_HANGUL_LVT_SYLLABLE")]
+    HangulLvtSyllable,
+    #[doc(alias = "G_UNICODE_BREAK_CLOSE_PARANTHESIS")]
+    CloseParanthesis,
```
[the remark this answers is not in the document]
bilelmoussaoui: You will have to manually implement the enum, sadly. Unless you go and fix gir :)

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+
+impl DecompositionKind {
+    fn into_bool(self) -> bool {
+        use DecompositionKind::*;
+        match self {
+            Compatibility => true,
+            Canonical => false,
+        }
+    }
+}
```
bilelmoussaoui, line 34: Why not use the From trait here?

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn is_zero_width(self) -> bool {
+        unsafe { bool::from_glib(ffi::g_unichar_iszerowidth(self as _)) }
+    }
+
+    fn to_titlecase(self) -> char {
+        from_unichar(unsafe { ffi::g_unichar_totitle(self as _) })
+    }
+
+    fn mirror_char(self) -> Option<char> {
+        let mut mirrored = 0u32;
```
bilelmoussaoui, line 179: We usually use mem::MaybeUninit here

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn script(self) -> UnicodeScript {
+        unsafe { UnicodeScript::from_glib(ffi::g_unichar_get_script(self as _)) }
+    }
+
+    fn combining_class(self) -> i32 {
+        unsafe { ffi::g_unichar_combining_class(self as _) }
+    }
+
+    fn is_mark(self) -> bool {
+        unsafe { bool::from_glib(ffi::g_unichar_ismark(self as _)) }
```
bilelmoussaoui, line 143:
```suggestion
        unsafe { from_glib(ffi::g_unichar_ismark(self as _)) }
```
should be good enough, same for others

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// The first character may decompose further, but the second cannot
+    Pair(char, char),
+}
+
+// rustdoc-stripper-ignore-next
+/// This trait provides access to Unicode character classification and manipulations functions
+/// provided by GLib that do not exist in the standard library
+#[doc(alias = "g_unichar")]
+pub trait Unichar: sealed::Sealed + Copy {
+    #[doc(alias = "g_unichar_type")]
```
bilelmoussaoui, line 55: All these functions should be marked as manual in the Gir.toml

glib/src/unichar.rs
RealKC: I can't really recall my thought process here, I'll go with whatever you prefer

glib/src/unichar.rs
RealKC: I don't think it'd make that much of a difference?

glib/src/auto/enums.rs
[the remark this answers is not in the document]
RealKC: I think I'll manually implement the enum then ^^

glib/src/unichar.rs
RealKC: Will do

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+};
+
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for char {}
+}
+
+// rustdoc-stripper-ignore-next
+/// The kind of decomposition to perform
+#[derive(Clone, Copy)]
```
sdroege, line 15: Why not `PartialEq`, `Eq` and `Debug`? Those are useful :)

glib/src/unichar.rs
sdroege: What is the use case for this?

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+        match self {
+            Compatibility => true,
+            Canonical => false,
+        }
+    }
+}
+
+// rustdoc-stripper-ignore-next
+/// The result of a single step of the Unicode canonical decomposition algorithm
+pub enum DecompositionResult {
```
sdroege, line 38: Please derive useful traits here

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+    fn break_type(self) -> UnicodeBreakType;
+
+    #[doc(alias = "g_unichar_get_script")]
+    #[doc(alias = "unichar_get_script")]
+    fn script(self) -> UnicodeScript;
+
+    #[doc(alias = "g_unichar_combining_class")]
+    #[doc(alias = "unichar_combining_class")]
+    // There's no enum defining named constants for this one as far as I can tell
+    fn combining_class(self) -> i32;
```
sdroege, line 70: You could create one :)

glib/src/unichar.rs
sdroege: Yes but I would go one step further. Use `Vec::with_capacity()` above to create the `Vec` (and don't zero-initialize it).

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+fn from_unichar(unichar: u32) -> char {
+    // It's a glib bug if they give us an invalid unicode codepoint from one of their manipulation
+    // functions. It's safest if we just panic in that case
+    char::from_u32(unichar).unwrap()
```
sdroege, line 238: That's just what `char::from_glib()` is already doing or not?

glib/src/unichar.rs
sdroege: By itself not: just avoids one instruction here (and in every inlined copy). But if everything did this, there would be measurable code bloat.

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+    #[doc(alias = "unichar_decompose")]
+    fn decompose(self) -> DecompositionResult;
+
+    #[doc(alias = "g_unichar_compose")]
+    #[doc(alias = "unichar_compose")]
+    fn compose(a: char, b: char) -> Option<char>;
+}
+
+impl Unichar for char {
+    fn unicode_type(self) -> UnicodeType {
```
sdroege, line 126: You probably want to add `#[inline]`  here in in other places for these trivial functions to give some more hints to the compiler.

glib/src/auto/enums.rs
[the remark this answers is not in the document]
RealKC: Oh that worked, thanks

glib/src/unichar.rs
RealKC: I ended up just removing and inlining the contents of this function in the one place where it is used (`fully_decompose`)

glib/src/unichar.rs
RealKC: That doesn't seem to exist

glib/src/unichar.rs
RealKC: There's quite a few of these ([UAX #44](https://www.unicode.org/reports/tr44/)], section 5.7.4 talks about them, [PropertyValueAliases.txt](https://www.unicode.org/Public/15.0.0/ucd/PropertyValueAliases.txt) (search for `# Canonical_Combining_Class (ccc)`) has a more complete list). Should we have an enum that has separate variants for the named ones, with the unnamed ones going into a generic `Ccc` variant? The gtk demos seem to assign more meaning to some combining classes but I'm not sure where they get that from. (see https://github.com/GNOME/gtk/blob/fa4ad1590dbbcaf1a0eb230d2a20d0a867f8b752/demos/gtk-demo/unicode-names.c#L101)

glib/src/unichar.rs
bilelmoussaoui: there is  char::try_from_glib though

RealKC:
Regarding the CI failures:
- can I tell typos to ignore either a specific file or specific line? I'd rather not ignore the paranthesis typo on the whole project (and where it is made, it is necessary)
- do I need to add doc aliases to the impls for the functions too of the trait too? Seems strange

glib/src/unichar.rs
RealKC: I don't think so? I can't find an impl anyway: https://github.com/search?q=repo%3Agtk-rs%2Fgtk-rs-core%20%22impl%20TryFromGlib%22&type=code

bilelmoussaoui: Oh it seems to use TryFrom instead, see https://github.com/gtk-rs/gtk4-rs/blob/db7af24294c706136f3bcba74caf284957b9b9c2/gtk4/src/entry.rs#L22

RealKC: hm I guess I could switch to using TryFrom, the main point of this function is to (a) have the comment in one place (b) not repeat the actual expression to go from g_unichar -> char everywhere it's needed

I guess I could mark it as `#[inline]` and `#[track_caller]` though?

bilelmoussaoui: Sounds like a good plan to me, but i would wait for @sdroege's opinion on this one :)

sdroege: Sounds like a plan

glib/src/unichar.rs
sdroege: Not sure we should do that. We would have to keep that enum up to date :) I wonder why this is not already done in GLib though...

Let's keep it a plain integer then for now (but maybe it should be unsigned?)

sdroege:
> do I need to add doc aliases to the impls for the functions too of the trait too? Seems strange

You only need to add them on plain functions and trait declarations, not trait implementations.

> can I tell typos to ignore either a specific file or specific line? I'd rather not ignore the paranthesis typo on the whole project (and where it is made, it is necessary)

Something for @bilelmoussaoui I guess :)

glib/src/unichar.rs
RealKC: I think the best type for this might be u8?

sdroege: If it's only a byte and can't be more than that, sure

RealKC:
Changes since last push:
 - rebased on top of master
 - addressed review comments

> You only need to add them on plain functions and trait declarations, not trait implementations.

As far as I can tell, CI is now failing because the aliases don't exist in the trait impl

RealKC:
I rewrote the trait so that CI would be happy, seems to have worked

glib/src/unichar.rs
```diff
@@ -0,0 +1,239 @@
...
+        let res = unsafe {
+            from_glib(ffi::g_unichar_decompose(
+                self.into(),
+                a.as_mut_ptr(),
+                b.as_mut_ptr(),
+            ))
+        };
+
+        // SAFETY: `a` and `b` will always be init after the g_unichar_decompose call returns
+        let (a, b) = unsafe { (a.assume_init(), b.assume_init()) };
```
bilelmoussaoui, line 210: Even if res is false ?

RealKC: > If ch is not decomposable, a is set to ch and b is set to zero.
> Return value
> TRUE if the character could be decomposed.

Based on these two statements, I'd say yes, but I can move the assignment in the `if res {` branch, it's not used on the else case anyway

bilelmoussaoui (dismissed):

glib/src/unichar.rs
sdroege: Yes if `FALSE` is returned than `a` is set to `ch` and `b` is set to `0`. Both values are always initialized.

.typos.toml
```diff
@@ -1,5 +1,5 @@
 [files]
-extend-exclude = ["auto", "sys", "gobject-sys", "*.svg"]
+extend-exclude = ["auto", "sys", "gobject-sys", "*.svg", "glib/Gir.toml"]
```
sdroege, line 2: @bilelmoussaoui Is this fine with you?

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+use std::mem::MaybeUninit;
+
+use crate::{
+    translate::{from_glib, IntoGlib},
+    UnicodeBreakType, UnicodeScript, UnicodeType,
+};
+
+mod sealed {
+    pub trait Sealed: Into<u32> {
+        fn from_unichar(unichar: u32) -> char;
```
sdroege, line 12: This doesn't make much sense. The `Sealed` trait is only for trait privacy reasons, it should not contain any functions or anything. It's just a marker trait.

How about implementing `glib::UnsafeFrom<u32> for char` and requiring that in the `Unichar` trait?

glib/src/unichar.rs
sdroege: Maybe `CharacterDecomposition`?

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+    // rustdoc-stripper-ignore-next
+    /// The first character may decompose further, but the second cannot
+    Pair(char, char),
+}
+
+// rustdoc-stripper-ignore-next
+/// This trait provides access to Unicode character classification and manipulations functions
+/// provided by GLib that do not exist in the standard library
+#[doc(alias = "g_unichar")]
+pub trait Unichar: sealed::Sealed + Copy {
```
sdroege, line 57:
```suggestion
pub trait Unichar: sealed::Sealed + Into<u32> + Copy {
```

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+    pub trait Sealed: Into<u32> {
+        fn from_unichar(unichar: u32) -> char;
+    }
+    impl Sealed for char {
+        #[inline]
+        #[track_caller]
+        fn from_unichar(unichar: u32) -> char {
+            // It's a glib bug if they give us an invalid unicode codepoint from one of their manipulation
+            // functions. It's safest if we just panic in that case
+            char::from_u32(unichar).expect("glib returned an invalid Unicode codepoint")
```
sdroege, line 20: FWIW I would go with `from_u32_unchecked()` here. If you really want the checks here then only enable them in `#[cfg(debug-assertions)]`.

The `expect()` here together with all the panic handling will considerably blow up the code size when working with these functions, and if glib gives us an invalid value then that's a bug in glib. We have to trust it to do the right thing in more critical places already.

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn to_titlecase(self) -> char {
+        Self::from_unichar(unsafe { ffi::g_unichar_totitle(self.into()) })
+    }
+
+    #[doc(alias = "g_unichar_get_mirror_char")]
+    #[doc(alias = "unichar_get_mirror_char")]
+    #[inline]
+    fn mirror_char(self) -> Option<char> {
+        let mut mirrored = MaybeUninit::uninit();
+        let res = unsafe {
```
sdroege, line 157: The whole function body should really be in an `unsafe` here. The three steps of creating uninitialized memory, calling the function and only assuming the memory is initialized when the function succeeds all belong together and are the same "safety promise". Just try writing a `SAFETY` comment for each of the individual `unsafe` blocks and you'll see what I mean (you'll have to refer to the other `unsafe` blocks!).

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+                self.into(),
+                // NOTE: This assumes that glib will only ever write valid unicode codepoints to the
+                //       buffer.
+                compat.into_glib(),
+                buffer.as_mut_ptr().cast(),
+                buffer_len,
+            )
+        };
+        debug_assert!(decomposition_length <= buffer_len);
+        unsafe { buffer.set_len(decomposition_length) };
```
sdroege, line 191: This `unsafe` block should be together with the other one *and* the assertion. Reasoning further above.

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+            from_glib(ffi::g_unichar_decompose(
+                self.into(),
+                a.as_mut_ptr(),
+                b.as_mut_ptr(),
+            ))
+        };
+
+        if res {
+            // SAFETY: `a` and `b` will always be init after the g_unichar_decompose call returns
+            let (a, b) = unsafe { (a.assume_init(), b.assume_init()) };
```
sdroege, line 211: Same about the `unsafe` block here

glib/src/unichar.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[doc(alias = "g_unichar_compose")]
+    #[doc(alias = "unichar_compose")]
+    #[inline]
+    fn compose(a: char, b: char) -> Option<char> {
+        let mut out = MaybeUninit::uninit();
+        let res =
+            unsafe { from_glib(ffi::g_unichar_compose(a.into(), b.into(), out.as_mut_ptr())) };
+
+        if res {
+            Some(Self::from_unichar(unsafe { out.assume_init() }))
```
sdroege, line 231: Same about the `unsafe` block here

.typos.toml
bilelmoussaoui: I would ask @Hofer-Julian

RealKC: I think the only alternative is ignoring "paranthasis" as a typo in general? Which I don't really like as an idea. I should add a comment explaining why we ignore that file though

Hofer-Julian: Yes, this looks fine to me

glib/src/unichar.rs
RealKC: Yeah I liked that name more

sdroege (approved):

PR #1154, opened 2023-08-05
Disentangle docsrs and features

pentamassiv:
Remaining cleanup from https://github.com/gtk-rs/gtk-rs-core/pull/1078

### DIFF
--- a/gio/src/lib.rs
+++ b/gio/src/lib.rs
@@ -34,7 +34,7 @@ mod dbus_node_info;
 #[cfg(feature = "v2_72")]
 #[cfg_attr(docsrs, doc(cfg(feature = "v2_72")))]
 mod debug_controller_dbus;
-#[cfg(any(all(not(windows), not(target_os = "macos")), docsrs))]
+#[cfg(all(not(windows), not(target_os = "macos")))]
 mod desktop_app_info;
 mod error;
 mod file;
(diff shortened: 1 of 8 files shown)


bilelmoussaoui (approved):

bilelmoussaoui:
Now that the gir changes have landed, it would be great to also update & regenerate the sys crate.

pentamassiv:
Are you asking me to do it and add it to the PR? I see that you approved it already, so I am not sure

bilelmoussaoui:
> Are you asking me to do it and add it to the PR? I see that you approved it already, so I am not sure

Yes, please

Just make sure the update submodule commit is separate from the regeneration, as that would simplify backporting

bilelmoussaoui:
Actually, I will do that along with more changes

gio/src/lib.rs
```diff
@@ -84,19 +84,19 @@ pub use socket_msg_flags::SocketMsgFlags;
 mod subprocess;
 mod subprocess_launcher;
 mod threaded_socket_service;
-#[cfg(any(unix, docsrs))]
+#[cfg(unix)]
```
sdroege, line 87: Is this actually correct? This way we won't get docs for this unless building on a unix-y system?

bilelmoussaoui: I think that is correct yes, but i would refer to @GuillaumeGomez here

sdroege: That seems like a problem then. Previously we got platform-specific docs when building docs on any platform. There's quite a bit of compat hackery for that :)

PR #1156, opened 2023-08-06
Added bindings for Gio.DBusObjectManager, Gio.DBusObjectManagerClientFlags

pentamassiv:
I noticed that the above mentioned objects are not generated. I am currently working on updated bindings for Gnome Online Accounts and some of the code can't be generated because they are missing from gio.

For some reason I get a ton of formatting changes after regenerating the code, so I first regenerated before doing changes. Then I added them to the Gir.toml file and regenerated the bindings again. There were only the expected lines to uncommented after that. I ran `cargo build` and `cargo test` without any issues.

Because I don't know why I had all these formatting changes, I did not include any of the regeneration commits.


bilelmoussaoui:
You have to re-run the generator as there are no bindings generated as you can tell.

pentamassiv:
Unfortunately I can't. I tried to explain that for some reason I get a ton of formatting changes when I regenerate the code and it becomes difficult to see the changes of this PR.
I added another commit for you to see what I mean. I switched to stable Rust and tried reinstalling gir from this repo. It still causes big changes :S

pentamassiv:
Sorry, nevermind. It works. I just ran gir directly and did not use the python script. It now works

glib/src/auto/enums.rs
```diff
@@ -284,20 +284,28 @@ impl FromGlib<ffi::GDateMonth> for DateMonth {
 #[non_exhaustive]
 #[doc(alias = "GDateWeekday")]
 pub enum DateWeekday {
+    // rustdoc-stripper-ignore-next-stop
```
bilelmoussaoui, line 287: No please don't add these; the CI complains about them due to an error in rustdoc-stripper (that no one took the time to debug yet)

pentamassiv: Sorry, I removed them again

bilelmoussaoui (approved):

PR #1158, opened 2023-08-07
Reduce size of the crates by excluding unnecessary files 

pentamassiv:
The size of the crates can be reduced by excluding the `tests` folder and the `Gir.toml`, `docs.md` and `versions.txt` files from being published to crates.io.

The size could be reduced even more by excluding the LICENSE and COPYRIGHT files but I did not do that.

The `sys` folders do not need to be excluded. Apparently cargo is smart enough to recognize that they are independent crates. The `gir-files` folder also does not need to be excluded, because it is one level higher than the crates so cargo does not try to include it anyways.

You can see which files will be included without publishing the crates by running `cargo package --list`.

### DIFF
--- a/gdk-pixbuf/Cargo.toml
+++ b/gdk-pixbuf/Cargo.toml
@@ -9,9 +9,7 @@ documentation = "https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gdk_pixbuf/"
 version = "0.19.0"
 description = "Rust bindings for the GdkPixbuf library"
 repository = "https://github.com/gtk-rs/gtk-rs-core"
-exclude = [
-    "gir-files/*",
-]
+exclude = ["tests", "Gir.toml", "docs.md", "versions.txt"]
 edition = "2021"
 rust-version = "1.70"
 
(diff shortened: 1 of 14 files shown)


gdk-pixbuf/Cargo.toml
```diff
@@ -9,9 +9,7 @@ documentation = "https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gdk_pixbuf/"
 version = "0.19.0"
 description = "Rust bindings for the GdkPixbuf library"
 repository = "https://github.com/gtk-rs/gtk-rs-core"
-exclude = [
-    "gir-files/*",
-]
+exclude = ["tests", "Gir.toml", "docs.md", "versions.txt"]
```
sdroege, line 12: Why do you want to exclude the tests (they're useful for `cargo test`) and `versions.txt` (useful so people know what version of everything this is based on so they can reproduce it).

sdroege: crates.io is not only used for pulling in dependencies but also as a source for e.g. Linux distro packagers (who need these two things, for example). Maybe that's the misconception here?

pentamassiv: Yes, that is where I was wrong. I did not know that.

So packagers fetch the crates even for libraries such as gtk-rs and run tests on them or check the versions file? Is there anywhere, where I can read more on this topic?

pentamassiv: No problem, I'll close both PRs. Thank you for letting me know

sdroege: No problem and thanks for trying to make these crates smaller :)

PR #1163, opened 2023-08-09
image: Switch to latest fedora stable

bilelmoussaoui:
rawhide can break pretty badly


bilelmoussaoui:
The image builds fine

Issue #55, opened 2020-03-16
Support for dynamic types


shutton:
glib_wrapper! doesn't appear to have any support for dynamic types (types that become available after being loaded from a dynamic library)

https://developer.gnome.org/gobject/2.64/GTypeModule.html#G-DEFINE-DYNAMIC-TYPE:CAPS

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/1169

PR #1169, opened 2023-08-11
add support of module types

fbrouille:
This PR aims to add support of module types and interfaces. See https://docs.gtk.org/gobject/class.TypeModule.html
The new macros `module_object_subclass` and `module_object_interface` allow to register types and interfaces that are part of a module (e.g. a shared library on linux). The macros mimic `object_subclass` and `object_interface` but generate code that support registration when the module (`TypeModule`) is loaded, following the behavior defined in Glib doc. However it is possible to postpone the registration at first use of the type (like for `object_subclass`) by explicitly setting the macro attribute `lazy_registration = true`.
Some examples in `glib-macros/tests/test.rs`, `glib/src/subclass/mod.rs` and `glib/src/subclass/type_module.rs`

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -645,6 +645,75 @@ pub fn object_subclass(_attr: TokenStream, item: TokenStream) -> TokenStream {
     }
 }
 
+/// Macro for boilerplate of [`ObjectSubclass`] implementations that are
+/// registered as dynamic types.
+///
+/// An object subclass must be explicitly registered as a dynamic type when the
+/// system loads its implementation (see [`TypePlugin`] and [`TypeModule`].
+/// Therefore, whereas an object subclass can be registered only once as a
+/// static type, it can be registered several times as a dynamic type.
+///
+/// An object subclass registered as a dynamic type is never unregistered. The
+/// system calls [`TypePluginExt::unuse`] to unload its implementation. If the
+/// [`TypePlugin`] subclass is a [`TypeModule`], the object subclass registered
+/// as a dynamic type is marked as unloaded and must be registered again when
+/// the module is reloaded.
+///
+/// This macro provides two behaviors when registering an object subclass as a
+/// dynamic type:
+///
+/// By default an object subclass is registered as a dynamic type when the
+/// system loads its implementation (e.g. when the module is loaded):
+/// ```ignore
+/// #[glib::dynamic_object_subclass]
+/// impl ObjectSubclass for MyType { ... }
+/// ```
+///
+/// Optionally setting the macro attribute `lazy_registration` to `true`
+/// postpones registration on the first use (when `type_()` is called for the
+/// first time), similarly to the [`macro@object_subclass`]
+/// macro:
+/// ```ignore
+/// #[glib::dynamic_object_subclass(lazy_registration = true)]
+/// impl ObjectSubclass for MyType { ... }
+/// ```
+///
+/// By default an object subclass is considered to be registered as a dynamic
+/// type within a [`TypeModule`] subclass. Optionally setting the macro
+/// attribute `plugin_type` allows to register an object subclass as a dynamic
+/// type within a given [`TypePlugin`] subclass:
+/// ```ignore
+/// #[glib::dynamic_object_subclass(plugin_type = MyPlugin)]
+/// impl ObjectSubclass for MyType { ... }
+/// ```
+///
+/// [`ObjectSubclass`]: ../glib/subclass/types/trait.ObjectSubclass.html
+/// [`TypePlugin`]: ../glib/gobject/type_plugin/struct.TypePlugin.html
+/// [`TypeModule`]: ../glib/gobject/type_module/struct.TypeModule.html
+/// [`TypePluginExt::unuse`]: ../glib/gobject/type_plugin/trait.TypePluginExt.html#method.unuse
+#[proc_macro_attribute]
+#[proc_macro_error]
+pub fn dynamic_object_subclass(attr: TokenStream, item: TokenStream) -> TokenStream {
+    use proc_macro_error::abort_call_site;
+    let attrs = match syn::parse::Parser::parse(
+        syn::punctuated::Punctuated::<syn::Expr, syn::Token![,]>::parse_terminated,
+        attr,
+    ) {
+        Ok(attrs)
+            if attrs
+                .iter()
+                .all(|attr| matches!(attr, syn::Expr::Assign(_))) =>
+        {
+            attrs
+        }
+        _ => abort_call_site!(object_subclass_attribute::WRONG_EXPRESSION_MSG),
+    };
+    match syn::parse::<syn::ItemImpl>(item) {
+        Ok(input) => object_subclass_attribute::impl_dynamic_object_subclass(&attrs, &input).into(),
+        Err(_) => abort_call_site!(object_subclass_attribute::WRONG_PLACE_MSG),
+    }
+}
+
 /// Macro for boilerplate of [`ObjectInterface`] implementations.
 ///
 /// This adds implementations for the `get_type()` method, which should probably never be defined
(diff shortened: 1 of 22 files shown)


sdroege:
This would fix https://github.com/gtk-rs/gtk-rs-core/issues/55, or would something still be missing?

glib-macros/src/lib.rs
```diff
@@ -645,6 +645,59 @@ pub fn object_subclass(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// #[glib::module_object_subclass(lazy_registration = true)]
+/// impl ObjectSubclass for MyType { ... }
+/// ```
+///
+/// [`ObjectSubclass`]: ../glib/subclass/types/trait.ObjectSubclass.html
+/// [`TypeModule`]: ../glib/gobject/auto/type_module/struct.TypeModule.html
+/// [`TypeModuleExt::unuse`]: ../glib/gobject/auto/type_module/trait.TypeModuleExt.html#method.unuse
+#[proc_macro_attribute]
+#[proc_macro_error]
+pub fn module_object_subclass(attr: TokenStream, item: TokenStream) -> TokenStream {
```
sdroege, line 683: Would it be possible to re-use the existing macro and just add configuration to it? `#[glib::object_subclass(dynamic)]` or so for example.

Or do you think having separate macros would be useful for better discoverability or so?

glib/src/gobject/auto/type_module.rs
```diff
@@ -0,0 +1,121 @@
...
+                self.as_ref().to_glib_none().0,
+                instance_type.into_glib(),
+                interface_type.into_glib(),
+                interface_info.to_glib_none().0,
+            );
+        }
+    }
+
+    #[doc(alias = "g_type_module_register_enum")]
+    fn register_enum(&self, name: &str, const_static_values: &EnumValue) -> crate::types::Type {
```
sdroege, line 47: The `add()` / `register()` API here, would it make more sense to hide that behind a macro? E.g. the `glib::Enum` macro with some additional configuration?

glib/src/gobject/interface_info.rs
```diff
@@ -0,0 +1,16 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+wrapper! {
+    #[derive(Debug)]
+    #[doc(alias = "GInterfaceInfo")]
+    pub struct InterfaceInfo(BoxedInline<gobject_ffi::GInterfaceInfo>);
```
sdroege, line 6: As this has no `GType` assigned and also otherwise no memory management functions, maybe this should be implemented fully manually? There's no point in ever cloning these, for example, is there? So it could just be a `&` reference returned from the functions

sdroege: (same for the other two types here)

glib/src/subclass/type_module.rs
```diff
@@ -0,0 +1,145 @@
...
+
+impl<T: TypeModuleImpl> TypeModuleImplExt for T {
+    fn parent_load(&self) -> bool {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().parent_class() as *const gobject_ffi::GTypeModuleClass;
+
+            let f = (*parent_class)
+                .load
+                .expect("No parent class implementation for \"load\"");
```
sdroege, line 40: These are optional or not? So if they don't exist you'd simply do nothing instead of panicking

glib/src/subclass/type_module.rs
```diff
@@ -0,0 +1,145 @@
...
+    type_module: *mut gobject_ffi::GTypeModule,
+) -> ffi::gboolean {
+    let instance = &*(type_module as *mut T::Instance);
+    let imp = instance.imp();
+
+    let res = imp.load();
+
+    if res {
+        unsafe {
+            gobject_ffi::g_object_ref(type_module as _);
```
sdroege, line 84: This should probably get some comment. Why do we ref here (and unref below)?

glib/src/subclass/type_plugin.rs
```diff
@@ -0,0 +1,186 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::translate::IntoGlib;
+use crate::translate::{FromGlib, ToGlibPtr, ToGlibPtrMut};
+use crate::{
+    subclass::prelude::*, Cast, Interface, InterfaceInfo, Type, TypeInfo, TypePlugin,
+    TypeValueTable,
+};
+
+pub trait TypePluginImpl: ObjectImpl + TypePluginImplExt {
```
sdroege, line 10: Can you add a minimal test for this too?

sdroege (commented):
Generally looks good to me but I don't really know the dynamic type API of GObject :)

glib-macros/src/lib.rs
sdroege: Not sure. Either seems fine to me really :) Maybe @bilelmoussaoui has an opinion?

glib/src/gobject/interface_info.rs
sdroege: You'd define a

```rust
#[repr(transparent)]
pub struct InterfaceInfo(gobject_ffi::GInterfaceInfo);
```

and then can do things like

```rust
let ptr: *const gobject_ffi::GInterfaceInfo;
let info: &InterfaceInfo = &*(ptr as *const InterfaceInfo);
```

As there's no memory management associated with these structs in C, inventing that on the bindings side is going to be problematic.

glib/src/subclass/type_module.rs
sdroege: I checked the C code and it indeed requires implementations of these. In that case, panicking here is correct.

glib/src/subclass/type_module.rs
sdroege: Panicking if zero sounds like a good approach if that solves the problem here.

With the ref/unref as before, wouldn't this cause the type module to never be disposed?

glib-macros/src/lib.rs
bilelmoussaoui: There are trade-offs for both options but i think having a separate macro might make more sense as very few people would end up using it in general and it would make the documentation of object_subclass just more complex. On the other hand, we will have more macros and people will have a harder time to figure out when to use what

sdroege: Let's keep it a separate macro then :)

glib/src/gobject/auto/type_plugin.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+mod sealed {
+    pub trait Sealed {}
+    impl<T: super::IsA<super::TypePlugin>> Sealed for T {}
+}
+
+pub trait TypePluginExt: IsA<TypePlugin> + sealed::Sealed + 'static {
+    #[doc(alias = "g_type_plugin_complete_interface_info")]
+    fn complete_interface_info(
```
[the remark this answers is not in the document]
sdroege: What I mean is that this PR adds quite a bit of public API that should probably never be called by user code but only by the macro, so maybe it can be moved directly into the macro instead and call the FFI API there.

`TypePlugin` / TypeModule` and their ext traits with the (useful) public API should still exists of course.

glib/src/subclass/type_module.rs
sdroege: So that means we should also just leak the instance here ourselves or not? You could simply call `g_object_ref()` once and never unref it again, that seems simpler IMHO.

glib/src/gobject/interface_info.rs
sdroege: You can't have `ToGlibPtr::to_glib_full()` for these and also not `FromGlibPtrNone` and `IntoGlib` and `FromGlib`. Those implementations conceptually make no sense.

For the time being you will have to implement the functions with these types manually to get the correct behaviour (borrowing).

sdroege: This is very basic infrastructure so it's quite expected that gir won't do the right thing here and manual bindings are needed in many places.

glib/src/subclass/type_module.rs
sdroege: That sounds reasonable

glib/src/gobject/type_info.rs
```diff
@@ -0,0 +1,6 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[derive(Debug)]
+#[doc(alias = "GTypeInfo")]
+#[repr(transparent)]
+pub struct TypeInfo(pub gobject_ffi::GTypeInfo);
```
sdroege, line 6:
```suggestion
pub struct TypeInfo(pub(crate) gobject_ffi::GTypeInfo);
```

Maybe, plus an `as_ptr()` function that returns the underlying FFI pointer?

glib/src/subclass/type_plugin.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+pub trait TypePluginImplExt: ObjectSubclass {
+    fn parent_use_plugin(&self);
+    fn parent_unuse_plugin(&self);
+    fn parent_complete_type_info(
+        &self,
+        type_: Type,
+        info: &mut TypeInfo,
```
sdroege, line 44: Why does this have to be mutable? There's nothing we (can) change in it, is there?

glib/src/gobject/type_plugin.rs
```diff
@@ -0,0 +1,83 @@
...
+            );
+        }
+    }
+
+    #[doc(alias = "g_type_plugin_complete_type_info")]
+    fn complete_type_info(
+        &self,
+        g_type: crate::types::Type,
+        info: &mut TypeInfo,
+        value_table: &mut TypeValueTable,
```
sdroege, line 49: Why are these two mutable? (also in the other places)

glib/src/gobject/type_value_table.rs
```diff
@@ -0,0 +1,6 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[derive(Debug)]
+#[doc(alias = "GTypeValueTable")]
+#[repr(transparent)]
+pub struct TypeValueTable(pub gobject_ffi::GTypeValueTable);
```
sdroege, line 6:
```suggestion
pub struct TypeValueTable(pub(crate) gobject_ffi::GTypeValueTable);
```

maybe

sdroege (commented):
Seems mostly fine to me now. If this is the best API design for this will have to be figured out by someone actually using it, but it works for now apparently :)

glib/src/gobject/interface_info.rs
```diff
@@ -3,4 +3,24 @@
 #[derive(Debug)]
 #[doc(alias = "GInterfaceInfo")]
 #[repr(transparent)]
-pub struct InterfaceInfo(pub gobject_ffi::GInterfaceInfo);
+pub struct InterfaceInfo(pub(crate) gobject_ffi::GInterfaceInfo);
```
bilelmoussaoui, line 6:
```suggestion
pub struct InterfaceInfo(gobject_ffi::GInterfaceInfo);
```

as you would be from_glib_ptr_borrow_mut to construct them anyway, right?

glib/src/subclass/type_plugin.rs
bilelmoussaoui: that sounds like this vfunc should be returning `(glib::TypeInfo, glib::TypeValueTable)` then

sdroege: My question was more how an implementation in Rust would even create or modify a `TypeInfo` right now. But I see, that makes sense for future extensibility then.

Is the `TypeInfo` passed in by GObject here already correctly initialized, or is the implementer supposed to return a new one? That would basically be what decides if the mutable references or the return tuple is the right choice.

glib/src/gobject/interface_info.rs
sdroege: Seems fine to me like this, but a constructor `pub(crate)` would also work and reduce the code a bit

glib/src/gobject/type_module.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl<O: IsA<TypeModule>> TypeModuleExt for O {}
+
+impl fmt::Display for TypeModule {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("TypeModule")
+    }
+}
```
bilelmoussaoui, line 119: Please remove all those Display implementations, we have dropped them in the main branch

bilelmoussaoui (commented):
looks good to me otherwise, thanks!

bilelmoussaoui (dismissed):

sdroege:
> Let me know if I have to modify this PR (needs some work) or if you prefer me to create a new one.

Seems fine in a separate PR.

glib/src/subclass/type_plugin.rs
```diff
@@ -0,0 +1,323 @@
...
+            fn register_dynamic_type(
+                &self,
+                parent_type: Type,
+                type_name: &str,
+                type_info: &TypeInfo,
+                flags: TypeFlags,
+            ) -> Type {
+                let type_ = unsafe {
+                    let type_name = CString::new(type_name).unwrap();
+                    let mut g_type = gobject_ffi::g_type_from_name(type_name.as_ptr());
```
sdroege, line 273:
```suggestion
                    let mut g_type = gobject_ffi::g_type_from_name(type_name.to_glib_none().0);
```

or

```suggestion
                    let mut g_type = type_name.run_with_gstr(|type_name| gobject_ffi::g_type_from_name(type_name.as_ptr()));
```

also elsewhere

glib/src/subclass/mod.rs
```diff
@@ -197,7 +197,220 @@
...
+//!         fn register_dynamic_enum(&self,_: &str, _: &glib::EnumValue) -> glib::Type {
+//!             unimplemented!()
+//!         }
+//!
+//!         fn register_dynamic_flags(&self,_: &str, _: &glib::FlagsValue) -> glib::Type {
+//!             unimplemented!()
+//!         }
+//!
+//!         fn register_dynamic_type(&self, parent_type: glib::Type, type_name: &str, type_info: &glib::TypeInfo, flags: glib::TypeFlags) -> glib::Type {
+//!             let type_ = unsafe {
```
sdroege, line 361: I wonder if there wouldn't be some more convenient (even if `unsafe`) API that wouldn't require user code to do FFI directly.

`g_type_from_name()` for example has safe wrappers, `g_type_register_dynamic()` should be possible to wrap in something a bit more convenient at least (taking non-FFI types, taking care of the C string conversion, ...)

sdroege:
Once this is ready, please also squash it all into fewer commits that are each one functional and useful change (i.e. in your case I guess a single commit)

glib/src/types.rs
```diff
@@ -223,6 +223,48 @@ impl Type {
         }
     }
 
+    #[doc(alias = "g_type_get_plugin")]
+    pub fn get_plugin(self) -> Option<TypePlugin> {
```
sdroege, line 227:
```suggestion
    pub fn plugin(self) -> Option<TypePlugin> {
```

sdroege (commented):
One last thing, otherwise this is IMHO ready. @bilelmoussaoui ?

glib/src/subclass/interface.rs
```diff
@@ -3,25 +3,28 @@
 use std::{marker, mem};
 
 use super::{InitializingType, Signal};
-use crate::{prelude::*, translate::*, Object, ParamSpec, Type};
+use crate::{
+    gobject::traits::DynamicObjectRegisterExt, prelude::*, translate::*, Object, ParamSpec, Type,
```
bilelmoussaoui, line 7:
```suggestion
    prelude::*, translate::*, Object, ParamSpec, Type,
```
glib's prelude should include gobject's one

bilelmoussaoui (commented):
otherwise lgtm, thanks

sdroege (approved):

bilelmoussaoui (approved):

PR #1171, opened 2023-08-18
Use associated type in memory managers

ranfdev:
It makes it clearer that an `XMemoryManager` only manages a single target type.


bilelmoussaoui (approved):

sdroege (approved):

sdroege:
I'd ideally like to get rid of the memory manager stuff btw. It doesn't really bring any advantage and just adds more things for the compiler to optimize away.

PR #1172, opened 2023-08-18
Pin the Fedora version in Dockerfile

chriskilding:
Pin the `fedora` base image version to `38` in the Dockerfile, to ensure that CI builds which use the `gtk-rs` Docker images are reproducible.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM fedora:latest
+FROM fedora:38
 
 RUN dnf update -y && \
     dnf install wget git meson cmake gcc gcc-c++ \


bilelmoussaoui:
We might need the most recent build tools in certain cases as we depend on bleeding edge everything to test the newest additions to the libraries we provide bindings for. So, I don't think this change makes sense.

bilelmoussaoui:
The image is only meant for bindings CI, using it for anything else is your own responsibility.

chriskilding:
Thanks for clarifying that. The idea was to use the gtk-rs image as the basis for Relm4 app builds on GitHub Actions; that way, these apps would not be limited to using the GTK / GNOME versions present on the current Ubuntu LTS release. I believe the same thing could work for building apps that use gtk-rs directly.

Would you be willing to support these use cases, either by enabling the gtk-rs image to be generally used for downstream app builds, or with a different solution e.g. by creating a parallel image with stable versions on it?

bilelmoussaoui:
> Thanks for clarifying that. The idea was to use the gtk-rs image as the basis for Relm4 app builds on GitHub Actions; that way, these apps would not be limited to using the GTK / GNOME versions present on the current Ubuntu LTS release. I believe the same thing could work for building apps that use gtk-rs directly.
> 
> Would you be willing to support these use cases, either by enabling the gtk-rs image to be generally used for downstream app builds, or with a different solution e.g. by creating a parallel image with stable versions on it?

You are free to build an image yourself, the dockerfile & the CI file are available for people to copy from but we have no power to get involved in such things.

bilelmoussaoui:
Let us close this; feel free to re-open if you have stronger motivations

PR #1177, opened 2023-09-06
glib: Remove `#[doc(hidden)]` from `once_cell` and `bitflags` re-export

RealKC:
The blogpost[1] mentions that these are re-exported, so to avoid causing confusion, make it so these appear in documentation.

[1]: https://gtk-rs.org/blog/2023/08/28/new-release.html#switch-to-bitflags-20

---

Alternatives:
- do nothing
- remove the note about the two crates being re-exported from the blogpost

### DIFF
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -4,7 +4,6 @@
 #![allow(clippy::missing_safety_doc)]
 #![doc = include_str!("../README.md")]
 
-#[doc(hidden)]
 pub use bitflags;
 pub use ffi;
 #[doc(hidden)]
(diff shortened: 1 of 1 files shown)


bilelmoussaoui (approved):
I am fine with "publicly" re-exporting them. Even if I am looking forward to Lazy being in the std so we can fully drop the once_cell usage.

RealKC:
> Even if I am looking forward to Lazy being in the std so we can fully drop the once_cell usage.

Yeah that's something I'm looking forward for my personal projects as well

sdroege (approved):

PR #1179, opened 2023-09-07
CI: add a cargo deny job

### LINKED ISSUE #1178: ci: Run cargo-deny on the CI (regularly) and maintain a deny.toml
This will make sure that we don't have duplicate dependencies in different versions, only depend on crates with suitable licenses (i.e. non-copyleft OSI licenses here) and also don't depend on anything that ships with dubious binaries (hello serde-derive).

In gstreamer-rs / gst-plugins-rs I'm already using that since a few years and it's very useful.

I don't know what Github Actions offers here but I guess someone already wrote an extension or even bot (like dependabot) for this.

CC @bilelmoussaoui @GuillaumeGomez

bilelmoussaoui:
Fixes #1178

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -138,6 +138,12 @@ jobs:
           command: fmt
           args: --all -- --check
 
+  cargo-deny:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: EmbarkStudios/cargo-deny-action@v1
+
   generator:
     name: regen check
     runs-on: ubuntu-latest
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,43 @@
+exclude = [
+    "gtk-rs-examples",
+]
+
+[advisories]
+db-path = "~/.cargo/advisory-db"
+db-urls = ["https://github.com/rustsec/advisory-db"]
+vulnerability = "deny"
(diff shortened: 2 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -138,6 +138,12 @@ jobs:
           command: fmt
           args: --all -- --check
 
+  cargo-deny:
+    runs-on: ubuntu-latest
```
sdroege, line 142: Is it possible to in addition run this every day or once a week or so?

bilelmoussaoui: It should be yes, I will look into it. But not sure if we will receive some form of notification when it fails

sdroege: I get a mail whenever a build on `master` fails, so I would expect that to happen. To be checked :)

sdroege: Should we just merge this for now and worry about the scheduled job independently? I don't know how hard that is to add.

bilelmoussaoui: It should be easy, I just didn't had the time for it yet. Hopefully this weekend

sdroege: Let's merge it as-is for now then? @bilelmoussaoui

bilelmoussaoui: yeah, sounds good to me

PR #1192, opened 2023-09-18
glib-macros: Mark property getters as #[must_use]

### LINKED ISSUE #1186: [BUG] property macro  #[must_use] since 0.18.2
**Bug description**

After a `cargo update` which bumped glib-macros to 0.18.2, I've got several of clippy warnings in Relm4:

```
warning: this method could have a `#[must_use]` attribute
   --> relm4/src/binding/bindings.rs:59:17
    |
59  |                 #[property(get, set)]
    |                 ^ help: add the attribute: `#[must_use] #`
...
101 | binding!(StringBinding, "StringBinding", String, imp_strin...
    | ------------------------------------------------------------ in this macro invocation
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate
    = note: this warning originates in the macro `binding` (in Nightly builds, run with -Z macro-backtrace for more info)
```

It seems like the macro is using `quote_spanned` since https://github.com/gtk-rs/gtk-rs-core/pull/1149, which makes clippy think that the user forget to add a `#[must_use]` if you enabled the `clippy::must_use_candidate` lint for your project. `quote_spanned` might help to improve the error reporting, but it can also trigger unexpected clippy lints.

To solve this, either `#[must_use]` should be added to the generated methods or the spans have to be adjusted carefully to make it clear for clippy that those tokens were produced by a macro.

**Steps to reproduce**

1. Use the property macro
2. Enable the `clippy::must_use_candidate` lint
3. Rust `cargo clippy`

AaronErhardt:
This PR should fix #1186 

I'm still not convinced that `quote_spanned` is correctly used (because there are obviously a lot of tokens which have nothing to do with the span of the attribute), but I don't know the reasoning behind the code, so adding `#[must_use]` to the codegen is the easiest fix and actually a useful addition even if you don't use `clippy::must_use_candidate`.

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -566,9 +566,11 @@ fn expand_impl_getset_properties(props: &[PropDesc]) -> Vec<syn::ImplItemFn> {
 
         let getter = p.get.is_some().then(|| {
             let span = p.attrs_span;
-            parse_quote_spanned!(span=> pub fn #ident(&self) -> <#ty as #crate_ident::Property>::Value {
-                 self.property::<<#ty as #crate_ident::Property>::Value>(#stripped_name)
-            })
+            parse_quote_spanned!(span=>
+                #[must_use]
+                pub fn #ident(&self) -> <#ty as #crate_ident::Property>::Value {
+                    self.property::<<#ty as #crate_ident::Property>::Value>(#stripped_name)
+                })
         });
 
         let setter = (p.set.is_some() && !p.is_construct_only).then(|| {


ranfdev:
`quote_spanned` is required mostly when dealing with user input.

In this piece of code, the cast `<#ty as #crate_ident::Property>` could fail, and without a span the user is left with a seemingly random error without any direction on what's causing it.

I added the span to the entire generated code without knowing about the possible consequences with clippy... I agree that `quote_spanned` should be used more granularly. 

I don't agree on adding `#[must_use]` on a getter though, because the returned value doesn't have to be used at all.

AaronErhardt:
> I don't agree on adding #[must_use] on a getter though, because the returned value doesn't have to be used at all.

What's the point of using a getter without using the value?

ranfdev:
A custom getter could trigger a side effect and you may not need the actual value.
It's an edge case but it's technically true.

Apart from that edge case, in the general case rust getters don't have a `#[must_use]` attribute. 

`#[must_use]` should only be used if discarding the value triggers a different behavior than using the value, as with a `Future`. If you use a Future, you get a side effect, if you don't use it, the Future is simply discarded.

In our case whether we use the returned value or not, nothing changes, so adding `#[must_use]` is incorrect and confusing.

AaronErhardt:
Alright, that makes sense. I just assumed that `#[must_use]` was supposed to be used more often, likely because I though the clippy lint should suggest idiomatic code, but it's under "pedantic" after all.

I guess a proper fix would then be about keeping the function declaration largely mixed site and only use the attribute span for the type cast.

ranfdev:
Ok, I've read more about `#[must_use]`.
Apparently it is adopted in more cases than I had anticipated, even in the stdlib (https://blog.rust-lang.org/2018/06/21/Rust-1.27.html#must_use-on-functions and https://std-dev-guide.rust-lang.org/policy/must-use.html).

Considering that, most of the times, the user calls a getter to actually get and use the returned value and not to produce a side effect, I assume adding `#[must_use]` on a getter is correct.

ranfdev (approved):

bilelmoussaoui (approved):

PR #1199, opened 2023-10-03
Remove explicit linking in param_spec.rs

jeroen:
See the issue in librsvg here: https://gitlab.gnome.org/GNOME/librsvg/-/issues/896

I think this line can be removed, which fixes the ability to statically link the stack on e.g. Windows.

### DIFF
--- a/glib/src/param_spec.rs
+++ b/glib/src/param_spec.rs
@@ -296,7 +296,6 @@ pub unsafe trait ParamSpecType:
 {
 }
 
-#[link(name = "gobject-2.0")]
 extern "C" {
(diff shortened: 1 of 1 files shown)


sdroege:
This was specifically added for MSVC on Windows, see https://github.com/gtk-rs/gtk-rs-core/commit/d14883cbb9f56a5b63af0ab6a600a76d5c6a49c2

sdroege:
https://github.com/gtk-rs/gir/pull/1475 is also kind of related here

bilelmoussaoui:
Let us close this one as it is not the correct solution as mentioned above.

PR #1203, opened 2023-10-11
glib: Implement object class methods via a trait instead of directly …

sdroege:
…on `Class<Object>`

This makes them callable directly from more places and is more in sync with how such methods are implemented in other crates that can't directly implement methods on `Class<T>`.

----

I wonder if this breaks something!

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -3232,12 +3232,13 @@ fn validate_signal_arguments(type_: Type, signal_query: &SignalQuery, args: &mut
     }
 }
 
-impl ObjectClass {
+/// Trait for class methods on `Object` and subclasses of it.
+pub unsafe trait ObjectClassExt {
     // rustdoc-stripper-ignore-next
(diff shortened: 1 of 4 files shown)


glib/src/object.rs
```diff
@@ -3232,12 +3232,13 @@ fn validate_signal_arguments(type_: Type, signal_query: &SignalQuery, args: &mut
     }
 }
 
-impl ObjectClass {
```
bilelmoussaoui, line 3235: Can you mark the type as manual and see if any extra-docs gets embedded?

sdroege: That seems very optimistic, but let's give it a try :)

sdroege: CI passed. Ok to merge?

bilelmoussaoui: let us do that

bilelmoussaoui (approved):

bilelmoussaoui (approved):

PR #1204, opened 2023-10-13
Implement ext trait on IsA<T>, don't generate overridden methods

ranfdev:
Without this, it's impossible to use the generated ext trait on a subclass.

When a property is overridden, the getter and setter is not generated, because that property should be accessed through the parent interface/trait. What do you think about that?

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -332,6 +332,9 @@ impl PropDesc {
             is_construct_only,
         })
     }
+    fn is_overriding(&self) -> bool {
+        self.override_class.is_some() || self.override_interface.is_some()
+    }
 }
 
 fn expand_param_spec(prop: &PropDesc) -> TokenStream2 {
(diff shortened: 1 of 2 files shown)


glib-macros/src/properties.rs
```diff
@@ -723,10 +716,10 @@ pub fn impl_derive_props(input: PropsMacroInput) -> TokenStream {
                 item
             });
         quote! {
-            #trait_def
-            impl #trait_ident for #wrapper_type {
-                #(#impls)*
+            pub trait #trait_ident: #crate_ident::IsA<#wrapper_type> {
```
bilelmoussaoui, line 719: Can we seal the traits similar to how we do for Ext/ExtManual traits?

ranfdev: Considering #1205, we don't have to

bilelmoussaoui (approved):

sdroege (approved):

PR #1210, opened 2023-10-21
glib: Allow variable expansion in format strings passed to bool_error & result_from_gboolean

### LINKED ISSUE #1209: `glib::bool_error!` etc don't allow for variable expansion in format strings
See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/490

Problem is that there's a special case of the macro for a single parameter where it doesn't go through `format!`. We can still have that fast path (in optimized builds) by always using `format_args!` and checking if `as_str()` returns `Some`.

RealKC:
Not sure if this is the best way to go about it

Fixes #1209

### DIFF
--- a/glib/src/error.rs
+++ b/glib/src/error.rs
@@ -177,26 +177,30 @@ macro_rules! bool_error(
 
 #[macro_export]
 macro_rules! result_from_gboolean(
-// Plain strings
-    ($ffi_bool:expr, $msg:expr) => {{
-        $crate::BoolError::from_glib(
-            $ffi_bool,
-            $msg,
-            file!(),
-            $crate::function_name!(),
-            line!(),
-        )
-    }};
-
-// Format strings
     ($ffi_bool:expr, $($msg:tt)*) =>  {{
-        $crate::BoolError::from_glib(
-            $ffi_bool,
-            format!($($msg)*),
-            file!(),
-            $crate::function_name!(),
-            line!(),
-        )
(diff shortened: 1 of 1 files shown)


glib/src/error.rs
```diff
@@ -152,49 +152,55 @@ pub trait ErrorDomain: Copy {
...
-
-// Format strings
     ($($msg:tt)*) =>  {{
-        $crate::BoolError::new(
-            format!($($msg)*),
-            file!(),
-            $crate::function_name!(),
-            line!(),
-        )
+        match ::std::format_args!($($msg)*) {
```
sdroege, line 156: The `match` because the return value of `format_args!` is otherwise annoying to work with?

RealKC: Yeah, I tried storing it in a normal variable using `let` and that caused borrow checker issues

sdroege: Nicer than my workaround for that :) I usually use a closure

sdroege (approved):

sdroege:
Thanks!

RealKC:
@bilelmoussaoui I don't believe this change can be backported as it is actually a breaking change, previously `bool_error!(format!(...))` or `bool_error!(some_variable_holding_a_string)` would've compiled but after this change they will no longer compile due to becoming the first argument to `format_args!` which has to be a literal string afaik

sdroege:
Yes, this can't be backported unfortunately.

PR #1214, opened 2023-10-29
Add spawn_future_with_priority and spawn_future_local_with_priority

alatiera:
Followup to #1201

Allow to set custom priority for the spawned futures.

### DIFF
--- a/glib/src/functions.rs
+++ b/glib/src/functions.rs
@@ -269,8 +269,23 @@ pub fn file_open_tmp(
 pub fn spawn_future<R: Send + 'static, F: std::future::Future<Output = R> + Send + 'static>(
     f: F,
 ) -> crate::JoinHandle<R> {
+    spawn_future_with_priority(crate::source::Priority::DEFAULT, f)
+}
+
+// rustdoc-stripper-ignore-next
+/// Spawn a new infallible `Future` on the main context, with a non-default priority.
+///
+/// This can be called from any thread and will execute the future from the thread
+/// where main context is running, e.g. via a `MainLoop`.
+pub fn spawn_future_with_priority<
+    R: Send + 'static,
+    F: std::future::Future<Output = R> + Send + 'static,
+>(
+    priority: crate::source::Priority,
+    f: F,
+) -> crate::JoinHandle<R> {
     let ctx = crate::MainContext::ref_thread_default();
-    ctx.spawn(f)
+    ctx.spawn_with_priority(priority, f)
 }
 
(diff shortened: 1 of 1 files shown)


ranfdev:
IMHO this is not an idiomatic way to handle the problem. We are polluting the library with a lot of combinations `do_x`, `do_x_with_priority` `do_x_local`, `do_x_with_priority_local`...

The same problems exists for timeouts, and other glib sources.

In #1033 I was trying to build a futures module where you can use a builder pattern to reduce the number of spawn combinations. For example, that PR introduced a `SpawnOption` builder... 

```
    SpawnOptions::new()
         .context(ctx)
         .priority(Priority::HIGH)  
         .spawn_local(async move {});
```

Note: the context is optional, the priority is optional. That's 1 API to handle all the possible spawn combinations. That also resembles the API of `OpenOptions` in the rust std library.

sdroege:
I like that approach @ranfdev

sdroege:
Yes I think these should be replaced by the builder too.

ranfdev:
I would keep the functions #1201 under the futures scope, eg: `glib::futures::spawn_local`.

Only because spawn_local is used 10x more than spawn_local_with_priority, so I think it's nice having a shorthand for the common case.

Also, you can `use` the functions in a module to shorten the call even more, but you cannot do that with a Builder method.

```
// Valid
use glib::futures::spawn_local;
spawn_local(async {});

// No way to shorten the builder as easily
SpawnOptions::new().local().spawn(async {});
```

ranfdev:
Ah, I didn't expect that. Can we make every `spawn_local` use the priority `default_idle` by default then? 

For sure I've spawned many futures with `spawn_local` in the past, without thinking they had a higher priority than GTK's rendering. I'm confident a lot of people are making the same mistake.

sdroege:
That's also a more general problem then. `g_timeout_add()` for example also uses `G_PRIORITY_DEFAULT`, as does `g_main_context_invoke()`, and basically every `GSource` in GLib except for the idle one.

If that's not suitable for GTK applications then something will have to change in GTK.

Hofer-Julian:
> Ah, I didn't expect that. Can we make every spawn_local use the priority default_idle by default then?

Like @sdroege said ideally that would be fixed on the GTK side.
At the very least, we should double-check with a GTK person that `default_idle` is even a good default.

sdroege:
It doesn't even seem like a good default for GTK. It depends on how you want your future to be ordered in relation to the drawing operations. It probably makes sense to export the GDK priority as a constant though.

Issue #1213, opened 2023-10-27
Consider removing channels from glib


Hofer-Julian:
As mentioned by @sdroege in https://github.com/gtk-rs/gtk4-rs/pull/1521#issuecomment-1774071454:

>Should we maybe remove the GLib channel? It mostly exists from the times before async was mature and because people instead used timeout callbacks to periodically check if a non-async channel contains something.

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/1216

p-kraszewski:
Splendid move! Could you consider using Rust's [deprecated](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) attribute _next time_, let's say, for 2 releases?

bilelmoussaoui:
It was deprecated during the release cycle of 0.18.x.

p-kraszewski:
Ok. My bad. 

* First note on deprecation (merged into 0.18.4): 2023.11.25  93f6a02a32d6e2a532cdb3a8922dbb9b31419a93
* Feature reaped (merged into 0.19 line): 2023.11.31 2cb10ec3f305f954ca30c708c752458e4af4eabb

sdroege:
First release without the API was on Sunday, so that gives a bit more than 2 months.

PR #1216, opened 2023-10-30
glib: Remove `MainContext::channel()`

sdroege:
Async code generally allows for more flexible message handling than the callback based MainContext channel.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1213


bilelmoussaoui (approved):
Less code to maintain

GuillaumeGomez (approved):

PR #1218, opened 2023-11-02
gio: fix UnixSocketAddress constructor with a path

pbor:
as_os_str().len() does not return the number of bytes in the string, but the full allocation size. Just pass -1 and let glib calculate the len.

### DIFF
--- a/gio/src/unix_socket_address.rs
+++ b/gio/src/unix_socket_address.rs
@@ -46,20 +46,21 @@ impl UnixSocketAddress {
         use self::UnixSocketAddressPath::*;
 
         let type_ = address_type.to_type();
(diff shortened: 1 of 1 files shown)


pbor:
Marked as draft because even with the length fixed by passing -1 to glib, the unit test still fails.
To check what is going on I tried to print `std::ffi::CStr::from_ptr(path.to_glib_none().0)` for the Path case and it shows garbage

pbor:
(Also I need to drop the gir and gir-files changes, but that is secondary)

pbor:
The problem was that the stash did not live long enough

sdroege (approved):

PR #1220, opened 2023-11-03
Add support of enums as dynamic types

fbrouille:
This PR aims to add support of enums registered as dynamic types. See [https://docs.gtk.org/gobject/class.TypeModule.html](https://docs.gtk.org/gobject/class.TypeModule.html)  and [https://docs.gtk.org/gobject/iface.TypePlugin.html](https://docs.gtk.org/gobject/iface.TypePlugin.html).
The new macro `dynamic_enum_derive` allows to register enums that are part of a module or of a plugin (e.g. a shared library on linux). The macro mimic `enum_derive` but generates code that support registration when the plugin (`TypePlugin`) is used, following the behavior defined in Glib doc. However it is possible to postpone the registration at first use of the enum (like within `enum_derive`) by explicitly setting the macro attribute `lazy_registration = true`.
Some examples in [glib-macros/tests/dynamic_enums.rs](https://github.com/fbrouille/gtk-rs-core/blob/68da73d2a5a1292ac9f0c1ebfc9a11e61feb0330/glib-macros/tests/dynamic_enums.rs)

### DIFF
--- a/glib-macros/src/enum_derive.rs
+++ b/glib-macros/src/enum_derive.rs
@@ -1,9 +1,9 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use heck::{ToKebabCase, ToUpperCamelCase};
+use heck::{ToKebabCase, ToShoutySnakeCase, ToUpperCamelCase};
 use proc_macro2::TokenStream;
 use proc_macro_error::abort_call_site;
-use quote::{quote, quote_spanned, ToTokens};
+use quote::{format_ident, quote, quote_spanned, ToTokens};
 use syn::{
     parse_quote, punctuated::Punctuated, spanned::Spanned, token::Comma, Data, ExprArray, Ident,
     Variant,
@@ -188,37 +188,44 @@ pub fn impl_dynamic_enum(input: &syn::DeriveInput) -> TokenStream {
         // registers the enum as a dynamic type on the first use (lazy registration).
         // a weak reference on the plugin is stored and will be used later on the first use of the enum.
         // this implementation relies on a static storage of a weak reference on the plugin and of the GLib type to know if the enum has been registered.
+
+        // the registration status type.
+        let registration_status_type = format_ident!("{}RegistrationStatus", name);
+        // name of the static variable to store the registration status.
+        let registration_status = format_ident!(
+            "{}",
+            registration_status_type.to_string().to_shouty_snake_case()
+        );
+        // name of the static array to store the enumeration values.
+        let enum_values_array = format_ident!("{}_VALUES", name.to_string().to_shouty_snake_case());
         quote! {
-            struct RegistrationStatus(<#plugin_ty as #crate_ident::clone::Downgrade>::Weak, #crate_ident::Type);
-            unsafe impl Send for RegistrationStatus {}
+            /// The registration status type: a tuple of the weak reference on the plugin and of the GLib type.
+            struct #registration_status_type(<#plugin_ty as #crate_ident::clone::Downgrade>::Weak, #crate_ident::Type);
+            unsafe impl Send for #registration_status_type {}
+
+            /// The registration status protected by a mutex guarantees so that no other threads are concurrently accessing the data.
+            static #registration_status: ::std::sync::Mutex<Option<#registration_status_type>> = ::std::sync::Mutex::new(None);
+
+            /// Array of `EnumValue` for the possible enumeration values.
+            static #enum_values_array: #enum_values;
 
             impl #name {
-                /// Returns a mutable reference to the registration status: a tuple of the weak reference on the plugin and of the GLib type.
-                /// This is safe because the mutable reference guarantees that no other threads are concurrently accessing the data.
-                #[inline]
-                fn get_registration_status_ref() -> &'static ::std::sync::Mutex<Option<RegistrationStatus>> {
-                    static REGISTRATION_STATUS: ::std::sync::Mutex<Option<RegistrationStatus>> = ::std::sync::Mutex::new(None);
-                    &REGISTRATION_STATUS
-                }
-
                 /// Registers the enum as a dynamic type within the plugin only once.
                 /// Plugin must have been used at least once.
                 /// Do nothing if plugin has never been used or if the enum is already registered as a dynamic type.
                 #[inline]
                 fn register_enum() -> #crate_ident::Type {
-                    let registration_status_ref = Self::get_registration_status_ref();
-                    let mut registration_status = registration_status_ref.lock().unwrap();
+                    let mut registration_status = #registration_status.lock().unwrap();
                     match ::std::ops::DerefMut::deref_mut(&mut registration_status) {
                         // plugin has never been used, so the enum cannot be registered as a dynamic type.
                         None => #crate_ident::Type::INVALID,
                         // plugin has been used and the enum has not been registered yet, so registers it as a dynamic type.
-                        Some(RegistrationStatus(type_plugin, type_)) if !type_.is_valid() => {
-                            static VALUES: #enum_values;
-                            *type_ = <#plugin_ty as glib::prelude::DynamicObjectRegisterExt>::register_dynamic_enum(type_plugin.upgrade().unwrap().as_ref(), #gtype_name, VALUES.as_ref());
+                        Some(#registration_status_type(type_plugin, type_)) if !type_.is_valid() => {
+                            *type_ = <#plugin_ty as glib::prelude::DynamicObjectRegisterExt>::register_dynamic_enum(type_plugin.upgrade().unwrap().as_ref(), #gtype_name, #enum_values_array.as_ref());
                             *type_
                         },
                         // plugin has been used and the enum has already been registered as a dynamic type.
-                        Some(RegistrationStatus(_, type_)) => *type_
+                        Some(#registration_status_type(_, type_)) => *type_
                     }
                 }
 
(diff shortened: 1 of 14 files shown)


pbor:
naive question: why not a `dynamic = true` attribute to the existing macro?

sdroege (commented):
That's a lot of code but I also don't see how to simplify that, right now :)

Can you also add the same for flags once we agreed on an API / implementation here?

@bilelmoussaoui What do you think about the API?

sdroege:

glib-macros/src/enum_derive.rs
```diff
@@ -3,24 +3,31 @@
...
 use proc_macro2::TokenStream;
 use proc_macro_error::abort_call_site;
-use quote::{quote, quote_spanned};
+use quote::{quote, quote_spanned, ToTokens};
 use syn::{punctuated::Punctuated, spanned::Spanned, token::Comma, Data, Ident, Variant};
 
 use crate::utils::{crate_ident_new, gen_enum_from_glib, parse_nested_meta_items, NestedMetaItem};
 
-// Generate glib::gobject_ffi::GEnumValue structs mapping the enum such as:
+// If `wrap` is `false`, generates glib::gobject_ffi::GEnumValue structs mapping the enum such as:
```
sdroege, line 11: For simplification, can't we generate the wrapped one always if that's needed somewhere? The non-wrapped one can be directly pulled out of it again if needed.

glib-macros/src/enum_derive.rs
```diff
@@ -81,10 +100,228 @@ pub fn impl_enum(input: &syn::DeriveInput) -> TokenStream {
...
+        // registers the enum as a dynamic type on the first use (lazy registration).
+        // a weak reference on the plugin is stored and will be used later on the first use of the enum.
+        // this implementation relies on a static storage of a weak reference on the plugin and of the glib type to know if the enum has been registered.
+        quote! {
+            impl #name {
+                /// Returns a mutable reference to the registration status: a tuple of the weak reference on the plugin and of the glib type.
+                /// This is safe because the mutable reference guarantees that no other threads are concurrently accessing the data.
+                #[inline]
+                fn get_registration_status_ref_mut() -> &'static mut Option<(<#plugin_ty as #crate_ident::clone::Downgrade>::Weak, #crate_ident::Type)> {
+                    static mut REGISTRATION_STATUS: ::std::sync::Mutex<Option<(<#plugin_ty as #crate_ident::clone::Downgrade>::Weak, #crate_ident::Type)>> = ::std::sync::Mutex::new(None);
```
sdroege, line 207: Could this be a `OnceCell`? No, because it can be registered multiple times?

glib-macros/src/enum_derive.rs
```diff
@@ -81,10 +100,228 @@ pub fn impl_enum(input: &syn::DeriveInput) -> TokenStream {
...
+                    }
+                }
+            }
+        }
+    } else {
+        // registers immediately the enum as a dynamic type.
+        quote! {
+            impl #name {
+                /// Returns a mutable reference to the glib type.
+                /// This is safe because the mutable reference guarantees that no other threads are concurrently accessing the atomic data.
```
sdroege, line 283: I don't think that's true. There's nothing preventing multiple threads from calling this function in parallel and each getting their own mutable reference.

This seems like it needs some changes

glib-macros/src/lib.rs
```diff
@@ -468,6 +468,75 @@ pub fn enum_derive(input: TokenStream) -> TokenStream {
     gen.into()
 }
 
+/// Derive macro for register a rust enum in the glib type system as a dynamic
```
sdroege, line 471:
```suggestion
/// Derive macro for register a Rust enum in the GLib type system as a dynamic
```

everywhere

glib-macros/src/lib.rs
```diff
@@ -468,6 +468,75 @@ pub fn enum_derive(input: TokenStream) -> TokenStream {
...
+/// ```ignore
+/// #[derive(Debug, Copy, Clone, PartialEq, Eq, glib::DynamicEnum)]
+/// #[enum_type(name = "MyEnum", lazy_registration = true)]
+/// enum MyEnum {
+///     ...
+/// }
+/// ```
+///
+/// By default an enum is considered to be registered as a dynamic type within
+/// a [`TypeModule`] subclass. Optionally setting the macro attribute
```
sdroege, line 517: Maybe add a bit more details / example for what it means to be "considered to be registered as a dynamic type within a TypeModule" :)

glib-macros/tests/dynamic_enums.rs
```diff
@@ -0,0 +1,494 @@
...
+
+        impl TypePluginImpl for MyModule {}
+
+        impl TypeModuleImpl for MyModule {
+            fn load(&self) -> bool {
+                // registers enums as dynamic types.
+                let my_module = self.obj();
+                let type_module: &glib::TypeModule = my_module.upcast_ref();
+                super::MyModuleEnum::on_implementation_load(type_module)
+                    && super::MyModuleEnumLazy::on_implementation_load(type_module)
```
sdroege, line 33: What happens if the user forgets to call these here but then something tries to use the enums?

glib-macros/tests/dynamic_enums.rs
```diff
@@ -0,0 +1,494 @@
...
+                super::MyPluginEnumLazy::on_implementation_unload(my_plugin.as_ref());
+                super::MyPluginEnum::on_implementation_unload(my_plugin.as_ref());
+            }
+
+            fn complete_type_info(
+                &self,
+                type_: glib::Type,
+            ) -> (glib::TypeInfo, glib::TypeValueTable) {
+                let enum_type_values = match type_ {
+                    type_ if type_ == super::MyPluginEnum::static_type() => {
```
sdroege, line 258: Same question here. What happens if the user forgets this and then the enum is tried to be used?

glib/src/enums.rs
```diff
@@ -187,6 +187,33 @@ impl EnumClass {
...
     }
+
+    // rustdoc-stripper-ignore-next
+    /// Complete `TypeInfo` for an enum with values.
+    /// This is an associated function. A method would result in a stack overflow due to a recurvice call:
+    /// callers should first create an `EnumClass` instance by calling `EnumClass::with_type()` which indirectly
+    /// calls `TypePluginRegisterImpl::register_dynamic_enum()` and `TypePluginImpl::complete_type_info()`
+    /// and one of them should call `EnumClass::with_type()` before calling this method.
+    #[doc(alias = "g_enum_complete_type_info")]
+    pub fn type_info(type_: Type, const_static_values: &'static [EnumValue]) -> Option<TypeInfo> {
```
sdroege, line 198:
```suggestion
    pub fn complete_type_info(type_: Type, const_static_values: &'static [EnumValue]) -> Option<TypeInfo> {
```

maybe?

glib/src/enums.rs
```diff
@@ -227,6 +254,12 @@ impl fmt::Debug for EnumValue {
 }
 
 impl EnumValue {
+    // rustdoc-stripper-ignore-next
+    /// Creates a new `EnumValue` containing `value`.
+    pub const fn new(value: gobject_ffi::GEnumValue) -> Self {
```
sdroege, line 259: I think this should at least be unsafe. You can create a `GEnumValue` with random integers for the pointers, and then safe usage of that would cause memory corruption :)

sdroege: Maybe we can make use of `glib::UnsafeFrom` here? Basically, this feels like it should get very similar treatment to how `BoxedInline` works

glib/src/gobject/type_module.rs
```diff
@@ -38,6 +38,21 @@ pub trait TypeModuleExt: IsA<TypeModule> + sealed::Sealed + 'static {
         }
     }
 
+    #[doc(alias = "g_type_module_register_enum")]
+    fn register_enum(
+        &self,
+        name: &str,
+        const_static_values: &'static [EnumValue],
```
sdroege, line 45: These functions all assume that the last `EnumValue` is the terminator, but that's not necessarily the case and if this is called otherwise you'll get a crash

glib/src/subclass/type_plugin.rs
```diff
@@ -184,7 +184,11 @@ pub trait TypePluginRegisterImpl: ObjectImpl + TypePluginImpl {
         _interface_type: Type,
         _interface_info: &InterfaceInfo,
     );
-
+    fn register_dynamic_enum(
+        &self,
+        _name: &str,
+        _const_static_values: &'static [EnumValue],
+    ) -> Type;
```
sdroege, line 191: Should this maybe get a default impl that just does `unimplemented!()`?

sdroege:
Sorry for taking so long to review this!

glib-macros/src/enum_derive.rs
sdroege: Looks like a good approach, thanks :) Feel free to send the PR for fixing that in the other cases now already

glib-macros/src/lib.rs
sdroege: Thanks!

glib-macros/tests/dynamic_enums.rs
sdroege: Can we only do that for dynamic enums? It seems like unnecessary code bloat otherwise (all the assertions inserted everywhere with their stack unwinding machinery). Otherwise seems like a good approach.

However, there's also nothing really wrong with having these traits work if the type is not registered yet, is there? Things only have to fail if a `GType` is needed, and that should already lead to assertions. So maybe that's just not necessary at all?

glib/src/enums.rs
sdroege: Sounds good to me

glib/src/gobject/type_module.rs
sdroege: Or alternatively the function would have to be `unsafe` but that would seems unfortunate.

Maybe you could just add an assertion here for the last element being the terminator?

glib-macros/tests/dynamic_enums.rs
sdroege: Maybe :sweat_smile: I think this behaviour makes more sense, but time will tell us!

glib/src/gobject/type_module.rs
sdroege: Hm, actually we would panic when trying to get the name or nick of such a `EnumValue`, wouldn't we? Maybe we can have the macro create some wrapper type that ensures the terminator and when dereffing would give a `&[EnumValue]` without the terminator?

sdroege: Sounds good

sdroege: `EnumValues` should probably implement `Deref<Target = [EnumValue]>` so that a slice without the terminator is returned

sdroege: And maybe not re-exported from the crate root but make the `enum` module public instead. There should rarely be a reason for a user to type out the name, so having it outside the crate root reduces some noise

sdroege: Otherwise looks good to me, thanks :)

sdroege (approved):
Good for me. @bilelmoussaoui @pbor last chance :) I'll merge it during the weekend if there are no objections.

sdroege:
> Maybe another PR to address it (and also dynamic objects, interfaces)

Let's do that separately, yes. Unless you prefer to wait? Let me know. Otherwise I would merge now

sdroege (approved):

PR #1226, opened 2023-11-17
Move from unmaintained winapi crate to windows-sys

sdroege:
windows-sys / windows is also used by the gdk4-win32 crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,18 +162,18 @@ dependencies = [
 
 [[package]]
 name = "clap"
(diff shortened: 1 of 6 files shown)


sdroege:
CI failures are unrelated. Need to figure out something for the new clippy, as usual...

sdroege:
@GuillaumeGomez @bilelmoussaoui Any opinions here?

GuillaumeGomez:
Seems like the right thing to do to me. :+1:

PR #1231, opened 2023-11-24
glib: Deprecate main context channel

sdroege:
It was already removed in git master and giving an advance warning seems useful so people already port now if possible.

### DIFF
--- a/glib/src/main_context_channel.rs
+++ b/glib/src/main_context_channel.rs
@@ -539,6 +539,7 @@ impl MainContext {
     /// will fail.
     ///
     /// The returned `Sender` behaves the same as `std::sync::mpsc::Sender`.
+    #[deprecated = "Use an async channel on the main context instead"]
     pub fn channel<T>(priority: Priority) -> (Sender<T>, Receiver<T>) {
         let channel = Channel::new(None);
         let receiver = Receiver(Some(channel.clone()), priority);
@@ -562,6 +563,7 @@ impl MainContext {
     /// will fail.
     ///
     /// The returned `SyncSender` behaves the same as `std::sync::mpsc::SyncSender`.
+    #[deprecated = "Use an async channel on the main context instead"]
     pub fn sync_channel<T>(priority: Priority, bound: usize) -> (SyncSender<T>, Receiver<T>) {
         let channel = Channel::new(Some(bound));
         let receiver = Receiver(Some(channel.clone()), priority);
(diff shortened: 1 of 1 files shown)


sdroege:
CI failures unrelated. clippy was updated once again :roll_eyes:

glib/src/main_context_channel.rs
```diff
@@ -539,6 +539,7 @@ impl MainContext {
     /// will fail.
     ///
     /// The returned `Sender` behaves the same as `std::sync::mpsc::Sender`.
+    #[deprecated = "Use an async channel on the main context instead"]
```
GuillaumeGomez, line 542: If there is a name for the function/method, that'd be quite helpful I think.

sdroege: There are like 5 different crates that people could use. Not sure we want to prescribe one here.

GuillaumeGomez: Also, why do we want to force people to use async when this code works perfectly fine and is much better to use (imo)? I don't remember the discussion about this.

sdroege: Because it's not much better to use and leads to worse code in practice. Ask all the GNOME apps that moved away from the main context channel.

Callbacks alone are already worse than async Rust, but in the context of GTK it makes everything a lot more awkward. But by us providing this API, people use that first and then wonder why it's all so awful :)

GuillaumeGomez: Intriguing but my opinion doesn't really matter in the end as I'm not that much of gtk-rs user...

GuillaumeGomez (approved):

upsuper:
> There are like 5 different crates that people could use. Not sure we want to prescribe one here.

I have to say I'm quite frustrated by this deprecation, as it is completely unclear what to do.

It would probably be good to have some examples to show how the functionality can be achieved without this API, or at least mention relevant crates and API to use.

For anyone else finding finding this PR, I ended up using `futures_channel`'s [unbounded mpsc channel](https://docs.rs/futures-channel/latest/futures_channel/mpsc/fn.unbounded.html) with `MainContext::default().spawn_local` to replace the use of `MainContext::channel`, which seems to work fine.

sdroege:
@upsuper Does https://github.com/gtk-rs/gtk-rs-core/pull/1242 help? If not, can you suggest an improvement for the deprecation message?

upsuper:
Yes, that's way more helpful than the deprecation message here. It would probably be good to publish that change to crates.io sooner than later, so that others wouldn't have to struggle with figuring out it again.

sdroege:
I'll try getting a bugfix release out one of these days, yes. I think there was another useful bugfix too.

If you have any suggestions what else to improve with that message, now would be a good time :)

PR #1239, opened 2023-12-06
gio: return NULL from spawn_blocking's underlying gtask

ZanderBrown:
This tells GIO that the task did in fact complete, and avoids a log message when the task is finalised

### DIFF
--- a/gio/src/task.rs
+++ b/gio/src/task.rs
@@ -433,9 +433,10 @@ where
     // use Cancellable::NONE as source obj to fulfill `Send` requirement
     let task = unsafe { Task::<bool>::new(Cancellable::NONE, Cancellable::NONE, |_, _| {}) };
     let (join, tx) = JoinHandle::new();
-    task.run_in_thread(move |_, _: Option<&Cancellable>, _| {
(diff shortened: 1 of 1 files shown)


gio/src/task.rs
```diff
@@ -433,10 +433,12 @@ where
     // use Cancellable::NONE as source obj to fulfill `Send` requirement
     let task = unsafe { Task::<bool>::new(Cancellable::NONE, Cancellable::NONE, |_, _| {}) };
     let (join, tx) = JoinHandle::new();
-    task.run_in_thread(move |_, _: Option<&Cancellable>, _| {
-        let res = panic::catch_unwind(panic::AssertUnwindSafe(func));
-        let _ = tx.send(res);
-    });
+    task.clone()
+        .run_in_thread(move |_, _: Option<&Cancellable>, _| {
```
sdroege, line 437:
```suggestion
    task.run_in_thread(move |task, _: Option<&Cancellable>, _| {
```

seems better or am I missing something?

bilelmoussaoui (approved):

PR #1249, opened 2023-12-12
glib: Add upgrade_or_insert{_with} to WeakRefs

ZanderBrown:
Avoids some boilerplate when implementing things like single-window apps, where you don't really care if the object was already set, you just need an object

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -3429,6 +3429,35 @@ impl<T: ObjectType> WeakRef<T> {
             }
         }
     }
+
+    // rustdoc-stripper-ignore-next
+    /// Try to upgrade this weak reference to a strong reference.
+    ///
+    /// If the stored object was already destroyed or no object was set, the
+    /// reference is updated to `value`, which is then returned
+    #[inline]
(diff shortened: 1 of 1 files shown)


glib/src/object.rs
```diff
@@ -3429,6 +3429,35 @@ impl<T: ObjectType> WeakRef<T> {
...
+        self.upgrade_or_insert_with(|| value)
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Try to upgrade this weak reference to a strong reference.
+    ///
+    /// If the stored object was already destroyed or no object was set, the
+    /// reference is updated to the result of `f`, which is then returned
+    #[inline]
+    pub fn upgrade_or_insert_with<F>(&self, f: F) -> T
```
sdroege, line 3449: This is unfortunately racy if called from multiple threads at once. That at least should be documented.

sdroege: I don't think so

bilelmoussaoui:
what is left to be done here? the new APIs looks good to me

PR #1250, opened 2023-12-12
glib: Add get_transformed to Value

ZanderBrown:
Like transform(), but actually gives you T instead of a new Value, which is probably what you wanted

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -566,6 +566,22 @@ impl Value {
     }
 
     // rustdoc-stripper-ignore-next
+    /// Tries to transform the value into the target type
+    #[doc(alias = "g_value_transform")]
+    pub fn get_transformed<T: ValueType>(&self) -> Result<T, crate::BoolError> {
+        self.transform::<T>().and_then(|v| {
+            v.get().map_err(|e| {
+                crate::bool_error!(
+                    "Can't transform value of type '{}' into '{}': {}",
+                    self.type_(),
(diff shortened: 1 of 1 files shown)


glib/src/value.rs
```diff
@@ -565,6 +565,22 @@ impl Value {
         }
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Tries to transform the value into the target type
+    #[doc(alias = "g_value_transform")]
+    pub fn get_transformed<T: ValueType>(&self) -> Result<T, crate::BoolError> {
```
sdroege, line 571: `to_transformed()` maybe? It's a conversion, not a getter.

@pbor what do you think?

pbor: Is it really more convenient than `v.transform().get()`?

pbor: Maybe if we want a shortcut we should implement `TryInto` ?

pbor: Another naming possibility is `transform_to` which has a precedent: `https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/struct.BindingGroupBuilder.html#method.transform_to` though that actually returns `Option<Value>`

bilelmoussaoui: the other option is to rename this function to `transform` and rename the other `transform` function to `transform_as_value` or something like that

bilelmoussaoui:
Any update on this one?

PR #1254, opened 2023-12-22
Fix / work around a couple of new clippy 1.75 warnings

sdroege:


cairo/src/lib.rs
```diff
@@ -228,8 +228,8 @@ pub use crate::{
     enums::*,
     error::{BorrowError, Error, IoError, Result},
     font::{
-        FontExtents, FontFace, FontOptions, FontSlant, FontType, FontWeight, Glyph, ScaledFont,
-        TextCluster, TextExtents, UserFontFace,
```
sdroege, line 232: I would've said that this should get a bugfix release but considering that nobody noticed the lack of those types in the public API for *years*....

pbor (approved):

sdroege:
@GuillaumeGomez Any idea what's with https://github.com/gtk-rs/gtk-rs-core/actions/runs/7300850123/job/19896312494?pr=1254 ? This fails regularly now since recently.

GuillaumeGomez:
Absolutely no idea. :-/

PR #1256, opened 2023-12-24
Fix concurrency issues

fbrouille:
This PR aims to fix the concurrency issues in code generated by the macro helper attributes `object_subclass_dynamic` and `object_interface_dynamic`

### DIFF
--- a/glib-macros/src/object_interface_attribute.rs
+++ b/glib-macros/src/object_interface_attribute.rs
@@ -1,8 +1,9 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use heck::ToShoutySnakeCase;
 use proc_macro2::TokenStream;
 use proc_macro_error::abort_call_site;
-use quote::{quote, ToTokens};
+use quote::{format_ident, quote, ToTokens};
 
 use crate::utils::{parse_optional_nested_meta_items, NestedMetaItem};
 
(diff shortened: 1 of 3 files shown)


glib-macros/src/object_interface_attribute.rs
```diff
@@ -209,29 +230,29 @@ fn register_object_interface_as_dynamic(
...
-                }
+            /// The GLib type which can be safely shared between threads.
+            static #gtype_status: ::std::sync::atomic::AtomicUsize = ::std::sync::atomic::AtomicUsize::new(#crate_ident::gobject_ffi::G_TYPE_INVALID);
 
+            impl #self_ty {
                 /// Do nothing as the object interface has been registered on implementation load.
                 #[inline]
                 fn register_interface() -> #crate_ident::Type {
-                    unsafe { <#crate_ident::Type as #crate_ident::translate::FromGlib<#crate_ident::ffi::GType>>::from_glib(*Self::get_type_mut()) }
+                    let gtype = #gtype_status.load(::std::sync::atomic::Ordering::Relaxed);
```
sdroege, line 245: Is this somewhere protected also via a mutex? Otherwise it would probably be safer to use `SeqCst` here

sdroege (commented):
Otherwise seems good to me

glib-macros/src/object_interface_attribute.rs
sdroege: Yes, `Acquire` / `Release` seems to be enough here

PR #1271, opened 2024-01-07
add support of flags registered as dynamic types

fbrouille:
This PR aims to add support of flags registered as dynamic types. See https://docs.gtk.org/gobject/class.TypeModule.html and https://docs.gtk.org/gobject/iface.TypePlugin.html.
The new macro `flags_dynamic` allows to register the flags that are part of a module or of a plugin (e.g. a shared library on linux). This helper macro generates code that support registration when the plugin (`TypePlugin`) is used, following the behavior defined in Glib doc. However it is possible to postpone the registration at first use of the flags by explicitly setting the macro attribute `lazy_registration = true`.

### DIFF
--- a/glib/Gir_GObject.toml
+++ b/glib/Gir_GObject.toml
@@ -25,6 +25,7 @@ manual = [
     "GObject.Object",
     "GObject.Value",
     "GObject.EnumValue",
+    "GObject.FlagsValue",
     "GObject.TypeValueTable",
     "GObject.ParamFlags",
     "GObject.ParamSpec",
--- a/glib/src/enums.rs
+++ b/glib/src/enums.rs
@@ -243,6 +243,7 @@ impl Clone for EnumClass {
 // rustdoc-stripper-ignore-next
 /// Representation of a single enum value of an `EnumClass`.
 #[doc(alias = "GEnumValue")]
+#[derive(Copy, Clone)]
 #[repr(transparent)]
 pub struct EnumValue(gobject_ffi::GEnumValue);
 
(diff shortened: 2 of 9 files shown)


glib/src/enums.rs
```diff
@@ -1182,6 +1082,94 @@ impl fmt::Display for InvalidFlagsError {
...
+/// enum (or flags) values by this way.
+/// Use `EnumClass::values()` or `EnumClass::value()` to get the enum values.
+/// Use `FlagsClass::values()` or `FlagsClass::value()` to get the flags values.
+#[repr(C)]
+pub struct EnumerationValuesStorage<E: EnumerationValue<E>, const S: usize>([E; S]);
+
+impl<E: EnumerationValue<E>, const S: usize> EnumerationValuesStorage<E, S> {
+    // rustdoc-stripper-ignore-next
+    /// creates a new `EnumerationValuesStorage` with the given values and a final zero value.
+    pub const fn new<const N: usize>(values: [E; N]) -> Self {
```
sdroege, line 1108: This fails compiling if `N + 1 != S`, right?

sdroege (approved):
Looks good to me otherwise. Thanks!

@pbor @bilelmoussaoui what do you think?

glib/Gir_GObject.toml
```diff
@@ -25,6 +25,7 @@ manual = [
     "GObject.Object",
     "GObject.Value",
     "GObject.EnumValue",
+    "GObject.FlagsValue",
```
pbor, line 28: not very important, but I guess this should always have been here also before this patch? `pub struct FlagsValue(gobject_ffi::GFlagsValue);` is already manually implemented

pbor (approved):
It is consistent with all that was already added by @fbrouille for enums etc, so I am good with getting this in

glib/Gir_GObject.toml
bilelmoussaoui: yes, glib/gobject lacks a lot of types in the manually implemented list.

PR #1286, opened 2024-01-21
Support generic types in `glib::object_subclass`

andy128k:
Support generic types in `glib::object_subclass` and `glib::object_interface` macros.

### DIFF
--- a/glib-macros/src/object_interface_attribute.rs
+++ b/glib-macros/src/object_interface_attribute.rs
@@ -22,16 +22,15 @@ pub fn impl_object_interface(input: &mut syn::ItemImpl) -> TokenStream {
         ..
     } = input;
 
-    let self_ty_as_ident = match &**self_ty {
-        syn::Type::Path(syn::TypePath { path, .. }) => path.require_ident(),
-        _ => Err(syn::Error::new(
+    let Some(self_ty_ident) = (match &**self_ty {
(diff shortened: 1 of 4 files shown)


bilelmoussaoui:
What would be the use case here?

andy128k:
Well, code reuse. Let's say, I want to write a generic widget once and re-use it in a different contexts.

glib-macros/src/object_interface_attribute.rs
```diff
@@ -22,16 +22,15 @@ pub fn impl_object_interface(input: &mut syn::ItemImpl) -> TokenStream {
         ..
     } = input;
 
-    let self_ty_as_ident = match &**self_ty {
```
sdroege, line 25: I didn't look too close but I think this is currently unsound. There's only a single type registered for every `T`, so you can currently safely cast between different `T`. Not sure how that can be solved nicely.

sdroege (changes requested):
Can you add some tests to actually make use of this at runtime and make sure the types work correctly?

glib-macros/src/object_interface_attribute.rs
sdroege: Unlike in Java this is unsound here though :) In the simplest case, you could have a `MyObj<T>` that stores a `T`, and then simply casting that between `T: String` to `T: u32` and the subclass couldn't know about that happening and accesses the value with the wrong type then.

PR #1289, opened 2024-01-21
Replace `once_cell` usage with std::sync::OnceLock

### LINKED ISSUE #30: Use once_cell::Lazy instead of std::sync::Once
See summary. Simpler and less easy to get wrong.

### LINKED ISSUE #1141: Replace once_cell with std
Currently, despite requiring 1.70, we are still depending on once_cell. It would be nice to completely remove that dependency and replace it with std where possible.

andy128k:
Closes #30 
Closes #1141 

See also ~~https://github.com/gtk-rs/gir/pull/1532~~ https://github.com/gtk-rs/gir/pull/1537

This change does not fully remove `once_cell`. Full removal will break a contract of `glib::Properties` macro.

### DIFF
--- a/examples/object_subclass/author.rs
+++ b/examples/object_subclass/author.rs
@@ -1,12 +1,12 @@
 // You can copy/paste this file every time you need a simple GObject
 // to hold some data
 
-use glib::once_cell::sync::Lazy;
 use glib::prelude::*;
 use glib::subclass::prelude::*;
 use glib::subclass::Signal;
 use glib::Properties;
 use std::cell::RefCell;
+use std::sync::OnceLock;
 
 mod imp {
     use super::*;
@@ -23,9 +23,8 @@ mod imp {
     #[glib::derived_properties]
     impl ObjectImpl for Author {
         fn signals() -> &'static [Signal] {
-            static SIGNALS: Lazy<Vec<Signal>> =
-                Lazy::new(|| vec![Signal::builder("awarded").build()]);
-            SIGNALS.as_ref()
+            static SIGNALS: OnceLock<Vec<Signal>> = OnceLock::new();
+            SIGNALS.get_or_init(|| vec![Signal::builder("awarded").build()])
         }
     }
 
--- a/gdk-pixbuf/src/subclass/pixbuf_animation.rs
+++ b/gdk-pixbuf/src/subclass/pixbuf_animation.rs
@@ -5,11 +5,11 @@
 
 use std::{
     mem::MaybeUninit,
+    sync::OnceLock,
     time::{Duration, SystemTime},
 };
 
 use glib::{prelude::*, subclass::prelude::*, translate::*};
-use once_cell::sync::Lazy;
 
 use crate::{Pixbuf, PixbufAnimation, PixbufAnimationIter};
 
@@ -160,9 +160,6 @@ unsafe extern "C" fn animation_get_size<T: PixbufAnimationImpl>(
     }
 }
 
-static STATIC_IMAGE_QUARK: Lazy<glib::Quark> =
-    Lazy::new(|| glib::Quark::from_str("gtk-rs-subclass-static-image"));
-
 unsafe extern "C" fn animation_get_static_image<T: PixbufAnimationImpl>(
     ptr: *mut ffi::GdkPixbufAnimation,
 ) -> *mut ffi::GdkPixbuf {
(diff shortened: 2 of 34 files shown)


glib-macros/src/properties.rs
```diff
@@ -398,11 +398,10 @@ fn expand_properties_fn(props: &[PropDesc]) -> TokenStream2 {
     quote!(
         fn derived_properties() -> &'static [#crate_ident::ParamSpec] {
```
sdroege, line 399: >  Full removal will break a contract of glib::Properties macro.

That's OK though :) Can you update it accordingly?

examples/object_subclass/author.rs
```diff
@@ -23,9 +23,8 @@ mod imp {
     #[glib::derived_properties]
     impl ObjectImpl for Author {
         fn signals() -> &'static [Signal] {
-            static SIGNALS: Lazy<Vec<Signal>> =
```
sdroege, line 26: I wonder if we shouldn't wait for stabilization of `LazyLock`. But this also seems simple enough, doesn't really make much difference the way it's used in most places.

sdroege (commented):
Looks good to me otherwise

examples/object_subclass/author.rs
andy128k: `LazyLock` will not make it any cleaner or shorter in such cases.

- You need two statements with `OnceLock`: a declaration and an acquiring with initialization.
- You need two statements with `LazyLock`: a declaration with initialization and an acquiring.

OTOH, Initialization of `Regex`es with `LazyLock` would be much cleaner.

glib-macros/src/properties.rs
andy128k: Done

andy128k:
I guess, this [PR](https://github.com/gtk-rs/gir/pull/1537) should be merged first.

examples/object_subclass/author.rs
bilelmoussaoui: I am fine with this change, at the end of the day, the ideal situation would be having a macro for generating the signals boilerplate.

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -160,8 +160,10 @@ unsafe extern "C" fn animation_get_size<T: PixbufAnimationImpl>(
     }
 }
 
-static STATIC_IMAGE_QUARK: Lazy<glib::Quark> =
-    Lazy::new(|| glib::Quark::from_str("gtk-rs-subclass-static-image"));
+fn static_image_quark() -> glib::Quark {
```
bilelmoussaoui, line 163: I am not really a fan of this change tbh, as at the end of the day we don't get rid of once_cell from our dependencies tree. I would keep such Lazy usages

bilelmoussaoui (approved):
As I mentioned above, not a fan of the GQuark change but no strong opinion. Otherwise lgtm.

gdk-pixbuf/src/subclass/pixbuf_animation.rs
andy128k: May you explain? This PR literally removes usage of `once_cell` crate.

bilelmoussaoui: Never mind, I got confused by the tests/examples deps pulling once_cell.

glib-macros/src/properties.rs
sdroege: What contract does that break though? That change is completely internal or not?

sdroege (approved):
Looks good to me

gdk-pixbuf/src/subclass/pixbuf_animation.rs
bilelmoussaoui: I actually don't think you need a function here at all, the quark is used only inside a single function which can define the static itself. See https://github.com/gtk-rs/gtk4-rs/pull/1580/commits/3b8e9b04808509278efb14ead45873e4921cf81b

glib-macros/src/properties.rs
andy128k: Users of the macro who use `once_cell` will need to change their properties to std cell too.

gdk-pixbuf/src/subclass/pixbuf_animation.rs
andy128k: Most of the quarks may be moved inside of a function. But there are a few places where they are used across multiple functions. E. g. `PRINT_HANDLER`.

glib-macros/src/properties.rs
bilelmoussaoui: That is fine

bilelmoussaoui (approved):
before landing this, someone should do the same for gstreamer as they use the re-exported once_cell crate from glib iirc

sdroege:
> before landing this, someone should do the same for gstreamer as they use the re-exported once_cell crate from glib iirc

But we have a `Cargo.lock` so I'll take care of that once this is merged :) I'll shortly go over this again and then merge it.

sdroege (approved):

PR #1297, opened 2024-02-01
Use workspace features for ffi types

bilelmoussaoui:
We still have a `ffi = ...` crates defined outside of the workspace Cargo.toml. If we would change that, we would have to go through all the manual files and add `use crate::ffi` which is way too much work.


sdroege (approved):

sdroege:
```
Run git diff --exit-code
diff --git a/glib/src/subclass/type_module.rs b/glib/src/subclass/type_module.rs
index b46a129..249708c 100644
--- a/glib/src/subclass/type_module.rs
+++ b/glib/src/subclass/type_module.rs
@@ -16,6 +16,7 @@ pub trait TypeModuleImpl: ObjectImpl + TypeModuleImplExt {
     /// Unloads the module (see [`TypeModuleExt::unuse`]).
     ///
     /// [`TypeModuleExt::unuse`]: ../../gobject/auto/type_module/trait.TypeModuleExt.html#method.unuse
     fn unload(&self);
 }
 
Error: Process completed with exit code 1.
```

Can you shortly add a commit for that? Then just merge :)

bilelmoussaoui:
> glib/src/subclass/type_module

There is already a `// rustdoc-stripper-ignore-next`, it is rustdoc-stripper misbehaving as usual

PR #1304, opened 2024-02-07
Document values of Continue and Break

A6GibKm:
Users who are used to developing against other bindings might find useful to have in the documentation to which values each enum maps to.


glib/src/control_flow.rs
```diff
@@ -10,6 +10,8 @@ use crate::{prelude::*, translate::*};
 /// `ControlFlow::Continue` keeps the closure assigned, to be rerun when appropriate.
 ///
 /// `ControlFlow::Break` disconnects and drops it.
+///
+/// `Continue` and `Break` map to `true` and `false`, respectively.
```
sdroege, line 14: `G_SOURCE_CONTINUE` / `G_SOURCE_REMOVE` might be also helpful here?

sdroege: Are there similar ones for the signal propagation? There's `GDK_EVENT_PROPAGATE` / `GDK_EVENT_STOP`, those are the ones, right?

bilelmoussaoui:
ping?

sdroege:
@A6GibKm ping?

A6GibKm:
Sorry, I forgot about this! I will look into it later today!

glib/src/control_flow.rs
A6GibKm: Done.

A6GibKm:
I am not sure if having the doc aliases make sense, since they reference GDK.

glib/src/control_flow.rs
```diff
@@ -10,7 +10,12 @@ use crate::{prelude::*, translate::*};
 /// `ControlFlow::Continue` keeps the closure assigned, to be rerun when appropriate.
 ///
 /// `ControlFlow::Break` disconnects and drops it.
+///
+/// `Continue` and `Break` map to `G_SOURCE_CONTINUE` (`true`) and
+/// `G_SOURCE_REMOVE` (`false`), respectively.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[doc(alias = "G_SOURCE_REMOVE")]
```
bilelmoussaoui, line 17: Add the doc aliases to the enum members instead

sdroege (approved):

PR #1312, opened 2024-02-14
glib: Add `Quark::from_static_str()`

sdroege:
This avoids an allocation if the string is already statically allocated.

### DIFF
--- a/glib/src/quark.rs
+++ b/glib/src/quark.rs
@@ -12,13 +12,13 @@ pub struct Quark(NonZeroU32);
 impl Quark {
     #[doc(alias = "g_quark_from_string")]
     #[allow(clippy::should_implement_trait)]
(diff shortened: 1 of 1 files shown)


glib/src/quark.rs
```diff
@@ -16,6 +16,12 @@ impl Quark {
         unsafe { s.run_with_gstr(|s| from_glib(ffi::g_quark_from_string(s.as_ptr()))) }
     }
 
+    #[doc(alias = "g_quark_from_static_string")]
+    #[allow(clippy::should_implement_trait)]
+    pub fn from_static_str(s: &'static GStr) -> Quark {
```
bilelmoussaoui, line 21:
```suggestion
    pub fn from_static_str(s: &'static GStr) -> Self {
```

bilelmoussaoui (approved):
Otherwise lgtm

Issue #1257, opened 2023-12-24
[BUG] Wrong signature for `gio::content_type_guess`


Schmiddiii:
**Bug description**

The signature of [`gio::content_type_guess`](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gio/fn.content_type_guess.html)  has the type of `data` as `&[u8]`. The same documentation, and also [the gio documentation](https://docs.gtk.org/gio/func.content_type_guess.html) mention that `data` can be `None`. So a better signature would set `data` to `Option<&[u8]>` or similar. Note that the current signature can lead to wrong results if one wants the content type to be guessed only based on the filetype, this is not possible based on the current signature.

This change would be breaking.

sdroege:
It looks like currently all array-like `nullable` parameters are not optional. Passing `&[]` might or might not do what you want, but this should probably be fixed and be made more explicit. It's also not guaranteed that empty / NULL give the same behaviour.

This would need a change in the code generator.

Schmiddiii:
Pretty sure `None` and `&[]` are not the same for this function, passing `&[]` is interpreted as "empty file" and leading to `application/x-empty` (at least in one case I had, it worked for one other case. Not sure what the exact rules are).

euclio:
Related: https://github.com/gtk-rs/gir/issues/1133

fengalin:
WIP PR: https://github.com/gtk-rs/gir/pull/1622

PR #1318, opened 2024-02-27
WIP: generate with fixed array params

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gir/pull/1551

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1257


bilelmoussaoui:
Closing for now as it seems like gir doesn't handle zero-terminated arrays properly so everything works by luck so far.

PR #1321, opened 2024-03-04
glib: Requires Upgrade on Downgrade::Weak type

bilelmoussaoui:


glib/src/clone.rs
```diff
@@ -12,6 +12,7 @@ use std::{
 pub trait Downgrade
 where
     Self: Sized,
+    Self::Weak: Upgrade
```
sdroege, line 15: Why not `type Weak: Upgrade;` below?

bilelmoussaoui: Because I didn't knew you can write it like this!

sdroege (approved):

PR #1323, opened 2024-03-07
glib: Embed docs for ParamSpec types

bilelmoussaoui:


glib/src/param_spec.rs
```diff
@@ -304,16 +304,6 @@ extern "C" {
 macro_rules! define_param_spec {
     ($rust_type:ident, $ffi_type:path, $rust_type_offset:expr) => {
         // Can't use get_type here as this is not a boxed type but another fundamental type
-        wrapper! {
```
sdroege, line 307: Sadness

bilelmoussaoui: tbh, nowadays we can even generate a big portion of that code I think. Will give it a try in a bit

bilelmoussaoui: nvm, i forgot the paramspec are defined in gobject and not in glib :P

Issue #1326, opened 2024-03-14
[BUG] Memory leak when aborting tasks


AaronErhardt:
The bug first appeared in https://github.com/Relm4/Relm4/issues/619, but seems to be located in the glib crate rather than Relm4.

**Bug description**

Spawning and then aborting an uncompleted task leaks memory (presumably the future of the task).

```rust
use std::time::Duration;
use glib::timeout_future;

fn main() {
    let c = glib::MainContext::default();
    let l = glib::MainLoop::new(Some(&c), false);

    let l_clone = l.clone();
    c.spawn_local(async move {
        for _ in 0..1000 {
            // Spawn a task
            let handle = glib::MainContext::ref_thread_default().spawn_local(async {
                let test: u128 = std::future::pending().await;
                println!("{test}");
            });

            // Abort task => LEAK!
            handle.abort();
        }

        // Wait 1s and then quit
        timeout_future(Duration::from_secs(1)).await;
        l_clone.quit();
    });

    l.run();
}
```

**Backtrace**

I'm using `valgrind --suppressions=/usr/share/glib-2.0/valgrind/glib.supp --suppressions=/usr/share/gtk-4.0/valgrind/gtk.supp --leak-check=full BINARY_NAME` to detect the leaks.

```
❯ valgrind   --suppressions=/usr/share/glib-2.0/valgrind/glib.supp   --suppressions=/usr/share/gtk-4.0/valgrind/gtk.supp --leak-check=full ../target/debug/memory_leak 
==71348== Memcheck, a memory error detector
==71348== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==71348== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==71348== Command: ../target/debug/memory_leak
==71348== 
==71348== 
==71348== HEAP SUMMARY:
==71348==     in use at exit: 142,440 bytes in 1,271 blocks
==71348==   total heap usage: 7,683 allocs, 6,412 frees, 519,309 bytes allocated
==71348== 
==71348== 104,000 bytes in 1,000 blocks are definitely lost in loss record 251 of 251
==71348==    at 0x484280F: malloc (vg_replace_malloc.c:442)
==71348==    by 0x116BDF: alloc (alloc.rs:98)
==71348==    by 0x116BDF: alloc::alloc::Global::alloc_impl (alloc.rs:181)
==71348==    by 0x116956: allocate (alloc.rs:241)
[...]

PR #1328, opened 2024-03-16
glib: Drop the main context future return value sender on finalize

sdroege:
Otherwise it's going to be leaked if the future is dropped/aborted before it yields a value.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1326

### DIFF
--- a/glib/src/main_context_futures.rs
+++ b/glib/src/main_context_futures.rs
@@ -109,6 +109,8 @@ impl TaskSource {
             }
         }
 
+        ptr::drop_in_place(&mut (*source).return_tx);
+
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (dismissed):

glib/src/main_context_futures.rs
```diff
@@ -807,4 +809,26 @@ mod tests {
             );
         });
     }
+
+    #[test]
+    fn test_spawn_abort() {
```
sdroege, line 814: This test already passed before the fix, but in valgrind it would show a leak.

PR #1343, opened 2024-03-30
glib: fix UB in VariantStrIter::impl_get

decathorpe:
Passing an immutable reference (&p) to a function that mutates the data behind the pointer violates Rust's invariants.

This causes multiple tests in the test suite to crash when compiling it with optimizations (either `--release` mode or with `opt-level` of 2 or 3) with recent Rust versions, which is easy to reproduce, especially with nightly Rust:

```
$ cargo +nightly test --release --package glib
(...)
error: test failed, to rerun pass `--lib`
Caused by:
  process didn't exit successfully: (...) (signal: 11, SIGSEGV: invalid memory reference)
```

Looks like this wasn't caught earlier because the wrapped C function is variadic and there's less type checking happening because of that.

### DIFF
--- a/glib/src/variant_iter.rs
+++ b/glib/src/variant_iter.rs
@@ -117,13 +117,13 @@ impl<'a> VariantStrIter<'a> {
 
     fn impl_get(&self, i: usize) -> &'a str {
         unsafe {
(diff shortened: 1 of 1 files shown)


sdroege (approved):

sdroege:
Thanks!

PR #1349, opened 2024-04-03
Update gir / gir-files and update for `time_t`

sdroege:


gio/sys/src/lib.rs
```diff
@@ -10391,11 +10394,11 @@ extern "C" {
     pub fn g_credentials_get_unix_pid(
         credentials: *mut GCredentials,
         error: *mut *mut glib::GError,
-    ) -> c_int;
+    ) -> pid_t;
```
sdroege, line 10397: All this API should only exist on UNIX platforms

```
error[E0412]: cannot find type `pid_t` in this scope
Error:      --> gio\sys\src\lib.rs:10397:10
      |
10397 |     ) -> pid_t;
      |          ^^^^^ not found in this scope

error[E0412]: cannot find type `uid_t` in this scope
Error:      --> gio\sys\src\lib.rs:10401:10
      |
10401 |     ) -> uid_t;
      |          ^^^^^ not found in this scope

error[E0412]: cannot find type `uid_t` in this scope
Error:      --> gio\sys\src\lib.rs:10414:14
      |
10414 |         uid: uid_t,
      |              ^^^^^ not found in this scope
```

glib/sys/Cargo.toml
```diff
@@ -93,7 +94,10 @@ version = "2.76"
 version = "2.77"
 
 [package.metadata.system-deps.glib_2_0.v2_80]
-version = "2.79"
+version = "2.80"
```
sdroege, line 97:
```
  > PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 C:\gtk-build\gtk\x64\release\bin\pkgconf.exe --libs --cflags glib-2.0 glib-2.0 >= 2.80

  The system library `glib-2.0` required by crate `glib-sys` was not found.
  The file `glib-2.0.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
  The PKG_CONFIG_PATH environment variable is not set.
```

It looks like Windows currently uses GLib 2.79.X? How do we update that @bilelmoussaoui ? :)

bilelmoussaoui: @pbor is the windows expert in this repo, i only know how to deal with msvc

sdroege: Maybe like this. Let's see

sdroege: Yes

sdroege:
@bilelmoussaoui This would be ready now. Please check :)

bilelmoussaoui (approved):

PR #1369, opened 2024-04-25
glib::wrapper: Add docs for impls generated by the wrapper macro

felinira:
Annotate the generated impls and types with doc comments. This makes it clear which types are copy-on-clone and which types are reference counted without looking at the source.

### DIFF
--- a/glib/src/boxed_inline.rs
+++ b/glib/src/boxed_inline.rs
@@ -10,6 +10,7 @@ macro_rules! glib_boxed_inline_wrapper {
     ([$($attr:meta)*] $visibility:vis $name:ident $(<$($generic:ident $(: $bound:tt $(+ $bound2:tt)*)?),+>)?, $ffi_name:ty
      $(, @type_ $get_type_expr:expr)?) => {
         $(#[$attr])*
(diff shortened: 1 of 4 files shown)


glib/src/boxed_inline.rs
```diff
@@ -10,6 +10,7 @@ macro_rules! glib_boxed_inline_wrapper {
     ([$($attr:meta)*] $visibility:vis $name:ident $(<$($generic:ident $(: $bound:tt $(+ $bound2:tt)*)?),+>)?, $ffi_name:ty
      $(, @type_ $get_type_expr:expr)?) => {
         $(#[$attr])*
+        #[doc = "\n\nGLib type: Inline allocated boxed type with heap copy semantics."]
```
sdroege, line 13:
```suggestion
        #[doc = "\n\nGLib type: Inline allocated boxed type with stack copy semantics."]
```

or not?

sdroege: For everything in this file

sdroege (commented):
Looks good to me otherwise. That's a very good idea :)

sdroege (approved):

PR #1372, opened 2024-04-26
glib: Re-add and rename manual Win32 API additions

sdroege:
These were temporarily auto-generated in 0.19 but then disappeared from the .gir file again. Re-add them with the auto-generated names, and add all missing ones.

### DIFF
--- a/glib/src/win32.rs
+++ b/glib/src/win32.rs
@@ -4,18 +4,22 @@ use std::path::PathBuf;
 
 use crate::{translate::*, GString, StrV};
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-pub enum OSType {
+#[doc(alias = "GWin32OSType")]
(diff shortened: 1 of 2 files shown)


glib/src/win32.rs
```diff
@@ -4,18 +4,22 @@ use std::path::PathBuf;
 
 use crate::{translate::*, GString, StrV};
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
-pub enum OSType {
```
bilelmoussaoui, line 8: maybe keep this enum as an alias?

sdroege: That seems like a bad idea. It's specifically a Windows OS type, why should we drop the prefix of it? :)

sdroege: I think we will later have it as `glib_win32::OSType` or so but that's for some other time.

bilelmoussaoui: We can leave it as is then

Issue #1373, opened 2024-04-28
from_glib_ptr_borrow : Consider taking a *mut pointer to object


sophie-h:
Since we usually use *mut for GObjects, it would make sense to take a `*const *mut` or a `&*mut`. The latter one is what I usually have in my code so far.

sdroege:
Should ideally use a reference as that makes the lifetime tracking work better and makes it harder to use it wrong. Not sure why I didn't do that. I'll look into that again :)

sdroege:
Can you try if https://github.com/gtk-rs/gtk-rs-core/pull/1375 makes usage easy for you?

PR #1375, opened 2024-04-29
glib: Use a reference to a pointer of correct mutability for from_gli…

sdroege:
…b_ptr_borrow()

This hopefully makes it easier to use and harder to get the returned lifetime wrong.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1373

----

CC @sophie-h

### DIFF
--- a/glib/src/boxed.rs
+++ b/glib/src/boxed.rs
@@ -47,14 +47,14 @@ macro_rules! glib_boxed_wrapper {
 
             #[doc = "Borrows the underlying C value."]
             #[inline]
-            pub unsafe fn from_glib_ptr_borrow<'a>(ptr: *const *const $ffi_name) -> &'a Self {
-                &*(ptr as *const Self)
+            pub unsafe fn from_glib_ptr_borrow(ptr: &*mut $ffi_name) -> &Self {
+                &*(ptr as *const *mut $ffi_name as *const Self)
             }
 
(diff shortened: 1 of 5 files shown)


sdroege:
Also need to update gstreamer-rs after merging this.

glib/src/boxed.rs
```diff
@@ -354,7 +354,7 @@ macro_rules! glib_boxed_wrapper {
                 debug_assert_eq!(std::mem::size_of::<Self>(), std::mem::size_of::<$crate::ffi::gpointer>());
                 let value = &*(value as *const $crate::Value as *const $crate::gobject_ffi::GValue);
                 debug_assert!(!value.data[0].v_pointer.is_null());
-                <$name $(<$($generic),+>)?>::from_glib_ptr_borrow(&value.data[0].v_pointer as *const $crate::ffi::gpointer as *const *const $ffi_name)
+                <$name $(<$($generic),+>)?>::from_glib_ptr_borrow(&*(&value.data[0].v_pointer as *const $crate::ffi::gpointer as *const *mut $ffi_name))
```
sdroege, line 357: These changes make me a bit nervous. Does this create a temporary because of the `&value.data[0].v_pointer`, and then we return a reference to some long gone stack frame?

sdroege: We do the same in other places already so it should be fine

Issue #1366, opened 2024-04-23
[BUG] ObjectBuilder property population gets slower after application startup


exi:
**Bug description**

I have these three variants used to construct 1 million GObject subclass instances each:

```rust
    pub fn new_builder(id: u64, name: String) -> Self {
        let obj: AccountObject = Object::builder()
            .property("id", id)
            .property("name", name)
            .build();
        obj
    }

    pub fn new_setter(id: u64, name: String) -> Self {
        let obj: AccountObject = Object::builder()
            .build();
        obj.set_name(name);
        obj.set_id(id);
        obj
    }

    pub fn new_imp(id: u64, name: String) -> Self {
        let obj: AccountObject = Object::builder()
            .build();
        {
            let mut data = obj.imp().data.borrow_mut();
            data.id = id;
            data.name = name;
        }
        obj
    }

```

One uses ObjectBuilder to set properties, one uses property-setters and one just modifies the data directly.
The results of the creation:
```
Creation with new_builder() took 16385ms
Creation with new_setter() took 10925ms
Creation with new_imp() took 3867ms
```

According to my profiler, most of the time in the new_builder variant is spent in:
```
Object::new_internal() -> ObjectClassExt::list_properties
```

This makes a noticable different when constructing large ListModels for big tables.

# EDIT
Reproduction repository here: https://github.com/exi/gtk-rs-obj-builder

The slowness of the ObjectBuilder variant only appears after the application is started up.

bilelmoussaoui:
That is actually very known. The problem is the various checks that are made by new_internal and we would need some (unsafe?) variant that doesn't go through them. 

That would be useful for builder patterns that are generated by gir and possibly in the future for a builder pattern generated by the Properties macro

exi:
Are there any downsides to directly inject properties into the imp()?
I'd guess we would miss some property-notify-event signals, but those would not be fired in the ObjectBuilder as well.
Or are there some other hazards I should be aware of?

sdroege:
Can you provide the full, runnable code for this? The numbers for the second variant are confusing and I'd like to know what's going on there.

exi:
@bilelmoussaoui @sdroege I boiled it down to a reproducible change, which was quite surprising:

https://github.com/exi/gtk-rs-obj-builder

ObjectBuilder gets significantly slower after the application is started up.
The code in the repo results in output like this:

```rust
Before startup
Setter took 9223ms
Builder took 9134ms
After startup
[...]

PR #1376, opened 2024-04-29
glib: Don't use `g_object_list_properties()` for setting properties

sdroege:
Listing all properties of a type needs to iterate over every single pspec that is registered in the process, which is rather slow. In comparison, finding a specific property for a type only requires one or at most number of class hierarchy depth hash table lookups.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1366


sdroege:
This makes object construction about 30% faster after GTK is initialized.

bilelmoussaoui (approved):
nice!

PR #1377, opened 2024-04-29
ObjectBuilder: add property_if(), property_if_some(), property_from_iter() ...

fengalin:
... & property_if_not_empty()

This commit adds an `ObjectBuilderPropertySetter` `trait` which implements
functions to improve usability when setting properties.

**`property_if()`** allows setting a property from an `Option` only if a given
`predicate` evaluates to `true`. Otherwise, the `Object`'s default value or
any previously set value is kept.

**`property_if_some()`** allows setting a property from an `Option` only if the
`Option` is `Some`. Otherwise, the `Object`'s default value or any previously
set value is kept.

For instance, assuming an `Args` `struct` which was built from the application's
CLI arguments:

```rust
let args = Args {
    name: Some("test"),
    anwser: None,
};
```

... without this change, we need to write something like this:

```rust
let mut builder = Object::builder::<SimpleObject>();

if let Some(ref name) = args.name {
    builder = builder.property("name", name)
}

if let Some(ref answer) = args.answer {
    builder = builder.property("answer", &args.answer)
}

let obj = builder.build();
```

With `property_if_some()`, we can write:

```rust
let obj = Object::builder::<SimpleObject>()
    .property_if_some("name", &args.name)
    .property_if_some("answer", &args.answer)
    .build();
```

**`property_from_iter()`** allows building a collection-like `Value` property
from a collection of items:

```rust
let obj = Object::builder::<SimpleObject>()
    .property_from_iter::<ValueArray>("array", ["val0", "val1"])
    .build();
```

**`property_if_not_empty()`** allows building a collection-like `Value` property
from a collection of items but does nothing if the provided collection if empty,
thus keeping the default value of the property, if any:

```rust
let obj = Object::builder::<SimpleObject>()
    .property_if_not_empty::<ValueArray>("array", ["val0", "val1"])
    .build();
```

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -1542,8 +1542,9 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
     }
 
     // rustdoc-stripper-ignore-next
-    /// Set property `name` to the given value `value`.
-    #[inline]
+    /// Sets property `name` to the given value `value`.
+    ///
+    /// Overrides any default or previously defined value for `name`.
     pub fn property(self, name: &'a str, value: impl Into<Value>) -> Self {
         let ObjectBuilder {
             type_,
@@ -1556,6 +1557,67 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
             type_,
             properties,
             phantom: PhantomData,
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` to the given inner value if the `predicate` evaluates to `true`.
+    ///
+    /// This has no effect if the `predicate` evaluates to `false`,
+    /// i.e. default or previous value for `name` is kept.
+    #[inline]
+    pub fn property_if(self, name: &'a str, value: impl Into<Value>, predicate: bool) -> Self {
+        if predicate {
+            self.property(name, value)
+        } else {
+            self
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` to the given inner value if `value` is `Some`.
+    ///
+    /// This has no effect if the value is `None`, i.e. default or previous value for `name` is kept.
+    #[inline]
+    pub fn property_if_some(self, name: &'a str, value: Option<impl Into<Value>>) -> Self {
+        if let Some(value) = value {
+            self.property(name, value)
+        } else {
+            self
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` using the given `ValueType` `V` built from `iter`'s the `Item`s.
+    ///
+    /// Overrides any default or previously defined value for `name`.
+    #[inline]
+    pub fn property_from_iter<V: ValueType + Into<Value> + FromIterator<Value>>(
+        self,
+        name: &'a str,
+        iter: impl IntoIterator<Item = impl Into<Value>>,
+    ) -> Self {
+        let iter = iter.into_iter().map(|item| item.into());
+        self.property(name, V::from_iter(iter))
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` using the given `ValueType` `V` built from `iter`'s Item`s,
+    /// if `iter` is not empty.
+    ///
+    /// This has no effect if `iter` is empty, i.e. previous property value for `name` is unchanged.
+    #[inline]
+    pub fn property_if_not_empty<V: ValueType + Into<Value> + FromIterator<Value>>(
+        self,
+        name: &'a str,
+        iter: impl IntoIterator<Item = impl Into<Value>>,
+    ) -> Self {
+        let mut iter = iter.into_iter().peekable();
+        if iter.peek().is_some() {
+            let iter = iter.map(|item| item.into());
+            self.property(name, V::from_iter(iter))
+        } else {
+            self
         }
     }
 
--- a/glib/src/subclass/object.rs
+++ b/glib/src/subclass/object.rs
@@ -350,11 +350,24 @@ mod test {
 
         impl ObjectImpl for ChildObject {}
 
-        #[derive(Default)]
         pub struct SimpleObject {
             name: RefCell<Option<String>>,
             construct_name: RefCell<Option<String>>,
             constructed: RefCell<bool>,
+            answer: RefCell<i32>,
+            array: RefCell<Vec<String>>,
+        }
+
+        impl Default for SimpleObject {
+            fn default() -> Self {
+                SimpleObject {
+                    name: Default::default(),
+                    construct_name: Default::default(),
+                    constructed: Default::default(),
+                    answer: RefCell::new(42i32),
+                    array: RefCell::new(vec!["default0".to_string(), "default1".to_string()]),
+                }
+            }
         }
 
         #[glib::object_subclass]
@@ -377,6 +390,10 @@ mod test {
                             .read_only()
                             .build(),
                         crate::ParamSpecObject::builder::<super::ChildObject>("child").build(),
+                        crate::ParamSpecInt::builder("answer")
+                            .default_value(42i32)
+                            .build(),
+                        crate::ParamSpecValueArray::builder("array").build(),
                     ]
                 })
             }
@@ -437,6 +454,20 @@ mod test {
                     "child" => {
                         // not stored, only used to test `set_property` with `Objects`
                     }
+                    "answer" => {
+                        let answer = value
+                            .get()
+                            .expect("type conformity checked by 'Object::set_property'");
+                        self.answer.replace(answer);
+                    }
+                    "array" => {
+                        let value = value
+                            .get::<crate::ValueArray>()
+                            .expect("type conformity checked by 'Object::set_property'");
+                        let mut array = self.array.borrow_mut();
+                        array.clear();
+                        array.extend(value.iter().map(|v| v.get().unwrap()));
+                    }
                     _ => unimplemented!(),
                 }
             }
@@ -446,6 +477,8 @@ mod test {
                     "name" => self.name.borrow().to_value(),
                     "construct-name" => self.construct_name.borrow().to_value(),
                     "constructed" => self.constructed.borrow().to_value(),
+                    "answer" => self.answer.borrow().to_value(),
+                    "array" => crate::ValueArray::new(self.array.borrow().iter()).to_value(),
                     _ => unimplemented!(),
                 }
             }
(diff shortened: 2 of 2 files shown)


bilelmoussaoui:
Haven't had the time to review the PR yet but at some point there was a discussion about adding the possibility to set a property only if a condition is truthful. The use case was setting a property only if a certain gstreamer version is available iirc

cc @sdroege, do you remember where was that needed? or maybe it was someone else.

fengalin:
Noteworthy annoyance with these changes is that we need to `use glib::prelude::*;` (or `gst::prelude::*;`) in some places were it wasn't before.

bilelmoussaoui:
> Noteworthy annoyance with these changes is that we need to `use glib::prelude::*;` (or `gst::prelude::*;`) in some places were it wasn't before.

That is ok, I already migrated a bunch of files to use prelude instead of including traits manually. Going forward in that direction is good from my point of view

bilelmoussaoui:
> I propose renaming `property_if_any()` to `property_if_not_empty()` to clarify its purpose and to be more similar to the common `is_empty()` collection method.

I would name `property_if_some` instead.

fengalin:
> I propose renaming property_if_any() to property_if_not_empty() to clarify its purpose and to be more similar to the common is_empty() collection method

Makes sense.

> I would name it property_if_some instead.

Already taken by the `Option<_>` variant :)

glib/src/object.rs
```diff
@@ -1525,9 +1525,78 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
...
+        if iter.peek().is_some() {
+            let iter = iter.map(|item| item.into());
+            ObjectBuilderPropertySetter::property(self, name, V::from_iter(iter))
+        } else {
+            self
+        }
+    }
+}
+
+impl<'a, O: IsA<Object> + IsClass> ObjectBuilderPropertySetter<'a> for ObjectBuilder<'a, O> {
```
sdroege, line 1597: Why is this an extension trait instead of just implementing it directly on `ObjectBuilder`?

glib/src/object.rs
```diff
@@ -1525,9 +1525,78 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
...
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` using the given `ValueType` `V` built from `iter`'s the `Item`s.
+    ///
+    /// Overrides any default or previously defined value.
     #[inline]
-    pub fn property(self, name: &'a str, value: impl Into<Value>) -> Self {
+    fn property_from_iter<V: ValueType + Into<Value> + FromIterator<SendValue>>(
+        self,
+        name: &'a str,
+        iter: impl IntoIterator<Item = impl Into<SendValue>>,
```
sdroege, line 1570:
```suggestion
        iter: impl IntoIterator<Item = impl Into<Value>>,
```

just like for the other property methods

glib/src/object.rs
```diff
@@ -1525,9 +1525,78 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
...
+        }
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` using the given `ValueType` `V` built from `iter`'s the `Item`s.
+    ///
+    /// Overrides any default or previously defined value.
     #[inline]
-    pub fn property(self, name: &'a str, value: impl Into<Value>) -> Self {
+    fn property_from_iter<V: ValueType + Into<Value> + FromIterator<SendValue>>(
```
sdroege, line 1567:
```suggestion
    fn property_from_iter<V: ValueType + Into<Value> + FromIterator<Value>>(
```

glib/src/object.rs
```diff
@@ -1525,9 +1525,78 @@ impl<'a, O: IsA<Object> + IsClass> ObjectBuilder<'a, O> {
...
+    // rustdoc-stripper-ignore-next
+    /// Sets property `name` using the given `ValueType` `V` built from `iter`'s Item`s,
+    /// if `iter` is not empty.
+    ///
+    /// This has no effect if `iter` is empty, i.e. previous property value is unchanged.
+    #[inline]
+    fn property_if_any<V: ValueType + Into<Value> + FromIterator<SendValue>>(
+        self,
+        name: &'a str,
+        iter: impl IntoIterator<Item = impl Into<SendValue>>,
```
sdroege, line 1585:
```suggestion
    fn property_if_any<V: ValueType + Into<Value> + FromIterator<Value>>(
        self,
        name: &'a str,
        iter: impl IntoIterator<Item = impl Into<Value>>,
```

sdroege (commented):
Looks good to me otherwise and the method names also make sense. @bilelmoussaoui @felinira do you prefer any different names, or would like to have something else changed?

sdroege:
> > I propose renaming property_if_any() to property_if_not_empty() to clarify its purpose and to be more similar to the common is_empty() collection method
> 
> Makes sense.

Also seems fine to me

glib/src/object.rs
fengalin: Because you suggested this :): https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/1431#note_2392609

I think it's convenient to have a single definition too: I can use this `trait` for `gst::ElementFactory`, implement `property()` and get all the other implementations for free.

I used a similar pattern `gst::StructureBuilderPropertySetter` which is applicable to `gst::StructureBuilder`, `gst::CapsBuilder`, `gst_audio::CapsBuilder`, gst_video::CapsBuilder`, ...

sdroege: I suggested this because I thought of adding this extension trait in gstreamer-rs, but here it could also be a normal impl block.

If it makes anything easier and you're going to implement the trait on other types too that's fine. (Btw, could you also add this to the appsrc/appsink builders then? I just noticed that they don't have the general `property` method)

sdroege: > I can only see one implementation here

But more in gstreamer-rs?

And here, I think all the autogenerated builders for specific types could make use of this. I think they all don't have a generic property method right now but that would be useful.

glib/src/object.rs
fengalin: Problem with that is `gst::Array` & `gst::List` use `SendValue`: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/b7b5352353e147fccc00ff684dffb65c21a6e6b1/gstreamer/src/value.rs#L884

sdroege: That's inconvenient but at this level here there is no reason to use `SendValue`. Nothing wrong with storing non-`Send` things in non-threadsafe objects

glib/src/object.rs
fengalin: I'm considering using a macro instead of a `trait`. It would be possible to add `xxx_if()`,  `xxx_if_some()`,  `xxx_from_iter()` &  `xxx_if_not_empty()` to any builder implementing `xxx()`. There would be a specific macro in GStreamer due to the `Send` constraint, but it would be applicable for `ElementFactory` (property), `Structure` (field),  `Caps` (field), `Event` (other_field) & `Message` (other_field).

sdroege: A macro that is exported from the glib crate? Not sure I like adding more macro public API. @bilelmoussaoui ?

fengalin: The way I see it: due to the `Value` / `SendValue` differences, I will handle this separately in `gtk-rs` & `GStreamer`.

* In `gtk-rs`: the new methods `property_*` can be added to the `ObjectBuilder` `impl` directly. For the generated `Builder`s, a `gir` update could add the additional methods on a per-property basis.
* In `GStreamer`, I think I can use the macro I was mentioning in my previous comment. It would be transparent to users since it would only be used in the bindings for the `ElementFactory` & the `Structure`-like `struct` builders, not something like `glib::clone!` for instance.

glib/src/object.rs
fengalin: See also discussion in https://github.com/gtk-rs/gtk-rs-core/pull/1377#pullrequestreview-2030316501

glib/src/object.rs
fengalin: See discussion in https://github.com/gtk-rs/gtk-rs-core/pull/1377#pullrequestreview-2030319735 & https://github.com/gtk-rs/gtk-rs-core/pull/1377#pullrequestreview-2030316501

glib/src/object.rs
fengalin: See discussion in https://github.com/gtk-rs/gtk-rs-core/pull/1377#pullrequestreview-2030319735 & https://github.com/gtk-rs/gtk-rs-core/pull/1377#pullrequestreview-2030316501

glib/src/object.rs
sdroege: Sounds like a plan

bilelmoussaoui: I don't see which other types that could use this. I think the only use case would be either gstreamer-rs or gegl-rs(which is not maintained anyways...). So do we really need this to be an extension trait?

bilelmoussaoui (commented):
otherwise lgtm, can't comment on the namings because i am bad at that

glib/src/value_array.rs
```diff
@@ -23,6 +24,25 @@ impl ValueArray {
         unsafe { from_glib_full(gobject_ffi::g_value_array_new(n_prealloced)) }
     }
 
+    pub fn from_items(values: impl IntoIterator<Item = impl Into<Value> + Send>) -> Self {
```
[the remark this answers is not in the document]
fengalin: ack

glib/src/object.rs
fengalin: We don't. See my comment: https://github.com/gtk-rs/gtk-rs-core/pull/1377#discussion_r1586068809

fengalin:
Sorry for the lack of progress on this one. fwiw, I'm planning on working on it during the hackfest.

glib/src/value_array.rs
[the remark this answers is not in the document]
fengalin: Was handled in https://github.com/gtk-rs/gtk-rs-core/pull/1411

sdroege (approved):

glib/src/object.rs
fengalin: Related gir issue: https://github.com/gtk-rs/gir/issues/1581

fengalin:
Should be ready now

andy128k:
Another option to keep ergonomic and flexible builder syntax is to use something similar to [Kotlin's scope functions](https://kotlinlang.org/docs/scope-functions.html).

```rust
trait BuilderScopeExt {
    fn apply(self, scope: &mut FnMut(&Self)) -> Self;
}

impl BuilderScopeExt {
    fn apply(mut self, scope: &mut FnMut(&Self)) -> Self {
        (scope)(&mut self);
        self
    }
}

let obj = Object::builder::<SimpleObject>()
     .apply(|builder| {
         if let Some(ref name) = args.name {
              builder.property("name", name);
         }
         if let Some(ref answer) = args.answer {
              builder.property("answer", &args.answer);
         }
     })
     .build();
```

andy128k:
@fengalin There will always be specific inconvenient cases. Now it is `Option`, later someone may ask for `property_if_ok`, or `property_if_err` etc. I'm just saying, instead of polluting builder scope with a variety of corner cases it could be a more universal approach, even if it is a bit more verbose.

PR #1378, opened 2024-04-29
examples: Add example for custom class structs and virtual methods

felinira:
Adds a very detailed example of virtual methods and interfaces. We already had one in gtk4, but not here. I also couldn't find any example for properly chaining up interface vfuncs.

The example is a bit elaborate, but I am convinced that every class here has some merit. It was for example unexpected to me that the interface had to be re-declared in the subclass a second time to override its vfunc, as this is not required for classes. I can however split it up if that is desired.

Some questions that came up:
- Should some of these `Ext` traits be sealed? They have a bound that is equal to their implementation, so I'm not entirely sure what that would accomplish.
-  All examples of subclassing I found put the class struct out of the implementation module. The few interface implementations I found in the wild don't do that, and instead put them in a separate module, calling the class struct the same as the type. What should be best practise here?

### DIFF
--- a/examples/virtual_methods/pet.rs
+++ b/examples/virtual_methods/pet.rs
@@ -1,0 +1,198 @@
+use glib::prelude::*;
+use glib::subclass::prelude::*;
+
+/// The ffi module includes the exported C API of the object.
+#[allow(dead_code)]
+pub mod ffi {
+    /// The instance pointer is used for references to the object.
+    #[repr(C)]
+    pub struct Instance(std::ffi::c_void);
+
+    /// Custom class struct for the [`Pet`](super::Pet) class.
+    ///
+    /// The class struct is used to implement the vtable for method dispatch. The first field
+    /// *must* be a pointer to the parent type. In our case, this is
+    /// [`GObjectClass`](glib::gobject_ffi::GObjectClass)
+    #[derive(Copy, Clone, Debug)]
+    #[repr(C)]
+    pub struct Class {
+        /// The first field in a class struct must always be the parent class struct
+        parent_class: glib::gobject_ffi::GObjectClass,
+
+        /// Virtual method for the [`PetImpl::pet`](super::PetImpl::pet) trait method
+        pub(super) pet: fn(&super::Pet) -> bool,
+
+        /// Virtual method for the [`PetImpl::feed`](super::PetImpl::feed) trait method
+        pub(super) feed: fn(&super::Pet),
+    }
+
+    /// Every class struct is required to implement the `ClassStruct` trait
+    ///
+    /// Safety: This impl is unsafe because it requires the struct to be `repr(C)` and
+    /// the first field must be the parent class.
+    unsafe impl glib::subclass::types::ClassStruct for Class {
+        type Type = super::imp::Pet;
+    }
+
+    /// Implement Deref to the parent class for convenience.
+    impl std::ops::Deref for Class {
+        type Target = glib::gobject_ffi::GObjectClass;
+
+        fn deref(&self) -> &Self::Target {
+            &self.parent_class
+        }
+    }
+}
+
+/// Private (implementation) module of the class. This is not exported.
+mod imp {
+    use glib::subclass::prelude::*;
+
+    /// The Pet implementation struct
+    #[derive(Default)]
+    pub struct Pet {}
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for Pet {
+        /// This name is exported to the gobject type system and must be unique between all loaded
+        /// shared libraries.
+        ///
+        /// Usually this is achieved by adding a short prefix to all names coming from a
+        /// particular app / library.
+        const NAME: &'static str = "Pet";
+
+        /// The [`Pet`](super::Pet) class is abstract and instances to it can't be created.
+        ///
+        /// If an instance of this class were instantiated it would panic.
+        const ABSTRACT: bool = true;
+
+        type Type = super::Pet;
+
+        /// Override the class struct with the custom [class](super::ffi::Class)
+        type Class = super::ffi::Class;
+
+        /// Initialize the [class struct](super::ffi::Class) with the default implementations of the
+        /// virtual methods.
+        fn class_init(klass: &mut Self::Class) {
+            klass.pet = |obj| obj.imp().pet_default();
+            klass.feed = |obj| obj.imp().feed_default();
+        }
+    }
+
+    impl ObjectImpl for Pet {}
+
+    impl Pet {
+        /// Default implementation of [`PetImpl::pet`](super::PetImpl::pet)
+        fn pet_default(&self) -> bool {
+            // The default behaviour is unsuccessful pets
+            println!("Pet::pet_default");
+            false
+        }
+
+        /// Default implementation of [`PetImpl::feed`](super::PetImpl::feed)
+        fn feed_default(&self) {
+            println!("Pet::feed_default");
+        }
+    }
+}
+
+glib::wrapper! {
+    /// The `Pet` class acts as a base class for pets and provides two virtual methods.
+    pub struct Pet(ObjectSubclass<imp::Pet>);
+}
+
+/// The `PetExt` trait contains public methods of all [`Pet`] objects
+///
+/// These methods need to call the appropriate vfunc from the vtable.
+pub trait PetExt: IsA<Pet> {
+    /// Calls the [`PetImpl::pet`] vfunc
+    fn pet(&self) -> bool {
+        let this = self.upcast_ref();
+        let class = this.class();
+
+        (class.as_ref().pet)(this)
+    }
+
+    /// Calls the [`PetImpl::feed`] vfunc
+    fn feed(&self) {
+        let this = self.upcast_ref();
+        let class = this.class();
+
+        (class.as_ref().feed)(this)
+    }
+}
+
+/// Implement PetExt for all [`Pet`] subclasses (and `Pet` itself)
+impl<T: IsA<Pet>> PetExt for T {}
+
+/// The `PetImpl` trait contains overridable virtual function definitions for [`Pet`] objects.
+pub trait PetImpl: ObjectImpl {
+    /// Default implementation of a virtual method.
+    ///
+    /// This always calls into the implementation of the parent class so that if
+    /// the subclass does not explicitly implement it, the behaviour of its
+    /// parent class will be preserved.
+    fn pet(&self) -> bool {
+        self.parent_pet()
+    }
+
+    /// Default implementation of a virtual method.
+    ///
+    /// This always calls into the implementation of the parent class so that if
+    /// the subclass does not explicitly implement it, the behaviour of its
+    /// parent class will be preserved.
+    fn feed(&self) {
+        self.parent_feed();
+    }
+}
+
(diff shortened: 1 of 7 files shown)


sdroege:
> * Should some of these `Ext` traits be sealed? They have a bound that is equal to their implementation, so I'm not entirely sure what that would accomplish.

Nothing, they shouldn't be sealed :) Do we have any left somewhere that are sealed in this situation? If so we should fix that. I thought I did that a while ago already.

> * All examples of subclassing I found put the class struct out of the implementation module. The few interface implementations I found in the wild don't do that, and instead put them in a separate module, calling the class struct the same as the type. What should be best practise here?

The struct should be outside the implementation module (unless it's a final type, etc). It's public API, even if you don't use it from Rust directly (usually).

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+
+mod imp {
+    use super::*;
+
+    /// The Pet implementation struct
+    #[derive(Default)]
+    pub struct Pet {}
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for Pet {
```
sdroege, line 12: Maybe want to make this one abstract to also show that aspect?

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+glib::wrapper! {
+    /// The `Pet` class acts as a base class for pets and provides two virtual methods
+    pub struct Pet(ObjectSubclass<imp::Pet>);
+}
+
+/// Custom class struct for the [`Pet`] class.
+///
+/// The class struct is used to implement the vtable for method dispatch. The first field *must*
+/// be a pointer to the parent type. In our case, this is `GTypeInterface`
+#[derive(Copy, Clone, Debug)]
```
sdroege, line 53: Do we need copy/clone?

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn parent_pet(&self) -> bool {
+        let data = Self::type_data();
+        let parent_class = unsafe { &*(data.as_ref().parent_class() as *const PetClass) };
+        let pet = parent_class
+            .pet
+            .expect("Pet::pet vfunc not set by class_init");
+
+        pet(self
+            .obj()
+            .dynamic_cast_ref()
```
sdroege, line 136: I'd do

```suggestion
            .unsafe_cast_ref()
```

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+        // `PetImpl` of the subclass.
+        //
+        // Note that this is only called for actual subclasses and not [`Pet`]
+        // itself: `Pet` does not implement [`PetImpl`] and handles this inside
+        // `ObjectSubclass::class_init()` for providing the default implementation
+        // of the virtual methods.
+        let klass = class.as_mut();
+
+        // Set the virtual methods
+        fn vpet_trampoline<Obj: ObjectSubclass + PetImpl>(obj: &Pet) -> bool {
```
sdroege, line 175: I would assign them inline as closures. That's a lot less noise

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+            // (Down-)cast the object as the concrete type
+            let this = obj
+                .dynamic_cast_ref::<<Obj as ObjectSubclass>::Type>()
```
sdroege, line 178:
```suggestion
                .unsafe_cast_ref::<<Obj as ObjectSubclass>::Type>()
```

also here

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    /// Virtual method for the [`PetImpl::pet`] trait method
+    pub(super) pet: Option<fn(&super::Pet) -> bool>,
+
+    /// Virtual method for the [`PetImpl::feed`] trait method
+    pub(super) feed: Option<fn(&super::Pet)>,
+}
+
+/// Every class struct is required to implement the `ClassStruct` trait
+unsafe impl glib::subclass::types::ClassStruct for PetClass {
```
sdroege, line 67: Usually also useful to implement `Deref<Target = ParentClass>`

examples/virtual_methods/pet.rs
sdroege: I guess it's good to have for consistency with the autogenerated FFI structs

examples/virtual_methods/purrable.rs
```diff
@@ -0,0 +1,115 @@
...
+        let data = Self::type_data();
+        let parent_class =
+            unsafe { &*(data.as_ref().parent_interface::<Purrable>() as *const iface::Purrable) };
+        let is_purring = parent_class
+            .is_purring
+            .expect("Purrable::is_purrable vfunc not set by class_init");
+
+        is_purring(
+            self.obj()
+                .dynamic_cast_ref()
```
sdroege, line 89:
```suggestion
                .unsafe_cast_ref()
```

etc

examples/virtual_methods/purrable.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// Custom class struct for the [`super::Purrable`] interface.
+    ///
+    /// The class struct is used to implement the vtable for method dispatch. The first field *must*
+    /// be a pointer to the parent type. In our case, this is `GTypeInterface`
+    #[derive(Copy, Clone)]
+    #[repr(C)]
+    pub struct Purrable {
+        /// The first field in a class struct must always be the parent class struct
+        parent_type: glib::gobject_ffi::GTypeInterface,
+        pub(super) is_purring: Option<fn(&super::Purrable) -> bool>,
```
sdroege, line 17: Could also be without the `Option`

examples/virtual_methods/purrable.rs
```diff
@@ -0,0 +1,115 @@
...
+/// The `PetImplExt` trait is implemented for all subclasses that have [`Pet`] in the class hierarchy
+impl<T: PurrableImpl> PurrableImplExt for T {}
+
+/// To make this interface implementable we need to implement [`IsImplementable`]
+unsafe impl<Obj: PurrableImpl> IsImplementable<Obj> for Purrable {
+    fn interface_init(iface: &mut glib::Interface<Self>) {
+        let klass = iface.as_mut();
+
+        // Set the virtual method
+        fn vis_purring<Obj: ObjectSubclass + PurrableImpl>(obj: &Purrable) -> bool {
```
sdroege, line 104: Closure would be shorter

sdroege (commented):
Nice example, and much better than https://github.com/sdroege/gobject-example-rs

Now only need to add C API stuff here and we can retire that one

examples/virtual_methods/purrable.rs
```diff
@@ -0,0 +1,115 @@
...
+mod iface {
+    use super::*;
+
+    /// Custom class struct for the [`super::Purrable`] interface.
+    ///
+    /// The class struct is used to implement the vtable for method dispatch. The first field *must*
+    /// be a pointer to the parent type. In our case, this is `GTypeInterface`
+    #[derive(Copy, Clone)]
+    #[repr(C)]
+    pub struct Purrable {
```
sdroege, line 14: This should be called `PurrableInterface`, just like the corresponding other one is called `PetClass`

sdroege: Like @pbor mentioned, should also get a `#[repr(C)] pub struct Purrable(std::ffi::c_void);` for the instance pointers

sdroege: Or could call them just `Class`, `Instance`, `Interface` etc because we get namespacing from the modules anyway

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/// The `PetExt` trait contains public methods of all [`Pet`] objects
+///
+/// These methods need to call the appropriate vfunc from the vtable.
+pub trait PetExt: IsA<Pet> {
+    /// Calls the [`PetImpl::pet`] vfunc
+    fn pet(&self) -> bool {
+        let this = self.upcast_ref();
+        let vtable = this.class();
```
sdroege, line 78:
```suggestion
        let class = this.class();
```

maybe? Closer to the GObject terminology

examples/virtual_methods/pet.rs
sdroege: The reason is apparently the `Box::new()` in `interface_init()` for the parent interface struct. For classes that's not required because we get it by pointer via `g_type_class_peek_parent()`.

 * I think we should have `Clone + Copy` on all these `repr(C)` structs for consistency with the sys crates
 * If I'm not confused, class structs must be `Copy` too because that's what GObject is going to do internally, so there's currently a little hole here. If we were exposing `base_init` this could be avoided
 * We could look at using a pointer for the interface in `interface_init()` too

Do you want to look at the second and third? If not we should create an issue at least for the second one as that seems potentially problematic :)

sdroege: Also with regards to `base_init()`, we could only require `Clone` and automatically call that in there I guess?

examples/virtual_methods/purrable.rs
sdroege: They would be used whenever you have FFI code calling into your interface. Here it's not necessary and I don't think you would actually encounter them without exposing a C API, but probably a good idea to have them here for consistency?

That's basically the equivalent of `gio::ffi::GInitable`

sdroege: Maybe this also tells us that these `repr(C)` structs should actually be in an `ffi` module? Not sure

examples/virtual_methods/pet.rs
```diff
@@ -0,0 +1,201 @@
...
+/// The class struct is used to implement the vtable for method dispatch. The first field *must*
+/// be a pointer to the parent type. In our case, this is `GTypeInterface`
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct PetClass {
+    /// The first field in a class struct must always be the parent class struct
+    parent_class: glib::gobject_ffi::GObjectClass,
+
+    /// Virtual method for the [`PetImpl::pet`] trait method
+    pub(super) pet: Option<fn(&super::Pet) -> bool>,
```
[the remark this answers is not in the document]
sdroege: We do the same in many places in gtk-rs btw, as otherwise all the `Option`s cause quite some code bloat

examples/virtual_methods/purrable.rs
sdroege: Question is whether the `Class` / `Interface` structs should also be in that module, I guess

sdroege: Also for `Pet` as an abstract class, you would use the `ObjectSubclass::Instance` struct

sdroege:
@felinira This needs to be rebased around https://github.com/gtk-rs/gtk-rs-core/pull/1384 now, and I guess then we can get it in as a first version and improve on it later as needed?

examples/virtual_methods/pet.rs
sdroege: Handled in https://github.com/gtk-rs/gtk-rs-core/issues/1381

sdroege (approved):

PR #1384, opened 2024-05-04
glib: Decouple ObjectInterface impl from interface class struct

### LINKED ISSUE #1382: Align ObjectInterface to ObjectSubclass
We discussed on matrix that we should decouple the ObjectInterface implementation from the `repr(C)` class struct and have an empty implementation struct instead. This makes them more similar to each other, and creates a clearer separation between `ffi` code and implementation.

felinira:
This makes ObjectInterface and ObjectSubclass more similar to each other and allows for cleaner separation between the unsafe ffi code and rust implementation.

There is some bikeshedding left to be done: `InterfaceClassStruct` is quite clumsy, but reusing `ClassStruct` is not possible, as it is tailored towards object subclasses. We could call it `InterfaceStruct` to align it with `IsClass` ->`IsInterface` and similar traits that omit `Class`. Then we should also name `ObjectInterface::Class` `ObjectInterface::Interface` I suppose.

Do we need something like `glib::subclass::basic::ClassStruct` for `InterfaceClassStruct`? I would say we don't, as most interfaces define vfuncs anyway.

Closes #1382

### DIFF
--- a/glib-macros/tests/object_subclass_dynamic.rs
+++ b/glib-macros/tests/object_subclass_dynamic.rs
@@ -11,13 +11,21 @@ mod static_ {
         // impl for an object interface to register as a static type.
         #[derive(Clone, Copy)]
         #[repr(C)]
-        pub struct MyStaticInterface {
+        pub struct MyStaticInterfaceClass {
             parent: glib::gobject_ffi::GTypeInterface,
         }
 
+        unsafe impl InterfaceStruct for MyStaticInterfaceClass {
+            type Type = MyStaticInterface;
+        }
+
+        pub enum MyStaticInterface {}
+
         #[glib::object_interface]
-        unsafe impl ObjectInterface for MyStaticInterface {
+        impl ObjectInterface for MyStaticInterface {
             const NAME: &'static str = "MyStaticInterface";
+
+            type Interface = MyStaticInterfaceClass;
         }
 
         pub trait MyStaticInterfaceImpl: ObjectImpl + ObjectSubclass {}
@@ -69,15 +77,22 @@ mod module {
         // impl for a object interface to register as a dynamic type and that extends `MyStaticInterface`.
         #[derive(Clone, Copy)]
         #[repr(C)]
-        pub struct MyModuleInterface {
+        pub struct MyModuleInterfaceClass {
             parent: glib::gobject_ffi::GTypeInterface,
         }
 
+        unsafe impl InterfaceStruct for MyModuleInterfaceClass {
+            type Type = MyModuleInterface;
+        }
+
+        pub enum MyModuleInterface {}
+
         #[glib::object_interface]
         #[object_interface_dynamic]
-        unsafe impl ObjectInterface for MyModuleInterface {
+        impl ObjectInterface for MyModuleInterface {
             const NAME: &'static str = "MyModuleInterface";
             type Prerequisites = (MyStaticInterface,);
+            type Interface = MyModuleInterfaceClass;
         }
 
         pub trait MyModuleInterfaceImpl: ObjectImpl + ObjectSubclass {}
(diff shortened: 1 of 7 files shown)


sdroege:
> `InterfaceClassStruct` is quite clumsy, but reusing `ClassStruct` is not possible, as it is tailored towards object subclasses. We could call it `InterfaceStruct` to align it with `IsClass` ->`IsInterface` and similar traits that omit `Class`. Then we should also name `ObjectInterface::Class` `ObjectInterface::Interface` I suppose.

Yes I think that makes sense, both. Having the word class in the context of interfaces is just confusing, even if some GObject API calls it like that (`g_type_class_peek()` etc)

> Do we need something like `glib::subclass::basic::ClassStruct` for `InterfaceClassStruct`? I would say we don't, as most interfaces define vfuncs anyway.

I think that would be nice, especially as it's just a bit of copy&paste. It is probably not useful often, but if you do a signals/properties-only interface then this would make it a bit less code to write (there are some such interfaces!).

glib-macros/tests/object_subclass_dynamic.rs
```diff
@@ -11,13 +11,21 @@ mod static_ {
...
-        pub struct MyStaticInterface {
+        pub struct MyStaticInterfaceClass {
             parent: glib::gobject_ffi::GTypeInterface,
         }
 
+        unsafe impl InterfaceClassStruct for MyStaticInterfaceClass {
+            type Type = MyStaticInterface;
+        }
+
+        pub struct MyStaticInterface;
```
sdroege, line 22: A value of this never appears anywhere so this could also be an uninhabited type or not?

```suggestion
        pub enum MyStaticInterface {}
```

glib/src/object.rs
```diff
@@ -1317,6 +1317,15 @@ macro_rules! glib_object_wrapper {
         );
     };
 
+    (@object_interface [$($attr:meta)*] $visibility:vis $name:ident $(<$($generic:ident $(: $bound:tt $(+ $bound2:tt)*)?),+>)?, $iface:ty,
```
sdroege, line 1320: There's probably one branch of the macro that is not in use anymore now?

glib/src/subclass/types.rs
```diff
@@ -305,22 +323,22 @@ unsafe extern "C" fn interface_init<T: ObjectSubclass, A: IsImplementable<T>>(
 ) where
     <A as ObjectType>::GlibClassType: Copy,
 {
-    let iface = &mut *(iface as *mut crate::Interface<A>);
+    let klass = &mut *(iface as *mut crate::Interface<A>);
```
sdroege, line 326: OOC, could we use `g_type_class_peek_parent()` here instead of making our own copy?

sdroege (commented):
Looks good to me but I only took a short look now :) I'll check again tomorrow or Monday.

glib/src/object.rs
sdroege: Mh ok! This macro grew too much :)

glib-macros/tests/object_subclass_dynamic.rs
sdroege: I like it because it makes it more explicit that there are never values of this type

sdroege:
> Alright, that apparently also needs an addition to `ObjectInterfaceType` then to define a `GlibType` and `GlibClassType`?

Also a good thing to make more consistent between interfaces and classes I guess :) `GlibInterfaceType` though.

For classes I used those associated types elsewhere in the past too so it might also become useful for something else.

sdroege:
>  Interfaces can also inherit from other interfaces, so we can't just assume GTypeInterface

Is that so? :exploding_head: I thought that worked differently via the prerequisites.

I'm fine with keeping this for a separate PR in any case. Don't have to do everything at once.

sdroege:
But now I'm wondering if that could actually be made to work :rofl:

sdroege:
@felinira Should we get this in then or do you want to change more as part of this PR? The boxed copy of the struct and the simple generic interface struct can also be done separately, don't really have to entangle it all and this here is IMHO complete and good as it is.

PR #1392, opened 2024-05-11
Fix/silence clippy warnings

sdroege:

PR #1393, opened 2024-05-11
Fix clippy warnings

felinira:
* Fix clippy warnings in glib-macros/tests
* Annotate automatically generated property getters and setters with `#[allow(dead_code]`

### DIFF
--- a/glib-macros/tests/test.rs
+++ b/glib-macros/tests/test.rs
@@ -281,12 +281,16 @@ fn subclassable() {
             }
 
             impl ObjectImpl for Foo {}
+
+            impl Foo {
+                pub(super) fn test(&self) {}
+            }
         }
 
         pub trait FooExt: IsA<Foo> + 'static {
             fn test(&self) {
-                let _self = self.as_ref().downcast_ref::<Foo>().unwrap().imp();
-                unimplemented!()
+                let imp = self.as_ref().upcast_ref::<Foo>().imp();
+                imp.test();
             }
         }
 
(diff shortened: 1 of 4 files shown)


sdroege:
See also https://github.com/gtk-rs/gtk-rs-core/pull/1392 :rofl:

sdroege:
Yours is better though. Can you pull in my first and third commit into here, and then let's merge it?

glib-macros/tests/test.rs
```diff
@@ -281,12 +281,17 @@ fn subclassable() {
             }
 
             impl ObjectImpl for Foo {}
+
+            impl Foo {
+                pub(super) fn test(&self) {}
+            }
         }
 
+        #[allow(dead_code)]
```
sdroege, line 290: Not dead anymore :)

bilelmoussaoui (approved):

glib/src/main_context_futures.rs
```diff
@@ -355,9 +355,9 @@ impl<T: 'static> futures_core::FusedFuture for JoinHandle<T> {
     }
 }
 
-/// Safety: We can't rely on the auto implementation because we are retrieving
-/// the result as a `Box<dyn Any + 'static>` from the [`Source`]. We need to
-/// rely on type erasure here, so we have to manually assert the Send bound too.
+// Safety: We can't rely on the auto implementation because we are retrieving
+// the result as a `Box<dyn Any + 'static>` from the [`Source`]. We need to
+// rely on type erasure here, so we have to manually assert the Send bound too.
```
bilelmoussaoui, line 360:
```suggestion
// rustdoc-stripper-ignore-next
// Safety: We can't rely on the auto implementation because we are retrieving
// the result as a `Box<dyn Any + 'static>` from the [`Source`]. We need to
// rely on type erasure here, so we have to manually assert the Send bound too.
```

bilelmoussaoui: ah nvm

PR #1406, opened 2024-05-20
glib/functions: add compute_checksum_for_string

pranjalkole:
Please let me know if this would be a useful addition - I am currently porting [libipuz](https://gitlab.gnome.org/jrb/libipuz) to Rust, and need this function to do so.

I haven't fully looked at the tests and everything, let me know what else I need to implement to get this merged.

### DIFF
--- a/glib/src/functions.rs
+++ b/glib/src/functions.rs
@@ -15,7 +15,7 @@ use std::ptr;
 // #[cfg(windows)]
 // #[cfg(feature = "v2_58")]
 // use std::os::windows::io::AsRawHandle;
-use crate::{translate::*, GStr};
+use crate::{translate::*, ChecksumType, GStr};
 #[cfg(not(windows))]
 use crate::{Error, Pid, SpawnFlags};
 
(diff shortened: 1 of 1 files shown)


glib/src/auto/functions.rs
```diff
@@ -97,6 +97,21 @@ pub fn compute_checksum_for_data(
     }
 }
 
+#[doc(alias = "g_compute_checksum_for_string")]
+pub fn compute_checksum_for_string(
+    checksum_type: ChecksumType,
+    str: &str,
```
sdroege, line 103: This should probably become

```suggestion
    str: impl IntoGStr,
```

to avoid copies/allocations if we can

bilelmoussaoui: This is in  the auto/ which is wrong btw

pranjalkole: Thanks, fixed

bilelmoussaoui: No, the fix is wrong. See other functions in glib/src/functions. That is where you should add the function.

sdroege: Could also add it to `src/checksum.rs` as an associated function to `Checksum`. Either works IMHO

glib/src/functions.rs
```diff
@@ -287,3 +287,18 @@ pub fn spawn_future_local<R: 'static, F: std::future::Future<Output = R> + 'stat
...
+    str: impl IntoGStr,
+) -> Option<crate::GString> {
+    let length = str.len() as _;
+    unsafe {
+        from_glib_full(ffi::g_compute_checksum_for_string(
+            checksum_type.into_glib(),
+            str.to_glib_none().0,
+            length,
+        ))
+    }
```
sdroege, line 303:
```suggestion
    unsafe {
        str.run_with_gstr(|str| {
        from_glib_full(ffi::g_compute_checksum_for_string(
            checksum_type.into_glib(),
            str.as_ptr(),
            str.len() as _,
        ))
        })
    }
```

or so. I don't think the current version compiles

glib/src/auto/functions.rs
pranjalkole: > No, the fix is wrong. See other functions in glib/src/functions. That is where you should add the function.

I'm unable to figure out where to add the function - the other functions don't seem to be in alphabetical order...

glib/src/functions.rs
pranjalkole: Thanks, fixed

PR #1409, opened 2024-05-23
gio: more ergonomic ActionEntry initialization

allan2:
This aligns the initialization with other `gio` and `gtk4` builders.

Before:
```rs
ActionEntry::builder("name").build()
```

Proposed:
```rs
ActionEntry::new("name")
ActionEntry::builder().name("name").build()
```

### DIFF
--- a/gio/src/action_entry.rs
+++ b/gio/src/action_entry.rs
@@ -22,6 +22,16 @@ impl<O> ActionEntry<O>
 where
     O: IsA<ActionMap>,
 {
+    pub fn new(name: &str) -> Self {
+        Self {
+            name: name.to_owned(),
+            parameter_type: Default::default(),
+            state: Default::default(),
+            activate: Default::default(),
+            change_state: Default::default(),
+        }
+    }
+
     pub fn name(&self) -> &str {
         &self.name
     }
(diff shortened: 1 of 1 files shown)


sdroege:
AFAIU the name is mandatory and can't be the empty string, and that's why it is a parameter to the builder constructor.

bilelmoussaoui:
> AFAIU the name is mandatory and can't be the empty string, and that's why it is a parameter to the builder constructor.

Yes that is the whole reason why. For now, the code generator always generates that pattern for builders but ultimately, once we have a way to know which properties are required, we would generate builder patterns that match them instead of not generating them at all.

allan2:
That makes sense. It was because I saw `UriLauncher::new("myurl")` and `UriLauncher::builder().uri("myurl").build()`, so I wasn't sure whether mandatory parameters should be in the `builder` call or not.

Could I still implement `ActionEntry::new` then?

bilelmoussaoui:
> That makes sense. It was because I saw `UriLauncher::new("myurl")` and `UriLauncher::builder().uri("myurl").build()`, so I wasn't sure whether mandatory parameters should be in the `builder` call or not.

yeah, we have to figure out how to improve those in the future.

> Could I still implement `ActionEntry::new` then?


Sure

sdroege:
What's the plan here?

bilelmoussaoui:
I don't think we should merge this, the action name is required and that is why I explicitly required it when creating the builder instead of passing it later on (which the user might forget to do).

PR #1413, opened 2024-05-26
glib: Add unsafe `Value::into_send_value()`

sdroege:
This allows converting a `Value` into a `SendValue` without statically knowing its type but is unsafe because the caller needs to ensure the contained type is actually `Send`.

This is slightly safer than going through `SendValue::unsafe_from(value.into_raw())`.

### DIFF
--- a/glib/src/value.rs
+++ b/glib/src/value.rs
@@ -666,6 +666,9 @@ impl Value {
         }
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Converts a `Value` into a `SendValue`. This fails if `self` does not store a value of type
+    /// `T`. It is required for `T` to be `Send` to call this function.
     #[inline]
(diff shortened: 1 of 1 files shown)


glib/src/value.rs
```diff
@@ -675,6 +678,17 @@ impl Value {
...
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Converts a `Value` into a `SendValue`.
+    ///
+    /// # Safety
+    ///
+    /// The type of the value contained in `self` must be `Send`.
+    #[inline]
+    pub unsafe fn unsafe_into_send_value(self) -> SendValue {
```
bilelmoussaoui, line 688: i would drop the unsafe prefix :)

sdroege: Ok!

PR #1417, opened 2024-05-28
gio: make DBusConnection::register_object take optional clousures

pbor:
The documentation states that these can be NULL so that you can decide to pass only some of them.

### DIFF
--- a/gio/src/dbus_connection.rs
+++ b/gio/src/dbus_connection.rs
@@ -22,87 +22,147 @@ pub struct FilterId(NonZeroU32);
 #[derive(Debug, Eq, PartialEq)]
 pub struct SignalSubscriptionId(NonZeroU32);
 
-impl DBusConnection {
-    #[doc(alias = "g_dbus_connection_register_object_with_closures")]
-    pub fn register_object<MethodCall, SetProperty, GetProperty>(
-        &self,
-        object_path: &str,
-        interface_info: &DBusInterfaceInfo,
-        method_call: MethodCall,
-        get_property: GetProperty,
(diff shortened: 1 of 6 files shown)


gio/src/dbus_connection.rs
```diff
@@ -28,80 +28,83 @@ impl DBusConnection {
         &self,
         object_path: &str,
         interface_info: &DBusInterfaceInfo,
-        method_call: MethodCall,
-        get_property: GetProperty,
-        set_property: SetProperty,
+        method_call: Option<MethodCall>,
+        get_property: Option<GetProperty>,
+        set_property: Option<SetProperty>,
```
sdroege, line 33: Did you try to pass a `None` here? You'll notice that's impossible because you can't name any type for the closure :)

It would either have to be a `Option<Box<Fn ...>>` or you need a separate variant of the function that always uses no closures (but considering that there are 3 here, adding 9 functions seems not very useful...)

bilelmoussaoui: Could make it use a builder pattern :) similar to how we do for properties bindings

pbor: Added the builder

sdroege (dismissed):
Looks good to me but clippy is angry

pbor:
> Looks good to me but clippy is angry

heh, I changed to not use map because I thought it was more readable... oh well, switched back to `map`

sdroege:
Feel free to add an override for that. I agree that not using map was slightly more readable

pbor:
Nah, let's merge it like this (unless you see other problems)

Issue #263, opened 2021-09-08
Functions like `bus_watch_name`, `bus_own_name`, `register_object` lack `_local` variants


ids1024:
Having `_local` versions of functions that take closures seems fairly important since GObjects generally can't be `Send` or `Sync`.

Implementing this manually is a pain; I wonder if there could be a way to make that easier...

sdroege:
That would make sense to have a configuration for optionally generating also `_local` versions of such functions.

PR #1422, opened 2024-05-31
gio: remove Send + Sync requirements from DBusConnection::register_ob…

pbor:
…ject

glib documentation says:
"Calls to functions in vtable (and user_data_free_func ) will happen in the thread-default main context of the thread you are calling this method from."

### DIFF
--- a/gio/src/dbus.rs
+++ b/gio/src/dbus.rs
@@ -13,9 +13,9 @@ pub struct WatcherId(NonZeroU32);
 
 fn own_closure<F>(f: F) -> glib::Closure
 where
-    F: Fn(DBusConnection, &str) + Send + Sync + 'static,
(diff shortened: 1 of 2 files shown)


sdroege (dismissed):
Did you check if this is the only place in GDbus?

pbor:
There are other places, eg `g_bus_own_name_with_closures` but there the documentation is not explicit about where they run

sdroege:
All GDbus callbacks should run from the main thread I think? Can you check that and update it all at once so we don't have to do it one by one?

pbor:
Added the other functions.

This fixes https://github.com/gtk-rs/gtk-rs-core/issues/263

sdroege (approved):

Issue #1394, opened 2024-05-11
Change clone/closure macro syntax to play better with rustfmt and other tooling


sdroege:
E.g. the following is treated correctly by rustfmt again because it's all valid Rust syntax. Using only valid Rust syntax would also allow us to make use of syn/quote, which hopefully makes it easier to preserve spans correctly, and make rust-analyzer and other tooling work better too.

```rust
    let x = clone!(
        #[weak(x)]
        #[strong(self, rename_to = "y")]
        move |a, b, c| {
            print!("{a} {b} {c} {x} {y}");
        }
    );
```

sdroege:
I'll probably look into this during the hackfest. In the meantime if anybody has syntax suggestions...

sophie-h:
@felinira and I played around with this ages ago. I think that option c is a bit special but also much faster to type for me.

```rust
        let x = clone!(
            // a
            #[weak(x)]
            #[strong(self, rename_to = "y")]
            // b
            #[weak]
            x,
            #[stong]
            self as y,
            // c
            &'weak x,
            self as &'strong y,
            // return
            // a
            #[default_return(false)]
            // b
            #[default_return]
            false,
            // c
            return false,
            move |a, b, c| {
                print!("{a} {b} {c} {x} {y}");
            }
        );
```
For completeness: This one works as well. But I think it might be a bit confusing.

```rust
        let x = clone!(
            #[return_default]
            true,
            move |#[weak] x, #[strong] y: self, a, b, c| {
                print!("{a} {b} {c} {x} {y}");
            }
        );
```

sdroege:
Interesting ideas, thanks! Just some initial impression / thoughts on these

I kind of like c for the references but it might indeed be a bit special while the attributes look more familiar. For the attributes, I think I would prefer b. Less verbose than a and equally descriptive.

For the return either a or b seem fine. c is strange because you have a `return` statement as a parameter, which you'd never have anywhere else.

The last one ("for completeness") is indeed to confusing IMHO. Those variables are captured from the environment, not parameters to the closure.

Maybe C++ lambda capture lists can also give some inspiration here but personally I think that syntax is too compressed and strange.

DaKnig:
to me the most natural is `a` and `a` because it's what I would expect when coming from rust.

sdroege:
[...]

PR #1424, opened 2024-06-01
Update `clone!` and `closure!` macro to new syntax

sdroege:
This works correctly with rustfmt now, and thanks to using syn correctly also gives much better user experience with rust-analyzer.

Also the `closure!` macro now supports plain weak references too.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1394

### DIFF
--- a/glib-macros/src/clone.rs
+++ b/glib-macros/src/clone.rs
@@ -1,952 +1,547 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::iter::Peekable;
-
-use proc_macro::{
-    token_stream::IntoIter as ProcIter, Delimiter, Group, Ident, Literal, Punct, Spacing, Span,
-    TokenStream, TokenTree,
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::{quote, ToTokens};
+use syn::{
+    parse::{Parse, ParseStream},
+    spanned::Spanned,
+    Attribute, Expr, ExprAsync, ExprClosure, Token,
 };
 
 use crate::utils::crate_ident_new;
 
-struct PeekableProcIter {
-    inner: Peekable<ProcIter>,
-    current_span: Option<Span>,
-    next_span: Option<Span>,
-}
-
-impl<'a> From<&'a Group> for PeekableProcIter {
-    fn from(f: &'a Group) -> Self {
-        let current_span = Some(f.span());
-        let mut inner = f.stream().into_iter().peekable();
-        let next_span = inner.peek().map(|n| n.span());
-        Self {
-            inner,
-            current_span,
-            next_span,
-        }
-    }
-}
-
-impl From<TokenStream> for PeekableProcIter {
-    fn from(f: TokenStream) -> Self {
-        let mut inner = f.into_iter().peekable();
-        let next_span = inner.peek().map(|n| n.span());
-        Self {
-            inner,
-            current_span: None,
-            next_span,
-        }
-    }
-}
-
-impl Iterator for PeekableProcIter {
-    type Item = TokenTree;
-
-    fn next(&mut self) -> Option<Self::Item> {
-        let next = self.inner.next()?;
-        self.current_span = Some(next.span());
-        self.next_span = self.inner.peek().map(|n| n.span());
-        Some(next)
-    }
-}
-
-impl PeekableProcIter {
-    fn peek(&mut self) -> Option<&TokenTree> {
-        self.inner.peek()
-    }
-
-    fn generate_error(&self, message: &str) -> TokenStream {
-        self.generate_error_with_span(message, self.current_span)
-    }
-
-    fn generate_error_with_next_span(&self, message: &str) -> TokenStream {
-        self.generate_error_with_span(message, self.next_span)
-    }
-
-    fn generate_error_with_span(&self, message: &str, span: Option<Span>) -> TokenStream {
-        let span = span.unwrap_or_else(Span::call_site);
-        // We generate a `compile_error` macro and assign it to the current span so the error
-        // displayed by rustc points to the right location.
-        let mut stream = TokenStream::new();
-        stream.extend(vec![TokenTree::Literal(Literal::string(message))]);
-
-        let mut tokens = vec![
-            TokenTree::Ident(Ident::new("compile_error", span)),
-            TokenTree::Punct(Punct::new('!', Spacing::Alone)),
-            TokenTree::Group(Group::new(Delimiter::Parenthesis, stream)),
-            TokenTree::Punct(Punct::new(';', Spacing::Alone)),
-        ];
-
-        for tok in &mut tokens {
-            tok.set_span(span);
-        }
-
-        let mut stream = TokenStream::new();
-        stream.extend(tokens);
-        let mut t = TokenTree::Group(Group::new(Delimiter::Brace, stream));
-        t.set_span(span);
-
-        let mut stream = TokenStream::new();
-        stream.extend(vec![t]);
-        stream
-    }
-}
-
-#[derive(Clone, Copy, Debug)]
-enum BorrowKind {
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum CaptureKind {
+    Watch,
     Weak,
     WeakAllowNone,
     Strong,
     ToOwned,
 }
 
-impl BorrowKind {
-    fn to_str(self) -> &'static str {
-        match self {
-            Self::Weak => "@weak",
-            Self::WeakAllowNone => "@weak-allow-none",
-            Self::Strong => "@strong",
-            Self::ToOwned => "@to-owned",
-        }
+impl<'a> TryFrom<&'a Ident> for CaptureKind {
+    type Error = syn::Error;
+
+    fn try_from(s: &Ident) -> Result<Self, Self::Error> {
+        Ok(match s.to_string().as_str() {
+            "watch" => CaptureKind::Watch,
+            "strong" => CaptureKind::Strong,
+            "weak" => CaptureKind::Weak,
+            "weak_allow_none" => CaptureKind::WeakAllowNone,
+            "to_owned" => CaptureKind::ToOwned,
+            _ => {
+                // This is actually never shown to the user but we need some kind of error type for
+                // TryFrom, () would be enough but then clippy complains.
+                //
+                // We'll keep it here in case it is useful somewhere at a later time.
+                return Err(syn::Error::new(
+                    s.span(),
+                    format!("unknown capture type `{s}`"),
+                ));
+            }
+        })
     }
 }
 
-enum WrapperKind {
-    DefaultPanic,
-    DefaultReturn(String),
+#[derive(Default)]
+pub(crate) enum UpgradeBehaviour {
+    #[default]
+    Unit,
+    Panic,
+    Default,
+    Expression(Expr),
+    Closure(ExprClosure),
 }
 
-impl WrapperKind {
-    fn to_str(&self) -> String {
-        match *self {
-            Self::DefaultPanic => "@default-panic".to_owned(),
-            Self::DefaultReturn(ref r) => format!("@default-return {r}"),
-        }
-    }
+impl UpgradeBehaviour {
+    pub(crate) fn maybe_parse(
+        attrs: &[Attribute],
+        input: ParseStream,
+    ) -> syn::Result<Option<Self>> {
+        // Caller checked for empty
+        let attr = &attrs[0];
+        attr.meta.require_path_only()?;
 
-    fn keyword(&self) -> &'static str {
-        match *self {
-            Self::DefaultPanic => "default-panic",
-            Self::DefaultReturn(_) => "default-return",
-        }
-    }
-}
-
-#[derive(Debug)]
-struct ElemToClone {
-    name: String,
-    alias: Option<String>,
-    borrow_kind: BorrowKind,
-}
-
-impl ElemToClone {
-    fn to_str_before(&self) -> String {
-        match self.borrow_kind {
-            BorrowKind::Weak | BorrowKind::WeakAllowNone => format!(
-                "let {} = {}::clone::Downgrade::downgrade(&{});",
-                if let Some(ref a) = self.alias {
-                    a
-                } else {
-                    &self.name
-                },
-                crate_ident_new(),
-                self.name,
-            ),
-            BorrowKind::Strong => format!(
-                "let {} = {}.clone();",
-                if let Some(ref a) = self.alias {
-                    a
-                } else {
-                    &self.name
-                },
-                self.name,
-            ),
-            BorrowKind::ToOwned => format!(
-                "let {} = ::std::borrow::ToOwned::to_owned(&*{});",
-                if let Some(ref a) = self.alias {
-                    a
-                } else {
-                    &self.name
-                },
-                self.name,
-            ),
-        }
-    }
-
-    fn to_str_after(&self, wrapper_kind: &Option<WrapperKind>) -> String {
-        let name = if let Some(ref a) = self.alias {
-            a
-        } else {
-            &self.name
+        let Some(attr_name) = attr.path().get_ident() else {
+            return Ok(None);
         };
-        match (self.borrow_kind, wrapper_kind) {
-            (BorrowKind::Weak, Some(WrapperKind::DefaultPanic)) => {
+
+        let upgrade_behaviour = match attr_name.to_string().as_str() {
+            "upgrade_or" => {
+                let expr = input.parse::<Expr>()?;
+                input.parse::<Token![,]>()?;
+                UpgradeBehaviour::Expression(expr)
+            }
+            "upgrade_or_else" => {
+                let closure = input.parse::<ExprClosure>()?;
+                if closure.asyncness.is_some() {
+                    return Err(syn::Error::new_spanned(
+                        &closure,
+                        "`upgrade_or_else` closure needs to be a non-async closure",
+                    ));
+                }
+                if !closure.inputs.is_empty() {
+                    return Err(syn::Error::new_spanned(
+                        &closure,
+                        "`upgrade_or_else` closure must not have any parameters",
+                    ));
+                }
+
+                input.parse::<Token![,]>()?;
+                UpgradeBehaviour::Closure(closure)
+            }
+            "upgrade_or_default" => UpgradeBehaviour::Default,
+            "upgrade_or_panic" => UpgradeBehaviour::Panic,
+            _ => {
+                return Ok(None);
+            }
+        };
+
+        if attrs.len() > 1 {
+            return Err(syn::Error::new_spanned(
+                &attrs[1],
                 format!(
-                    "\
-let {0} = match {1}::clone::Upgrade::upgrade(&{0}) {{
-    Some(val) => val,
-    None => panic!(
-        \"failed to upgrade `{0}` (if you don't want to panic, use @default-return)\",
-    ),
-}};",
-                    name,
-                    crate_ident_new(),
-                )
-            }
-            (BorrowKind::Weak, Some(WrapperKind::DefaultReturn(ref r))) => {
-                let not_unit_ret = r.chars().any(|c| c != '(' && c != ')' && c != ' ');
+                    "upgrade failure attribute must not be followed by any other attributes. Found {} more attribute{}",
+                    attrs.len() - 1,
+                    if attrs.len() > 2 { "s" } else { "" },
+            )));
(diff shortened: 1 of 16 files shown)


sdroege:
I'll also update the closure macro in a bit once this is cleaned up (and re-use part of the code here I guess).

glib-macros/src/lib.rs
```diff
@@ -153,36 +173,45 @@ use utils::{parse_nested_meta_items_from_stream, NestedMetaItem};
...
 /// # use glib_macros::clone;
 /// # use std::rc::Rc;
 /// # let v = Rc::new(1);
-/// let closure = clone!(@weak v => @default-panic, move |x| {
-///     println!("v: {}, x: {}", v, x);
-///     true
-/// });
+/// let closure = clone!(
+///     #[weak] v,
+///     #[default_return(panic)]
```
sdroege, line 209: Are there opinions on this? Should it be called something like `#[default_on_weak_upgrade_failure(panic)]` and `#[default_on_weak_upgrade_failure(return)]` or so instead?

GuillaumeGomez: Simpler is better I think so as is looks good to me.

pbor: Not strong opinion, but I would prefer to have `clone_or!(false, ...` and `clone_or_panic!` variants and do not allow `#[weak]` in the simple `clone!`.

sdroege:
closure macro is also covered now

glib-macros/src/clone_old.rs
```diff
@@ -0,0 +1,965 @@
...
+    // This part is creating the TokenStream using the variables that needs to be cloned (from the
+    // @weak and @strong annotations).
+    let mut inner: Vec<TokenTree> = Vec::new();
+
+    {
+        let stream = "{
+#[deprecated = \"Using old-style clone! syntax\"]
+macro_rules! clone { () => {}; }
+clone!();
+}"
```
sdroege, line 858: `clone_old.rs` and `closure_old.rs` are exactly the same as the previous versions plus the addition of this for generating deprecation warnings.

github unfortunately doesn't detect that they're basically the same files.

With the next major release (0.21) we would drop the old code.

glib-macros/src/lib.rs
sdroege: Having separate variants seems like something we'll repeatedly have to tell people about when they ask why `#[weak]` does not work

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
-impl BorrowKind {
-    fn to_str(self) -> &'static str {
-        match self {
-            Self::Weak => "@weak",
-            Self::WeakAllowNone => "@weak-allow-none",
-            Self::Strong => "@strong",
-            Self::ToOwned => "@to-owned",
-        }
+impl CaptureKind {
+    pub(crate) fn kinds() -> &'static [&'static str] {
```
GuillaumeGomez, line 22: Instead of checking like that, why not implement `TryFrom<str>` (or `TryFromStr`, never remember the name)? Like that you can make `Capture::new` take `CaptureKind` argument.

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
-                            "Unexpected `.` after `{}`",
-                            borrow_kind.to_str()
-                        )));
+        let mut alias = None;
+        if attr.meta.require_list().is_ok() {
+            attr.parse_nested_meta(|meta| {
+                if meta.path.is_ident("rename_to") {
+                    let value = meta.value()?;
+                    let lit = value.parse::<LitStr>()?;
+                    let id = lit.parse::<Ident>()?;
```
GuillaumeGomez, line 111: Wouldn't it be simpler to ask for an ident directly?

```rust
#[rename_to = something]
```

doesn't seem that bad either.

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
+                    let #alias = ::core::convert::AsRef::as_ref(&#alias);
                 }
-                ret.push_str(&i.to_string());
             }
-            Some(TokenTree::Punct(p)) if p.to_string() == "-" => {
-                if !prev_is_ident {
-                    break;
+            CaptureKind::Weak => match weak_upgrade_failure_kind {
+                DefaultReturn::Panic => {
+                    let err_msg = format!("Failed to upgrade `{}` (if you don't want to panic, use `#[default_return]`", alias);
```
GuillaumeGomez, line 216: You forgot to close the paren in your error message. ;)

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
+
+impl Parse for Clone {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut captures: Vec<Capture> = vec![];
+        let mut default_return: DefaultReturn = DefaultReturn::default();
+
+        loop {
+            // Parse none, one or more attributes.
+            //
+            // There must either be one or no attributes here. Multiple attributes are not
```
GuillaumeGomez, line 294: It contradicts your sentence just above which says "none, one or more attributes".

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
+                        &attrs[1],
+                        "Variable capture attributes must be followed by an identifier",
+                    ));
+                }
+                captures.push(Capture::new(attr, input)?);
+            } else if attr_name.is_ident("default_return") {
+                if attrs.len() > 1 {
+                    return Err(syn::Error::new_spanned(
+                        &attrs[1],
+                        "The `default_return` attribute must not be followed by more attributes",
```
GuillaumeGomez, line 326:
```suggestion
                        &format!("The `default_return` attribute must be followed by 0 or 1 attributes, found {}", attrs.len()),
```

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
-            }
-        }
-        Delimiter::Bracket => {
-            if open {
-                "["
+                let next_attrs = &input.call(Attribute::parse_outer)?;
+                if !next_attrs.is_empty() {
+                    return Err(syn::Error::new_spanned(
+                        &attrs[0],
+                        "The `default_return` attribute must not be followed by more attributes",
```
GuillaumeGomez, line 336: Same as above.

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
-                } else if p_s == "," && total == 0 {
-                    return Ok(ret);
+        // Following is a closure or async block
+        let expr = input.parse::<Expr>()?;
+        let body = match expr {
+            Expr::Async(async_) => {
+                if async_.capture.is_none() {
+                    return Err(syn::Error::new_spanned(
+                        async_,
+                        "Async blocks with the `move` keyword required",
```
GuillaumeGomez, line 356:
```suggestion
                        "Async blocks must be moved. Use the `move` keyword to do so",
```

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
+                ClosureOrAsync::Async(async_)
             }
-            Some(x) => {
-                if total == 0 && !ret.ends_with(':') {
-                    return Ok(ret);
+            Expr::Closure(closure) => {
+                if closure.capture.is_none() {
+                    return Err(syn::Error::new_spanned(
+                        closure,
+                        "Closures with the `move` keyword required",
```
GuillaumeGomez, line 366: Same as above. You should maybe write a small common function to only write the message in one place (same for the wrong number of attributes above).

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,487 @@
...
+        });
+
+        let default_return_closure =
+            if let DefaultReturn::Return(Some(ref closure)) = self.default_return {
+                Some(quote! {
+                    let #default_return_closure_ident = #closure;
+                })
+            } else {
+                None
+            };
```
GuillaumeGomez, line 418: I think a `match` would look a bit better:

```suggestion
        let default_return_closure = match self.default_return {
             DefaultReturn::Return(Some(ref closure)) => Some(quote! {
                 let #default_return_closure_ident = #closure;
            }),
            _ => None,
        };
```

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,98 @@
...
         let mut captures: Vec<Capture> = vec![];
-        if input.peek(Token![@]) {
-            loop {
-                let capture = input.parse::<Capture>()?;
+        let mut default_return: DefaultReturn = DefaultReturn::default();
+
+        loop {
+            // Parse none, one or more attributes.
+            //
+            // There must either be one or no attributes here. Multiple attributes are not
```
GuillaumeGomez, line 31: Contradicts above comment.

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,98 @@
...
-                    if !input.peek(Token![@]) {
-                        break;
-                    }
-                } else {
-                    break;
+            } else if attr_name.is_ident("default_return") {
+                if attrs.len() > 1 {
+                    return Err(syn::Error::new_spanned(
+                        &attrs[1],
+                        "The `default_return` attribute must not be followed by more attributes",
```
GuillaumeGomez, line 68: Same as mentioned for the other file.

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,98 @@
...
+                    ));
                 }
+
+                default_return = DefaultReturn::new(attr, input)?;
+
+                let next_attrs = &input.call(Attribute::parse_outer)?;
+                if !next_attrs.is_empty() {
+                    return Err(syn::Error::new_spanned(
+                        &attrs[0],
+                        "The `default_return` attribute must not be followed by more attributes",
```
GuillaumeGomez, line 78: Same as mentioned for the other file.

glib-macros/src/lib.rs
```diff
@@ -87,17 +92,21 @@ use utils::{parse_nested_meta_items_from_stream, NestedMetaItem};
...
+/// );
 ///
 /// closure(3);
 /// ```
 ///
 /// #### Allowing a nullable weak reference
 ///
 /// In some cases, even if the weak references can't be retrieved, you might want to still have
-/// your closure called. In this case, you need to use `@weak-allow-none`:
+/// your closure called. In this case, you need to use `#[weak_allow_none]` instead of `#[weak]:
```
GuillaumeGomez, line 109:
```suggestion
/// your closure called. In this case, you need to use `#[weak_allow_none]` instead of `#[weak]`:
```

glib-macros/src/clone.rs
sdroege: I didn't know that was possible. Better, thanks!

glib-macros/src/clone.rs
sdroege: The block is not moved, the captured variables are

glib-macros/src/clone.rs
sdroege: Both messages are actually different though

glib-macros/src/clone.rs
GuillaumeGomez: Proof that the error message needs improvement. :wink:

sdroege:
@GuillaumeGomez Updated everything, thanks for the review :)

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,495 @@
...
-    });
-    Ok(())
-}
+                default_return = DefaultReturn::new(attr, input)?;
+
+                let next_attrs = &input.call(Attribute::parse_outer)?;
+                if !next_attrs.is_empty() {
+                    return Err(syn::Error::new_spanned(
+                        &next_attrs[0],
+                        format!(
```
GuillaumeGomez, line 342: It's definitely worth having a common function since it has the same text and the logic as the error above.

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,107 @@
...
+                    ));
                 }
+
+                default_return = DefaultReturn::new(attr, input)?;
+
+                let next_attrs = &input.call(Attribute::parse_outer)?;
+                if !next_attrs.is_empty() {
+                    return Err(syn::Error::new_spanned(
+                        &next_attrs[0],
+                        format!(
```
GuillaumeGomez, line 83: Same as mentioned above, common function please. :)

glib-macros/src/closure_old.rs
```diff
@@ -0,0 +1,300 @@
...
+                    name,
+                    "Can't use `self` as variable name. Try storing it in a temporary variable or \
+                    rename it using `as`.",
+                ));
+            }
+            if name.to_string().contains('.') {
+                return Err(syn::Error::new(
+                    name.span(),
+                    format!(
+                        "`{}`: Field accesses are not allowed as is, you must rename it!",
```
GuillaumeGomez, line 136: You could use (not a blocker, just a style thing, I personally have no preference, just wanted to mention it):

```suggestion
                        "`{name}`: Field accesses are not allowed as is, you must rename it!",
```

glib-macros/src/closure_old.rs
```diff
@@ -0,0 +1,300 @@
...
+    fn parse(input: syn::parse::ParseStream) -> syn::Result<Self> {
+        let mut captures: Vec<Capture> = vec![];
+        if input.peek(Token![@]) {
+            loop {
+                let capture = input.parse::<Capture>()?;
+                if capture.kind == CaptureKind::Watch {
+                    if let Some(existing) = captures.iter().find(|c| c.kind == CaptureKind::Watch) {
+                        return Err(syn::Error::new(
+                            existing.start,
+                            "Only one `@watch` capture is allowed per closure",
```
GuillaumeGomez, line 168: Didn't see any test for this error. Please add one.

glib-macros/src/closure_old.rs
GuillaumeGomez: Please ignore, old file.

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,107 @@
...
+            if let Some(capture_kind) = attr_name
+                .get_ident()
+                .and_then(|i| CaptureKind::try_from(i).ok())
+            {
+                if capture_kind == CaptureKind::Watch {
                     if let Some(existing) = captures.iter().find(|c| c.kind == CaptureKind::Watch) {
                         return Err(syn::Error::new(
                             existing.start,
-                            "Only one `@watch` capture is allowed per closure",
+                            "Only one watch capture is allowed per closure",
```
GuillaumeGomez, line 51: Didn't see any test for this error. Please add one.

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,495 @@
...
+        Ok(match s.to_string().as_str() {
+            "watch" => CaptureKind::Watch,
+            "strong" => CaptureKind::Strong,
+            "weak" => CaptureKind::Weak,
+            "weak_allow_none" => CaptureKind::WeakAllowNone,
+            "to_owned" => CaptureKind::ToOwned,
+            _ => {
+                return Err(syn::Error::new(
+                    s.span(),
+                    format!("Unknown capture type `{s}`"),
```
GuillaumeGomez, line 34: Didn't see any test for this error. Please add one.

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,495 @@
...
+        if attr.meta.require_list().is_ok() {
+            attr.parse_nested_meta(|meta| {
+                if meta.path.is_ident("panic") {
+                    panic = true;
                 } else {
-                    &self.name
-                },
-                self.name,
-            ),
+                    return Err(meta.error("Unsupported `default_return` property"));
```
GuillaumeGomez, line 63: Didn't see any test for this error. Please add one.

glib-macros/src/closure.rs
GuillaumeGomez: Also, please put `watch` between backticks.

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,107 @@
...
                         ));
                     }
                 }
+                if attrs.len() > 1 {
+                    return Err(syn::Error::new_spanned(
+                        &attrs[1],
+                        "Variable capture attributes must be followed by an identifier",
```
GuillaumeGomez, line 58: Didn't see any test for this error. Please add one.

glib-macros/src/closure.rs
```diff
@@ -1,193 +1,107 @@
...
-                        break;
-                    }
-                } else {
-                    break;
+            } else if attr_name.is_ident("default_return") {
+                if attrs.len() > 1 {
+                    return Err(syn::Error::new_spanned(
+                        &attrs[1],
+                        format!(
+                            "The `default_return` attribute must not be followed by any other attributes. Found {} more attribute{}",
```
GuillaumeGomez, line 70: Didn't see any test for this error. Please add one.

glib-macros/src/clone.rs
sdroege: Except that it's not :rofl: It's looking at different arrays, different indices, does the length handling different.

glib-macros/src/clone.rs
sdroege: My message says exactly what is required: people have to add the `move` keyword.

Do you have a suggestion for a better message?

glib-macros/src/clone.rs
GuillaumeGomez: I mean for the error, you can pass value, slice and everything.

glib-macros/src/lib.rs
DaKnig: how about `#[on_upgrade_failure(return)]` ? `default` doesnt mean anything in this context...

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,495 @@
...
                 }
-                // This is to prevent to push `-` if the next item isn't an ident.
-                prev_is_ident = false;
-                stored = true;
-            }
-            _ => break,
+                DefaultReturn::Return(Some(_)) => {
+                    let err_msg = format!("Failed to upgrade `{}`", alias);
+                    quote! {
+                        let #alias = match #crate_ident::clone::Upgrade::upgrade(&#alias) {
```
DaKnig, line 230: let ... else

glib-macros/src/lib.rs
sdroege: I'd prefer something shorter. We'll have some time for bikeshedding here before this is merged :)

glib-macros/src/closure.rs
sdroege: There are no tests for `closure!` error messages at all

sdroege: I.e. I'll add those

GuillaumeGomez (approved):

glib-macros/src/lib.rs
sdroege: I prefer the variant you list under `my preference`. Also over what is implemented here right now. Unless there are more opinions, I'll change it to that and implement the two other variants you mentioned.

sdroege: As discussed on Matrix that might be too implicit and cause `false` to be returned unintentionally.

sdroege: `upgrade_or_panic` seems most descriptive to me

glib-macros/src/clone.rs
```diff
@@ -1,952 +1,505 @@
...
+                let #alias = #crate_ident::object::Watchable::watched_object(&#name);
+            },
+            CaptureKind::Weak | CaptureKind::WeakAllowNone => quote! {
+                let #alias = #crate_ident::clone::Downgrade::downgrade(&#name);
+            },
+            CaptureKind::Strong => quote! {
+                let #alias = #name.clone();
+            },
+            CaptureKind::ToOwned => quote! {
+                let #alias = ::std::borrow::ToOwned::to_owned(&*#name);
```
DaKnig, line 194:
```suggestion
                let #alias = std::borrow::ToOwned::to_owned(&*#name);
```

sdroege: Why?

glib-macros/src/lib.rs
sdroege: So let's summarize: `#[upgrade_or_else]` for the closure, `#[upgrade_or]` for a plain value (needs to be `Copy`!), `#[upgrade_or_default]` for returning `Default::default()`, `#[upgrade_or_panic]` for panicking and if nothing is specified then `()` is returned (and it will fail to compile if the closure is supposed to return a different type).

sdroege:
Updated everything accordingly and will merge once the CI is happy. Everything else can happen on top of this

PR #1425, opened 2024-06-02
Regenerate with properties/async attributes support

bilelmoussaoui:


gio/src/auto/application.rs
```diff
@@ -443,11 +454,6 @@ pub trait ApplicationExt: IsA<Application> + sealed::Sealed + 'static {
         }
     }
 
-    #[doc(alias = "action-group")]
```
bilelmoussaoui, line 446: This actually makes to remove. As the getter was deprecated upstream in 2.3x

gio/src/auto/dbus_connection.rs
```diff
@@ -798,10 +801,6 @@ impl DBusConnection {
         }
     }
 
-    pub fn get_property_flags(&self) -> DBusConnectionFlags {
```
bilelmoussaoui, line 801: It has a public getter that was added in 2.60. We could add a getter manually for not(v2_60) but that is annoying to do manually... I was thinking of adding that to gir but is it worth the work? 

there are some more cases in a gstreamer crate and also in gtk4 but not that many...

bilelmoussaoui: this is now fixed

gio/src/auto/application.rs
bilelmoussaoui: But the property wasn't marked as deprecated, see https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4101

sdroege:
This is still waiting on the gir file update, right?

bilelmoussaoui:
Yes, should trigger an update of the gir files and see if that landed already

gio/src/auto/application.rs
sdroege: That's there now, you can regenerate this :) Also lots of gtk4-related changes that look relevant

sdroege:
This can be regenerated now

gio/src/auto/file.rs
```diff
@@ -268,30 +268,12 @@ pub trait FileExt: IsA<File> + sealed::Sealed + 'static {
...
-                io_priority,
-                cancellable.map(|p| p.as_ref()).to_glib_none().0,
-                progress_callback_closure.to_glib_none().0,
-                ready_callback_closure.to_glib_none().0,
-            );
-        }
-    }
+    //#[cfg(feature = "v2_82")]
+    //#[cfg_attr(docsrs, doc(cfg(feature = "v2_82")))]
+    //#[doc(alias = "g_file_copy_async_with_closures")]
```
bilelmoussaoui, line 273: The fix for this still hasn't landed on the gir file? cc @A6GibKm

A6GibKm: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4099 I think it is only in main, I don't think I have updated this gir file on gir-files.

A6GibKm: Latest update to the gir at gir-files was 2 weeks ago, patch was merged a week ago, so no, it is not in there.

bilelmoussaoui: I see it here `   <method name="copy_async_with_closures" c:identifier="g_file_copy_async_with_closures" shadows="copy_async" version="2.82" glib:finish-func="copy_finish">`

bilelmoussaoui: From a quick look, it seems like the finish-func annotation is wrongly set in that function. It probably needs a manual implementation to turn it into a future instead of still using closures.

sdroege: @bilelmoussaoui can you ping ebassi about that? Would be good to get this in to not have this PR bitrot

sdroege:
This here also updates `gir-files`. So that's still needed

PR #1427, opened 2024-06-03
ci: bump gvsbuild

pbor:
I think we are still using a pre-release version of glib... Update to the latest version.


sdroege (approved):
Good for me once the CI passes

PR #1429, opened 2024-06-08
gio: Properly export Win32InputStream / Win32OutputStream traits

sdroege:
And de-duplicate them.


gio/src/prelude.rs
```diff
@@ -27,6 +27,10 @@ pub use crate::unix_input_stream::UnixInputStreamExtManual;
 pub use crate::unix_output_stream::UnixOutputStreamExtManual;
 #[cfg(unix)]
 pub use crate::unix_socket_address::{UnixSocketAddressExtManual, UnixSocketAddressPath};
+#[cfg(windows)]
+pub use crate::win32_input_stream::Win32InputStreamExt;
```
bilelmoussaoui, line 31: Make sure the gir.toml contains those as manual traits while you are at it

bilelmoussaoui (dismissed):

gio/src/prelude.rs
sdroege: Good point, thanks and done :)

PR #1431, opened 2024-06-13
Derive TransparentPtrType trait for Boxed

pbor:


bilelmoussaoui (approved):

sdroege:
I don't think this is correct, at least not for the shared case. `from_raw()` and `into_raw()` are not just returning `&Arc<T> as *const T` or so but do some offsetting.

Can you add tests for both cases somewhere (using a pointer slice or something else that requires the traits) to ensure this is actually working correctly? Also if these traits are implemented you can implement `FromValue` for references like `glib::wrapper!` does.

pbor:
Thanks for catching that SharedBoxed was a bad idea... indeed I did it by cut&paste without thinking :(

> Can you add tests for both cases somewhere (using a pointer slice or something else that requires the traits) to ensure this is actually working correctly? 
I added a small test for the boxed case, but please check if you think it is enough to exercise this.

> you can implement FromValue for references like glib::wrapper! does
As far as I can see that is already there: https://github.com/gtk-rs/gtk-rs-core/blob/master/glib-macros/src/boxed_derive.rs#L56

sdroege (approved):

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/1559. This is going to be changed to `TransparentType` as `TransparentPtrType` was simply wrong.

PR #1439, opened 2024-06-17
Bump MSRV to 1.74

bilelmoussaoui:
Silencing the new clap (or something else?) requirement

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,7 +18,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - "1.70.0"
+          - "1.74.0"
         conf:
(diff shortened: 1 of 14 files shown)


sdroege:
I've pushed c69f816d53925c5e7d17bd09ad8ef33aecb4947d to downgrade it for now in Cargo.lock. `cargo update -p clap@4.5.7 --precise 4.4.18` FWIW.

I'm not against updating the MSRV though, but we should do this coordinated and not just to a random version. Ideally I'd like us to go to 1.80 so we can make use of `LazyLock` etc but that's after our targetted next release date. Is there anything in 1.75 or so that would be useful for us?

bilelmoussaoui:
Let us close this one, until 1.80 is out and then we can reevaluate

PR #1441, opened 2024-06-18
Backport: Update `clone!` and `closure!` macro to new syntax

pbor:
This works correctly with rustfmt now, and thanks to using syn correctly also gives much better user experience with rust-analyzer.

Also the `closure!` macro now supports plain weak references too.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1394

### DIFF
--- a/glib-macros/src/clone_old.rs
+++ b/glib-macros/src/clone_old.rs
@@ -1,0 +1,965 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::iter::Peekable;
+
+use proc_macro::{
+    token_stream::IntoIter as ProcIter, Delimiter, Group, Ident, Literal, Punct, Spacing, Span,
+    TokenStream, TokenTree,
+};
+
+use crate::utils::crate_ident_new;
+
+struct PeekableProcIter {
+    inner: Peekable<ProcIter>,
+    current_span: Option<Span>,
+    next_span: Option<Span>,
+}
+
+impl<'a> From<&'a Group> for PeekableProcIter {
+    fn from(f: &'a Group) -> Self {
+        let current_span = Some(f.span());
+        let mut inner = f.stream().into_iter().peekable();
+        let next_span = inner.peek().map(|n| n.span());
+        Self {
+            inner,
+            current_span,
+            next_span,
+        }
+    }
+}
+
+impl From<TokenStream> for PeekableProcIter {
+    fn from(f: TokenStream) -> Self {
+        let mut inner = f.into_iter().peekable();
+        let next_span = inner.peek().map(|n| n.span());
+        Self {
+            inner,
+            current_span: None,
+            next_span,
+        }
+    }
+}
+
+impl Iterator for PeekableProcIter {
+    type Item = TokenTree;
+
+    fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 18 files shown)


glib-macros/src/closure_old.rs
```diff
@@ -0,0 +1,300 @@
...
+            }
+        } else {
+            quote! {
+                {
+                    #[deprecated = "Using old-style closure_local! syntax"]
+                    macro_rules! closure_local { () => {}; }
+                    closure_local!();
+                }
+            }
+        };
```
sdroege, line 264: Should remove this part for stable

glib-macros/src/clone_old.rs
```diff
@@ -0,0 +1,965 @@
...
+        let stream = "{
+#[deprecated = \"Using old-style clone! syntax\"]
+macro_rules! clone { () => {}; }
+clone!();
+}"
+        .parse::<TokenStream>()
+        .expect("can't parse deprecation");
+
+        inner.extend(stream);
+    }
```
sdroege, line 863: Should remove this part for stable

glib-macros/src/lib.rs
```diff
@@ -275,16 +323,31 @@ use utils::{parse_nested_meta_items_from_stream, NestedMetaItem};
...
+    // Check if this is an old-style clone macro invocation.
+    // These always start with an '@' punctuation.
+    let Some(first) = item.clone().into_iter().next() else {
+        return syn::Error::new(Span::call_site(), "expected a closure or async block")
+            .to_compile_error()
+            .into();
+    };
+
+    match first {
+        TokenTree::Punct(ref p) if p.to_string() == "@" => clone_old::clone_inner(item),
```
sdroege, line 348: glib should get an `unstable-clone-syntax` feature flag, that maps to the same flag in `glib-macros`. If that flag is *not* enabled, then this and the other functions are directly mapping to `clone_old` / `closure_old` instead of doing anything else like they do now

sdroege:
Also need to include https://github.com/gtk-rs/gtk-rs-core/pull/1443 once merged

glib-macros/src/lib.rs
```diff
@@ -336,17 +336,21 @@ use utils::{parse_nested_meta_items_from_stream, NestedMetaItem};
...
 
-    match first {
-        TokenTree::Punct(ref p) if p.to_string() == "@" => clone_old::clone_inner(item),
-        _ => clone::clone_inner(item),
+        match first {
+            TokenTree::Punct(ref p) if p.to_string() == "@" => clone_old::clone_inner(item),
+            _ => clone::clone_inner(item),
+        }
+    } else {
+        clone::clone_inner(item)
```
sdroege, line 303: `clone_old::clone_inner(item)`

You're always calling the new one now :) Same for closure/closure_local below

glib-macros/tests/clone.rs
```diff
@@ -4,6 +4,169 @@ use std::rc::Rc;
 
 use glib::clone;
 
+#[test]
+#[cfg(feature = "unstable-clone-syntax")]
```
sdroege, line 8: These tests should probably continue to test the old implementation. Also tests/closure.rs

Otherwise looks good to me

sdroege (approved):

PR #1443, opened 2024-06-20
glib-macros: Don't produce unnecessary braces in `clone!(async move {…

sdroege:
… x })`

This would wrongly suggest to remove the braces, which is invalid syntax, because the macro already adds another layer of braces around the body of the async block.

To avoid this, don't output the braces of the async block in the macro expansion and instead only use the braces added by the macro.

### DIFF
--- a/glib-macros/src/clone.rs
+++ b/glib-macros/src/clone.rs
@@ -524,6 +524,11 @@ impl ToTokens for Clone {
                     block,
                 } = a;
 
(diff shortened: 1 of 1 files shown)


sdroege:
This also affects the old macro but I can't be bothered to fix it there

bilelmoussaoui (approved):

pbor (approved):

Issue #1064, opened 2023-03-22
gio::SocketService: Add a `Stream` implementation for the incoming connections


sdroege:
See title

jf2048:
I think this would be similar to https://github.com/gtk-rs/gtk-rs-core/issues/431

carlosmn:
Would it have to be a `Stream`? I've been looking into it but it looks like the regular way you'd do this with tokio or async-std has the shape of 

```rust
while let Ok(stream) = listener.accept().await {
    accept_connection(stream)
}
```

which seems more like a one-shot bridge between the signal and the async world. We can of course still have it as

```rust
let incoming = service.incoming();
while let Some(stream) = incoming.next() {
   accept_connection(stream)
}
```

but it might be beneficial for it to look more like the other async world. Or it might be more beneficial to look more like other glib stuff?

The underlying `SocketListener` has a whole host of `accept*` methods which would make this a bit harder to differentiate (including `accept_async`).

carlosmn:
Actually async-std does provide an `incoming()` method so either way there is prior art.

But while working on this it looks like a bigger issue is the inversion of control between the callback method and the Stream. On a Stream, you choose when to accept the next connection, whereas with callbacks, the `SocketService` decides when to call.

The naïve implementation of passing along connections from the `connect_incoming` into a channel that the caller can listen on works well enough for most cases, but the `SocketService` is going to choose how quickly new connections are added to that list, and they've already been accepted.

A more complex solution here I think would wait for a `poll_next` and then subscribe to `connect_incoming` for one call and pass back a connection whenever it gets it. Looking through the `SockerService` code, I'm not sure if this would stop it from accepting connections we're not ready for, but even if it does, it seems wasteful how much setup and teardown we're doing.

Given that "all" it does is call `SocketListener`'s `accept_async` and emit the signal when that completes, it might make more sense for this `Stream` to exist as part of `SocketListener` so that on `poll_next` it calls `SocketListener::accept_async` so we can call the waker when that completes. If the inheritance rules work out, that would also make it available on `SocketService`(and `ThreadedSocketService` though it would not very useful if we do want threads).

All this to say, it feels like `SocketService` is the callback version that makes `SocketListener` less awkward in single-threaded code and an async `Stream` would be the solution in the async world, so building it on top of it is going to be super awkward.

carlosmn:
Given #1454 is this done now? Or is there more?

sdroege:
[...]

PR #1454, opened 2024-07-08
gio: Add a method to get a stream of incoming connections to SocketListener

carlosmn:
This object provides a `Stream` over which we can iterate in order to accept connections in an async environment.

This is analogous to the one provided by async-std's `TcpListener`.

This is an attempt at #1064. As I mention there, I think `SocketService` is the wrong abstraction to use in async so this is implemented for `SocketListener`, so that this `incoming()` is by itself an alternative to `SocketService`.

The type that we return here feels pretty awkward but it's what this returns. The pattern with the source object feels like another instance of something that makes sense without easy async but with async Rust you'd create multiple objects? But maybe I don't see other use-cases for it here.

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -118,7 +118,6 @@ generate = [
     "Gio.SocketConnectable",
     "Gio.SocketConnection",
     "Gio.SocketFamily",
-    "Gio.SocketListener",
     "Gio.SocketListenerEvent",
     "Gio.SocketProtocol",
     "Gio.SocketService",
@@ -1332,6 +1331,11 @@ manual_traits = ["SocketControlMessageExtManual"]
     # writes to a void pointer
     manual = true
     doc_trait_name = "SocketControlMessageExtManual"
+
+[[object]]
+name = "Gio.SocketListener"
+status = "generate"
+manual_traits = ["SocketListenerExtManual"]
 
 [[object]]
 name = "Gio.Subprocess"
--- a/gio/src/lib.rs
+++ b/gio/src/lib.rs
@@ -82,6 +82,7 @@ mod simple_proxy_resolver;
 mod socket;
 pub use socket::{InputMessage, InputVector, OutputMessage, OutputVector, SocketControlMessages};
 mod socket_control_message;
+mod socket_listener;
 mod socket_msg_flags;
 pub use socket_msg_flags::SocketMsgFlags;
 mod subprocess;
--- a/gio/src/socket_listener.rs
+++ b/gio/src/socket_listener.rs
@@ -1,0 +1,56 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::{pin::Pin, task::ready};
+
+use futures_core::{
+    stream::Stream,
+    task::{Context, Poll},
+    Future,
+};
+
+use crate::{prelude::SocketListenerExt, SocketConnection, SocketListener};
+use glib::{prelude::*, Error, Object};
+
+pub struct Incoming {
+    listener: SocketListener,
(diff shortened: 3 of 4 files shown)


gio/src/socket_listener.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        let fut = self.fut.as_mut().unwrap();
+        let res = ready!(Pin::new(fut).poll(ctx));
+        self.fut.take();
+
+        Poll::Ready(Some(res))
+    }
+}
+
+pub trait SocketListenerIncomingStream: SocketListenerExt {
```
sdroege, line 35: Rename it to `SocketListenerExtManual`  instead, and also add it to `Gir.toml` as `manual_trait`

sdroege (commented):
Looks good otherwise, thanks!

gio/src/socket_listener.rs
carlosmn: Ah gotcha, I was under the impression that name was for things that already existed but were implemented manually.

carlosmn:
Alright, updated. I *think* the change in Gir.toml is right, but it's the first time I've had to touch something like that.

gio/Gir.toml
```diff
@@ -1333,6 +1332,15 @@ manual_traits = ["SocketControlMessageExtManual"]
     manual = true
     doc_trait_name = "SocketControlMessageExtManual"
 
+[[object]]
+name = "Gio.SocketListener"
+status = "generate"
+manual_traits = ["SocketListenerExtManual"]
+    [[object.function]]
+    name = "incoming"
```
sdroege, line 1340: Is this function does not exist in C, you don't have to list it here :) It has no effect

carlosmn: I see. So just the `manual_traits` bit is what's important?

gio/src/socket_listener.rs
```diff
@@ -0,0 +1,55 @@
...
+        let fut = self.fut.as_mut().unwrap();
+        let res = ready!(Pin::new(fut).poll(ctx));
+        self.fut.take();
+
+        Poll::Ready(Some(res))
+    }
+}
+
+pub trait SocketListenerExtManual: SocketListenerExt {
+    /// Returns a stream of incoming connections
```
sdroege, line 36:
```suggestion
    // rustdoc-stripper-ignore-next
    /// Returns a stream of incoming connections
```

and then the CI will also be happy :)

gio/src/lib.rs
```diff
@@ -82,6 +82,8 @@ mod simple_proxy_resolver;
 mod socket;
 pub use socket::{InputMessage, InputVector, OutputMessage, OutputVector, SocketControlMessages};
 mod socket_control_message;
+mod socket_listener;
+pub use socket_listener::SocketListenerExtManual;
```
bilelmoussaoui, line 86: prelude only for traits in general

carlosmn: I see, yes, the other `ExtManual` traits are in prelude, which makes sense.

gio/Gir.toml
sdroege: Not sure how this came back

carlosmn: :/ I think I switched computers and started working off of the wrong base

carlosmn:
Well after way too much trial-and-error, this should be good to go now, but I don't understand that "docs embed" check as it seems to be about some bit flags.

sdroege:
> but I don't understand that "docs embed" check as it seems to be about some bit flags.

There's something broken with that check. Every now and then it reports errors like this for no reason.

PR #1470, opened 2024-08-05
Implement std::iter::Extend for glib::List

### LINKED ISSUE #1468: [FEATURE REQUEST] Implement `std::iter::Extend` for `glib::List`
A standard library has some useful functions which require `Extend`. E.g. `std::iter::unzip`, `std::iter::partition` and some other like `std::iter::collect_into` in nightly.

andy128k:
Closes #1468

### DIFF
--- a/glib/src/collections/list.rs
+++ b/glib/src/collections/list.rs
@@ -486,6 +486,25 @@ impl<T: TransparentPtrType> std::iter::IntoIterator for List<T> {
     }
 }
 
+impl<T: TransparentPtrType> std::iter::Extend<T> for List<T> {
+    #[inline]
+    fn extend<I: IntoIterator<Item = T>>(&mut self, iter: I) {
+        let list = iter.into_iter().collect::<Self>();
+        if list.is_empty() {
+            return;
+        }
+        match self.ptr.map(|p| p.as_ptr()) {
+            Some(ptr1) => {
+                let ptr2 = list.into_raw();
+                let _ = unsafe { ffi::g_list_concat(ptr1, ptr2) };
+            }
+            None => {
+                self.ptr = ptr::NonNull::new(list.into_raw());
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


glib/src/collections/list.rs
```diff
@@ -486,6 +486,25 @@ impl<T: TransparentPtrType> std::iter::IntoIterator for List<T> {
     }
 }
 
+impl<T: TransparentPtrType> std::iter::Extend<T> for List<T> {
```
sdroege, line 489: Could you add the same for `SList`? Basically same implementation

glib/src/collections/list.rs
```diff
@@ -486,6 +486,25 @@ impl<T: TransparentPtrType> std::iter::IntoIterator for List<T> {
...
+    #[inline]
+    fn extend<I: IntoIterator<Item = T>>(&mut self, iter: I) {
+        let list: Self = iter.into_iter().collect();
```
sdroege, line 492:
```suggestion
        let list = iter.into_iter().collect::<Self>();
```

glib/src/collections/slist.rs
```diff
@@ -480,6 +480,25 @@ impl<T: TransparentPtrType> std::iter::IntoIterator for SList<T> {
     }
 }
 
+impl<T: TransparentPtrType> std::iter::Extend<T> for SList<T> {
+    #[inline]
+    fn extend<I: IntoIterator<Item = T>>(&mut self, iter: I) {
+        let list: Self = iter.into_iter().collect();
```
sdroege, line 486:
```suggestion
        let list = iter.into_iter().collect::<Self>();
```

glib/src/collections/list.rs
andy128k: Done

glib/src/collections/slist.rs
andy128k: Done

sdroege (approved):

Issue #1236, opened 2023-11-29
[FEATURE REQUEST] Properties macro: Document properties better


sophie-h:
I'm using cargo docs online for people to work on Loupe and I'm sometimes using it myself.

With using the `Properties` macro the docs got a bit worse. Before, getters also had documentation. Now I have to lookup the field in the struct to see it's documentation.

Is it possible to either link to the field in getter/setter docs or do create an overview of all properties for the object or maybe both?

SeaDve:
I also have the same issue with exposing properties to a public object in a library. Specifically, I had to resort to manually implementing the properties in https://github.com/SeaDve/plotters-gtk4/blob/main/src/paintable.rs to improve the documentation.

I'm wondering if the field documentation on the `imp` struct can be forwarded to the generated setters and getters docs. For example,

```rust
#[derive(glib::Properties)]
#[properties(wrapper_type = Paintable)]
pub struct PaintableImp {
    /// The width of the paintable.
    #[property(get, set)]
    pub(super) width: OnceCell<u32>,
    /// The height of the paintable.
    #[property(get, set)]
    pub(super) height: OnceCell<u32>,
}
```

will generate the following getters:

```rust
impl Paintable {
    /// The width of the paintable.
    pub fn width(&self) -> u32 {
        ...
    }

    /// The height of the paintable.
    pub fn height(&self) -> u32 {
        ...
    }
}
```

Not sure though of a great API to specialize the documentation for the setters.

carlosmn:
Unfortunately I don't think you can make the properties get documented in the wrapper type itself as the `struct` block is implemented by a different macro, but it should be possible to copy the doc from the field into the methods. If we could have a single macro that knows all about the obj and imp all together, then that would be able to do it.

Given that docs are one attribute per line, I wonder if you could add a marker attribute to indicate you're switching to the setter and then the rest of the docs could be for the setter. This is probably terrible stylistically but it could probably work. Something like

```rust
#[derive(glib::Properties)]
#[properties(wrapper_type = MyType)]
pub struct MyTypeImp {
    /// The width of the object in my type.
    #[setter_doc]
    /// Set the width of the object. This has a million side effects
    #[property(get, set)]
    pub(super) width: RefCell<u32>,
}
```

that then becomes

```rust
impl MyType {
    /// The width of the paintable.
    pub width() -> u32 {
        ...
    }
    /// Set the width of the object. This has a million side effects
    pub set_width(v: u32) {
        ...
[...]

PR #1490, opened 2024-08-26
Copy docs to getter in Properties macro

carlosmn:
Spurred by #1236 though there's probably more we can do here.

The change here is that we take anything that's a doc before a `#[property]` and consider that to be the docs for the getter, as it seems like the getter and the property would be the most likely to share doc comments.

There is of course an open question of what one could do to get docs for the setter that aren't a pain to write, but at least for the setter we can make something that, I think, makes intuitive sense.

The big issue here is that you wouldn't get docs copied over if you add a doc to the field but after the `#[property]`. This is probably something we could allow, but I don't know how important it would be.

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -12,7 +12,7 @@ use syn::parse::Parse;
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
 use syn::Token;
-use syn::{parse_quote_spanned, LitStr};
+use syn::{parse_quote_spanned, Attribute, LitStr};
 
 pub struct PropsMacroInput {
     wrapper_ty: syn::Path,
@@ -255,6 +255,7 @@ struct PropDesc {
     field_ident: syn::Ident,
     ty: syn::Type,
     name: syn::LitStr,
+    comments: Vec<Attribute>,
     override_class: Option<syn::Type>,
     override_interface: Option<syn::Type>,
     nullable: bool,
@@ -271,6 +272,7 @@ impl PropDesc {
         attrs_span: proc_macro2::Span,
         field_ident: syn::Ident,
         field_ty: syn::Type,
+        comments: Vec<Attribute>,
         attrs: ReceivedAttrs,
     ) -> syn::Result<Self> {
         let ReceivedAttrs {
@@ -321,6 +323,7 @@ impl PropDesc {
             field_ident,
             ty,
             name,
+            comments,
             override_class,
             override_interface,
             nullable,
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
I am fine with the changes but we need to first come up with at least an approach on how to handle the setters/notify functions docs. Could we just auto-generate them based on `Set $this_property`  or something like that?

carlosmn:
By `Set $property name` do you mean if a line starts with that text, we'd consider the rest of the doc comments to be part of the setter? That is maybe a bit subtle but it does let you write a normal comment.

I wonder if a pseudo-attribute would work, something like

```rust
/// Here I am writing my comment for the getter
///
/// #[setter]
///
/// And now I'm writing docs for the setter.
```

Then again this works the other way around to how you write doc comments so maybe it's the other way around that matches how you would expect it to be in actual Rust. Or there's also adding more actual attributes for us to read, though I haven't checked how awkward it would be to parse this, something like

```rust
/// Comments for the getter
#[getter]
/// And for the setter
#[setter]
#[property(get, set)]
```

though it makes it more awkward to write.

I don't know how possible it is (and it would be really ugly, I think) but you could also put the comments where you define that you have a getter and setter

```rust
#[property(
/// Some comment for the getter
get,
/// And a comment for the setter
set]
```

but maybe that's just too ugly and awkward a syntax to ask people to write.

bilelmoussaoui:
Tbh, I don't have any ideas here that is why I didn't reply earlier. I just don't want people to refactor a bunch of code in case the to be added setter support makes the getter docs behave differently.

carlosmn:
Yeah, there's no obviously good answer unfortunately.

A variant I just though of is using `/// # Setter` as the context change (or allow `/// # Getter` too, whatever we end up with). This would a little bit like the `/// # Safety` block that clippy will warn you about for unsafe functions. There is https://github.com/gtk-rs/gtk-rs-core/pull/1490#issuecomment-2484952817 and then also https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_safety_doc for when it's unnecessary. That sets a bit of a precedent that the tooling can take markdown headers and treat the them as machine-readable.

Of course that's only a little bit like this as clippy isn't doing anything beyond seeing that the line exists, and it doesn't affect how the docs generate anything. So it's not the strongest precedent, but having

```rust
/// Comments for the getter
///
/// # Setter
///
/// And for the setter
#[property(get, set)]
```

seems like it flows when writing it and it's something you can parse and relate to other cases (though maybe not all that many people are actually writing unsafe functions).

bilelmoussaoui:
I would say using # Getter and # Setter would be good. As we also have # Notify for the generted notify_$property. And leave any comment before that relevant to the field as it might be some in code explanation

carlosmn:
Ok that seems reasonable, so we'd end up like

```rust
/// Some comments that remain internal
///
/// # Getter
///
/// Docs that land in the getter
///
/// # Setter
///
/// And docs for the setter
#[property(get, set)]
```

carlosmn:
Ok well I finally found time/motivation to look back into this and just ended up taking a few hours.

This now implements what we discussed as a reasonable compromise which includes an example so you can check it while messing around with the code.

If this does seem like we want then these comment copy rules and example should end up in `lib.rs` too. I might tackle that that in the next few days.

glib-macros/src/properties.rs
```diff
@@ -559,6 +569,50 @@ fn strip_raw_prefix_from_name(name: &LitStr) -> LitStr {
...
+    for attr in comments {
+        if let syn::Meta::NameValue(meta) = &attr.meta {
+            if let syn::Expr::Lit(expr) = &meta.value {
+                if let syn::Lit::Str(lit_str) = &expr.lit {
+                    // Now that we have the one line of comment, see if we need
+                    // to switch a particular section to be the active one (via
+                    // the header syntax) or add the current line to the active
+                    // section.
+                    match lit_str.value().trim() {
+                        "# Getter" => {
```
bilelmoussaoui, line 593: Would add Notify support as well while at it

bilelmoussaoui (approved):

bilelmoussaoui:
Otherwise this looks good to me!

glib-macros/src/properties.rs
carlosmn: That makes sense yeah. Though I think it'll have to be in a different function because we call this one from within the loop inside `expand_impl_getset_properties` and the notify code is generated from a different one.

carlosmn: I added `# Notify` to copy the docs into `notify_$property` but now I'm not sure if that's what you mean or if it was supposed to be the `connect_$property_notify` as this one is maybe more likely to be something that a user of a class would need documentation about when that triggers or what the data would mean there.

bilelmoussaoui: I would auto-generate the connect_property_notify docs based on the notify one but not sure

carlosmn: How often does one want to document the `notify_$property` method itself? Wouldn't that always just say to notify the property? I suppose you could mention when to call it as a user of the class, but I don't know how common that is for public types.

But the `connect_$property_notify` method I think would benefit much more commonly from docs, no? You might want to specify non-obvious times it would get called, maybe values that depend on other values?

I don't have experience with maintaining a public package, but looking at the docs for gtk itself, e.g. https://docs.gtk.org/gdk4/class.Surface.html (not that we have to match C but consistency is nice) the docs for the signals seem like the `connect_` side of the signals rather than the emitting side. Maybe it should be a `# Signal` heading instead? That would also make it match tighter if we end up making it so there's some sort of list of signals in the docs.

carlosmn: It looks like I was conflating explicitly-defined signals and property change notifications, as the latter is not spelled out in the docs, so now I'm not sure.

carlosmn: OK, I've removed the handling for notify and added generic comments that just mention the notification name. We can always add some more "instructions" to the comment if this is something people want to do.

bilelmoussaoui:
thanks!

PR #1491, opened 2024-08-27
glib: Ensure NULL terminator is present after creating allocation in `{StrV,PtrSlice}::reserve`

kawadakk:
After creating a new memory allocation, [`StrV::reserve`](https://gtk-rs.org/gtk-rs-core/stable/0.20/docs/glib/collections/strv/struct.StrV.html#method.reserve) didn't properly null-terminate the underlying array. This could cause an undefined behavior in `g_strfreev` if the `StrV` was immediately dropped.


sdroege (dismissed):
Thanks! Same bug in `PtrSlice::reserve()` AFAICS. Can you add a commit fixing it there too?

kawadakk:
Done, though technically `PtrSlice` should be fine since `drop()` uses its own `len` to clean up elements (instead of iterating up to null).

sdroege (approved):

sdroege:
ありがとうございます！

Issue #1485, opened 2024-08-25
[BUG] FromGlibBorrow etc implemented on the wrong types for interfaces


pbor:
**Bug description**

See https://github.com/sdroege/gobject-example-rs/pull/22 for the context.

in the `ffi` module it uses:

```
pub type ExNameable = std::ffi::c_void;
```

However it would be preferrable to use 

```
  #[repr(C)]
  pub struct ExNameable(std::ffi::c_void);
```

But that fails because `FromGlibBorrow` etc are not implemented

From the chat in the channel:

> slomo@slomo:matrix.org: i think there's a bug in the macro there, it's implemented the traits on the wrong types it seems. maybe Fina has an idea. otherwise please just create an issue with code to reproduce that and i'll take a look one of these days

sdroege:
This would implement it. It's not a breaking change (:partying_face:) but before making a PR for this I'd have to add some unsafe trait for the instance struct (see last part of the diff).

Any opinions here?

CC @felinira 

```diff
diff --git a/examples/virtual_methods/purrable.rs b/examples/virtual_methods/purrable.rs
index 170ad10d10e..d900b480a01 100644
--- a/examples/virtual_methods/purrable.rs
+++ b/examples/virtual_methods/purrable.rs
@@ -43,6 +43,7 @@ mod iface {
     impl ObjectInterface for Purrable {
         const NAME: &'static str = "Purrable";
 
+        type Instance = super::ffi::Instance;
         type Interface = super::ffi::Interface;
 
         /// Initialize the class struct with the default implementations of the
diff --git a/glib-macros/src/object_impl_attributes/interface.rs b/glib-macros/src/object_impl_attributes/interface.rs
index c9e79bb5cd9..1f5b2f34c63 100644
--- a/glib-macros/src/object_impl_attributes/interface.rs
+++ b/glib-macros/src/object_impl_attributes/interface.rs
@@ -32,11 +32,14 @@ pub fn impl_object_interface(input: super::Input) -> TokenStream {
     };
 
     let mut has_prerequisites = false;
+    let mut has_instance = false;
     for item in items.iter() {
         if let syn::ImplItem::Type(type_) = item {
             let name = type_.ident.to_string();
             if name == "Prerequisites" {
                 has_prerequisites = true;
+            } else if name == "Instance" {
+                has_instance = true;
             }
         }
     }
[...]

PR #1495, opened 2024-08-27
glib: Add `ObjectInterface::Instance` for symmetry with `ObjectSubcla…

sdroege:
…ss::Instance`

This defaults to `std::ffi::c_void` but can be used to provide a newtype wrapper around it instead for additional type-safety in FFI code.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1485

----

CC @felinira

### DIFF
--- a/glib/src/subclass/interface.rs
+++ b/glib/src/subclass/interface.rs
@@ -212,6 +212,8 @@ unsafe extern "C" fn interface_init<T: ObjectInterface>(
 ///
 /// [`object_interface!`]: ../../macro.object_interface.html
 pub fn register_interface<T: ObjectInterface>() -> Type {
+    assert_eq!(mem::size_of::<T>(), 0);
+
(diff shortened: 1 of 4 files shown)


glib/src/subclass/interface.rs
```diff
@@ -109,6 +109,13 @@ pub trait ObjectInterface: ObjectInterfaceType + Sized + 'static {
     /// in case of interfaces.
     type Prerequisites: PrerequisiteList;
 
+    // rustdoc-stripper-ignore-next
+    /// The C instance struct. This is usually either `std::ffi::c_void` or a newtype wrapper
+    /// around it.
+    ///
+    /// Optional
+    type Instance;
```
sdroege, line 117: I was checking and I can't find any way how one could cause problems here by providing some random other type, so I didn't add a new unsafe trait

PR #1497, opened 2024-08-30
glib: Do not use `end_unsafe` in `<Variant as From<VariantDict>>::from`

kawadakk:
While this method consumes a given `VariantDict`, the underlying object could still be accessed through other clones because of the reference counted clone semantics, which means it's unsound to use [`VariantDict::end_unsafe`](https://gtk-rs.org/gtk-rs-core/stable/0.20/docs/glib/struct.VariantDict.html#method.end_unsafe) here.


sdroege (approved):
Thanks!

Issue #1504, opened 2024-09-07
Regression in FilenameCollateKey change


ebassi:
Turns out that #1329 breaks code that does:

```rust
// f is a gio::File and set elsewhere
let basename = f.basename().unwrap();
let key = glib::FilenameCollationKey::from(basename.to_string_lossy());
```

which now generates a debugging assertion:

```
assertion failed: cstr.to_str().is_ok()
```

and a backtrace like this one taken from Amberol:

```
   0:     0x558f2b400cf5 - std::backtrace_rs::backtrace::libunwind::trace::h58eed11393533053
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:     0x558f2b400cf5 - std::backtrace_rs::backtrace::trace_unsynchronized::h6af9bae28ebb6388
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x558f2b400cf5 - std::sys_common::backtrace::_print_fmt::hb6748916642a4fb2
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x558f2b400cf5 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h3692694645b1bb6a
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x558f2b4283eb - core::fmt::rt::Argument::fmt::h7aa93977ba74ae0f
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/fmt/rt.rs:165:63
   5:     0x558f2b4283eb - core::fmt::write::h5131d80b4c69b88d
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/core/src/fmt/mod.rs:1168:21
   6:     0x558f2b3fd97f - std::io::Write::write_fmt::h1fb327a7d8b0eb36
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/io/mod.rs:1835:15
   7:     0x558f2b400ace - std::sys_common::backtrace::_print::he6ebb7b9d89f4456
[...]

PR #1505, opened 2024-09-09
glib: Collation keys are not necessary valid UTF-8

sdroege:
All that is guaranteed about them is that they give the correct order for the original strings when passed to strcmp().

Also add a separate integration test for this that calls setlocale() to get the default system locale. This makes sure that not just the "C" locale is selected, which returns literally the same string as collation key.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1504


sdroege:
The strings used in the tests should sort the same in any possible locale, I hope. If not, we'll find out!

PR #1507, opened 2024-09-14
Fix cargo instructions in build.rs function

swsnr:
I'm not sure if it makes a difference, but according to [Cargo's documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) cargo instructions in build scripts must be prefixed by `cargo::`, i.e. two colons, not one.


swsnr:
Ah never mind, I wasn't aware that this had changed in Rust.  Sorry for the noise :innocent:

Issue #1515, opened 2024-09-23
[BUG] Impl traits don't have enough trait bounds


felinira:
**Bug description**

`XyzImpl` traits don't actually check whether they are implemented on classes that derive / implement the specific class. The Impl traits themselves only provide implementations, so usually they would be fine to implement wherever. However, almost all bindings also offer to call parent implementations, or other ffi methods as part of some Ext traits.

In practise this means that you can call unrelated methods on any object, thus being able to cause UB.

Examples:

- Implement `gtk::WidgetImpl` on any gobject subclass and get access to things like `set_css_name` without the object actually having to be derived from GtkWidget. This causes a runtime critical with UB afterwards.
- Implement `gio::InitableImpl` on a gobject subclass that doesn't implement gio::Initable and call `parent_init`. This panics here: https://github.com/gtk-rs/gtk-rs-core/blob/b97554d03ae9ae8f1c56e0e61f55d2612e1be271/glib/src/subclass/types.rs#L496

There are probably more ways to make this lead to more serious and harder to detect behaviour.

**Solution**

Adding trait bounds like this should probably resolve all these issues:

> `pub trait WidgetImpl: ObjectSubclass where <Self as ObjectSubclass>::Type: IsA<Widget>`

sdroege:
As discussed on Matrix yesterday, adding such trait bounds will require them to be added recursively in more places. E.g. `ButtonImpl` will need both `IsA<Widget>` and `IsA<Button>`.

Also it means that the `glib::wrapper!` usage must list all super-classes and can't just skip one. That part will probably break a bit of code.

This seems like the right thing to do though.

PR #1519, opened 2024-09-24
Add additional type bounds to Impl traits

felinira:
See #1515

### DIFF
--- a/gdk-pixbuf/src/subclass/pixbuf_animation.rs
+++ b/gdk-pixbuf/src/subclass/pixbuf_animation.rs
@@ -13,7 +13,10 @@ use glib::{prelude::*, subclass::prelude::*, translate::*};
 
 use crate::{ffi, Pixbuf, PixbufAnimation, PixbufAnimationIter};
 
-pub trait PixbufAnimationImpl: ObjectImpl {
+pub trait PixbufAnimationImpl: ObjectImpl
+where
+    <Self as ObjectSubclass>::Type: IsA<PixbufAnimation>,
+{
     fn is_static_image(&self) -> bool {
         self.parent_is_static_image()
     }
@@ -31,12 +34,10 @@ pub trait PixbufAnimationImpl: ObjectImpl {
     }
 }
 
-mod sealed {
-    pub trait Sealed {}
-    impl<T: super::PixbufAnimationImplExt> Sealed for T {}
-}
-
-pub trait PixbufAnimationImplExt: sealed::Sealed + ObjectSubclass {
+pub trait PixbufAnimationImplExt: ObjectSubclass + PixbufAnimationImpl
+where
+    <Self as ObjectSubclass>::Type: IsA<PixbufAnimation>,
+{
     fn parent_is_static_image(&self) -> bool {
         unsafe {
             let data = Self::type_data();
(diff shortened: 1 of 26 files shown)


gio/src/subclass/action_group.rs
```diff
@@ -74,11 +77,18 @@ pub trait ActionGroupImpl: ObjectImpl {
...
+        <T as ObjectSubclass>::Type: IsA<super::ActionGroup>
+    {
+    }
 }
 
-pub trait ActionGroupImplExt: sealed::Sealed + ObjectSubclass {
+pub trait ActionGroupImplExt: sealed::Sealed + ObjectSubclass
+where
+    <Self as ObjectSubclass>::Type: IsA<ActionGroup>,
+{
```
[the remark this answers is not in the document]
sdroege: Not required if the compiler allows you to omit them

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -32,11 +35,18 @@ pub trait PixbufAnimationImpl: ObjectImpl {
...
     pub trait Sealed {}
-    impl<T: super::PixbufAnimationImplExt> Sealed for T {}
+    impl<T: super::PixbufAnimationImplExt> Sealed for T where
+        <T as ObjectSubclass>::Type: IsA<super::PixbufAnimation>
+    {
+    }
 }
 
-pub trait PixbufAnimationImplExt: sealed::Sealed + ObjectSubclass {
+pub trait PixbufAnimationImplExt: sealed::Sealed + ObjectSubclass
```
sdroege, line 46: I wonder if the sealing is still necessary like this

gdk-pixbuf/src/subclass/pixbuf_animation.rs
```diff
@@ -132,7 +148,10 @@ unsafe impl<T: PixbufAnimationImpl> IsSubclassable<T> for PixbufAnimation {
 
 unsafe extern "C" fn animation_is_static_image<T: PixbufAnimationImpl>(
     ptr: *mut ffi::GdkPixbufAnimation,
-) -> glib::ffi::gboolean {
+) -> glib::ffi::gboolean
+where
+    <T as ObjectSubclass>::Type: IsA<PixbufAnimation>,
```
sdroege, line 153: Why is this not implied by `T: PixbufAnimationImpl`? Is that a compiler bug?

gdk-pixbuf/src/subclass/pixbuf_animation.rs
sdroege: Also now that I think about it, is `pub trait PixbufAnimationExt: IsA<PixbufAnimation>` required to be sealed? It's implemented over all `IsA<PixbufAnimation>` so there can't really be any other implementation, can there?

Removing the sealing would probably reduce compile times because lot fewer traits?

sdroege:
```suggestion
pub trait PixbufAnimationImplExt: sealed::Sealed + ObjectSubclass + PixbufAnimationImpl
```

too maybe?

gio/src/subclass/action_group.rs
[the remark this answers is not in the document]
sdroege: I mean, this should be `pub trait ActionGroupImplExt: ActionGroupImpl` and then via `ActionGroupImpl` requiring `IsA<ActionGroup` it should be all covered

sdroege (commented):
Basically the same comments to everything :)

sdroege:
> I removed all sealed traits except for the one for ObjectImplExt, which I am unsure about. But I don't see much harm in that one.

What's different with that one? Seems like the others to me and should be fine with just the `: ObjectImpl + ObjectSubclass` trait bound

sdroege (dismissed):

sdroege:
Some failing tests :)

sdroege (approved):

PR #1521, opened 2024-09-28
glib-macros: add test macro for async-glib-context tests

BiagioFesta:
Added a procedural macro for async tests, wrapping them in `glib::MainContext`. 
Not sure if it's useful for others, but I found myself needing this occasionally. Feel free to reject if it's not a good fit.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -1,5 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+mod async_test;
 mod boxed_derive;
 mod clone;
 mod closure;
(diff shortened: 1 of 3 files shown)


glib/tests/log.rs
```diff
@@ -1,6 +1,10 @@
 use std::sync::{Arc, Mutex};
 
-use glib::*;
+use glib::{
+    clone, g_critical, g_debug, g_info, g_message, g_warning, log_remove_handler,
+    log_set_default_handler, log_set_fatal_mask, log_set_handler, log_unset_default_handler,
+    LogLevel, LogLevels,
+};
```
sdroege, line 7: Why this change?

glib-macros/src/test.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let gen_attr = quote::quote! {
+        #[::core::prelude::v1::test]
+    };
+
+    let body = &item_fn.block;
+
+    item_fn.block = syn::parse2(quote::quote! {
+        {
+            let main_ctx = glib::MainContext::default();
```
sdroege, line 36: You probably want each test to run in their own main context. Otherwise running multiple tests in parallel will easily cause chaos.

glib-macros/src/lib.rs
```diff
@@ -1567,3 +1568,22 @@ pub fn derive_value_delegate(input: TokenStream) -> TokenStream {
...
+/// An attribute macro for writing asynchronous test functions.
+///
+/// This macro is designed to wrap an asynchronous test function and ensure that
+/// it runs within a `glib::MainContext`. It helps in writing async tests that
+/// require the use of an event loop for the asynchronous execution.
+///
+/// # Example
+///
+/// ```
+/// #[glib::test]
```
sdroege, line 1581:
```suggestion
/// #[glib::async_test]
```

seems more descriptive

glib/tests/log.rs
BiagioFesta: being the macro name `test` importing `glib::*` was overriding """std""" test.

BiagioFesta: anyway, renaming the macro `test -> async_test` solves this

bilelmoussaoui (approved):

sdroege (approved):

Issue #1526, opened 2024-09-30
[BUG] gtlib-macros: testsuote started failing with 0.20.3


werdahias:
**Bug description**
All autopkgtests in Debian for glib-macros started to fail with 0.20.3:

```
722s error[E0437]: type `Instance` is not a member of trait `ObjectInterface`
722s    --> src/subclass/object.rs:507:9
722s     |
722s 507 |         #[glib::object_interface]
722s     |         ^^^^^^^^^^^^^^^^^^^^^^^^^ not a member of trait `ObjectInterface`
722s     |
722s     = note: this error originates in the attribute macro `glib::object_interface` (in Nightly builds, run with -Z macro-backtrace for more info)
722s 

```
Today I updated glib-macros to 0.20.4 and got a different error this time:

```
error[E0716]: temporary value dropped while borrowed
  --> tests/properties.rs:96:43
   |
85 |         #[derive(Properties, Default)]
   |                  ----------
   |                  |        |
   |                  |        temporary value is freed at the end of this statement
   |                  argument requires that borrow lasts for `'static`
...
96 |             #[property(get, set, builder(&<Option<i32>>::static_variant_type()))]
   |                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use

For more information about this error, try `rustc --explain E0716`.
error: could not compile `glib-macros` (test "properties") due to 1 previous error

```
This happens for all features. Let me know if you need more information

sdroege:
You need to update `glib` to 0.20.3 or newer too. The dependency update in Cargo.toml for that is missing. Can you send a PR for the 0.20 branch for that?

sdroege:
https://github.com/gtk-rs/gtk-rs-core/pull/1527

werdahias:
Thanks, will cherrypick this commit

werdahias:
Forcing glib-macros to depend on glib 0.20.4 and doing the same (inverted) for glib fixed this for me.

PR #1527, opened 2024-10-01
glib: Add glib-macros / glib 0.20.3 dependency

sdroege:
Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1526

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ bitflags = "2.6"
 gir-format-check = "^0.1"
 glib-sys = { path = "glib/sys", version = "0.20" }
 gobject-sys = { path = "glib/gobject-sys", version = "0.20" }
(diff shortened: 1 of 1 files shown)


sdroege:
Unfortunately this can't be added to just the dependencies in glib/glib-macros: `gtk-rs-core/glib/Cargo.toml: unused manifest key: dependencies.glib-macros.version`

bilelmoussaoui:
> Unfortunately this can't be added to just the dependencies in glib/glib-macros: `gtk-rs-core/glib/Cargo.toml: unused manifest key: dependencies.glib-macros.version`

Well, the goal is to never set versions there anyways otherwise the usage of workspace dependencies would be useless as cargo tells in the error message.

bilelmoussaoui (approved):

PR #1529, opened 2024-10-05
Use rust 1.79 implied supertraits to simplify trait bounds

felinira:
This removes a lot of the boilerplate introduced by #1519 by using supertraits instead of where clauses. Supertraits are allowed to imply their trait bounds since rust 1.79.

This bumps MSRV to 1.80.

IMO it's easier to review this against d8dfa3a26184e7c6010744233fe5cc516310438e than master, as it reverts most of what that's doing.

### DIFF
--- a/glib-macros/tests/object_subclass_dynamic.rs
+++ b/glib-macros/tests/object_subclass_dynamic.rs
@@ -28,10 +28,8 @@ mod static_ {
             type Interface = MyStaticInterfaceClass;
         }
 
-        pub trait MyStaticInterfaceImpl: ObjectImpl + ObjectSubclass
-        where
-            <Self as ObjectSubclass>::Type: IsA<glib::Object>,
-            <Self as ObjectSubclass>::Type: IsA<super::MyStaticInterface>,
+        pub trait MyStaticInterfaceImpl:
+            ObjectImpl + ObjectSubclass<Type: IsA<super::MyStaticInterface>>
         {
         }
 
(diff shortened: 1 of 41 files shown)


sdroege:
> IMO it's easier to review this against [d8dfa3a](https://github.com/gtk-rs/gtk-rs-core/commit/d8dfa3a26184e7c6010744233fe5cc516310438e) than master, as it reverts most of what that's doing.

https://github.com/felinira/gtk-rs-core/compare/d8dfa3a26184e7c6010744233fe5cc516310438e...wip/implied-supertraits

glib-macros/tests/object_subclass_dynamic.rs
```diff
@@ -374,10 +347,8 @@ pub mod plugin {
             type Interface = MyPluginInterfaceClass;
         }
 
-        pub trait MyPluginInterfaceImpl: ObjectImpl + ObjectSubclass
-        where
-            <Self as ObjectSubclass>::Type: IsA<glib::Object>,
-            <Self as ObjectSubclass>::Type: IsA<super::MyPluginInterface>,
+        pub trait MyPluginInterfaceImpl:
```
sdroege, line 350: Should do the same for all the other extension traits I guess. You look into this afterwards?

sdroege (approved):

glib-macros/tests/object_subclass_dynamic.rs
sdroege: I just added the comment at a random place. I mean for example `pub trait FileExt: IsA<File> + sealed::Sealed + 'static` could lose the sealing AFAICS, what we talked about on matrix 2 weeks or so ago

sdroege: That kind of change should be easy. But I can also take a look at it one of these days :)

bilelmoussaoui: if there is something that needs to be done on gir side, just let me know what is the change you want to have. I am a bit familiar with the codebase already, no need for extra-suffering ;)

sdroege: @bilelmoussaoui Instead of having the sealing it is sufficient for the extension traits to require `IsA<TheType>`

sdroege: And when regenerating, the sealing also should be dropped from the manual extension traits like that. I hope that slightly improves compile times too :)

PR #1539, opened 2024-10-13
Pass GLIB_DOMAIN as nul-terminated string

### LINKED ISSUE #1538: [BUG] Structured logging must pass GLIB_DOMAIN as nul-terminated string
**Bug description**

Currently, `log_structured!` passes `GLIB_DOMAIN` as a log field with byte value and explicit length.  However, GLib expects `GLIB_DOMAIN` to be a nul-terminated C string, see [`should_drop_message`](https://gitlab.gnome.org/GNOME/glib/-/blob/a8dbd7cad5636e5064ca6477dc6d7227675384a1/glib/gmessages.c?page=3#L2728) which calls [`domain_found`](https://gitlab.gnome.org/GNOME/glib/-/blob/a8dbd7cad5636e5064ca6477dc6d7227675384a1/glib/gmessages.c?page=3#L2637) which in turn uses `strlen` on the domain value.

This breaks debug logging for Rust domains via `$G_MESSAGES_DEBUG`, because GLib fails to find the domain in the environment variable, see https://github.com/gtk-rs/gtk-rs-core/discussions/1534

I'm only superficially familiar with C, and haven't checked the entire call stack, but it also seems that passing `GLIB_DOMAIN` as field with explicit length, i.e. omitting the final NUL byte from the field value, probably even leads to out-of-bounds memory access in GLib :thinking: 

I'm not sure whether this constitutes an issue in GLib, but it can be fixed on the Rust side by passing the domain as nul-terminated string, at the cost of one allocation.  I'll make a pull request.

swsnr:
Fixes #1538

### DIFF
--- a/glib/src/log.rs
+++ b/glib/src/log.rs
@@ -365,6 +365,9 @@ pub struct LogField<'a>(ffi::GLogField, std::marker::PhantomData<&'a GStr>);
 impl<'a> LogField<'a> {
     // rustdoc-stripper-ignore-next
     /// Creates a field from a borrowed key and value.
+    ///
+    /// Adds `value` as byte field with explicit length.  See [`new_str`] for
+    /// means to add field values as nul-terminated strings.
     pub fn new(key: &'a GStr, value: &[u8]) -> Self {
         let (value, length) = if value.is_empty() {
             // Use an empty C string to represent empty data, since length: 0 is reserved for user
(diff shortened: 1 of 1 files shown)


swsnr:
Tested with the following example:

```rust
const G_LOG_DOMAIN: &str = "LOGTEST";

fn main() {
    static GLIB_LOGGER: glib::GlibLogger = glib::GlibLogger::new(
        glib::GlibLoggerFormat::Structured,
        glib::GlibLoggerDomain::CrateTarget,
    );
    rs_log::set_logger(&GLIB_LOGGER).unwrap();
    rs_log::set_max_level(rs_log::LevelFilter::Debug);

    glib::debug!("DEBUG");
    glib::info!("INFO");
    glib::warn!("WARN");
}
```

With this PR `G_MESSAGES_DEBUG` shows all messages:

```console
$ G_MESSAGES_DEBUG=LOGTEST cargo run --features log,log_macros --package glib --example logtest
   Compiling glib v0.21.0 (/home/basti/Code/gtk-rs/gtk-rs-core/glib)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.40s
     Running `target/debug/examples/logtest`
(process:66204): LOGTEST-DEBUG: 09:36:13.777: DEBUG
LOGTEST-INFO: 09:36:13.777: INFO

(process:66204): LOGTEST-WARNING **: 09:36:13.777: WARN
```

Previously, it'd only show the warning message.

If it's any help I can also include the example in the pull request.

sdroege:
As mentioned in https://github.com/gtk-rs/gtk-rs-core/issues/1538#issuecomment-2410405552 this is not the way to go. It would incur an unnecessary performance penalty and shouldn't be necessary

PR #1543, opened 2024-10-18
Regenerate with dropped sealed auto traits

bilelmoussaoui:


sdroege:
You'll want to do the same for the manual extension traits

bilelmoussaoui:
The CI will continue to fail until we update gir-files, so merging this already

PR #1544, opened 2024-10-18
Regenerate with latest gir-files

bilelmoussaoui:


gio/src/auto/app_launch_context.rs
```diff
@@ -59,11 +59,15 @@ pub trait AppLaunchContextExt: IsA<AppLaunchContext> + 'static {
 
     #[doc(alias = "g_app_launch_context_get_startup_notify_id")]
     #[doc(alias = "get_startup_notify_id")]
-    fn startup_notify_id(&self, info: &impl IsA<AppInfo>, files: &[File]) -> Option<glib::GString> {
+    fn startup_notify_id(
+        &self,
+        info: Option<&impl IsA<AppInfo>>,
```
sdroege, line 64: Breaking change but as we discussed we'd backport these now because of correctness

bilelmoussaoui: Yep, it would be nice to backport it actually as it blocks some work we have in ashpd :)

gio/sys/src/lib.rs
```diff
@@ -17313,7 +17321,21 @@ extern "C" {
     pub fn g_unix_mount_is_system_internal(mount_entry: *mut GUnixMountEntry) -> gboolean;
     pub fn g_unix_mount_points_changed_since(time: u64) -> gboolean;
     pub fn g_unix_mount_points_get(time_read: *mut u64) -> *mut glib::GList;
+    #[cfg(feature = "v2_82")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v2_82")))]
+    pub fn g_unix_mount_points_get_from_file(
```
sdroege, line 17326: This needs to be conditional for non-Windows

gio/sys/src/lib.rs
```diff
@@ -17313,7 +17321,21 @@ extern "C" {
...
+        table_path: *const c_char,
+        time_read_out: *mut u64,
+        n_points_out: *mut size_t,
+    ) -> *mut *mut GUnixMountPoint;
     pub fn g_unix_mounts_changed_since(time: u64) -> gboolean;
     pub fn g_unix_mounts_get(time_read: *mut u64) -> *mut glib::GList;
+    #[cfg(feature = "v2_82")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v2_82")))]
+    pub fn g_unix_mounts_get_from_file(
```
sdroege, line 17335: This needs to be conditional for non-Windows

glib/src/auto/main_context.rs
```diff
@@ -85,6 +85,13 @@ impl MainContext {
         unsafe { from_glib(ffi::g_main_context_pending(self.to_glib_none().0)) }
     }
 
+    //#[cfg(feature = "v2_64")]
+    //#[cfg_attr(docsrs, doc(cfg(feature = "v2_64")))]
+    //#[doc(alias = "g_main_context_pusher_new")]
+    //pub fn pusher_new(&self) -> /*Unknown conversion*//*Unimplemented*/MainContextPusher {
```
sdroege, line 91: Funny, that's basically our `MainContextAcquireGuard`

gio/sys/src/lib.rs
sdroege: Hm we don't do that for any others so it probably doesn't matter after all

gio/src/auto/app_launch_context.rs
sdroege: Will do

gio/sys/src/lib.rs
bilelmoussaoui: Shouldn't those be in the gio-unix gi file nowadays anyway?

sdroege: I think that was not done for backwards compatibility reasons

PR #1558, opened 2024-10-28
Add some quality-of-life gdbus helpers

swsnr:
Add a helper to return a standard result type from an invocation, automatically wrapping the return value into a tuple if it isn't a tuple yet.

Add another helper to return an async result from a future.

Update the gdbus example to demo the new async helper, and show how to simplify parameter handling with auto-derived variant structs.


examples/gio_dbus_register_object/main.rs
```diff
@@ -22,25 +58,24 @@ fn on_startup(app: &gio::Application, tx: &Sender<gio::RegistrationId>) {
...
-                match method {
-                    "Hello" => {
-                        if let Some((name,)) = <(String,)>::from_variant(&params) {
-                            let greet = format!("Hello {name}!");
-                            println!("{greet}");
-                            invocation.return_value(Some(&(greet,).to_variant()));
-                        } else {
-                            invocation.return_error(gio::IOErrorEnum::Failed, "Invalid parameters");
-                        }
+        .method_call(move |_, _, _, _, method, params, invocation| {
```
bilelmoussaoui, line 61: I think this could be improved quite a bit by introducing a new trait that would implement the parsing and you could just call method_call::<T> and it would pass you the parsed method already

swsnr: Sounds like a good idea. I'll try to add this 🙏

swsnr: @bilelmoussaoui I've pushed a new commit which provides an API around handling "parsed" calls.

The builder interface became a bit more intricate to "hide" the closure type parameter, but I think that's an acceptable price for the added convenience :innocent: 

I've also added overloads which hide the `invocation` object and instead return a result from the closure, which is a lot safer, as you can't forget to call `invocation.return_*` anymore :slightly_smiling_face:

examples/gio_dbus_register_object/main.rs
```diff
@@ -8,10 +11,48 @@ const EXAMPLE_XML: &str = r#"
...
+    delay: u32,
+}
+
+#[derive(Debug)]
+enum HelloMethod {
+    Hello(Hello),
+    SlowHello(SlowHello),
+}
+
+impl DBusMethodCall for HelloMethod {
```
swsnr, line 40: In theory, this trait could get a derive macro, to automatically create the `match` from the names of the enum variants.

But I must admit that I'm not at all familiar with derive macros in Rust; I've never written one, and I'm not sure I'm going to find the time to learn that in near future.

So you folks consider this pull request a valuable API already, I'd appreciate if it was merged without a derive macro, as I'd like to leave that for a much later time, or perhaps even someone else :innocent: 

Hence,

examples/gio_dbus_register_object/main.rs
sdroege: That looks like an improvement indeed. A derive macro for this would also be relatively simple to write.
I'm not sure about the method name (`parse_method()` at that place seems a bit out of place).

examples/gio_dbus_register_object/main.rs
```diff
@@ -8,10 +11,48 @@ const EXAMPLE_XML: &str = r#"
...
+    fn parse_call(
+        _obj_path: &str,
+        _interface: &str,
+        method: &str,
+        params: glib::Variant,
+    ) -> Result<Self, glib::Error> {
+        match method {
+            "Hello" => Ok(params.get::<Hello>().map(HelloMethod::Hello)),
+            "SlowHello" => Ok(params.get::<SlowHello>().map(HelloMethod::SlowHello)),
```
bilelmoussaoui, line 49:
```suggestion
            "Hello" => Ok(params.get::<Hello>().map(Self::Hello)),
            "SlowHello" => Ok(params.get::<SlowHello>().map(Self::SlowHello)),
```

gio/src/lib.rs
```diff
@@ -28,8 +28,8 @@ mod dbus;
 pub use self::dbus::*;
 mod dbus_connection;
 pub use self::dbus_connection::{
-    ActionGroupExportId, FilterId, MenuModelExportId, RegistrationId, SignalSubscriptionId,
-    WatcherId,
+    ActionGroupExportId, DBusMethodCall, FilterId, MenuModelExportId, RegistrationId,
```
bilelmoussaoui, line 31: DBusMethodCall is a trait, it should be re-exported in prelude instead.

examples/gio_dbus_register_object/main.rs
swsnr: Yeah, the name 😬 Just couldn't come up with any better 🤔

Would `build_method_call` be a better choice?

swsnr: As for the derive macro, I'm not sure I'd be able to write this any time soon.  I've never done this before, see my other comment :)

swsnr:
Accidentally committed some left over change, I'll fix this tomorrow.

swsnr:
@sdroege Pipeline passes, but the question about the name still stands... should I rename the method?

examples/gio_dbus_register_object/main.rs
```diff
@@ -8,10 +11,48 @@ const EXAMPLE_XML: &str = r#"
...
+}
+
+#[derive(Debug)]
+enum HelloMethod {
+    Hello(Hello),
+    SlowHello(SlowHello),
+}
+
+impl DBusMethodCall for HelloMethod {
+    fn parse_call(
```
bilelmoussaoui, line 41:
```suggestion
    fn typed_method_call(
```
as a suggestion for a name?

swsnr: I've changed the method name, but I'm going to leave this thread unresolved, in case e.g. @sdroege has another suggestion.

Other than that this PR is ready as far as I'm concerned.

sdroege: I like this name more :)

sdroege: `typed_method_call()` as compared to the old `method_call()` but with proper types instead of strings and freeform variants

sdroege: But to be clear: for the trait method `parse_call()` sounds better to me

sdroege (approved):
@bilelmoussaoui ?

bilelmoussaoui (approved):

swsnr:
Thank you for reviewing and merging 🙏

examples/gio_dbus_register_object/main.rs
```diff
@@ -22,25 +63,25 @@ fn on_startup(app: &gio::Application, tx: &Sender<gio::RegistrationId>) {
...
+                    HelloMethod::Hello(Hello { name }) => {
+                        let greet = format!("Hello {name}!");
+                        println!("{greet}");
+                        Ok(Some(greet.to_variant()))
+                    }
+                    HelloMethod::SlowHello(SlowHello { name, delay }) => {
+                        glib::timeout_future(Duration::from_secs(delay as u64)).await;
+                        let greet = format!("Hello {name} after {delay} seconds!");
+                        println!("{greet}");
+                        Ok(Some(greet.to_variant()))
```
sdroege, line 80: We should probably also figure out a way to make the return value more type safe

swsnr: Type safety as in "the return value matches the signature of the Dbus interface"?

sdroege: Yes. Maybe the structs in the enum need to implement a trait with an associated type for that, or so. I think that would all require a bit of reorganization though

Issue #1554, opened 2024-10-26
[BUG] glib-macros started regressing with rustc 1.82.0 on 32-bit arches


werdahias:
**Bug description**
In Debian we enabled rustc to trigger the tests of crates which directly depend on it. glib-macros passed before but has regressed now on all 32-bit arches.

**Backtrace**

```
 running 12 tests
379s test derive_enum ... ok
379s test derive_boxed_nullable ... ok
379s test derive_boxed ... ok
379s test attr_flags_with_default ... ok
379s test attr_flags ... ok
379s thread 'boxed_transparent_ptr' panicked at core/src/panicking.rs:221:5:
379s unsafe precondition(s) violated: Layout::from_size_align_unchecked requires that align is a power of 2 and the rounded-up allocation size does not exceed isize::MAX
379s stack backtrace:
379s    0: rust_begin_unwind
379s              at /usr/src/rustc-1.82.0/library/std/src/panicking.rs:662:5
379s    1: core::panicking::panic_nounwind_fmt::runtime
379s              at /usr/src/rustc-1.82.0/library/core/src/panicking.rs:112:18
379s    2: core::panicking::panic_nounwind_fmt
379s              at /usr/src/rustc-1.82.0/library/core/src/panicking.rs:122:5
379s    3: core::panicking::panic_nounwind
379s              at /usr/src/rustc-1.82.0/library/core/src/panicking.rs:221:5
379s    4: core::alloc::layout::Layout::from_size_align_unchecked::precondition_check
379s              at /usr/src/rustc-1.82.0/library/core/src/ub_checks.rs:70:21
379s    5: core::alloc::layout::Layout::from_size_align_unchecked
379s              at /usr/src/rustc-1.82.0/library/core/src/ub_checks.rs:77:17
379s    6: alloc::raw_vec::RawVecInner<A>::current_memory
379s              at /usr/src/rustc-1.82.0/library/alloc/src/raw_vec.rs:534:30
379s    7: alloc::raw_vec::RawVecInner<A>::deallocate
379s              at /usr/src/rustc-1.82.0/library/alloc/src/raw_vec.rs:749:38
379s    8: <alloc::raw_vec::RawVec<T,A> as core::ops::drop::Drop>::drop
379s              at /usr/src/rustc-1.82.0/library/alloc/src/raw_vec.rs:409:18
379s    9: core::ptr::drop_in_place<alloc::raw_vec::RawVec<u8>>
[...]

PR #1559, opened 2024-10-29
glib-macros: Derived boxed types are not TransparentPtrType but `TransparentType`

sdroege:
I.e. they're equal to their FFI representation, not equal to a pointer of their FFI representation.

Fixes https://github.com/gtk-rs/gtk-rs-core/issues/1554

----

@pbor you introduced this in 84c951ef5a62. Sorry for not catching that. I'll backport this breaking change to the 0.20 branch and include it in the next bugfix release because this was simply wrong and could've only worked by chance (and usually involve memory corruption).


bilelmoussaoui (approved):

sdroege:
Will do a bugfix release some time later today

PR #1572, opened 2024-11-11
Update gir files

sdroege:


gio/sys/src/lib.rs
```diff
@@ -17301,6 +17361,17 @@ extern "C" {
...
+    #[cfg(feature = "v2_84")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v2_84")))]
+    pub fn g_unix_mount_entries_get(time_read: *mut u64) -> *mut glib::GList;
+    #[cfg(feature = "v2_84")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v2_84")))]
+    pub fn g_unix_mount_entries_get_from_file(
+        table_path: *const c_char,
+        time_read_out: *mut u64,
+        n_entries_out: *mut size_t,
+    ) -> *mut *mut GUnixMountEntry;
```
sdroege, line 17374: Should probably bind these

glib/src/auto/enums.rs
```diff
@@ -1514,6 +1514,20 @@ pub enum UnicodeScript {
...
+    #[doc(alias = "G_UNICODE_SCRIPT_TULU_TIGALARI")]
+    TuluTigalari,
+    #[doc(alias = "G_UNICODE_SCRIPT_SUNUWAR")]
+    Sunuwar,
+    #[doc(alias = "G_UNICODE_SCRIPT_GURUNG_KHEMA")]
+    GurungKhema,
+    #[doc(alias = "G_UNICODE_SCRIPT_KIRAT_RAI")]
+    KiratRai,
+    #[doc(alias = "G_UNICODE_SCRIPT_OL_ONAL")]
+    OlOnal,
```
sdroege, line 1530: Need versions

gio/src/auto/unix_mount_entry.rs
```diff
@@ -3,17 +3,227 @@
...
 
 glib::wrapper! {
     #[derive(Debug)]
     pub struct UnixMountEntry(Boxed<ffi::GUnixMountEntry>);
 
     match fn {
-        copy => |ptr| glib::gobject_ffi::g_boxed_copy(ffi::g_unix_mount_entry_get_type(), ptr as *mut _) as *mut ffi::GUnixMountEntry,
-        free => |ptr| glib::gobject_ffi::g_boxed_free(ffi::g_unix_mount_entry_get_type(), ptr as *mut _),
+        copy => |ptr| ffi::g_unix_mount_entry_copy(mut_override(ptr)),
+        free => |ptr| ffi::g_unix_mount_entry_free(ptr),
```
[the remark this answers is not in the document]
sdroege: Could add fallback implements for these two functions in `src/manual.rs` I guess?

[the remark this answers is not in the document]
sdroege: Ah no, those are the same as `g_unix_mount_copy()` etc. We could provide fallbacks for everything here with older versions.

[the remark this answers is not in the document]
sdroege: @bilelmoussaoui ?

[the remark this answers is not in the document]
bilelmoussaoui: > Ah no, those are the same as g_unix_mount_copy() etc. We could provide fallbacks for everything here with older versions.

That sounds like the good thing to do here then :)

gio/src/auto/unix_mount_entry.rs
```diff
@@ -3,17 +3,227 @@
...
         type_ => || ffi::g_unix_mount_entry_get_type(),
     }
 }
 
+impl UnixMountEntry {
+    #[cfg(feature = "v2_84")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v2_84")))]
+    #[doc(alias = "g_unix_mount_entry_compare")]
+    fn compare(&mut self, mount2: &mut UnixMountEntry) -> i32 {
```
bilelmoussaoui, line 28: all of those seems to take &mut unnecessarily

gio/src/auto/unix_mount_entry.rs
```diff
@@ -3,17 +3,227 @@
...
+#[cfg(feature = "v2_84")]
+#[cfg_attr(docsrs, doc(cfg(feature = "v2_84")))]
+impl PartialEq for UnixMountEntry {
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        self.compare(other) == 0
+    }
+}
+
+impl Eq for UnixMountEntry {}
```
bilelmoussaoui, line 210: This will fail as PartialEq is behind a guard, maybe we should fix gir here:)

glib/sys/src/lib.rs
```diff
@@ -797,6 +804,8 @@ pub const G_URI_RESERVED_CHARS_GENERIC_DELIMITERS: &[u8] = b":/?#[]@\0";
 pub const G_URI_RESERVED_CHARS_SUBCOMPONENT_DELIMITERS: &[u8] = b"!$&'()*+,;=\0";
 pub const G_USEC_PER_SEC: c_int = 1000000;
 pub const G_WIN32_MSG_HANDLE: c_int = 19981206;
+pub const g_macro__has_attribute_ifunc: c_int = 0;
+pub const g_macro__has_attribute_no_sanitize_address: c_int = 0;
```
bilelmoussaoui, line 808: can we ignore those?

sdroege: We probably should

gio/src/auto/unix_mount_entry.rs
sdroege: Yes, but we can have `PartialEq` without the guard too :)

sdroege:
Now it passes, yay

sdroege:
@bilelmoussaoui All OK for you?

gio/Gir.toml
```diff
@@ -1565,6 +1565,92 @@ concurrency = "send+sync"
 cfg_condition = "unix"
     [[object.derive]]
     name = "Debug"
+    [[object.function]]
+    name = "compare"
```
bilelmoussaoui, line 1569: As this file is already too huge, I would suggest using pattern instead of name to combine those similar configs

glib/Gir.toml
```diff
@@ -911,6 +911,27 @@ status = "generate"
     [[object.member]]
     name = "nag_mundari"
     version = "2.74"
+    [[object.member]]
```
bilelmoussaoui, line 914: same here

bilelmoussaoui (commented):
otherwise lgtm

gio/Gir.toml
sdroege: Putting them all into the same line with `|`?

bilelmoussaoui: yep, that would be easier & can be documented as well

PR #1581, opened 2024-11-27
Signal/accumulator

### LINKED ISSUE #1580: [FEATURE REQUEST] Make `SignalBuilder::accumulator` closure easier to use.
The current function definition for the `SignalHandler::accumulator` is `Fn(&SignalInvocationHint, &mut Value, &Value) -> bool`. The returned bool decides on the control flow of the signal emission. Could we migrate to a `Fn(&SignalInvocationHint, &Value, &Value) -> std::ops::ControlFlow<Value, Value>`?

This would allow for nicer bindings in e.g. #214 and I'd like it to be in the same release as #1575 since `class_handler` and `accumulator` are tightly related.

@sdroege What do you think?

PJungkamp:
Closes #1580

### DIFF
--- a/glib/src/subclass/object.rs
+++ b/glib/src/subclass/object.rs
@@ -422,6 +422,19 @@ mod test {
                             .build(),
                         super::Signal::builder("create-string")
                             .return_type::<String>()
+                            .accumulator(|_hint, acc, val| {
+                                // join all strings from signal handlers by newline
+                                let mut acc = acc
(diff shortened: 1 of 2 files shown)


sdroege (approved):
Looks good to me. Why is it marked as draft? :)

PJungkamp:
> Why is it marked as draft? :)

By accident! :D

PR #1584, opened 2024-11-30
gio: Fix nullability of various DBus method call related parameters

sdroege:
CC @sophie-h

### DIFF
--- a/gio/tests/dbus_peer.rs
+++ b/gio/tests/dbus_peer.rs
@@ -1,0 +1,157 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[cfg(unix)]
+#[test]
+fn test_gdbus_peer_connection() {
+    use gio::{
+        glib::{self, VariantTy},
+        prelude::*,
+        DBusConnection, DBusConnectionFlags, DBusNodeInfo, Socket,
+    };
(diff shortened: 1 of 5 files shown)


gio/tests/dbus_peer.rs
```diff
@@ -0,0 +1,169 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[test]
```
bilelmoussaoui, line 3: Wouldn't moving this to test::run be enough? so you don't need the duplicate function :)  otherwise lgtm

sdroege: Indeed, this can be untangled considerably. Thanks :)

gio/tests/dbus_peer.rs
```diff
@@ -0,0 +1,159 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[cfg(unix)]
+#[test]
+fn test_gdbus_peer_connection() {
+    use gio::{
+        glib::{self, VariantTy},
+        prelude::SocketExt,
```
bilelmoussaoui, line 8: You have the prelude below, replace this with prelude::* and remove the other one

sdroege: Done

PR #1590, opened 2024-12-02
Draft: Use `Into<Option<_>>` in argument position where applicable

fengalin:
See [the gir PR](https://github.com/gtk-rs/gir/pull/1620) for details.

# TODO

- [ ] rebase, update gir versions and re-generate when gir PR is merged


fengalin:
In my tests so far, I only encountered rare cases such as the one fixed in https://github.com/gtk-rs/gtk-rs-core/pull/1590/commits/8c09634464f3c06e49c86620fbfe8101897e22ea, in response to the following error:

```
error[E0277]: the trait bound `Option<&str>: From<Option<&gio::glib::GString>>` is not satisfied
   --> gio/tests/dbus_peer.rs:34:13
    |
32  |         let connection = DBusConnection::new_future(
    |                          -------------------------- required by a bound introduced by this call
33  |             &socket_connection,
34  |             Some(&guid),
    |             ^^^^^^^^^^^ the trait `From<Option<&gio::glib::GString>>` is not implemented for `Option<&str>`, which is required by `Option<&gio::glib::GString>: Into<Option<&str>>`
    |
    = help: the following other types implement trait `From<T>`:
              `Option<&T>` implements `From<&Option<T>>`
              `Option<&mut T>` implements `From<&mut Option<T>>`
              `Option<T>` implements `From<T>`
    = note: required for `Option<&gio::glib::GString>` to implement `Into<Option<&str>>`
note: required by a bound in `DBusConnection::new_future`
   --> /home/fengalin/Projects/gtk-rs-core/gio/src/auto/dbus_connection.rs:947:20
    |
945 |     pub fn new_future<'a>(
    |            ---------- required by a bound in this associated function
946 |         stream: &(impl IsA<IOStream> + Clone + 'static),
947 |         guid: impl Into<Option<&'a str>>,
    |                    ^^^^^^^^^^^^^^^^^^^^^ required by this bound in `DBusConnection::new_future`
```

sdroege:
> `error[E0277]: the trait bound `Option<&str>: From<Option<&gio::glib::GString>>` is not satisfied`

Hm, why not and should we maybe add that? :)

> In my tests so far, I only encountered rare cases

Can you highlight the others ones in the different PRs/MRs together with the compiler error?

fengalin:
> > `error[E0277]: the trait bound `Option<&str>: From<Option<&gio::glib::GString>>` is not satisfied`
> Hm, why not and should we maybe add that? :)

Take a look at [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021) for an illustration. AIUI, the problem is not really that `Option<&str>` can't be built from `Option<&gio::glib::GString>`, but rustc doesn't seem to be able to handle this behind an `impl Into<_>` bound. A way to handle this would be the 3d case, but it adds more inconvenience that it solves imo. However:

> > In my tests so far, I only encountered rare cases
> Can you highlight the others ones in the different PRs/MRs together with the compiler error?

In the 3 main repositories, that was the only error that needed a fix. I pushed a [`gst-plugins-rs` branch](https://gitlab.freedesktop.org/fengalin/gst-plugins-rs/-/tree/more-into-option-args) patched to use the 3 repos and got only one `LazyLock` call site which needed [a fix](https://gitlab.freedesktop.org/fengalin/gst-plugins-rs/-/commit/2eeea47cb862c800b179ba9e57db31cc72f0fb42). Also I pushed [a preview](https://gitlab.freedesktop.org/fengalin/gst-plugins-rs/-/commit/ac8e6680e62edf4987cfec5bb0d0645db8a9acfc) in another branch of what to expect if we'd used this usability improvement throughout the code (based on a previous iteration and not complete: in particular, this doesn't take advantage of the availability of `Into<Option<_>>` args in gtk-rs-core nor in gtk4).

sdroege:
That `Some(&stringly_value)` doesn't compile anymore in many cases and requires an `as_str()` or otherwise an explicit conversion to `&str` seems a bit ugly. AFAIU that's because previously this was covered by auto-deref but that doesn't trigger anymore in the generic context? It feels like I created a rustc issue about that long ago.

These are all cases of `impl Into<Option<impl IntoGStr>>` or `impl Into<Option<&str>>` I assume? One could add a new `IntoOptionStr` trait or so for this but that seems more confusing.

One concern I have is that people might do `as_str()` for the `impl IntoGStr` case, while `as_gstr()` would be more efficient but less known.

Otherwise this seems fine. The same problem with `as_str()` also exists already on `glib::GString` if you want to `match` on them, for example. The compiler errors are also less bad than I remember them in the past.

@bilelmoussaoui Do you have opinions on this?

fengalin:
I agree this case is suboptimal. Now that the generator implementation has settled down, I can take some time to figure out if something could improve this use case.

Worse case scenario, we could remove support for `Into<Option<& T>>` and keep the others which don't cause problems afaict: `Into<Option<T>>` and `Into<Option<IsA<T>>`.

sdroege:
> Worse case scenario, we could remove support for `Into<Option<& T>>` and keep the others which don't cause problems afaict: `Into<Option<T>>` and `Into<Option<IsA<T>>`.

`impl Into<Option<impl IsA<T>>>` AFAIU already has the problem right now with `Option<impl IsA<T>>` alone, so yes, that doesn't make things worse :)

Considering the minimal amount of changes you had to do and it only affecting strings, I'm not sure keeping it like this is really a problem.

fengalin:
Postponing these changes to a future Rust bindings generator rewrite

PR #1596, opened 2024-12-09
[backport] windows: use a range for windows-rs dependency

pbor:


sdroege:
I'll backport that with everything else before the release. Unless you want to do that now :)

pbor:
I added it here because it required manual changes, as you prefer

PR #1600, opened 2024-12-10
gio: Work around GLib memory leak in g_dbus_connection_register_objec…

sdroege:
…t_with_closures()

Assume that the invocation is passed as `transfer full` into the closure.

This workaround is not going to break with future versions of GLib as fixing the bug was considered a breaking API change.

See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4427

----

CC @sophie-h


sdroege:
This breaks on the CI until GLib >= 2.82.4 or < 2.82.3 is available there.

sophie-h:
What a mess overall. I think the upstream function should be deprecated when there is a new one.

sdroege:
That's the plan, probably: https://gitlab.gnome.org/GNOME/glib/-/issues/3560

PR #1604, opened 2024-12-16
Operator overloads for Graphene

jgcodes2020:
Currently includes:

- vector addition, subtraction, scaling, component-wise multiply/divide
- matrix scaling, matrix multiplication, matrix-vector multiplication
- quaternion addition, multiplication

Semantics follow GLSL: `vector * vector` is component-wise multiplication, `scalar * vector` is scaling.

Only thing I haven't implemented is dividing a vector by a scalar, since Graphene for whatever reason doesn't have that as a function.


sdroege:
> Only thing I haven't implemented is dividing a vector by a scalar, since Graphene for whatever reason doesn't have that as a function.

That's just `1.0/scalar * vector` or not?

graphene/src/matrix.rs
```diff
@@ -236,3 +236,45 @@ mod tests {
...
+
+impl std::ops::Mul<Matrix> for Matrix {
+    type Output = Matrix;
+
+    fn mul(self, rhs: Matrix) -> Self::Output {
+        (&self).multiply(&rhs)
+    }
+}
+
+impl std::ops::Mul<Vec4> for Matrix {
```
sdroege, line 248: Could also add scalar multiplication here

sdroege (commented):
Looks good to me in general. @bilelmoussaoui ?

sdroege:
> Sure that works, but it can lose accuracy because of the limited precision of floating-point numbers.

Indeed. Maybe better to keep that explicit then that there's no direct division and leave the choice to the caller

sdroege:
The clippy warnings need some fixing

graphene/src/vec2.rs
```diff
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::fmt;
+use std::{fmt, ops};
```
bilelmoussaoui, line 3: No need to import std::ops here, you didnt do so in the other file

jgcodes2020: I wanted to maintain consistency with importing `std::fmt` for `fmt::Display`. I've fixed `Matrix` to implement that.

graphene/src/matrix.rs
jgcodes2020: Did that in the last commit.

jgcodes2020:
> The clippy warnings need some fixing

Do I just explicitly specify the method names (i.e. `Self::add(&self, &rhs)`?) If I remove the borrow, Rust selects `add` from `std::ops::Add` instead of `add` from `graphene::Vec2`, since it prefers to take by value.

sdroege:
> Do I just explicitly specify the method names (i.e. Self::add(&self, &rhs)?

I think that would make it most clear

sdroege:
clippy is still unhappy :)

jgcodes2020 (commented):
> clippy is still unhappy :) 

That should be fixed now.

sdroege (approved):

PR #1621, opened 2025-01-07
glib: Remove SendWeakRef::new() because implementation is unsound

### LINKED ISSUE #1617: [BUG] SendWeakRef is unsound
**Bug description**

Calling `SendWeakRef.set()` does not update the thread id.

```rust
let w = gtk::Window::new();
w.present();
let r = glib::SendWeakRef::new();
// set() does not set the thread id in SendWeakRef:
r.set(Some(&w));
std::thread::spawn(move || {
    // So now this does not panic:
    let w = r.upgrade().unwrap();
    // And we can call window methods from the wrong thread, potentially causing segfaults:
    w.destroy();
});
```

quo:
Fixes #1617

Not sure whether it would be better to remove the Default impl too, I've kept it for now.

### DIFF
--- a/glib/src/object.rs
+++ b/glib/src/object.rs
@@ -3583,18 +3583,13 @@ impl<T: ObjectType> PartialOrd for WeakRef<T> {
 /// where it was created on will panic but dropping or cloning can be done
 /// safely from any thread.
 #[derive(Debug)]
-pub struct SendWeakRef<T: ObjectType>(WeakRef<T>, Option<usize>);
+pub struct SendWeakRef<T: ObjectType>(WeakRef<T>, usize);
 
 impl<T: ObjectType> SendWeakRef<T> {
     #[inline]
-    pub fn new() -> SendWeakRef<T> {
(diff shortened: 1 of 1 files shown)


sdroege (approved):

sdroege:
> Not sure whether it would be better to remove the Default impl too, I've kept it for now.

That seems OK and simplifies initializing struct fields.

sdroege:
Do you also want to add a `impl<T: ObjectType> From<T> for SendWeakRef<T>` or some other kind of convenience constructor? Seems useful to me at least

quo:
`SendWeakRef::from(o.downgrade())` is good enough for me. Like you said, usually async channels are cleaner, so `SendWeakRef` probably shouldn't see all that much use.

If you do want to add a nicer API, I would make it similar to the `WeakRef` API. So something like `obj.downgrade_send()`, or `obj.downgrade().send()`, or maybe `obj.ref_send()`?

sdroege:
Ok, let's keep it like this then :)

PR #1634, opened 2025-01-18
Expose dbus_register and dbus_unregister vfuncs

swsnr:
Using this vfunc makes exporting objects on the app's bus name much easier and more reliable.

I've rewritten the corresponding example to use these two methods, as the GIO docs recommend to use these vfuncs for object registration.  Also, registering an object in startup is prone to race conditions wrt to bus ownership as the object gets registered only after the application claimed the bus name.

### DIFF
--- a/examples/gio_dbus_register_object/main.rs
+++ b/examples/gio_dbus_register_object/main.rs
@@ -1,126 +1,162 @@
-use gio::{prelude::*, IOErrorEnum};
-use std::{
-    sync::mpsc::{channel, Receiver, Sender},
-    time::Duration,
-};
+use gio::prelude::*;
 
-const EXAMPLE_XML: &str = r#"
-  <node>
-    <interface name='com.github.gtk_rs.examples.HelloWorld'>
-      <method name='Hello'>
-        <arg type='s' name='name' direction='in'/>
-        <arg type='s' name='greet' direction='out'/>
-      </method>
-      <method name='SlowHello'>
-        <arg type='s' name='name' direction='in'/>
-        <arg type='u' name='delay' direction='in'/>
-        <arg type='s' name='greet' direction='out'/>
-      </method>
-    </interface>
-  </node>
+glib::wrapper! {
+    pub struct SampleApplication(ObjectSubclass<imp::SampleApplication>)
+        @extends gio::Application,
+        @implements gio::ActionGroup, gio::ActionMap;
+}
+
+impl Default for SampleApplication {
+    fn default() -> Self {
+        glib::Object::builder()
+            .property(
+                "application-id",
+                "com.github.gtk-rs.examples.RegisterDBusObject",
+            )
+            .build()
+    }
+}
+
+mod imp {
+    use std::cell::RefCell;
+    use std::time::Duration;
+
+    use gio::prelude::*;
+    use gio::subclass::prelude::*;
+    use gio::{DBusConnection, IOErrorEnum};
+
+    const EXAMPLE_XML: &str = r#"
+<node>
+  <interface name='com.github.gtk_rs.examples.HelloWorld'>
+    <method name='Hello'>
+      <arg type='s' name='name' direction='in'/>
+      <arg type='s' name='greet' direction='out'/>
+    </method>
+    <method name='SlowHello'>
+      <arg type='s' name='name' direction='in'/>
+      <arg type='u' name='delay' direction='in'/>
+      <arg type='s' name='greet' direction='out'/>
+    </method>
+  </interface>
+</node>
 "#;
 
-#[derive(Debug, glib::Variant)]
-struct Hello {
-    name: String,
-}
-
-#[derive(Debug, glib::Variant)]
-struct SlowHello {
-    name: String,
-    delay: u32,
-}
-
-#[derive(Debug)]
-enum HelloMethod {
-    Hello(Hello),
-    SlowHello(SlowHello),
-}
-
-impl DBusMethodCall for HelloMethod {
-    fn parse_call(
-        _obj_path: &str,
-        _interface: Option<&str>,
-        method: &str,
-        params: glib::Variant,
-    ) -> Result<Self, glib::Error> {
-        match method {
-            "Hello" => Ok(params.get::<Hello>().map(Self::Hello)),
-            "SlowHello" => Ok(params.get::<SlowHello>().map(Self::SlowHello)),
-            _ => Err(glib::Error::new(IOErrorEnum::Failed, "No such method")),
-        }
-        .and_then(|p| p.ok_or_else(|| glib::Error::new(IOErrorEnum::Failed, "Invalid parameters")))
+    #[derive(Debug, glib::Variant)]
+    struct Hello {
+        name: String,
     }
-}
 
-fn on_startup(app: &gio::Application, tx: &Sender<gio::RegistrationId>) {
-    let connection = app.dbus_connection().expect("connection");
+    #[derive(Debug, glib::Variant)]
+    struct SlowHello {
+        name: String,
+        delay: u32,
+    }
 
-    let example = gio::DBusNodeInfo::for_xml(EXAMPLE_XML)
-        .ok()
-        .and_then(|e| e.lookup_interface("com.github.gtk_rs.examples.HelloWorld"))
-        .expect("Example interface");
+    #[derive(Debug)]
+    enum HelloMethod {
+        Hello(Hello),
+        SlowHello(SlowHello),
+    }
 
-    if let Ok(id) = connection
-        .register_object("/com/github/gtk_rs/examples/HelloWorld", &example)
-        .typed_method_call::<HelloMethod>()
-        .invoke_and_return_future_local(|_, sender, call| {
-            println!("Method call from {sender:?}");
-            async {
-                match call {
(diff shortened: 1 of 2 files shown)


swsnr:
This is my first time doing glib FFI, so I'd ask for a careful review, especially around ownership of pointers.

If this is acceptable, I'd very much appreciate a timely backport and release, as I need this to reliably export an object on the bus in my code.

swsnr:
Mmh, did I cause this CI error?

sdroege:
> Mmh, did I cause this CI error?

No, the docs job is just randomly failing.

gio/src/subclass/application.rs
```diff
@@ -110,6 +110,10 @@ pub trait ApplicationImpl:
     fn handle_local_options(&self, options: &VariantDict) -> ExitCode {
         self.parent_handle_local_options(options)
     }
+
+    fn dbus_register(&self, connection: &DBusConnection, object_path: &str) -> Result<bool, Error> {
```
sdroege, line 114: This should return `Result<(), Error>` AFAICS. The C function is supposed to either return `FALSE` and set the error, or return `TRUE` and not set the error.

gio/src/subclass/application.rs
```diff
@@ -390,6 +425,30 @@ unsafe extern "C" fn application_handle_local_options<T: ApplicationImpl>(
...
+) -> glib::ffi::gboolean {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    match imp.dbus_register(
+        &from_glib_borrow(connection),
+        &glib::GString::from_glib_borrow(object_path),
+    ) {
+        Ok(true) => glib::ffi::GTRUE,
+        Ok(false) => glib::ffi::GFALSE,
```
sdroege, line 462: And same here

examples/gio_dbus_register_object/main.rs
```diff
@@ -94,24 +97,33 @@ mod imp {
...
-                                println!("{greet}");
-                                Ok(Some(greet.to_variant()))
-                            }
-                            HelloMethod::SlowHello(SlowHello { name, delay }) => {
-                                glib::timeout_future(Duration::from_secs(delay as u64)).await;
-                                let greet = format!("Hello {name} after {delay} seconds!");
-                                println!("{greet}");
-                                Ok(Some(greet.to_variant()))
+                .invoke_and_return_future_local(glib::clone!(
+                    #[strong(rename_to = app)]
```
sdroege, line 101: I think this causes a reference cycle. Should probably be a weak reference at least

sdroege (commented):
Looks good to me otherwise, thanks :)

gio/src/subclass/application.rs
```diff
@@ -286,6 +339,8 @@ unsafe impl<T: ApplicationImpl> IsSubclassable<T> for Application {
         klass.shutdown = Some(application_shutdown::<T>);
         klass.startup = Some(application_startup::<T>);
         klass.handle_local_options = Some(application_handle_local_options::<T>);
+        klass.dbus_register = Some(application_dbus_register::<T>);
```
bilelmoussaoui, line 342: please add name_lost as well while you are at it, that would leave only add_platform_data missing which is not really useful as we don't have GVariantBuilder support

gio/src/subclass/application.rs
swsnr: Thanks, I'll change this. I must admit that I didn't entirely understand the semantics here 😇

gio/src/subclass/application.rs
swsnr: Okay, will do.

examples/gio_dbus_register_object/main.rs
swsnr: Sure, can change this, though I'm not really following where the cycle would come from, or why it'd be harmful given that the app usually lives for the entire lifetime of the process 🤔

But I'm also only superficially familiar with all this, so I trust your judgement 😅

gio/src/subclass/application.rs
sdroege: Me neither, the documentation was confusing and I had to look at the implementation :)

swsnr: 😅

swsnr:
@bilelmoussaoui @sdroege I think I addressed everything.  I added name_lost too; I hope I did it right, because I didn't really understand what the return value does and it doesn't seem to be documented :shrug:

gio/src/subclass/application.rs
```diff
@@ -390,6 +462,54 @@ unsafe extern "C" fn application_handle_local_options<T: ApplicationImpl>(
...
+        &glib::GString::from_glib_borrow(object_path),
+    );
+}
+
+unsafe extern "C" fn application_name_lost<T: ApplicationImpl>(
+    ptr: *mut ffi::GApplication,
+) -> glib::ffi::gboolean {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.imp();
+    if imp.name_lost() {
```
bilelmoussaoui, line 506: Use imp.name_lost().into_glib() directly

swsnr: Yeah, right :see_no_evil: I'll also fix the other conditional that broke the build, but probably not before tomorrow evening; it's getting late here :slightly_smiling_face:

gio/src/subclass/application.rs
```diff
@@ -266,6 +278,63 @@ pub trait ApplicationImplExt: ApplicationImpl {
...
+        }
+    }
+
+    fn parent_name_lost(&self) -> bool {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *mut ffi::GApplicationClass;
+            let f = (*parent_class)
+                .name_lost
+                .expect("No parent class implementation for \"name_lost\"");
```
swsnr, line 330: Is this correct?  Most code around here uses `.expect` to access the parent function, but `handle_local_option` properly unpacks the `Option`.  I'm not sure when to use either, so I'd appreciate if you'd check this.

gio/src/subclass/application.rs
swsnr: Fixed.

gio/src/subclass/application.rs
bilelmoussaoui: Unfortunately you have to check the C code and see if the Object has a default implementation, meaning there would always be a function to chain up.

sdroege: Not only that, but also whether this function is mandatory. It might not have a default implementation but the caller might just call it always without checking if there's one, in which case it's mandatory

sdroege: Checking the code, it looks mandatory so unwrapping/expect here is OK.

gio/src/subclass/application.rs
```diff
@@ -266,6 +270,33 @@ pub trait ApplicationImplExt: ApplicationImpl {
...
+                .expect("No parent class implementation for \"dbus_register\"");
+            let mut err = ptr::null_mut();
+            let res = f(
+                self.obj().unsafe_cast_ref::<Application>().to_glib_none().0,
+                connection.to_glib_none().0,
+                object_path.to_glib_none().0,
+                &mut err,
+            );
+            if res == glib::ffi::GFALSE {
+                debug_assert!(!err.is_null());
```
sdroege, line 301: That assertion is already in `from_glib_full()`, but you might want to check the in the `TRUE`-case it is always `NULL`

gio/src/subclass/application.rs
bilelmoussaoui: I am curious though if this should return one of the two glib enums we have that internally maps to a boolean type

sdroege: Return value should be a `glib::Propagation` I think

examples/gio_dbus_register_object/main.rs
```diff
@@ -94,24 +97,33 @@ mod imp {
...
+                                    Ok(Some(greet.to_variant()))
+                                }
+                                HelloMethod::SlowHello(SlowHello { name, delay }) => {
+                                    glib::timeout_future(Duration::from_secs(delay as u64)).await;
+                                    let greet = format!("Hello {name} after {delay} seconds!");
+                                    println!("{greet}");
+                                    Ok(Some(greet.to_variant()))
+                                }
+                                HelloMethod::GoodBye => {
+                                    app.inspect(|app| app.quit());
```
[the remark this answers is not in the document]
swsnr: Mmh, isn't the whole point and purpose of `inspect` in contrast to `map` to run side effects?  I always use it that war 😅

But I'm fine to follow whatever style you prefer so I'll change it this evening.

gio/src/subclass/application.rs
swsnr: @sdroege Thanks for checking and saving me the trip to C 🙏

I'll change the return value to `Propagation`

swsnr (commented):
I've addressed all remarks, changed the return value of `name_lost` to `Propagation`, and rebased onto main.

swsnr:
While I was at it I've pushed another commit which replaces the IOErrorEnum's in the DBus example with `DBusError` which seemed more appropriate.

sdroege (approved):

PR #1640, opened 2025-01-27
Allow implementing a GAction interface

andy128k:
Closes !1636


gio/src/subclass/action.rs
```diff
@@ -0,0 +1,254 @@
...
+    prelude::*,
+    subclass::prelude::*,
+    translate::*,
+    GString, Variant, VariantType,
+};
+
+use crate::{ffi, Action};
+
+pub trait ActionImpl: ObjectImpl + ObjectSubclass<Type: IsA<Action>> {
+    fn name(&self) -> GString {
```
sdroege, line 14: Is it valid for this to change, or should this maybe just be an associated constant?

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,254 @@
...
+    imp.name().to_glib_none().0
+}
+
+unsafe extern "C" fn action_get_parameter_type<T: ActionImpl>(
+    actionptr: *mut ffi::GAction,
+) -> *const GVariantType {
+    let instance = &*(actionptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    imp.parameter_type().to_glib_none().0
```
sdroege, line 197: Are all the methods really `transfer none`? That complicates things considerably. Right now you're potentially causing use-after-frees here (the returned pointer might just be freed here now). Maybe this should return a `&VariantType` etc?

andy128k: Hmm. I am not sure I can progress from here. If I declare `param_type` and `state_type` as returning references then I have an issue with implementing proxy action which calls another action underneath cause it already [returns an owned object](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gio/prelude/trait.ActionExt.html#method.parameter_type).

gio/src/subclass/action.rs
andy128k: Well, a name is a property of the action instance, not an action class.

andy128k:
Another issue I have is that `GAction` implies that an implementor had some properties. 

```
(process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'state-type' from interface 'GAction'

(process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'state' from interface 'GAction'

(process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'parameter-type' from interface 'GAction'

(process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'enabled' from interface 'GAction'
```

I have no clue what would be an idiomatic way to use `g_object_class_override_property` in Rust.

gio/src/subclass/action.rs
sdroege: Indeed

gio/src/subclass/action.rs
sdroege: Then this would have be handled like in e.g. `IOStreamImpl`

bilelmoussaoui:
> Another issue I have is that `GAction` implies that an implementor had some properties. 
> 
> ```
> (process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'state-type' from interface 'GAction'
> 
> (process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'state' from interface 'GAction'
> 
> (process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'parameter-type' from interface 'GAction'
> 
> (process:527349): GLib-GObject-CRITICAL **: 20:51:28.369: Object class ExampleRenamedAction doesn't implement property 'enabled' from interface 'GAction'
> ```
> 
> I have no clue what would be an idiomatic way to use `g_object_class_override_property` in Rust.

GtkEditable provides a method to automatically add the properties for you along with the get_property and set_property. Either the example implementing a custom Action would have to do that manually or you could add helpers that would do it automatically for you, see the EditableImpl in the gtk4-rs repo.

examples/gio_action_impl/action.rs
```diff
@@ -0,0 +1,77 @@
+use gio::{prelude::*, subclass::prelude::*};
+use glib;
+
+mod imp {
+    use super::*;
+    use std::cell::OnceCell;
+
+    #[derive(glib::Properties, Default)]
+    #[properties(wrapper_type = super::RenamedAction)]
+    pub struct RenamedAction {
```
bilelmoussaoui, line 10: The properties macro allows overriding properties by the way

gio/src/subclass/action.rs
andy128k: Thanks! I did my best (copying from GActionGroup, he-he) but I have concerns about my implementation of `action_get_name`. It works but looks ugly.

andy128k:
@bilelmoussaoui Thanks for pointing a direction! Unfortunately Gio does not provide helpers similar to `gtk_editable_*_delegate_*`, so I needed to replicate C code myself.

---

I am not sure if I fully understand the subject but [this line](https://github.com/gtk-rs/gtk4-rs/blob/main/gtk4/src/subclass/editable.rs#L273) seems wrong. [Gtk's doc](https://docs.gtk.org/gtk4/iface.Editable.html#implementing-gtkeditable) uses `NUM_PROPERTIES` as a second parameter which is usually not an amount of properties but `amount + 1` (props are numbered from 1). [gtk_editable_install_properties](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkeditable.c?ref_type=heads#L968) does not add `1`.

examples/gio_action_impl/action.rs
andy128k: I gave it a try but received an error 

```
(process:718176): GLib-GObject-CRITICAL **: 22:59:07.010: When installing property: type 'ExampleRenamedAction' already has a property named 'name'
```

I guess it's because of the sequence of initialization. When I define a class and implement an interface then class installs properties first and interface fails to install them.

examples/gio_action_impl/action.rs
```diff
@@ -0,0 +1,91 @@
...
+    impl ObjectImpl for RenamedAction {
+        fn properties() -> &'static [glib::ParamSpec] {
+            Self::derived_properties()
+        }
+
+        fn set_property(&self, id: usize, value: &glib::Value, pspec: &glib::ParamSpec) {
+            let _ = self.delegate_set_property(id, value, pspec) || {
+                self.derived_set_property(id, value, pspec);
+                true
+            };
```
sdroege, line 35:
```suggestion
             if !self.delegate_set_property(id, value, pspec) {
                self.derived_set_property(id, value, pspec);
            }
```

reads less confusing

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    fn change_state(&self, value: Variant) {
+        self.parent_change_state(value);
+    }
+
+    fn activate(&self, parameter: Option<Variant>) {
+        self.parent_activate(parameter);
+    }
+
+    fn delegate_get_property(
```
sdroege, line 48: These two delegates should probably go into the `Ext` trait. They're not supposed to be overridden, are they?

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,427 @@
...
+
+            let mut first_prop = std::mem::MaybeUninit::uninit();
+            let properties = glib::gobject_ffi::g_object_class_list_properties(
+                object_class,
+                first_prop.as_mut_ptr(),
+            );
+            glib::ffi::g_free(properties as *mut _);
+            let first_prop = first_prop.assume_init() + 1;
+
+            set_first_prop(instance_type, first_prop as usize);
```
sdroege, line 281: That's a nice hack for the properties :)

examples/gio_action_impl/action.rs
andy128k: Indeed. I just fiddled with this to see how it will work for the case of multiple interfaces. Then it would be a single expression w/o deep nesting. Anyway, here it looks ugly.

examples/gio_action_impl/action.rs
sdroege: The way how you do it now should work or not?

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,427 @@
...
+
+            glib::gobject_ffi::g_object_class_override_property(
+                object_class,
+                first_prop + ActionProperty::Name as u32,
+                b"name\0".as_ptr() as *const _,
```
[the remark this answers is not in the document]
andy128k: Oh, I totally forgot that c-strings exist. Thanks.

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,427 @@
...
+        self.parent_activate(parameter);
+    }
+
+    fn delegate_get_property(
+        &self,
+        prop_id: usize,
+        _pspec: &glib::ParamSpec,
+    ) -> Option<glib::Value> {
+        let type_: glib::Type = self.obj().type_();
+        let property = ActionProperty::from_type(type_, prop_id)?;
```
sdroege, line 54: If you pass in `Self::type_()` here then you don't need to do all the parent type walking inside `from_type()` and always have directly the correct type

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+}
+
+unsafe extern "C" fn action_get_parameter_type<T: ActionImpl>(
+    actionptr: *mut ffi::GAction,
+) -> *const GVariantType {
+    let instance = &*(actionptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    if let Some(parameter_type) = imp.parameter_type() {
```
sdroege, line 341: This should probably be moved into the `set_qdata()` call. Now you'd get inconsistent behaviour if the function returns `Some` on the first call and `None` afterwards

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,427 @@
...
+    }
+}
+
+unsafe extern "C" fn action_get_state_type<T: ActionImpl>(
+    actionptr: *mut ffi::GAction,
+) -> *const GVariantType {
+    let instance = &*(actionptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    if let Some(state_type) = imp.state_type() {
```
sdroege, line 368: Same here and probably elsewhere

gio/src/subclass/action.rs
sdroege: Ugly but looks correct except for that one comment

sdroege: Can be slightly less ugly if you use `LazyLock` instead of `OnceLock`

gio/src/subclass/action.rs
andy128k: That's what I've found in `GtkEditable`'s [code](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkeditable.c?ref_type=heads#L1078-1081).

sdroege: @bilelmoussaoui Why does it do things so complicated, do you remember? :)

examples/gio_action_impl/action.rs
andy128k: It does not work. I guess It works as expected in the cases like 

```
class A implements IFace

class B extends A implements IFace {
   override property x of IFace
}
```

where initialization sequence is: `A`, `IFace`, `B`. So, when `B` is initializing `IFace` is already there and could be overriden.

but does not work for 

```
class B implements IFace {
   override property x of IFace
}
```

where initialization sequence is: `B`, `IFace`. So, `B` has nothing to override yet.

gio/src/subclass/action.rs
andy128k: It's not directly forbidden by Gio's docs, but I think it could be insane if the type of a parameter would change during the action's lifetime.

A similar assumption is made in [`ActionGroup`](https://github.com/gtk-rs/gtk-rs-core/blob/main/gio/src/subclass/action_group.rs#L433).

gio/src/subclass/action.rs
andy128k: Moved

gio/src/subclass/action.rs
andy128k: My concern is rather around moving `GString` to qdata and querying it back.

sdroege: That's fine. The returned pointer after storing it in qdata is stable

gio/src/subclass/action.rs
sdroege: My personal approach here would be to either only call the function once to don't give the user a chance to do it wrong, or otherwise at least to add an assertion :)

andy128k: I've changed this and other places (`name`, `parameter_type` and `state_type`) in the way that the function is called once.

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+        Some(match property {
+            ActionProperty::Name => self.name().to_value(),
+            ActionProperty::ParameterType => self.parameter_type().to_value(),
+            ActionProperty::StateType => self.state_type().to_value(),
+            ActionProperty::Enabled => self.is_enabled().to_value(),
+            ActionProperty::State => self.state().to_value(),
+        })
+    }
+
+    fn delegate_set_property(
```
bilelmoussaoui, line 181: good call of adding this even if it just returns false; it would allow adding some attribute to glib::derived_properties to automatically use the `delegate_get_property` / `delegate_set_property` calls, simplifying things a bit.

bilelmoussaoui:
what is left to be done here btw?

andy128k:
> what is left to be done here btw?

I'd say its ready, he-he. But I have a concern about a potential off-by-one error (see above) and waited for your judgement.

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<Action>() as *const ffi::GActionInterface;
+
+            let func = (*parent_iface)
+                .activate
+                .expect("no parent \"activate\" implementation");
+            func(
+                self.obj().unsafe_cast_ref::<Action>().to_glib_none().0,
+                parameter.to_glib_none().0,
```
sdroege, line 160: As this is `to_glib_none()`, maybe the parameter should be a reference (same for others above)

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<Action>() as *const ffi::GActionInterface;
+
+            let func = (*parent_iface)
+                .get_state
+                .expect("no parent \"get_state\" implementation");
+            let ret = func(self.obj().unsafe_cast_ref::<Action>().to_glib_none().0);
+            from_glib_none(ret)
```
sdroege, line 129: Could make this return an `Option<&glib::Variant>` with a bit of trickery maybe

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+    let imp = instance.imp();
+
+    imp.state_hint().to_glib_full()
+}
+
+unsafe extern "C" fn action_get_enabled<T: ActionImpl>(actionptr: *mut ffi::GAction) -> gboolean {
+    let instance = &*(actionptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    imp.is_enabled() as gboolean
```
sdroege, line 385:
```suggestion
    imp.is_enabled().into_glib()
```

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+}
+
+unsafe extern "C" fn action_get_state_hint<T: ActionImpl>(
+    actionptr: *mut ffi::GAction,
+) -> *mut GVariant {
+    let instance = &*(actionptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    imp.state_hint().to_glib_full()
```
sdroege, line 378:
```suggestion
    imp.state_hint().into_glib_ptr()
```

also elsewhere

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+
+unsafe extern "C" fn action_activate<T: ActionImpl>(
+    actionptr: *mut ffi::GAction,
+    parameterptr: *mut GVariant,
+) {
+    let instance = &*(actionptr as *mut T::Instance);
+    let imp = instance.imp();
+    let parameter: Option<glib::Variant> = from_glib_none(parameterptr);
+
+    imp.activate(parameter);
```
sdroege, line 416:
```suggestion
    let parameter: Option<glib::Variant> = from_glib_borrow(parameterptr);

    imp.activate(parameter.as_ref());
```

or so

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+            _ => None,
+        }
+    }
+
+    fn from_type(mut type_: glib::Type, id: usize) -> Option<Self> {
+        loop {
+            if let Some(first_prop) = get_first_prop(type_) {
+                break ActionProperty::from_id(first_prop, id);
+            }
+            type_ = type_.parent()?;
```
sdroege, line 247: Is this actually correct to look at all the parents, can you add a comment why?
Shouldn't this just look at the just the type of `T`?

Maybe a test with multiple levels of `GAction` inheritance would be useful to make sure this works and continues working

gio/src/subclass/action.rs
sdroege: Hm no, we'd need a `*const *const glib::ffi::GVariant` but don't have one :)

gio/src/subclass/action.rs
```diff
@@ -0,0 +1,417 @@
...
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<Action>() as *const ffi::GActionInterface;
+
+            let func = (*parent_iface)
+                .get_name
+                .expect("no parent \"get_name\" implementation");
+            let ret = func(self.obj().unsafe_cast_ref::<Action>().to_glib_none().0);
+            from_glib_none(ret)
```
sdroege, line 59: This is rather suboptimal as it copies strings all the time whenever the name is retrieved. This should be possible to make a `&glib::GStr` or not?

examples/gio_action_impl/action.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    #[glib::derived_properties]
+    impl ObjectImpl for RenamedAction {
+        fn properties() -> &'static [glib::ParamSpec] {
+            Self::derived_properties()
+        }
+
+        fn set_property(&self, id: usize, value: &glib::Value, pspec: &glib::ParamSpec) {
+            if !self.delegate_set_property(id, value, pspec) {
```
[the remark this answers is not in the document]
bilelmoussaoui: So far there are only Actions and Editable.i rather keep it this way until we find more cases like this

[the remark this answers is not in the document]
sdroege: You'd already have this problem then if you implement both interfaces in the same type :)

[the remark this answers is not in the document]
bilelmoussaoui: There is 0 use case for that. You would rather implement GActionMap instead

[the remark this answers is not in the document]
sdroege: Sure, but if we want to solve this problem later we need to change API in even more places so it seems useful to figure this out from the beginning. I'm not going to block this PR on this though, just don't tell me in a few years that I didn't warn you :P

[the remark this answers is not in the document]
bilelmoussaoui: My goal is to basically make the derived_properties macro automatically generate those bits for you. I will send a patch for it once this PR lands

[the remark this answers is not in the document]
sdroege: Sounds good

Issue #1638, opened 2025-01-25
`gio::SubprocessLauncher::set_environ` and `gio::ApplicationCommandLine::environ` don't compose well


pjungkamp:
I noticed that `gio::ApplicationCommandLine::environ` returns a `Vec<OsString>` and that `gio::SubprocessLauncher::set_environ` only takes a `&[&Path]` as argument.

This seems rather unfortunate. My application should share a primary instance for multiple windows, but still needs to spawn processes with the environment that spawned a new window on the primary instance. This currently needs an intermediate collection to collect the `Path` references that can be passed to `gio::SubprocessLauncher::set_environ`.

```rust
let cmd: gio:ApplicationCommandLine; // from the parameter of the `gio::Application::command_line` signal
let launcher = gio::SubprocessLauncher::new();

launcher.set_environ(cmd.environ().iter().map(<OsString as AsRef<Path>>::as_ref).collect::<Vec<&Path>>().as_slice());
```

---

This could be generic over all [`ToGlibContainerFromSlice`](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/translate/trait.ToGlibContainerFromSlice.html) types that coerce properly like `&OsStr`, `&str`, `&Path` and their owned variants `OsString`, `String`, `PathBuf`.

I think an intermediary trait for all kinds of `OsStr`-like values could help to make the `gio::SubprocessLauncher::set_environ` method more usable.

Maybe something like `fn set_environ<'a>(&self, envs: &'a [impl AsOsStr<'a>])`? Where `AsOsStr<'a>` is a subtrait of `glib::translate::ToGlibContainerFromSlice<'a, *const *const i8>` so the `.to_glib_none()` in the generated bindings still works?

```rust
trait AsOsStr<'a> : glib::translate::ToGlibContainerFromSlice<'a, *const *const i8> {}
impl<'a> AsOsStr<'a> for T where T: AsRef<OsStr> + glib::translate::ToGlibContainerFromSlice<'a, *const *const i8> {}
```

On a side-note: Why doesn't `CString` and `&CStr` implement `ToGlibContainerFromSlice`?

---

It's just a minor inconvenience but I wanted to start some discussion on these `OsStr`-like slices in function arguments.

sdroege:
> I noticed that gio::ApplicationCommandLine::environ returns a Vec<OsString> and that gio::SubprocessLauncher::set_environ only takes a &[&Path] as argument.

[...]

PR #1642, opened 2025-01-27
gio: Use manual env: &[OsString] parameter for SubprocessLauncher::set_environ

pjungkamp:
See #1638

This should be better than the existing bindings, but changing the API to something that accepts `&[impl AsRef<OsStr>]` would be preferable.

### DIFF
--- a/gio/src/auto/subprocess_launcher.rs
+++ b/gio/src/auto/subprocess_launcher.rs
@@ -79,13 +79,6 @@ impl SubprocessLauncher {
         }
     }
 
-    #[doc(alias = "g_subprocess_launcher_set_environ")]
-    pub fn set_environ(&self, env: &[&std::path::Path]) {
-        unsafe {
-            ffi::g_subprocess_launcher_set_environ(self.to_glib_none().0, env.to_glib_none().0);
-        }
-    }
-
     #[doc(alias = "g_subprocess_launcher_set_flags")]
     pub fn set_flags(&self, flags: SubprocessFlags) {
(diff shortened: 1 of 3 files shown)


gio/src/auto/subprocess_launcher.rs
```diff
@@ -80,7 +80,7 @@ impl SubprocessLauncher {
     }
 
     #[doc(alias = "g_subprocess_launcher_set_environ")]
-    pub fn set_environ(&self, env: &[&std::path::Path]) {
+    pub fn set_environ(&self, env: &[&std::ffi::OsStr]) {
```
sdroege, line 83: As discussed, update to use `&[OsString]` (and making this a manual binding)

gio/src/subprocess_launcher.rs
```diff
@@ -19,6 +19,13 @@ pub trait IntoRawFd: Sized {
 }
 
 impl SubprocessLauncher {
+    #[doc(alias = "g_subprocess_launcher_set_environ")]
+    pub fn set_environ(&self, env: &[std::ffi::OsString]) {
+        unsafe {
+            ffi::g_subprocess_launcher_set_environ(self.to_glib_none().0, env.to_glib_none().0);
```
sdroege, line 25: Remove the two `cfg(unit)`  at the top, then it should be OK

sdroege (approved):
Strictly speaking a breaking API change but `Path` was just completely wrong so this should be fine to backport

PR #1643, opened 2025-01-28
Signal macro

jgcodes2020:
Implements a macro for signals. This is still really rough around the edges and doesn't support things like extension traits. I also need to figure out how to write tests for this thing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -495,6 +495,7 @@ dependencies = [
 name = "glib-macros"
 version = "0.21.0"
 dependencies = [
+ "bitflags",
  "glib",
  "heck",
  "proc-macro-crate",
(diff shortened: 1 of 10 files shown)


bilelmoussaoui:
Have you seen https://github.com/gtk-rs/gtk-rs-core/pull/1577 and the comments left there?

jgcodes2020:
> Have you seen #1577 and the comments left there?

Just took a look; I'm not sure if they're still working on it. Where my implementation differs (as far as I can tell) is being able to specify no class handler using:
```rs
#[glib::signals]
impl Base {
    #[signal(run_first)]
    fn one(&self);
}
```

Without the macro, `rust-analyzer` will correctly parse the function signature before erroring saying it doesn't have a body.

bilelmoussaoui:
any plans for pushing this forward?

Issue #1649, opened 2025-02-07
[BUG] CancelledHandlerId is private


Hawk777:
**Bug description**
The type `CancelledHandlerId` is publicly visible (because `CancellableExtManual`’s `connect_foo` methods return it), but the type itself is in a private module and not re-exported anywhere, so it can’t be named.

sdroege:
Indeed, do you want to provide a PR for making it public?

Hawk777:
Sure, I think this should be sufficient.

PR #1652, opened 2025-02-10
Re-export CancelledHandlerId

Hawk777:
This type is effectively public, because it’s returned by `Cancellable::connect_cancelled`, but because it’s not re-exported it can’t be named, making it awkward to e.g. save it in a struct field.

Fixes GH-1649.


Hawk777:
I think this is ready; there is one CI check failing but it’s pointing nowhere near anything I modified and I don’t understand what it’s complaining about.

bilelmoussaoui (approved):

Issue #1663, opened 2025-02-21
[FEATURE REQUEST] Remove `unsafe` for into pointer conversion (rust edition 2024)


BiagioFesta:
The [IntoGlibPtr::into_glib_ptr](https://docs.rs/glib/latest/glib/translate/trait.IntoGlibPtr.html#tymethod.into_glib_ptr) method is currently marked as unsafe. We should consider removing this safety restriction, especially since it's part of a trait. In Rust, conversions into pointers are generally considered safe operations (it's the dereferencing of pointers that requires unsafe blocks).

While the trait consumes its input operand, potential memory leaks in Rust are still considered safe code from a language perspective.

This change would align well with the Rust 2024 edition, which requires explicit unsafe blocks for unsafe operations in FFI code (failing to do so produces a warning: https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html)

BiagioFesta:
Just found out now (sorry for not noticed this before) the same issue has been already open: https://github.com/gtk-rs/gtk-rs-core/issues/1483

PR #1666, opened 2025-02-24
Remove unsafe interface requirement for into ptr conversion

BiagioFesta:
Fixes: https://github.com/gtk-rs/gtk-rs-core/issues/1663

### For reviewers to focus on
Beside "`sed` changes", manual changes also include:
 * `impl IntoGlibPtr<*mut c_char> for GString` requires inner `unsafe` block as calling `extern "C"` functions
 * `pub unsafe trait RefCounted::into_raw` also changed removing `unsafe` requirement.
 * This is a API (breaking/major) change


sdroege (approved):

sdroege:
Thanks!

PR #1669, opened 2025-03-03
glib: Add bindings for `g_intern_string()` and `g_intern_static_string()` to `GStr`

sdroege:
@bilelmoussaoui Do you think naming (and place) makes sense like this?


bilelmoussaoui:
hmm no strong opinions here, seems good enough for me

Issue #1680, opened 2025-03-10
[BUG] pango::shape_full does not work when passing paragraph_text


LeoSchae:
**Bug description**

The wrapper `pango::shape_full` around `pango::ffi::pango_shape_full` seems to incorrectly pass the text and paragraph_text. The function `pango::shape_with_flags` seems to have the same problem.

When `use_safe` is set to true in the below code example, the image will not contain any text and the log contains:
```(process:215897): Pango-CRITICAL **: 17:28:02.640: pango_shape_internal: assertion 'paragraph_text + paragraph_length >= item_text + item_length' failed```
Calling pango::ffi::pango_shape_full directly (i.e. setting `use_safe` to false) and passing the texts and lengths as pointers fixes the problem.

Code Example (using pango, cairo and pangocairo):
```rs
fn test() {
    let s = cairo::ImageSurface::create(Format::ARgb32, 100, 100).unwrap();
    let c = cairo::Context::new(&s).unwrap();
    let pc = create_context(&c);
    let mut fd = pango::FontDescription::new();
    fd.set_size(13 * pango::SCALE);
    pc.set_font_description(Some(&fd));
    
    let text = "Hallo!";
    let glyphs = pango::itemize(&pc, text, 0, text.len() as i32, &AttrList::new(), None);
    let glyphs = glyphs.into_iter().map(|i| {
        let mut glyphs = pango::GlyphString::new();
        let item_text = &text[i.offset() as usize..(i.offset() + i.length()) as usize];
        
        let use_safe = true;
        if use_safe {
            // This branch does not produce the correct image
            pango::shape_full(item_text, Some(text), &i.analysis(), &mut glyphs);
        } else {
            // This works as expected
            unsafe {
                pango::ffi::pango_shape_full(
                    std::mem::transmute(item_text.as_ptr()),
                    item_text.len() as i32,
                    std::mem::transmute(text.as_ptr()),
                    text.len() as i32,
                    i.analysis().as_ptr(),
                    glyphs.as_ptr()
                );
            }
        }
        
        (i, glyphs)
    });
    
    for (i, mut glyphs) in glyphs {
        let font = i.analysis().font();
        c.move_to(10.0, 75.0);
        pangocairo::functions::show_glyph_string(&c, &font, &mut glyphs);
    }

    // save to png
    let mut file = File::create("test.png").unwrap();
    s.write_to_png(&mut file).unwrap();
}
```

**Remarks**

[...]

PR #1681, opened 2025-03-10
Fix bug in bindings of pango::shape_full and pango::shape_with_flags

LeoSchae:
Fix for issue #1680.

The implementation of `&str::to_glib_none()` copies the string and causes an error, as the `item_text` argument is required to be a sub sequence of paragraph_text (if paragraph_text is not null). This is no longer the case after copying both strings.

The new wrapper uses `&str::as_bytes()` to prevent the copy. Both `pango_shape_full` and `pango_shape_with_flags` do not rely on null terminators when passing a non-negative length as argument, so this does not cause undefined behavior.

### DIFF
--- a/pango/src/functions.rs
+++ b/pango/src/functions.rs
@@ -1,7 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use glib::translate::*;
-use std::ffi::c_char;
+use std::{ffi::c_char, ptr};
 
 pub use crate::auto::functions::*;
 #[cfg(feature = "v1_44")]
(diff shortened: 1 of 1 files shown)


pango/src/functions.rs
```diff
@@ -27,13 +28,15 @@ pub fn shape_full(
...
     } as i32;
-    let paragraph_text = paragraph_text.to_glib_none();
     let item_length = item_text.len() as i32;
     unsafe {
+        // The function does not take null-terminated strings when a length is provided.
+        // It also requires item_text to point to a subsequence of paragraph_text.
+        // Using to_glib_none() on &str will copy the string and cause problems.
         ffi::pango_shape_full(
-            item_text.to_glib_none().0,
+            item_text.as_bytes().to_glib_none().0 as *const c_char,
```
sdroege, line 37:
```suggestion
            item_text.as_ptr() as *const c_char,
```

seems simpler. Can you update that? Otherwise seems all correct, thanks!

LeoSchae:
I changed it to use as_ptr in all 4 cases (and made both functions use the same local variable definitions).

pango/src/functions.rs
```diff
@@ -23,17 +24,17 @@ pub fn shape_full(
...
-        Some(s) => s.len(),
-        None => 0,
-    } as i32;
-    let paragraph_text = paragraph_text.to_glib_none();
     let item_length = item_text.len() as i32;
+    let paragraph_length = paragraph_text.map(|t| t.len() as i32).unwrap_or_default();
+    let paragraph_ptr = paragraph_text.map_or(ptr::null(), |t| t.as_ptr() as *const c_char);
     unsafe {
+        // The function does not take null-terminated strings when a length is provided.
+        // It also requires item_text to point to a subsequence of paragraph_text.
```
sdroege, line 32: Should there be an assertion then that `item_text` is a subsequence of `paragraph_text`?

LeoSchae: Pango has a check for this built in, but will not crash. When calling with pointers that do not satisfy the condition it will print `Pango-CRITICAL **: 17:28:02.640: pango_shape_internal: assertion 'paragraph_text + paragraph_length >= item_text + item_length' failed`.

We could add an assertion if we want a more useful message (or want to cause a panic instead). I am unsure what the best idea is here.

sdroege: Generally I prefer assertions over `g_critical()`s where then random things happen afterwards

sdroege: But probably doesn't really matter here, nobody reviewed the Pango bindings for such things.

PR #1687, opened 2025-03-19
Port some functions to safe-io: pt I

A6GibKm:
This is the first and second commit of https://github.com/gtk-rs/gtk-rs-core/pull/871.

### DIFF
--- a/gio/src/socket.rs
+++ b/gio/src/socket.rs
@@ -1,7 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #[cfg(unix)]
-use std::os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
 #[cfg(windows)]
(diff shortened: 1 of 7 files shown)


gio/src/socket.rs
```diff
@@ -18,15 +18,18 @@ use crate::{ffi, Cancellable, Socket, SocketAddress, SocketControlMessage};
...
+    pub fn from_fd(fd: OwnedFd) -> Result<Socket, glib::Error> {
         let fd = fd.into_raw_fd();
         let mut error = ptr::null_mut();
-        let ret = ffi::g_socket_new_from_fd(fd, &mut error);
-        if error.is_null() {
-            Ok(from_glib_full(ret))
-        } else {
-            Err(from_glib_full(error))
+        unsafe {
+            let ret = ffi::g_socket_new_from_fd(fd, &mut error);
```
sdroege, line 26: Does this actually work? You'd pass the fd to the socket, but then the `OwnedFd` is dropped at the end of this scope and that would close the fd or not, leaving the socket with a closed fd.

sdroege: Same question also elsewhere, so let's figure this out before more review :)

A6GibKm: We use https://doc.rust-lang.org/std/os/fd/trait.IntoRawFd.html#tymethod.into_raw_fd right before this call. This should be fine.

sdroege (approved):

Issue #1592, opened 2024-12-09
Papercuts in handle_local_options's return parameter


A6GibKm:
The virtual method `handle_local_options` returns a `glib::ExitCode`, it is special cased so that the application exits when the return value is positive (e.g. ExitCode::FAILUE, ExitCode::SUCCESS) and continues to run when when negative. 

Returning `(-1).into()` is an option, but it is not very nice.

sdroege:
@A6GibKm Want to provide a PR for this? Docs say this:

```
   * Returns: an exit code. If you have handled your options and want
   * to exit the process, return a non-negative option, 0 for success,
   * and a positive value for failure. To continue, return -1 to let
   * the default option processing continue.
```

Now this just needs to be wrapped into meaningful types. All negative values are the same (default processing), 0 is success but no default processing, all values > 0 are errors and returned like that

A6GibKm:
What I dont follow is when it is actually a return code? I will test later if 0 is a return code. 

It could be an enum

```
LocalOptionsHandling {
    ExitCode(glib::ExitCode),  //  >= 0
    DefaultProcessing,  // -1
}
```

or something like that.

pjungkamp:
@A6GibKm Can we mark this fixed now?

sdroege:
I think so. If something is still missing, a more specific issue for that would be useful.

PR #1694, opened 2025-03-30
gio: Return `glib::ExitCode` from `gio::Application` signal handlers

### LINKED ISSUE #1691: [FEATURE REQUEST] `gio::Application` signal handlers should return `glib::ExitCode` instead of `i32`
The return value for `gio::Application` signal handler like the closures attached using `connect_command_line` or `connect_handle_local_options` describe an `glib::ExitCode` but actually return an `i32`. I find myself writing `glib::ExitCode::SUCCESS.value()` in these signal handlers, which doesn't seem to be ideal...

pjungkamp:
Add a manual override to use `glib::ExitCode` instead of the gir generated `i32` return value for `connect_command_line` and `connect_handle_local_options`.

This fixes #1691

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -310,6 +310,14 @@ generate_builder = true
     name = "open"
     manual = true
     doc_trait_name = "ApplicationExtManual"
+    [[object.signal]]
+    name = "command-line"
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.signal]]
+    name = "handle-local-options"
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
     [[object.function]]
     name = "run"
     manual = true
--- a/gio/README.md
+++ b/gio/README.md
@@ -6,7 +6,7 @@ GIO __2.56__ is the lowest supported version for the underlying library.
 
 ## Minimum supported Rust version
 
-Currently, the minimum supported Rust version is `1.80.0`.
+Currently, the minimum supported Rust version is `1.83.0`.
 
 ## Documentation
 
--- a/gio/src/application.rs
+++ b/gio/src/application.rs
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::{boxed::Box as Box_, mem::transmute};
+use std::{boxed::Box as Box_, mem::transmute, ops::ControlFlow};
 
 use glib::{
     prelude::*,
@@ -9,13 +9,14 @@ use glib::{
     ExitCode, GString,
 };
 
-use crate::{ffi, Application, File};
+use crate::{ffi, Application, ApplicationCommandLine, File};
 
 pub trait ApplicationExtManual: IsA<Application> {
     #[doc(alias = "g_application_run")]
     fn run(&self) -> ExitCode {
         self.run_with_args(&std::env::args().collect::<Vec<_>>())
     }
+
     #[doc(alias = "g_application_run")]
     fn run_with_args<S: AsRef<str>>(&self, args: &[S]) -> ExitCode {
         let argv: Vec<&str> = args.iter().map(|a| a.as_ref()).collect();
@@ -23,8 +24,10 @@ pub trait ApplicationExtManual: IsA<Application> {
         let exit_code = unsafe {
             ffi::g_application_run(self.as_ref().to_glib_none().0, argc, argv.to_glib_none().0)
         };
-        ExitCode::from(exit_code)
+        ExitCode::try_from(exit_code).unwrap()
     }
+
+    #[doc(alias = "open")]
     fn connect_open<F: Fn(&Self, &[File], &str) + 'static>(&self, f: F) -> SignalHandlerId {
         unsafe extern "C" fn open_trampoline<P, F: Fn(&P, &[File], &str) + 'static>(
             this: *mut ffi::GApplication,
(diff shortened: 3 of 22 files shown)


gio/src/application.rs
```diff
@@ -25,6 +26,8 @@ pub trait ApplicationExtManual: IsA<Application> {
         };
         ExitCode::from(exit_code)
     }
+
+    #[doc(alias = "open")]
```
pjungkamp, line 30: I added this alias on `connect_open` and some spacing between functions as a little cleanup.

gio/Gir.toml
```diff
@@ -310,6 +310,14 @@ generate_builder = true
...
     manual = true
     doc_trait_name = "ApplicationExtManual"
+    [[object.signal]]
+    name = "command-line"
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
+    [[object.signal]]
+    name = "handle-local-options"
+    manual = true
+    doc_trait_name = "ApplicationExtManual"
```
bilelmoussaoui, line 320: i don't think the doc_trait_name is needed anymore

bilelmoussaoui (dismissed):

gio/Gir.toml
pjungkamp: I just followed the style of the `[[object.signal]]` override for the "open" just above these. Should I remove the `doc_trait_name` for that one too?

gio/src/application.rs
```diff
@@ -56,6 +59,72 @@ pub trait ApplicationExtManual: IsA<Application> {
...
+                Some(std::mem::transmute::<*const (), unsafe extern "C" fn()>(
+                    command_line_trampoline::<Self, F> as *const (),
+                )),
+                Box_::into_raw(f),
+            )
+        }
+    }
+
+    #[doc(alias = "handle-local-options")]
+    fn connect_handle_local_options<F: Fn(&Self, &glib::VariantDict) -> ExitCode + 'static>(
```
sdroege, line 96: See https://github.com/gtk-rs/gtk-rs-core/issues/1592

Also CC @A6GibKm

pjungkamp: Could we add another associated constant to `glib::ExitCode`, e.g. `glib::ExitCode::CONTINUE`? I dont feel comfortable with an `enum` or e.g. `Option` because `Some(glib::ExitCode::from(-1))` and `None` would be two representations for the effectively the same value.

pjungkamp: I just checked the documentation for the return value of the `handle-local-options` signal.

> An exit code. If you have handled your options and want to exit the process, return a non-negative option, 0 for success, and a positive value for failure. To continue, return -1 to let the default option processing continue.

It seems like the definition of `glib::ExitCode` should be:

- positive means failure
- zero means success
- negative means special value

This means that I'm really not comfortable with the implementation of `Termination` for `glib::ExitCode`. Any value that does not fit into `u8` shoudl probably be mapped to e.g. `u8::MAX`.

I've pushed an additional commit that adds a `glib::ExitCode::CONTINUE` associated constant and some initial documentation for the `glib::ExitCode` struct.

gio/Gir.toml
pjungkamp: I think it's still used to generate the documentation on the `connect_...` functions in `ApplicationExtManual`. It's still used here: https://github.com/gtk-rs/gir/blob/491114ad76bde3d0c2c30312638782c6638c00c2/src/codegen/doc/mod.rs#L537-L561

gio/src/application.rs
A6GibKm: > It seems like the definition of glib::ExitCode should be:

That proposal only makes sense if the negative values have a meaning whenever they are used. As far as I have seen it is only a handful of fns that do something special with negative values.

sdroege: Like @A6GibKm says, the main problem here is that `handle_local_options()` is special. It should probably have a different type than the other cases.

pjungkamp: I wouldn't like to have multiple representations for the same handling. If we'd use `Option<glib::ExitCode>` there shouldn't be a way to say `Some(glib::ExitCode::from(-1))` and `None` which would lead to the same effect.

We could constrain the `glib::ExitCode` type though.

- Remove `impl From<i32> for glib::ExitCode`
- Add `impl From<u8> for glib::ExitCode` instead
- Add `impl TryFrom<i32> for glib::ExitCode`
- Add `struct InvalidExitCode(i32)` as the `TryFrom::Error` type

We could then modify interfaces that currently return a `glib::ExitCode` which may originate from foreign code, e.g. `gio::ApplicationExtManual::run` to return a `Result<glib::ExitCode, glib::InvalidExitCode>` instead. While we'd enforce good exit codes from Rust code through the traits and conversion methods. Using `Option<glib::ExitCode>` or `std::ops::ControlFlow<(), glib::ExitCode>` would then be fine for `handle-local-options`.

pjungkamp:
I've now changed the `handle-local-options` signal and `local-command-line` virtual function to use `std::ops::ControlFlow<ExitCode>`. `ExitCode` now contains a `u8` and only implements `TryFrom<i32>` instead of `From<i32>`.

One benefit of `std::ops::ControlFlow<ExitCode>` over `Option<ExitCode>` is the handling with the `?` operator.

This required me to bump the MSRV to `1.83`.

#### Using `Option<ExitCode>`

```rs
impl ApplicationExt for MyApplication {
    fn local_command_line(&self, arguments: &mut ArgumentList) -> Option<ExitCode> {
        // The `?` operator won't work here since it returns on `None` 
        if let Some(exit_code) = self.parent_local_command_line() {
            return Some(exit_code);
        }
        
        // Handle my own options
        
        None // Continue with default handling
    }
}
```

#### Using `ControlFlow<ExitCode>`

```rs
impl ApplicationExt for MyApplication {
    fn local_command_line(&self, arguments: &mut ArgumentList) -> ControlFlow<ExitCode> {
        // This will return if the parent class has handled the entire command line. 
        self.parent_local_command_line()?;
        
        // Handle my own options
        
        ControlFlow::Continue(()) // Continue with default handling
    }
}
```

glib/src/exit_code.rs
```diff
@@ -1,33 +1,87 @@
...
 
-impl From<ExitCode> for i32 {
-    fn from(value: ExitCode) -> Self {
-        value.0
+impl TryFrom<i32> for ExitCode {
+    type Error = InvalidExitCode;
+
+    fn try_from(code: i32) -> Result<Self, Self::Error> {
+        match code {
+            ExitCode::MIN..ExitCode::MAX => Ok(Self(code as _)),
```
sdroege, line 63:
```suggestion
            ExitCode::MIN..=ExitCode::MAX => Ok(Self(code as _)),
```

gio/README.md
```diff
@@ -6,7 +6,7 @@ GIO __2.56__ is the lowest supported version for the underlying library.
 
 ## Minimum supported Rust version
 
-Currently, the minimum supported Rust version is `1.80.0`.
+Currently, the minimum supported Rust version is `1.83.0`.
```
sdroege, line 9: Which part exactly needs the MSRV bump?

gio/src/application.rs
sdroege: I think this implementation makes sense. @A6GibKm what do you think?

gio/README.md
pjungkamp: The `ControlFlow::break_value` function, Rust 1.83 adds some functions to make `ControlFlow` more convenient.

sdroege: That's a simple match so we could also just not use it here. Updating the MSRV seems fine to me too though.

@bilelmoussaoui what do you think?

sdroege: Ok, then just waiting for @A6GibKm in the other thread, then I'll merge this.

glib/src/exit_code.rs
sdroege: Well, and this one needs taken care of too

sdroege:
Let's get this in then

sdroege (approved):

sdroege (approved):

sdroege:
That sounds good to me. Do you want to provide a PR @pjungkamp ?

PR #1696, opened 2025-03-30
 Port some functions to safe-io: pt II

A6GibKm:
Continuation of https://github.com/gtk-rs/gtk-rs-core/pull/1687. This includes up to the 5th commit of https://github.com/gtk-rs/gtk-rs-core/pull/871.

### DIFF
--- a/gio/src/unix_fd_list.rs
+++ b/gio/src/unix_fd_list.rs
@@ -1,12 +1,12 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #[cfg(unix)]
-use std::os::unix::io::{AsRawFd, IntoRawFd, RawFd};
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
 use std::{mem, ptr};
 
 use glib::{prelude::*, translate::*};
 #[cfg(all(not(unix), docsrs))]
-use socket::{AsRawFd, IntoRawFd, RawFd};
+use socket::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd};
 
 use crate::{ffi, UnixFDList};
 
@@ -25,12 +25,12 @@ impl UnixFDList {
 
 pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
     #[doc(alias = "g_unix_fd_list_append")]
-    fn append<T: AsRawFd>(&self, fd: T) -> Result<i32, glib::Error> {
+    fn append(&self, fd: impl AsFd) -> Result<i32, glib::Error> {
         unsafe {
             let mut error = ptr::null_mut();
             let ret = ffi::g_unix_fd_list_append(
                 self.as_ref().to_glib_none().0,
-                fd.as_raw_fd(),
+                fd.as_fd().as_raw_fd(),
                 &mut error,
             );
             if error.is_null() {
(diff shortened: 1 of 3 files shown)


gio/src/unix_fd_list.rs
```diff
@@ -42,39 +42,41 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
...
 
     #[doc(alias = "g_unix_fd_list_get")]
-    fn get(&self, index_: i32) -> Result<RawFd, glib::Error> {
+    fn get(&self, index_: i32) -> Result<OwnedFd, glib::Error> {
         unsafe {
             let mut error = ptr::null_mut();
-            let ret = ffi::g_unix_fd_list_get(self.as_ref().to_glib_none().0, index_, &mut error);
+            let raw_fd =
+                ffi::g_unix_fd_list_get(self.as_ref().to_glib_none().0, index_, &mut error);
+            let fd = OwnedFd::from_raw_fd(raw_fd);
```
sdroege, line 50: This needs to be moved into the `error.is_null()` branch or otherwise you might create an `OwnedFd` out of `-1`

gio/src/unix_fd_list.rs
```diff
@@ -42,39 +42,41 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
...
                 length.assume_init() as usize,
-            );
-            ret
+            )
         }
     }
 
     #[doc(alias = "g_unix_fd_list_steal_fds")]
-    fn steal_fds(&self) -> Vec<RawFd> {
+    fn steal_fds(&self) -> Vec<OwnedFd> {
```
sdroege, line 72: I think this invalidates the return value of `peek_fds()`, so you'd be left with a dangling pointer

A6GibKm: Would it work to make peek_fds return a RawFd (again) and mark it as unsafe?

sdroege: Doesn't have to be `unsafe` but yes, returning `Vec<RawFd>` seems like the only sensible choice here

A6GibKm: Done.

gio/src/unix_fd_list.rs
A6GibKm: Done.

gio/src/unix_fd_list.rs
```diff
@@ -42,27 +42,29 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
...
-    fn peek_fds(&self) -> Vec<RawFd> {
+    fn peek_fds(&self) -> &[BorrowedFd<'_>] {
         unsafe {
             let mut length = mem::MaybeUninit::uninit();
-            let ret = FromGlibContainer::from_glib_none_num(
-                ffi::g_unix_fd_list_peek_fds(self.as_ref().to_glib_none().0, length.as_mut_ptr()),
+            std::slice::from_raw_parts(
+                ffi::g_unix_fd_list_peek_fds(self.as_ref().to_glib_none().0, length.as_mut_ptr())
+                    as *const BorrowedFd,
                 length.assume_init() as usize,
```
sdroege, line 66:
```suggestion
                length.as_mut_ptr(),
```

This could've never worked before :)

gio/src/unix_fd_list.rs
```diff
@@ -42,27 +42,29 @@ pub trait UnixFDListExtManual: IsA<UnixFDList> + Sized {
...
+                Ok(fd)
             } else {
                 Err(from_glib_full(error))
             }
         }
     }
 
     #[doc(alias = "g_unix_fd_list_peek_fds")]
-    fn peek_fds(&self) -> Vec<RawFd> {
+    fn peek_fds(&self) -> &[BorrowedFd<'_>] {
```
sdroege, line 60: This must return a `Vec<RawFd>` too because this type here has inner mutability...

gio/src/unix_fd_list.rs
sdroege: Nevermind.

PR #1699, opened 2025-04-05
gio: Use `OwnedSocket` / `BorrowedSocket` for Windows APIs on GSocket

sdroege:
CC @A6GibKm

### DIFF
--- a/cairo/src/surface.rs
+++ b/cairo/src/surface.rs
@@ -78,14 +78,16 @@ impl Surface {
     #[doc(alias = "cairo_surface_get_mime_data")]
     #[doc(alias = "get_mime_data")]
     pub fn mime_data(&self, mime_type: &str) -> Option<Vec<u8>> {
-        let data_ptr: *mut u8 = ptr::null_mut();
+        let mut data_ptr: *mut u8 = ptr::null_mut();
         let mut length: c_ulong = 0;
+        // The function actually needs a mutable pointer
+        #[allow(clippy::unnecessary_mut_passed)]
         unsafe {
             let mime_type = CString::new(mime_type).unwrap();
             ffi::cairo_surface_get_mime_data(
                 self.to_raw_none(),
                 mime_type.as_ptr(),
-                &data_ptr,
+                &mut data_ptr,
                 &mut length,
             );
             if !data_ptr.is_null() && length != 0 {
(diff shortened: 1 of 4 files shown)


sdroege:
CI failure is a false positive compiler error with latest nightly...

gio/src/socket.rs
```diff
@@ -27,22 +29,24 @@ impl Socket {
             if error.is_null() {
                 Ok(from_glib_full(ret))
             } else {
-                libc::close(fd);
+                let _ = OwnedFd::from_raw_fd(fd);
```
A6GibKm, line 32: Ultimately, `Drop` on OwnedFd does libc::close, no strong opinion though.

https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#182

gio/src/socket.rs
```diff
@@ -27,22 +29,24 @@ impl Socket {
...
                 Err(from_glib_full(error))
             }
         }
     }
     #[cfg(windows)]
     #[cfg_attr(docsrs, doc(cfg(windows)))]
-    #[allow(clippy::missing_safety_doc)]
-    pub unsafe fn from_socket(socket: impl IntoRawSocket) -> Result<Socket, glib::Error> {
+    pub fn from_socket(socket: OwnedSocket) -> Result<Socket, glib::Error> {
         let socket = socket.into_raw_socket();
```
A6GibKm, line 40: Did we implement `Into<OwnedFd>` for Socket?

gio/src/socket.rs
sdroege: It does, yes. But it seems more symmetric to close it via `OwnedFd` if we got it from there :)

gio/src/socket.rs
sdroege: No, we can't really. We can't steal the fd from the socket.

cairo/src/surface.rs
```diff
@@ -78,14 +78,14 @@ impl Surface {
...
-        let data_ptr: *mut u8 = ptr::null_mut();
+        let mut data_ptr: *mut u8 = ptr::null_mut();
         let mut length: c_ulong = 0;
         unsafe {
             let mime_type = CString::new(mime_type).unwrap();
             ffi::cairo_surface_get_mime_data(
                 self.to_raw_none(),
                 mime_type.as_ptr(),
-                &data_ptr,
+                &mut data_ptr,
```
sdroege, line 88: Actually not a false positive. That code was completely broken

sdroege:
CI finally passing

gdk-pixbuf/src/pixbuf.rs
```diff
@@ -35,7 +35,7 @@ impl Pixbuf {
         let bits_per_sample = bits_per_sample as usize;
 
         let n_channels = if has_alpha { 4 } else { 3 };
-        let last_row_len = width * ((n_channels * bits_per_sample + 7) / 8);
+        let last_row_len = width * (n_channels * bits_per_sample).div_ceil(8);
```
A6GibKm, line 38: This is missing a 7 isn't it?

sdroege: No, why?

A6GibKm: There was a +7 before and that change seemed unrelated to clippy fixes.

sdroege: That `(x + 7) / 8` is the same as dividing by 8 and rounding up (which is what `div_ceil()` does)

Issue #1512, opened 2024-09-22
[FEATURE REQUEST] IntoFuture for gio::Cancellable


felinira:
Could we implement [IntoFuture](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) for `&IsA<gio::Cancellable>` or at least for `&gio::Cancellable`? It's the standardized way to convert something into a future. We could even deprecate `.future()` in the former case.

It would allow to await cancellable directly, and not have to call a future method.

Maybe there are also other objects that could use an IntoFuture implementation, could be worth it to have a look.

sdroege:
Sure, why not? Want to make a PR?

PR #1709, opened 2025-04-27
Wrap IntoFuture around Cancellable::future

swsnr:
Implement `IntoFuture` for `Cancellable` as a dead-simple wrapper around the existing `Cancellable::future`.

See #1512


swsnr:
I missed this, and found #1512 tracking this.

Opening as draft, because #1512 suggests to make `future` obsolete in favour of `IntoFuture` for `&Cancellable`.  Should I add that impl as well, and deprecate or remove `Cancellable::future`?

sdroege (commented):
Like this seems OK for now. We can deprecate it at a later time when this has proven itself.

swsnr:
@sdroege Okay, I added an impl for `&Cancellable`, and marked it as ready.

sdroege (approved):

PR #1713, opened 2025-05-03
Implement gio::File::set_attribute

fbrouille:


gio/src/file.rs
```diff
@@ -1111,6 +1113,98 @@ pub trait FileExtManual: IsA<File> + Sized {
 
         (fut, Box::pin(receiver))
     }
+    #[doc(alias = "g_file_set_attribute")]
+    fn set_attribute<T>(
+        &self,
+        attribute: &str,
+        type_: FileAttributeType,
+        value: FileAttributeTypeValue<T>,
```
sdroege, line 1121: There should be type-system matching from the file attribute to the value type probably. If you e.g. pass a u32 where a u64 was expected this will read random stack memory.

gio/src/file.rs
```diff
@@ -1111,6 +1113,98 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+
+impl From<u32> for FileAttributeTypeValue<u32> {
+    fn from(value: u32) -> Self {
+        Self(&value as *const _ as _, value)
+    }
+}
+
+impl From<i32> for FileAttributeTypeValue<i32> {
+    fn from(value: i32) -> Self {
+        Self(&value as *const _ as _, value)
```
sdroege, line 1196: This and the others are unsound. You store a pointer to some random, temporary stack frame of this function in the `FileAttributeValue`.

gio/src/file.rs
```diff
@@ -1111,6 +1113,137 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+                FileAttributeTypeValueStorage::Object(ToGlibPtr::<
+                    <glib::Object as glib::translate::GlibPtrDefault>::GlibType,
+                >::to_glib_none(value))
+            }
+            Self::Stringv(value) => FileAttributeTypeValueStorage::Stringv(
+                ToGlibContainerFromSlice::<*mut *mut libc::c_char>::to_glib_none_from_slice(value)
+                    .1,
+            ),
+        };
+        Stash(s.as_ptr(), s)
```
sdroege, line 1207: Here you get the pointer into `s` (stored on this functions stack), then return (i.e. move) `s` and afterwards the pointer points to this stack frame still while `s` is somewhere else

gio/src/file.rs
```diff
@@ -1111,6 +1113,137 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+        }
+    }
 }
 
 impl<O: IsA<File>> FileExtManual for O {}
+
+#[derive(Debug)]
+// checker-ignore-item
+pub enum FileAttributeTypeValue<'a> {
+    Pointer(FileAttributeType, glib::ffi::gpointer),
```
sdroege, line 1151: This allows creating values of arbitrary pointers from safe code, which is probably not a good idea

gio/src/file.rs
```diff
@@ -1111,6 +1113,137 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+
+impl FileAttributeTypeValueStorage<'_> {
+    pub fn as_ptr(&self) -> glib::ffi::gpointer {
+        match self {
+            Self::Pointer(s) => *s,
+            Self::String(s) => s.as_ptr() as _,
+            Self::ByteString(s) => s.as_ptr() as _,
+            Self::Boolean(s) => s as *const _ as _,
+            Self::Uint32(s) => *s as *const _ as _,
+            Self::Int32(s) => *s as *const _ as _,
```
sdroege, line 1242: Some more types instead of `_` would make it clearer whether this is correct or not

```suggestion
            Self::Int32(s) => *s as *const i32 as _,
```

if that doesn't compile then the code is wrong

gio/src/file.rs
```diff
@@ -1111,6 +1113,137 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+            }
+        }
+    }
 }
 
 impl<O: IsA<File>> FileExtManual for O {}
+
+#[derive(Debug)]
+// checker-ignore-item
+pub enum FileAttributeTypeValue<'a> {
```
sdroege, line 1150: I think I would gone with a trait but this also works. Should probably have `From<T>` impls to easily convert from plain types to this

gio/src/file.rs
```diff
@@ -1111,6 +1113,137 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+    Pointer(FileAttributeType, glib::ffi::gpointer),
+    String(&'a str),
+    ByteString(&'a str),
```
sdroege, line 1153: Isn't this supposed to be a `&[u8]`?

gio/src/file.rs
sdroege: Could then also use `impl Into<FileAttributeTypeValue>` in the function.

Also I wonder if the same type could be used for the other file attribute API

bilelmoussaoui:
i guess you could also easily add https://docs.gtk.org/gio/method.FileInfo.set_attribute.html ?

gio/src/file.rs
```diff
@@ -1111,6 +1113,177 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+                Err(from_glib_full(error))
+            }
+        }
+    }
 }
 
 impl<O: IsA<File>> FileExtManual for O {}
+
+#[derive(Debug)]
+// checker-ignore-item
```
bilelmoussaoui, line 1149: what is this?

gio/src/file.rs
```diff
@@ -1111,6 +1113,177 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+pub struct FileAttributeTypeValue<'a>(FileAttributeTypeValueInner<'a>);
+
+impl<'a> FileAttributeTypeValue<'a> {
+    // new_pointer is private to gio crate to not allow creating values of arbitrary pointers from safe code
+    #[allow(dead_code)] // TODO remove this allow attribute when new_pointer will be used by this crate
+    pub(crate) fn new_pointer(type_: FileAttributeType, ptr: glib::ffi::gpointer) -> Self {
+        Self(FileAttributeTypeValueInner::Pointer(type_, ptr))
+    }
```
bilelmoussaoui, line 1157: i would remove this and move the dead_code to the enum if clippy complains, can be added later if needed

gio/src/file.rs
sdroege: Not very different from `object.set_property()` and other such API though

gio/src/file.rs
sdroege: It's actually a NUL-terminated string of some sort, not a literal byte string where there can be 0-bytes in the middle?

gio/src/file.rs
```diff
@@ -1111,6 +1113,171 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+                Err(from_glib_full(error))
+            }
+        }
+    }
 }
 
 impl<O: IsA<File>> FileExtManual for O {}
+
+#[derive(Debug)]
+pub struct FileAttributeTypeValue<'a>(FileAttributeTypeValueInner<'a>);
```
sdroege, line 1149:
```suggestion
pub struct FileAttributeValue<'a>(FileAttributeValueInner<'a>);
```

the type seems a bit redundant to me, but no strong opinion

gio/src/file.rs
```diff
@@ -1111,6 +1113,171 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+
+    pub fn new_int64(value: i64) -> Self {
+        Self(FileAttributeTypeValueInner::Int64(value))
+    }
+
+    pub fn new_object<T: AsRef<glib::Object>>(value: &'a T) -> Self {
+        Self(FileAttributeTypeValueInner::Object(value.as_ref()))
+    }
+
+    pub fn new_stringv(value: &'a Vec<&'a str>) -> Self {
```
sdroege, line 1184: There's also `glib::collections::StrV` or `&[&str]` or so that would work here

gio/src/file.rs
```diff
@@ -1111,6 +1113,151 @@ pub trait FileExtManual: IsA<File> + Sized {
...
+    fn set_attribute<'a>(
+        &self,
+        attribute: &str,
+        value: impl Into<FileAttributeValue<'a>>,
+        flags: FileQueryInfoFlags,
+        cancellable: Option<&impl IsA<Cancellable>>,
+    ) -> Result<(), glib::Error> {
+        unsafe {
+            let mut error = std::ptr::null_mut();
+            let value: FileAttributeValue<'a> = value.into();
```
sdroege, line 1127:
```suggestion
            let value = FileAttributeValue::from(value);
```

seems clearer to me

gio/src/file.rs
```diff
@@ -1111,6 +1113,151 @@ pub trait FileExtManual: IsA<File> + Sized {
...
 
 impl<O: IsA<File>> FileExtManual for O {}
+
+#[derive(Debug)]
+pub struct FileAttributeValue<'a>(FileAttributeValueInner<'a>);
+
+impl From<&str> for FileAttributeValue<'_> {
+    fn from(value: &str) -> Self {
+        Self(FileAttributeValueInner::String(
+            ToGlibPtr::<*mut libc::c_char>::to_glib_none(value).1,
```
sdroege, line 1155: This creates a dangling pointer. `ToGlibPtr` creates a new temporary NUL-terminated string that is then freed when the `Stash` is dropped

sdroege: You could e.g. internally store a `glib::GString` instead and on `as_ptr()` use that directly

sdroege: Oh you're actually storing the `Stash` here. `.1` and not `.0`. Nevermind then :)

sdroege (dismissed):
Looks good to me then, is there any other place where this new type can be used?

bilelmoussaoui:
> i guess you could also easily add https://docs.gtk.org/gio/method.FileInfo.set_attribute.html ?

@sdroege see ^

gio/src/file.rs
```diff
@@ -1,13 +1,13 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use std::{cell::RefCell, ffi::CStr, mem, pin::Pin, ptr};
+use std::{cell::RefCell, mem, pin::Pin, ptr};
```
bilelmoussaoui, line 3: Can you add a closes  https://github.com/gtk-rs/gtk-rs-core/issues/40 to this commit?

gio/src/file_attribute_value.rs
```diff
@@ -0,0 +1,126 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
```
bilelmoussaoui, line 1: This should be done this way in the first commit, to avoid the noise in this commit.

sdroege (approved):

sdroege:
If you rebase this on top of https://github.com/gtk-rs/gtk-rs-core/pull/1724 (i.e. latest `main`) then the CI should pass

sdroege:
Actually let me just do that now.

PR #1721, opened 2025-05-19
Implement `glib_build_tools::compile_schemas`

shahradelahi:


sdroege:
Please squash the commits and provide a meaningful commit message (and PR description) for what this is actually doing.

glib-build-tools/src/lib.rs
```diff
@@ -57,3 +62,67 @@ pub fn compile_resources<P: AsRef<Path>>(source_dirs: &[P], gresource: &str, tar
...
+pub fn compile_schemas(schemas_dir: &[&str]) {
+    // Target Directory
+    //  Unix - $HOME/.local/share/glib-2.0/schemas/
+    //  Windows - C:/ProgramData/glib-2.0/schemas/
+
+    let prefix = if cfg!(windows) {
+        PathBuf::from("C:/ProgramData")
+    } else {
+        glib::user_data_dir()
+    };
```
bilelmoussaoui, line 85: This hardcoding the prefix to use is meh. Either the prefix will have to be passed as a param or this function won't land as is.

glib-build-tools/src/lib.rs
```diff
@@ -57,3 +62,73 @@ pub fn compile_resources<P: AsRef<Path>>(source_dirs: &[P], gresource: &str, tar
...
+
+// rustdoc-stripper-ignore-next
+/// Call to run `glib-compile-schemas` to generate compiled gschemas from `.gschema.xml` schemas
+/// files from specified directories and store `gschemas.compiled` into `glib-2.0` schema cache
+/// directory.
+///
+/// ```no_run
+/// glib_build_tools::compile_schemas(
+///     &["schemas"],
+///     None
```
bilelmoussaoui, line 74: Please document the default target_dir.

glib-build-tools/src/lib.rs
```diff
@@ -57,3 +62,73 @@ pub fn compile_resources<P: AsRef<Path>>(source_dirs: &[P], gresource: &str, tar
...
+pub fn compile_schemas(schemas_dir: &[&str], target_dir: Option<&str>) {
+    // Target Directory
+    //  Unix - $HOME/.local/share/glib-2.0/schemas/
+    //  Windows - C:/ProgramData/glib-2.0/schemas/
+
+    let target_dir = match target_dir {
+        Some(base) => PathBuf::from(base),
+        None => {
+            let prefix = if cfg!(windows) {
+                PathBuf::from("C:/ProgramData")
```
bilelmoussaoui, line 86: Why a hardcoded path on windows?

shahradelahi: I looked everywhere, but I couldn’t find any information on where the target directory is on Windows. The paths are based on the gtk-rs.org documentation.

https://gtk-rs.org/gtk4-rs/stable/latest/book/settings.html
https://man.archlinux.org/man/core/glib2/glib-compile-schemas.1.en

sdroege: Should also be `g_get_user_data_dir()` on Windows too or not?

sdroege: Also there's the extra complication that on Windows IIRC these things are relative to the executable (or library?) path. This needs some further investigation in any case.

glib-build-tools/src/lib.rs
```diff
@@ -57,3 +62,73 @@ pub fn compile_resources<P: AsRef<Path>>(source_dirs: &[P], gresource: &str, tar
...
+            } else {
+                glib::user_data_dir()
+            };
+
+            Path::new(&prefix).join("glib-2.0").join("schemas")
+        }
+    };
+
+    // Ensure target_dir exists
+    std::fs::create_dir_all(&target_dir).expect("Failed to create target directory");
```
sdroege, line 96: Installing things as part of `build.rs` is an anti-pattern and unexpected. You should not place (or modify) anything outside of `OUT_DIR`, see https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

glib-build-tools/src/lib.rs
```diff
@@ -57,3 +57,64 @@ pub fn compile_resources<P: AsRef<Path>>(source_dirs: &[P], gresource: &str, tar
...
+
+    let target_dir = out_dir.join("gschemas");
+
+    // Ensure target_dir exists
+    std::fs::create_dir_all(&target_dir).expect("Failed to create target directory");
+
+    println!(
+        "cargo:rustc-env=GSETTINGS_SCHEMA_DIR={}",
+        target_dir.to_str().unwrap()
+    );
```
shahradelahi, line 82: https://docs.gtk.org/gio/overview.html#running-gio-applications

> GSETTINGS_SCHEMA_DIR. This variable can be set to the names of directories to consider when looking for compiled schemas for [GSettings](https://docs.gtk.org/gio/class.Settings.html), in addition to the glib-2.0/schemas subdirectories of the XDG system data directories. To specify multiple directories, use G_SEARCHPATH_SEPARATOR_S as a separator.

glib-build-tools/src/lib.rs
```diff
@@ -65,36 +60,27 @@ pub fn compile_resources<P: AsRef<Path>>(source_dirs: &[P], gresource: &str, tar
...
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let out_dir = Path::new(&out_dir);
+
+    let target_dir = out_dir.join("gschemas");
 
     // Ensure target_dir exists
     std::fs::create_dir_all(&target_dir).expect("Failed to create target directory");
 
+    println!(
+        "cargo:rustc-env=GSETTINGS_SCHEMA_DIR={}",
```
sdroege, line 80: How would this be used?

glib-build-tools/src/lib.rs
sdroege: But this is passing the environment variable to rustc, it doesn't make your application run with it

sdroege: I don't think this makes sense really. The correct way for all this is to install the schemas in the right location and the environment variable is meant for uninstalled usage or special setups.

cargo has no mechanism for installing anything but an executable, so the usual solution for this is to put another, proper build system around cargo. E.g. having a Makefile (which hopefully does things correctly enough), or using meson or cmake.

I don't think what you're trying to achieve here (installing schemas) is possible with cargo in its current form but would need more features in cargo first.

PR #1725, opened 2025-05-21
Add gio::FileMonitor subclass

fbrouille:
Allow custom implementation of gio::FileMonitor

### DIFF
--- a/gio/src/subclass/file_monitor.rs
+++ b/gio/src/subclass/file_monitor.rs
@@ -1,0 +1,254 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{prelude::*, subclass::prelude::*, translate::*};
+
+use crate::{ffi, File, FileMonitor, FileMonitorEvent};
+
+// Support custom implementation of virtual functions defined in `gio::ffi::GFileMonitorClass`.
+pub trait FileMonitorImpl: ObjectImpl + ObjectSubclass<Type: IsA<FileMonitor>> {
+    fn changed(&self, file: &File, other_file: Option<&File>, event_type: FileMonitorEvent) {
+        self.parent_changed(file, other_file, event_type)
+    }
+
+    fn cancel(&self) {
+        self.parent_cancel()
+    }
+}
+
+// Support parent implementation of virtual functions defined in `gio::ffi::GFileMonitorClass`.
+pub trait FileMonitorImplExt: FileMonitorImpl {
+    fn parent_changed(&self, file: &File, other_file: Option<&File>, event_type: FileMonitorEvent) {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GFileMonitorClass;
+
+            if let Some(f) = (*parent_class).changed {
+                f(
+                    self.obj().unsafe_cast_ref::<FileMonitor>().to_glib_none().0,
+                    file.to_glib_none().0,
+                    other_file.to_glib_none().0,
+                    event_type.into_glib(),
+                );
+            }
+        }
+    }
+
+    fn parent_cancel(&self) {
+        unsafe {
+            let data = Self::type_data();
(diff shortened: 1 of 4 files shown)


gio/Gir.toml
```diff
@@ -916,6 +915,17 @@ status = "generate"
     # use strv
     manual = true
 
+[[object]]
+name = "Gio.FileMonitor"
+status = "generate"
+    [[object.function]]
+    name = "emit_event"
+        [[object.function.parameter]]
+        name = "other_file"
```
bilelmoussaoui, line 924: this should be set as nullable upstream

gio/tests/file_monitor.rs
```diff
@@ -0,0 +1,161 @@
...
+glib::wrapper! {
+    pub struct MyFileMonitor(ObjectSubclass<imp::MyFileMonitor>) @extends FileMonitor;
+}
+
+pub trait MyFileMonitorImpl:
+    ObjectImpl + ObjectSubclass<Type: IsA<MyFileMonitor> + IsA<FileMonitor>>
+{
+}
+
+unsafe impl<T: MyFileMonitorImpl + FileMonitorImpl> IsSubclassable<T> for MyFileMonitor {}
```
bilelmoussaoui, line 58: those bits are not useful, no?

gio/Gir.toml
bilelmoussaoui: I triggered the process chain to get new GIR files & changes in this repo, you can rebase once that is over.

gio/tests/file_monitor.rs
bilelmoussaoui: Where in the test that is needed?

gio/src/subclass/file_monitor.rs
```diff
@@ -0,0 +1,56 @@
...
+        let res = self.parent_cancel();
+        // cancel should always return true as specified in documentation
+        // https://docs.gtk.org/gio/vfunc.FileMonitor.cancel.html
+        debug_assert!(res, "FileMonitor.cancel should always return true");
+        res
+    }
+}
+
+// Support parent implementation of virtual functions defined in `gio::ffi::GFileMonitorClass`.
+pub trait FileMonitorImplExt: FileMonitorImpl {
```
sdroege, line 19: You probably want to provide something here for emitting the changed signal in a useful way. And probably also provide changed as a vfunc?

gio/tests/file_monitor.rs
```diff
@@ -0,0 +1,162 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
```
sdroege, line 1: Here and elsewhere, why do you provide these tests as external integration tests instead of just as part of the implementation? The latter is faster compilation-wise

gio/src/subclass/file_monitor.rs
```diff
@@ -0,0 +1,56 @@
...
+use glib::{prelude::*, subclass::prelude::*, translate::*};
+
+use crate::{ffi, FileMonitor};
+
+// Support custom implementation of virtual functions defined in `gio::ffi::GFileMonitorClass`.
+pub trait FileMonitorImpl: ObjectImpl + ObjectSubclass<Type: IsA<FileMonitor>> {
+    fn cancel(&self) -> bool {
+        let res = self.parent_cancel();
+        // cancel should always return true as specified in documentation
+        // https://docs.gtk.org/gio/vfunc.FileMonitor.cancel.html
```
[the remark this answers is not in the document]
bilelmoussaoui: Yeah... I am not sure it is useful

gio/tests/file_monitor.rs
sdroege: You could add a `src/test_utils.rs` module or so that is only included with `#[cfg(test)]`

sdroege:
Also looks good to me except for the two open discussions

bilelmoussaoui (approved):

PR #1726, opened 2025-05-21
Add gio::Vfs subclass

fbrouille:
Allow custom implementation of gio::Vfs

### DIFF
--- a/gio/src/subclass/vfs.rs
+++ b/gio/src/subclass/vfs.rs
@@ -1,0 +1,341 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use std::path::PathBuf;
+
+use glib::{prelude::*, subclass::prelude::*, translate::*, GString, StrVRef};
+
+use libc::c_char;
+
+use crate::{ffi, File, Vfs};
+
+// Support custom implementation of virtual functions defined in `gio::ffi::GVfsClass`.
+pub trait VfsImpl: ObjectImpl + ObjectSubclass<Type: IsA<Vfs>> {
+    fn is_active(&self) -> bool {
+        self.parent_is_active()
+    }
+
+    fn get_file_for_path(&self, path: &std::path::Path) -> File {
+        self.parent_get_file_for_path(path)
+    }
+
+    fn get_file_for_uri(&self, uri: &str) -> File {
+        self.parent_get_file_for_uri(uri)
+    }
+
+    fn get_supported_uri_schemes(&self) -> &'static StrVRef {
+        self.parent_get_supported_uri_schemes()
+    }
+
+    fn parse_name(&self, parse_name: &str) -> File {
+        self.parent_parse_name(parse_name)
+    }
+}
+
+// Support parent implementation of virtual functions defined in `gio::ffi::GVfsClass`.
+pub trait VfsImplExt: VfsImpl {
+    fn parent_is_active(&self) -> bool {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .is_active
+                .expect("No parent class implementation for \"is_active\"");
+
+            let res = f(self.obj().unsafe_cast_ref::<Vfs>().to_glib_none().0);
+            from_glib(res)
+        }
+    }
+
+    fn parent_get_file_for_path(&self, path: &std::path::Path) -> File {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .get_file_for_path
+                .expect("No parent class implementation for \"get_file_for_path\"");
+
+            let res = f(
+                self.obj().unsafe_cast_ref::<Vfs>().to_glib_none().0,
+                path.to_glib_none().0,
+            );
+            from_glib_full(res)
+        }
+    }
+
+    fn parent_get_file_for_uri(&self, uri: &str) -> File {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .get_file_for_uri
+                .expect("No parent class implementation for \"get_file_for_uri\"");
+
+            let res = f(
+                self.obj().unsafe_cast_ref::<Vfs>().to_glib_none().0,
+                uri.to_glib_none().0,
+            );
+            from_glib_full(res)
+        }
+    }
+
+    fn parent_get_supported_uri_schemes(&self) -> &'static StrVRef {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .get_supported_uri_schemes
+                .expect("No parent class implementation for \"get_supported_uri_schemes\"");
+
+            let res = f(self.obj().unsafe_cast_ref::<Vfs>().to_glib_none().0);
+            StrVRef::from_glib_borrow(res)
+        }
+    }
+
+    fn parent_parse_name(&self, parse_name: &str) -> File {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .parse_name
+                .expect("No parent class implementation for \"parse_name\"");
+
+            let res = f(
+                self.obj().unsafe_cast_ref::<Vfs>().to_glib_none().0,
+                parse_name.to_glib_none().0,
+            );
+            from_glib_full(res)
+        }
+    }
+}
+
+impl<T: VfsImpl> VfsImplExt for T {}
+
+// Implement virtual functions defined in `gio::ffi::GVfsClass`.
+unsafe impl<T: VfsImpl> IsSubclassable<T> for Vfs {
+    fn class_init(class: &mut ::glib::Class<Self>) {
+        Self::parent_class_init::<T>(class);
+
+        let klass = class.as_mut();
+        klass.is_active = Some(is_active::<T>);
+        klass.get_file_for_path = Some(get_file_for_path::<T>);
+        klass.get_file_for_uri = Some(get_file_for_uri::<T>);
+        klass.get_supported_uri_schemes = Some(get_supported_uri_schemes::<T>);
+        klass.parse_name = Some(parse_name::<T>);
+    }
+}
+
+unsafe extern "C" fn is_active<T: VfsImpl>(vfs: *mut ffi::GVfs) -> glib::ffi::gboolean {
+    let instance = &*(vfs as *mut T::Instance);
+    let imp = instance.imp();
+
+    let res = imp.is_active();
+
+    res.into_glib()
+}
+
+unsafe extern "C" fn get_file_for_path<T: VfsImpl>(
+    vfs: *mut ffi::GVfs,
+    path: *const c_char,
+) -> *mut ffi::GFile {
+    let instance = &*(vfs as *mut T::Instance);
+    let imp = instance.imp();
+
+    let file = imp.get_file_for_path(&PathBuf::from_glib_none(path));
+
+    file.into_glib_ptr()
+}
+
+unsafe extern "C" fn get_file_for_uri<T: VfsImpl>(
+    vfs: *mut ffi::GVfs,
+    uri: *const c_char,
+) -> *mut ffi::GFile {
+    let instance = &*(vfs as *mut T::Instance);
+    let imp = instance.imp();
+
+    let file = imp.get_file_for_uri(&GString::from_glib_borrow(uri));
+
+    file.into_glib_ptr()
+}
+
+unsafe extern "C" fn get_supported_uri_schemes<T: VfsImpl>(
+    vfs: *mut ffi::GVfs,
+) -> *const *const c_char {
+    let instance = &*(vfs as *mut T::Instance);
+    let imp = instance.imp();
+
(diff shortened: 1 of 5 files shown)


gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,184 @@
...
+// Support parent implementation of virtual functions defined in `gio::ffi::GVfsClass`.
+pub trait VfsImplExt: VfsImpl {
+    fn parent_is_active(&self) -> bool {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .is_active
+                .expect("No parent class implementation for \"is_active\"");
```
bilelmoussaoui, line 41: Are all those vfuncs required or is there a default impl?

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,184 @@
...
+
+use crate::{ffi, File, Vfs};
+
+// Support custom implementation of virtual functions defined in `gio::ffi::GVfsClass`.
+pub trait VfsImpl: ObjectImpl + ObjectSubclass<Type: IsA<Vfs>> {
+    fn is_active(&self) -> bool {
+        self.parent_is_active()
+    }
+
+    fn get_file_for_path(&self, path: impl AsRef<std::path::Path>) -> File {
```
sdroege, line 15:
```suggestion
    fn get_file_for_path(&self, path: &std::path::Path) -> File {
```

would work equally well. `impl trait` in parameter position in the `FooImpl` traits is rather useless and only causes code bloat :)

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,184 @@
...
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GVfsClass;
+
+            let f = (*parent_class)
+                .get_supported_uri_schemes
+                .expect("No parent class implementation for \"get_supported_uri_schemes\"");
+
+            let res = f(self.obj().unsafe_cast_ref::<Vfs>().to_glib_none().0);
+            FromGlibPtrContainer::from_glib_none(res)
```
sdroege, line 92: This is suboptimal as it copies strings all the time. Maybe this should use a `&glib::StrV`?

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,184 @@
...
+unsafe extern "C" fn get_file_for_uri<T: VfsImpl>(
+    vfs: *mut ffi::GVfs,
+    uri: *const c_char,
+) -> *mut ffi::GFile {
+    let instance = &*(vfs as *mut T::Instance);
+    let imp = instance.imp();
+
+    let file = imp.get_file_for_uri(&GString::from_glib_borrow(uri));
+
+    file.to_glib_full()
```
sdroege, line 160:
```suggestion
    file.into_glib_ptr()
```

also elsewhere

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+unsafe extern "C" fn get_supported_uri_schemes<T: VfsImpl>(
+    vfs: *mut ffi::GVfs,
+) -> *const *const c_char {
+    let instance = &*(vfs as *mut T::Instance);
+    let imp = instance.imp();
+
+    let supported_uri_schemes = imp.get_supported_uri_schemes();
```
sdroege, line 169: For the same VFS this is not supposed to change, so needs to be static of some sort. Maybe a `&'static glib::StrV` should be used here. Also because ...

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    supported_uri_schemes.to_glib_full()
```
sdroege, line 171: This should be transfer none, not transfer full

gio/tests/vfs.rs
```diff
@@ -0,0 +1,138 @@
...
+use glib::translate::ToGlibPtr;
+
+// Binding of existing GIO type GLocalVfs.
+mod ffi {
+    use gio::ffi;
+
+    #[derive(Copy, Clone)]
+    #[repr(C)]
+    pub struct GLocalVfs {
+        pub parent_instance: ffi::GVfs,
```
sdroege, line 17: You could use the default instance/class structs here

gio/tests/vfs.rs
```diff
@@ -0,0 +1,138 @@
...
+    match fn {
+        type_ => || {
+            use std::sync::Once;
+            static ONCE: Once = Once::new();
+
+            // ensure type is initialized by calling `gio::auto::File::for_path` to create a `GLocalFile` instance.
+            ONCE.call_once(|| unsafe {
+                let _ = File::for_path("path");
+            });
+            glib::gobject_ffi::g_type_from_name("GLocalVfs".to_glib_none().0)
```
sdroege, line 40: Why? You could use the `ObjectSubclass` variant of the `glib::wrapper!` macro here

sdroege: Oh, nevermind

gio/tests/vfs.rs
sdroege: Nevermind. But these could just be `glib::gpointer`s instead of actual structs

gio/tests/vfs.rs
```diff
@@ -0,0 +1,138 @@
...
+    impl ObjectSubclass for MyLocalVfs {
+        const NAME: &'static str = "MyLocalVfs";
+        type Type = super::MyLocalVfs;
+        type ParentType = LocalVfs;
+    }
+
+    impl ObjectImpl for MyLocalVfs {}
+
+    // Implements `VfsImpl` with default implementation, which calls the parent's implementation.
+    impl VfsImpl for MyLocalVfs {}
```
sdroege, line 66: Wouldn't it be interesting to also check if overriding them works as expected?

gio/src/subclass/vfs.rs
sdroege: The implementation can return a different one on every call, and you'd need to copy it to add the NULL-terminator (see above).

sdroege:
Looks good to me except for the two open discussions about the string array.

gio/src/subclass/vfs.rs
sdroege: `parent_get_supported_uri_schemes()` will already return a `&'static StrV`, you can just pass it further without any copying or anything else.

Actually, I see your problem. Yes, this actually has to return a `&'static [glib::GStringPtr]`. You can get that from `StrV::from_glib_borrow()` for example. User implementations can build a `&'static StrV` (e.g. via `LazyLock`) and that derefs into `&'static [glib::GStringPtr]`

sdroege: Indeed. Maybe a reason to make it `&'static StrV`. You can get a stable pointer out of that :)

glib/src/collections/strv.rs
```diff
@@ -1359,6 +1373,49 @@ impl<const N: usize> IntoStrV for [&'_ String; N] {
     }
 }
 
+// rustdoc-stripper-ignore-next
+/// Representation of a borrowed `NULL`-terminated C array of `NULL`-terminated UTF-8 strings.
+///
+/// It can be constructed safely from a `&StrV` and unsafely from a pointer to a C array.
+/// This type is very similar to `[GStringPtr]`, but with one added constraint: the underlying C array must be `NULL`-terminated.
+#[repr(transparent)]
+pub struct CStrV {
```
sdroege: The name seems inconsistent with everything else. This is basically the `String`/`PathBuf` (`StrV`) vs. `str`/`Path` (`CStrV`) duality. Maybe this one should've actually been `StrV` and the other one `StringV`. Using `Buf` here seems wrong.

In other places (e.g. in `gstreamer`) there's a pattern of `Foo` and `FooRef`, so maybe we can just call this one `StrVRef`? @bilelmoussaoui what do you think?

Also `GStringPtr` is an ugly name, should probably reconsider that before the next release. Its difference with `GStr` is that it doesn't store the length and is equivalent with the raw C pointer. `GStrLite` or so... doesn't have to be solved as part of this PR but in case someone has ideas while we talk about naming anyway.

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,397 @@
...
+
+    fn get_file_for_path(&self, path: &std::path::Path) -> File {
+        self.parent_get_file_for_path(path)
+    }
+
+    fn get_file_for_uri(&self, uri: &str) -> File {
+        self.parent_get_file_for_uri(uri)
+    }
+
+    unsafe fn get_supported_uri_schemes(&self) -> &'static CStrV {
```
sdroege:
```suggestion
    fn get_supported_uri_schemes(&self) -> &'static CStrV {
```

no need to for this to be unsafe anymore (but also if it was necessary then the *trait* impl would have to be unsafe: unsafe trait is unsafe to implement, unsafe trait function is unsafe to call)

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,397 @@
...
+        match fn {
+            type_ => || {
+                use std::sync::Once;
+                static ONCE: Once = Once::new();
+
+                // ensure type is initialized by calling `gio::auto::File::for_path` to create a `GLocalFile` instance.
+                ONCE.call_once(|| unsafe {
+                    let _ = File::for_path("path");
+                });
+                glib::gobject_ffi::g_type_from_name("GLocalVfs".to_glib_none().0)
```
sdroege: Not that it really matters because this is a test, but why not call `g_type_from_name()` also inside the `Once` (and make it a `LazyLock`)?

gio/src/subclass/vfs.rs
```diff
@@ -0,0 +1,397 @@
...
+
+        impl ObjectImpl for MyVfs {}
+
+        // Implements `VfsImpl` with specific implementation of `get_supported_uri_schemes`.
+        impl VfsImpl for MyVfs {
+            unsafe fn get_supported_uri_schemes(&self) -> &'static CStrV {
+                static SUPPORTED_URI_SCHEMES: std::sync::OnceLock<glib::StrV> =
+                    std::sync::OnceLock::new();
+                SUPPORTED_URI_SCHEMES
+                    .get_or_init(|| glib::StrV::from(["myvfs"]))
```
sdroege: Why not `LazyLock`?

glib/src/collections/strv.rs
```diff
@@ -123,6 +123,20 @@ impl std::ops::Deref for StrV {
     }
 }
 
+impl AsRef<CStrV> for StrV {
```
sdroege: Should probably also `Deref` into `&CStrV` (which in turn `Deref` into `&[GStringPtr]`).

glib/src/collections/strv.rs
```diff
@@ -1359,6 +1373,49 @@ impl<const N: usize> IntoStrV for [&'_ String; N] {
...
+/// Representation of a borrowed `NULL`-terminated C array of `NULL`-terminated UTF-8 strings.
+///
+/// It can be constructed safely from a `&StrV` and unsafely from a pointer to a C array.
+/// This type is very similar to `[GStringPtr]`, but with one added constraint: the underlying C array must be `NULL`-terminated.
+#[repr(transparent)]
+pub struct CStrV {
+    inner: [GStringPtr],
+}
+
+impl CStrV {
```
sdroege: Should probably get all the same trait impls that `StrV` has (`PartialEq`, etc)

glib/src/collections/strv.rs
fbrouille: `CStrV` because of `std::ffi::CStr`. However I agree that name should be consistent with others gtk-rs types. I think also that `CStrV` should be `StrV` and `StrV` should be `StringV` or if you don't want breaking changes, maybe rename`CStrV` to `GStrV` and rename `StrV` to `GStringV` and create a type alias (deprecated) `StrV = GStringV`.

Anyway let me know for `CStrV` new name.

gio/src/subclass/vfs.rs
fbrouille: done

gio/src/subclass/vfs.rs
fbrouille: No it does not work. Calling `File::for_path`is the only way to initialize `GLocalVfs`.

Whatever, I have removed binding to existing (private) gio types and refactor the tests.

gio/src/subclass/vfs.rs
fbrouille: Just to be consistent with initialization way of other static variables in the gtk-rs. `LazyLock` is not (yet) used in the project.

Whatever, I changed it.

glib/src/collections/strv.rs
fbrouille: done

glib/src/collections/strv.rs
fbrouille: I tried to implement the same traits except `Clone` and `Drop` because `CStrV` does not own the null-terminated array.

gio/src/subclass/vfs.rs
sdroege: I meant doing both the `File::for_path()` and `g_type_from_name()` inside the `LazyLock` (or `OnceCell`)

gio/src/subclass/vfs.rs
sdroege: Also seems fine to me. I think I should go over the other code and update it one of these days :) Thanks for the reminder!

glib/src/collections/strv.rs
sdroege: I'd still vote for `StrV` (stays as it is) and `StrVRef`. But waiting for @bilelmoussaoui to opine :)

bilelmoussaoui: StrvRef / Strv sounds good to me

fbrouille: done

sdroege (approved):

PR #1728, opened 2025-05-22
Don't skip `Option` for nullable collections

fengalin:
Consequence of https://github.com/gtk-rs/gir/pull/1622


gio/src/auto/functions.rs
```diff
@@ -181,9 +181,9 @@ pub fn content_type_get_symbolic_icon(type_: &str) -> Icon {
 #[doc(alias = "g_content_type_guess")]
 pub fn content_type_guess(
     filename: Option<impl AsRef<std::path::Path>>,
-    data: &[u8],
+    data: Option<&[u8]>,
```
sdroege, line 184: Needs rebasing as this is now manually implemented. Can be autogenerated again :)

sdroege: It was only reverted in the 0.20 branch. main still has it and should have it with the `Option`.

gdk-pixbuf/src/auto/pixbuf_format.rs
```diff
@@ -29,9 +29,9 @@ impl PixbufFormat {
 
     #[doc(alias = "gdk_pixbuf_format_get_extensions")]
     #[doc(alias = "get_extensions")]
-    pub fn extensions(&self) -> Vec<glib::GString> {
+    pub fn extensions(&self) -> Option<Vec<glib::GString>> {
         unsafe {
-            FromGlibPtrContainer::from_glib_full(ffi::gdk_pixbuf_format_get_extensions(
+            TryFromGlibPtrContainer::try_from_glib_full(ffi::gdk_pixbuf_format_get_extensions(
```
bilelmoussaoui: I would instead name it, MaybeFromGlibPtrContainer or so. Try as a prefix assumes we would get a Result back instead of an Option

bilelmoussaoui: I am also not sure if all those returned `Option<Vec<T>>` make sense or some of them needs annotations fixes to return just `Vec<T>` that would be empty if there are no items.

fengalin: > I would instead name it, MaybeFromGlibPtrContainer or so. Try as a prefix assumes we would get a Result back instead of an Option

Indeed. done.

> I am also not sure if all those returned Option<Vec<T>> make sense or some of them needs annotations fixes to return just Vec<T> that would be empty if there are no items.

Yes, I went looking at the documentation for these functions and spotted some discrepancies last time I tackled this. I'll force annotations with my conclusions in a bit. I just wanted to rebase & push now that I have something usable as a whole (last time I paused, I had just figured out I had to implement all those `Maybe` traits).

fengalin:
Postponing these changes to a future Rust bindings generator rewrite

PR #1732, opened 2025-06-09
Document requirement for explicit default in `Enum` properties

nozwock:
When using the `Properties` macro with enum types that derive `Enum`, the default value must be explicitly set using `#[property(builder(...))]`.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -525,6 +525,10 @@ pub fn closure_local(item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
+/// When using the [`Properties`] macro with enums that derive [`Enum`], the default value must be
+/// explicitly set via the `builder` parameter of the `#[property]` attribute. See
+/// [here](Properties#supported-types) for details.
+///
 /// An enum can be registered as a dynamic type by setting the derive macro
 /// helper attribute `enum_dynamic`:
 ///
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -1408,9 +1408,10 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 ///     nick: String,
 /// }
 ///
-/// #[derive(Debug, Copy, Clone, PartialEq, Eq, glib::Enum)]
+/// #[derive(Debug, Copy, Clone, PartialEq, Eq, glib::Enum, Default)]
 /// #[enum_type(name = "MyEnum")]
 /// enum MyEnum {
+///     #[default]
 ///     Val,
 ///     OtherVal
 /// }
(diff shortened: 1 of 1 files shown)


glib-macros/src/lib.rs
```diff
@@ -525,6 +525,35 @@ pub fn closure_local(item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
+/// When using [`Properties`] with an enum that derives [`Enum`], the default
```
sdroege: This is not specific to enums that are derived via `glib::Enum` but also other existing enums, e.g. provided by C libraries.

I think it would make more sense to move this to the documentation of the `Properties` macro or what do you think?

nozwock: Hmm, yeah makes sense, it's more of a quirk of Properties regarding these "FFI" enums.

Don't see any issue with moving this over to docs of Properties. I'll do so in a while.

sdroege: Thanks :)

And maybe leave a small paragraph here that points to the Properties docs? "When used with the properties macro, you need to blablablabla link to the docs of that"

bilelmoussaoui: The same applies to flags btw

nozwock: > The same applies to flags btw

I had tried with Flags but it worked fine on my end with just the Default trait impl, didn't need to use the `builder` param.

sdroege: Flags default to "no flags set" so they're not affected indeed.

sdroege (approved):

glib-macros/src/lib.rs
```diff
@@ -1426,6 +1440,8 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 ///         optional: RefCell<Option<String>>,
 ///         #[property(get, set)]
 ///         smart_pointer: Rc<RefCell<String>>,
+///         #[property(get, set, builder(MyEnum::Val))]
+///         my_enum: Cell<MyEnum>,
```
sdroege, line 1445:
```
error[E0277]: the trait bound `MyEnum: std::default::Default` is not satisfied
  --> glib-macros/src/lib.rs:1445:9
   |
27 |     #[derive(Properties, Default)]
   |                          ------- in this derive macro expansion
...
48 |         my_enum: Cell<MyEnum>,
   |         ^^^^^^^^^^^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `MyEnum`
   |
   = help: the trait `std::default::Default` is implemented for `std::cell::Cell<T>`
   = note: required for `std::cell::Cell<MyEnum>` to implement `std::default::Default`
```

nozwock: Ah, Idk why this didn't happen when I was building the docs locally.

nozwock: Should be fixed now?

sdroege (approved):

PR #1740, opened 2025-06-17
Add gio::FileEnumerator subclass

fbrouille:
Allow custom implementation of gio::FileEnumerator

### DIFF
--- a/gio/Gir.toml
+++ b/gio/Gir.toml
@@ -891,6 +891,11 @@ manual_traits = ["FileEnumeratorExtManual"]
     name = "iterate"
     # better with rust iterators
     ignore = true
+    [[object.function]]
+    name = "close"
+    # Return value does not indicate success or error: despite the documentation, this function could return TRUE and an error in same time.
+    # see https://gitlab.gnome.org/GNOME/glib/-/blob/dc6f9447cd2a01012731422afed6facff1eea81a/gio/gfileenumerator.c#L290-297
+    manual = true
 
 [[object]]
 name = "Gio.FileInfo"
--- a/gio/src/auto/file_enumerator.rs
+++ b/gio/src/auto/file_enumerator.rs
@@ -20,24 +20,6 @@ impl FileEnumerator {
 }
 
 pub trait FileEnumeratorExt: IsA<FileEnumerator> + 'static {
-    #[doc(alias = "g_file_enumerator_close")]
-    fn close(&self, cancellable: Option<&impl IsA<Cancellable>>) -> Result<(), glib::Error> {
-        unsafe {
-            let mut error = std::ptr::null_mut();
-            let is_ok = ffi::g_file_enumerator_close(
-                self.as_ref().to_glib_none().0,
-                cancellable.map(|p| p.as_ref()).to_glib_none().0,
-                &mut error,
-            );
-            debug_assert_eq!(is_ok == glib::ffi::GFALSE, !error.is_null());
-            if error.is_null() {
-                Ok(())
-            } else {
-                Err(from_glib_full(error))
-            }
-        }
-    }
-
     #[doc(alias = "g_file_enumerator_close_async")]
     fn close_async<P: FnOnce(Result<(), glib::Error>) + 'static>(
         &self,
--- a/gio/src/subclass/file_enumerator.rs
+++ b/gio/src/subclass/file_enumerator.rs
@@ -1,0 +1,415 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{prelude::*, subclass::prelude::*, translate::*};
+
+use crate::{ffi, prelude::*, Cancellable, FileEnumerator, FileInfo, IOErrorEnum};
+
+// Support custom implementation of virtual functions defined in `gio::ffi::GFileEnumeratorClass` except pairs `xxx_async/xxx_finish` for which GIO provides a default implementation (which should be ok).
+// TODO: overriding these default implementations might still be useful for subclasses (if they can do something better than blocking IO).
+pub trait FileEnumeratorImpl: ObjectImpl + ObjectSubclass<Type: IsA<FileEnumerator>> {
+    fn next_file(
+        &self,
+        cancellable: Option<&Cancellable>,
+    ) -> Result<Option<FileInfo>, glib::Error> {
+        self.parent_next_file(cancellable)
+    }
+
+    // rustdoc-stripper-ignore-next
+    /// Closes the enumerator (see [`FileEnumeratorExt::close`]).
+    ///
+    /// NOTE: If the enumerator has not been explicitly closed, GIO closes it when the object is dropped.
+    /// But GIO does it by calling `close` vfunc in `finalize`, which is not safe and could lead to undefined behavior,
+    /// such as accessing freed memory or resources, which can cause crashes or other unexpected behavior.
+    ///
+    /// An issue has been opened in GLib to address this: <https://gitlab.gnome.org/GNOME/glib/-/issues/3713> and a MR has been opened to fix it: <https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4672>.
+    ///
+    /// Until this is fixed, it is unsafe to rely on the enumerator being closed when the object is dropped.
+    /// It is recommended to close the enumerator explicitly before dropping it, by calling [`FileEnumeratorExt::close`],
+    /// or to implement the [`ObjectImpl::dispose`] method and call [`FileEnumeratorExt::close`] there (it is safe to access the object there):
+    /// ```ignore
+    /// pub struct MyFileEnumerator();
+    ///
+    /// #[glib::object_subclass]
+    /// impl ObjectSubclass for MyFileEnumerator { ... }
+    ///
+    /// impl ObjectImpl for MyFileEnumerator {
+    ///     fn dispose(&self) {
+    ///         // close the enumerator here is safe and avoids `finalize` to call close.
+    ///         let _ = self.obj().close(Cancellable::NONE);
+    ///     }
+    /// }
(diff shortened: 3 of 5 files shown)


gio/src/subclass/file_enumerator.rs
```diff
@@ -0,0 +1,395 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{prelude::*, subclass::prelude::*, translate::*, Error};
+
+use crate::{ffi, traits::FileEnumeratorExt, Cancellable, FileEnumerator, FileInfo, IOErrorEnum};
```
bilelmoussaoui: use prelude here as well, we don't go through traits::* as it is an implementation detail that can change at any point.

gio/src/subclass/file_enumerator.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        #[glib::object_subclass]
+        impl ObjectSubclass for MyFileEnumerator {
+            const NAME: &'static str = "MyFileEnumerator";
+            type Type = super::MyFileEnumerator;
+            type ParentType = FileEnumerator;
+
+            fn new() -> Self {
+                Self(Cell::new(0))
+            }
```
bilelmoussaoui: you already derive default, so this one is useless

gio/src/subclass/file_enumerator.rs
```diff
@@ -0,0 +1,395 @@
...
+
+        let klass = class.as_mut();
+        klass.next_file = Some(next_file::<T>);
+        klass.close_fn = Some(close_fn::<T>);
+        // `GFileEnumerator` already implements `xxx_async/xxx_finish` vfuncs and this should be ok.
+        // TODO: when needed, override the `GFileEnumerator` implementation of the following vfuncs:
+        // klass.next_files_async = Some(next_files_async::<T>);
+        // klass.next_files_finish = Some(next_files_finish::<T>);
+        // klass.close_async = Some(close_async::<T>);
+        // klass.close_finish = Some(close_finish::<T>);
```
bilelmoussaoui, line 129: Are those two vfuncs pairs not important to override/implement in general?  i am just curious, as i think they can easily be added, see https://github.com/gtk-rs/gtk4-rs/blob/main/gdk4/src/subclass/content_provider.rs for an example

gio/src/auto/file_enumerator.rs
```diff
@@ -21,17 +21,16 @@ impl FileEnumerator {
 
 pub trait FileEnumeratorExt: IsA<FileEnumerator> + 'static {
     #[doc(alias = "g_file_enumerator_close")]
-    fn close(&self, cancellable: Option<&impl IsA<Cancellable>>) -> Result<(), glib::Error> {
+    fn close(&self, cancellable: Option<&impl IsA<Cancellable>>) -> Result<bool, glib::Error> {
```
sdroege: Should be `(bool, Option<glib::Error>)` as return type or not? If it can return an error in both the `true` and `false` case like your comment suggests

gio/src/subclass/file_enumerator.rs
```diff
@@ -0,0 +1,395 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::{prelude::*, subclass::prelude::*, translate::*, Error};
+
+use crate::{ffi, traits::FileEnumeratorExt, Cancellable, FileEnumerator, FileInfo, IOErrorEnum};
+
+// Support custom implementation of virtual functions defined in `gio::ffi::GFileEnumeratorClass` except pairs `xxx_async/xxx_finish` for which GIO provides a default implementation.
```
sdroege: Overriding these default impls might still be useful for subclasses (if they can do something better than blocking IO). That can be added in another PR but the comment sounds right now like it's not useful.

gio/src/subclass/file_enumerator.rs
```diff
@@ -0,0 +1,395 @@
...
+    fn parent_close(&self, cancellable: Option<&Cancellable>) -> Result<bool, Error> {
+        unsafe {
+            let data = Self::type_data();
+            let parent_class = data.as_ref().parent_class() as *const ffi::GFileEnumeratorClass;
+
+            let f = (*parent_class)
+                .close_fn
+                .expect("No parent class implementation for \"close_fn\"");
+
+            // get the corresponding object instance without checking the reference count because the object might just be finalized.
```
sdroege: `close()` is called during `finalize()`?! That means that this can't be provided as a normal virtual method then as it's unsound to access the full object during finalization (parts might already be freed, you can easily cause memory safety issues)

gio/src/auto/file_enumerator.rs
fbrouille: ok, with to manual implementation for `close`

gio/src/subclass/file_enumerator.rs
fbrouille: Added a TODO in comment.

gio/src/subclass/file_enumerator.rs
fbrouille: I suggest to call `close` in `dispose` because in this case it is safe to access the object. See tests for an example.

I also added a comment in `closet notify developers they have to implement `dispose`.

gio/src/subclass/file_enumerator.rs
fbrouille: done

gio/src/subclass/file_enumerator.rs
fbrouille: done

gio/Gir.toml
```diff
@@ -891,6 +891,11 @@ manual_traits = ["FileEnumeratorExtManual"]
     name = "iterate"
     # better with rust iterators
     ignore = true
+    [[object.function]]
+    name = "close"
+    # Return value does not indicate success or error: despite the documentation, this function could return TRUE and an error in same time.
+    # see https://gitlab.gnome.org/GNOME/glib/-/blob/dc6f9447cd2a01012731422afed6facff1eea81a/gio/gfileenumerator.c#L290-297
+    ignore = true
```
bilelmoussaoui: should be marked as maunal = true so it is properly documented

gio/src/subclass/file_enumerator.rs
```diff
@@ -0,0 +1,411 @@
...
+    ///         let _ = self.obj().close(Cancellable::NONE);
+    ///     }
+    /// }
+    ///
+    /// impl FileEnumeratorImpl for MyFileEnumerator { ... }
+    /// ```
+    ///
+    /// [`FileEnumeratorExt::close`]: ../auto/file_enumerator/trait.FileEnumeratorExt.html#method.close
+    /// [`ObjectImpl::dispose`]: ../../glib/subclass/object/trait.ObjectImpl.html#method.dispose
+    // rustdoc-stripper-ignore-next-stop
```
bilelmoussaoui:
```suggestion
```

gio/src/subclass/file_enumerator.rs
fbrouille: IMO providing a coherent implementation of `gio::Vfs`, `gio::File`, `gio::FileEnumerator` and `gio::FileMonitor` without vfuncs pairs is a first (and big) step.
Of course there is only 2 vfuncs pairs in `gio::FileEnumerator`, but you will notice there is more than 20 vfuncs pairs in `gio::File`, that is why I think it would be better to implement all of them in a second step (one or more PR)

gio/Gir.toml
fbrouille: oh! thanks for the tip! rustdoc-stripper causes headaches to me :)

gio/src/subclass/file_enumerator.rs
fbrouille: done

gio/src/subclass/file_enumerator.rs
sdroege: We can't depend on developers implementing dispose for safe usage though. Calling it in dispose seems fine but you'd have to make sure this happens automatically then.

fbrouille: This could be done by adding a new method `IsSubclassable::instance_dispose` with default implementation which calls `ObjectImpl::dispose`.
Then implementation of `IsSubclassable` for `FileEnumerator` can overrides `instance_dispose` to call `close`.
Obviously `extern "C" fn dispose` must be modified to call `IsSubclassable::instance_dispose` instead of `ObjectImpl::dispose`
What do you think ?

bilelmoussaoui: >  We can't depend on developers implementing dispose for safe usage though. Calling it in dispose seems fine but you'd have to make sure this happens automatically then.

We don't have a mechanism for injecting a call inside ObjectImpl's dispose. As currently for gtk4-rs composite templates, you have to call `dispose_template` in dispose yourself.

So I rather we don't block on this given the low usage of the API and just make sure it is clearly documented. We can revisit this once we have a clear mechanism for achieving that.

sdroege: Sounds good to me but also maybe in gio itself it should already call this in dispose instead of finalize.

sdroege: Can someone create an issue, also make it very clear in the docs here, and then we can get it in I guess

fbrouille: issue: https://gitlab.gnome.org/GNOME/glib/-/issues/3713
MR: https://gitlab.gnome.org/GNOME/glib/-/issues/3713

sdroege: Can you also add a link to that to the docs here?

fbrouille: done

sdroege (approved):

PR #1745, opened 2025-06-21
unix_fd_add: Port to safe-io

A6GibKm:
From https://github.com/gtk-rs/gtk-rs-core/pull/871.

### DIFF
--- a/glib/src/source.rs
+++ b/glib/src/source.rs
@@ -1,12 +1,10 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #[cfg(unix)]
-use std::os::unix::io::RawFd;
+use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd};
 use std::{cell::RefCell, mem::transmute, num::NonZeroU32, time::Duration};
 
 use crate::ffi::{self, gboolean, gpointer};
-#[cfg(all(not(unix), docsrs))]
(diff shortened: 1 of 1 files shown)


A6GibKm:
From https://github.com/gtk-rs/gtk-rs-core/pull/871#discussion_r1631987238:

> @sdroege: I think these functions all need to be unsafe actually. impl AsFd is correct but it's up to the caller to ensure that the fd is valid long enough

In that case would it not be better to pass an OwnedFd? The docs for unix_fd_add or unix_fd_add_full don't say much.

sdroege (dismissed):

sdroege:
Also, shouldn't the `GSource` trigger when it's closed and be destroyed then? From inside GLib I mean.

A6GibKm:
> > In that case would it not be better to pass an OwnedFd? The docs for unix_fd_add or unix_fd_add_full don't say much.
> 
> Good question. But can you get an `OwnedFd` for stdin, for example?

If you mean std::io::Stdin then it implements AsFd from which you can call https://doc.rust-lang.org/std/os/fd/struct.BorrowedFd.html#method.try_clone_to_owned.

sdroege:
I think the main question here is what happens with the `GSource` if the fd is closed. Random things? Does the `GSource` get destroyed cleanly?

sdroege:
@A6GibKm This would need to be handled in the next weeks if it should end up in the next release

Issue #930, opened 2023-01-30
properties: Figure out how to make use of the default value of enum properties automatically


sdroege:
See title

ranfdev:
To clarify the problem here...

Ideally we would do:
```rust
// Specialized impl for Default types
impl HasParamSpec for MyEnumType where Self: Default {
  type BuilderFn = fn(&str) -> ParamSpecEnumBuilder;
  fn param_spec_builder() -> Self::BuilderFn {
    // first argument, the name, is provided by the macro. Default value filled using the trait 
    |name| ParamSpecEnum::builder_with_default(name, Self::default())
  }
}

impl HasParamSpec for MyEnumType where Self: Default {
  type BuilderFn = fn(&str, Self) -> ParamSpecEnumBuilder;
  fn param_spec_builder() -> Self::BuilderFn {
    // first argument, the name, is provided by the macro. Default value provided by the user
    |name, default_value| ParamSpecEnum::builder_with_default(name, default_value)
  }
}
```

Sadly rust doesn't support [specialization](https://github.com/rust-lang/rust/issues/31844) yet and I can't find an alternative way to achieve this.

pbor:
I have not looked at the new macro implementation, so I guess I am just being naive, but can't we just check if a `default = ...` is provided in the attributes? If yes, we generate `builder_with_default`, if not, we generate `builder` and then if the enum does not have a default it will fail to compile

ranfdev:
That requires adjusting the macro specifically to support Enums. That means other external types requiring a default value will not take advantage of this.

But yeah, that could work!

ranfdev:
Problem: the macro doesn't know at compile time if the property is an Enum, so I can't write

```
if type.is_enum() && there_is_default_attribute {
 call_builder_with_default(default_attribute_value)
}
```

So we are again without possible solutions.

sdroege:
We could do something like `#[property(..., default)]` and if it finds that it uses a new `HasParamSpecWithDefault` trait, which can also be implemented for everything else that has a default value and uses `Default::default()`

ranfdev:
Inspired by that idea, I've tried to do
```rust
// unless a custom `default` attribute is specified, the macro will use this trait. 
pub trait HasParamSpecDefaulted: HasParamSpec + Default {
    type BuilderFnDefaulted = Self::BuilderFn;
    fn param_spec_builder_defaulted() -> Self::BuilderFnDefaulted {
[...]

PR #1761, opened 2025-07-04
Allow specifying a default via `Default` in a field within `Properties`

carlosmn:
Given a struct like

```rust
#[derive(Copy, Default, Clone, Debug, PartialEq, Eq, glib::Enum)]
#[enum_type(name = "SimpleEnum")]
pub enum SimpleEnum {
    #[default]
    One,
    Two,
}
```

you currently have to specify the default value as part of the param spec builder, e.g.

```rust
            #[property(get, set, builder(SimpleEnum::One))]
            fenum: RefCell<SimpleEnum>,
```

With this PR, we allow for that to change to 

```rust
            #[property(get, set, default)]
            fenum: RefCell<SimpleEnum>,
```

as discussed in #930. This started with what @ranfdev wrote though it changed a little bit.

It's still a bit messier than I was hoping as you still have a few different ways of setting the default value, you can have

    #[property(default)]
    #[property(builder(Type::Default))]

which is fine for compat and because not every enum has a built-in default, but you could also do things like

    #[property(builder(Type::Default).default_value(Type::OtherDefault))]
    #[property(default, builder().default_value(Type::OtherDefault))]

though you shouldn't and probably wouldn't because why would you. It comes out of the generic nature of this and it's probably not worth catching this and erroring out, but it doesn't make the whole thing easier to understand.

The main thing here are `Enum`s but I also added it for `char`s because they do have a default value in Rust. This might however not make as much sense in glib? I'm not sure if there's an actually-correct answer there.

As I mentioned in https://github.com/gtk-rs/gtk-rs-core/issues/930#issuecomment-3033592982 I don't think you can make this an actual blanket implementation because you need to know a lot about the types and the builder function, which is arbitrary. But covering enums sounds like it's 90% of the use-case, and other libraries can still implement it themselves.

### DIFF
--- a/glib-macros/src/lib.rs
+++ b/glib-macros/src/lib.rs
@@ -525,9 +525,12 @@ pub fn closure_local(item: TokenStream) -> TokenStream {
 /// }
 /// ```
 ///
-/// When using the [`Properties`] macro with enums that derive [`Enum`], the default value must be
-/// explicitly set via the `builder` parameter of the `#[property]` attribute. See
-/// [here](Properties#supported-types) for details.
+/// When using the [`Properties`] macro with enums that derive [`Enum`], the
+/// default value can be explicitly set via the `builder` parameter of the
+/// `#[property]` attribute. If the enum implements or derives
+/// `Default`, you can specify that should be the default value
+/// via the `default` parameter. See [here](Properties#supported-types) for
+/// details.
 ///
 /// An enum can be registered as a dynamic type by setting the derive macro
 /// helper attribute `enum_dynamic`:
@@ -1333,7 +1336,8 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 /// | `construct` | Specify that the property is construct property. Ensures that the property is always set during construction (if not explicitly then the default value is used). The use of a custom internal setter is supported. | | `#[property(get, construct)]` or `#[property(get, set = set_prop, construct)]` |
 /// | `construct_only` | Specify that the property is construct only. This will not generate a public setter and only allow the property to be set during object construction. The use of a custom internal setter is supported. | | `#[property(get, construct_only)]` or `#[property(get, set = set_prop, construct_only)]` |
 /// | `builder(<required-params>)[.ident]*` | Used to input required params or add optional Param Spec builder fields | | `#[property(builder(SomeEnum::default()))]`, `#[builder().default_value(1).minimum(0).maximum(5)]`, etc.  |
-/// | `default` | Sets the `default_value` field of the Param Spec builder | | `#[property(default = 1)]` |
+/// | `default` | Sets the param spec builder field to the default value | | `#[property(default)]` |
+/// | `default = expr` | Sets the `default_value` field of the Param Spec builder | | `#[property(default = 1)]` |
 /// | `<optional-pspec-builder-fields> = expr` | Used to add optional Param Spec builder fields | | `#[property(minimum = 0)` , `#[property(minimum = 0, maximum = 1)]`, etc. |
 /// | `<optional-pspec-builder-fields>` | Used to add optional Param Spec builder fields | | `#[property(explicit_notify)]` , `#[property(construct_only)]`, etc. |
 ///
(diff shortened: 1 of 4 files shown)


glib-macros/src/lib.rs
```diff
@@ -1333,7 +1336,8 @@ pub fn cstr_bytes(item: TokenStream) -> TokenStream {
 /// | `construct` | Specify that the property is construct property. Ensures that the property is always set during construction (if not explicitly then the default value is used). The use of a custom internal setter is supported. | | `#[property(get, construct)]` or `#[property(get, set = set_prop, construct)]` |
 /// | `construct_only` | Specify that the property is construct only. This will not generate a public setter and only allow the property to be set during object construction. The use of a custom internal setter is supported. | | `#[property(get, construct_only)]` or `#[property(get, set = set_prop, construct_only)]` |
 /// | `builder(<required-params>)[.ident]*` | Used to input required params or add optional Param Spec builder fields | | `#[property(builder(SomeEnum::default()))]`, `#[builder().default_value(1).minimum(0).maximum(5)]`, etc.  |
-/// | `default` | Sets the `default_value` field of the Param Spec builder | | `#[property(default = 1)]` |
+/// | `default` | Sets the param spec builder field to the default value | | `#[property(default)]` |
```
sdroege, line 1339: This should probably do this in general for all types for consistency?

sdroege: OTOH that doesn't really make sense as for all other types you get the default by not providing anything

glib/src/param_spec.rs
```diff
@@ -2174,6 +2174,35 @@ pub trait HasParamSpec {
...
+    > HasParamSpecDefaulted for T
+{
+    type BuilderFnDefaulted = fn(name: &str) -> ParamSpecEnumBuilder<T>;
+    fn param_spec_builder_defaulted() -> Self::BuilderFnDefaulted {
+        |name| Self::ParamSpec::builder(name)
+    }
+}
+
+// Manually implement the trait for chars
+impl HasParamSpecDefaulted for char {
```
sdroege, line 2199: But for chars you can also just not provide `default` and it will default to 0 or so inside GLib

carlosmn: `ParamSpec::BuilderFn` for `char` does require an explicit value though. So you can either use `builder('x')` or `default`, but you cannot leave it out (I think it would make sense the implementation did default to NUL instead)

sdroege: Oh I see. That makes sense, yes

glib-macros/src/lib.rs
carlosmn: Indeed most or really all types covered by this crate save for `char` do have a default value, and `HasParamSpec::BuilderFn` accepts just the name, setting the value to the default. it's just enums that don't necessarily have a default value and need the second parameter for the generic case.

sdroege: For flags it might also make sense. Their `Default` impl might be different from the paramspec default for flags (=> empty flags)

bilelmoussaoui:
> I think this generally makes sense. thanks! @bilelmoussaoui ?

lgtm to me as well!

glib-macros/src/lib.rs
sdroege: @carlosmn Want to add this for flags here or should we just get it in for now?

carlosmn: The flags builder, if I'm looking at the right code, also only accepts the name and uses its own concept of a default in https://github.com/gtk-rs/gtk-rs-core/blob/7507de59b316bb4579c15f7d50b16e23d37c17e7/glib/src/param_spec.rs#L1276 so it doesn't have a way to fill in the default in the way enums or char had. So it feels like that would be a different kind of default-setting that might have larger repercussions for other types as well.

All that to say that I think flags getting a default should be a different PR.

sdroege: Ok, let's get it in then. Thanks!

Issue #1730, opened 2025-05-27
[BUG] when returning a primitive type from a C callback for a task task::propagate expects a GValue


Ytsejam76:
given a C code callback returning a primitive typed result (such as a boolean)  with, e.g., `g_task_return_boolean` the Gtk-rs bindings fail to propagate the result, since the generic `ffi::g_task_propagate_value()` (or ` `ffi::g_task_propagate_pointer`, depending on the version) is called under the hood by Task::propagate() and the call fails at the:
```
  g_return_val_if_fail (task->result_destroy == value_free, FALSE);
````

The workaround is to use `g_task_propagate_value()` even for primitive types:

````
  void my_task_callback(....) {
            GValue ret_value = { 0 };

/* My callback logic */

            g_value_init(&ret_value, G_TYPE_BOOLEAN);
            g_value_set_boolean(&ret_value, TRUE);
            g_task_return_value(task, &ret_value);
            g_value_unset(&ret_value);
}

sdroege:
Can you provide a runnable and minimal testcase that demonstrates the problem?

Ytsejam76:
Do you really need it? 
If you check [task.rs](https://github.com/gtk-rs/gtk-rs-core/blob/main/gio/src/task.rs) you see that `propagate()` either calls [task.rs:287: ffi::g_task_propagate_value()](https://github.com/gtk-rs/gtk-rs-core/blob/main/gio/src/task.rs#L287), or  [task.rs:302: ffi::g_task_propagate_pointer()](https://github.com/gtk-rs/gtk-rs-core/blob/main/gio/src/task.rs#L302)

Both expect the [gtask.c: 2305: task->result_destroy](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gtask.c?ref_type=heads#L2305) not to be NULL, which it is if a primitive value is returned as is the case of [gtask.c:2007: g_task_return_boolean()](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gtask.c?ref_type=heads#L2007)

Ytsejam76:
There we go:

https://github.com/Ytsejam76/gtk_rs_minimal_example.git

running `cargo test` should be sufficient.

Ytsejam76:
This is an attempt at a solution for the boolean case only. If deemed fine I could follow up along the same lines for integers.
[...]

PR #1762, opened 2025-07-05
gio: add methods for returning and propagating boolean and int in/from a `Task`

carlosmn:
When using Rust we prefer to use a `Value` for everything but gio also lets you return explicitly a `gboolean` or `gssize` which avoid boxing.

Usage of these functions to return these kinds of values requires a call to the corresponding propagate function and you cannot mix them. Trying to extract a `Value` when either of them were used leads to an incorrect value.

Expose methods in `Task` and `LocalTask` to return and propagate these kinds instead of assuming everything can go through a `Value`.

This is meant to address #1730 

Given how you can just get some nonsense back out of any of these, I wonder if they should be considered `unsafe`. You can definitely try to grab the `gssize` represented by the pointer and `GTask` isn't going to stop you.


sdroege:
We should probably get this in before the release on Tuesday

bilelmoussaoui:
Not sure about the suggested names but i also don't have any better alternatives. Otherwise lgtm

PR #1775, opened 2025-08-05
cairo: write-to-stream callbacks should accept const pointers to data

### LINKED ISSUE #1771: Why does the PNG surface's write_func (and cairo_write_func_t) use mutable data?
https://github.com/gtk-rs/gtk-rs-core/blob/94c3909e6fa3a979b5ffeb1e4aa7882802525ead/cairo/sys/src/lib.rs#L264-L265
https://github.com/gtk-rs/gtk-rs-core/blob/94c3909e6fa3a979b5ffeb1e4aa7882802525ead/cairo/src/surface_png.rs#L58-L62

The [corresponding C type](https://www.cairographics.org/manual/cairo-PNG-Support.html#cairo-write-func-t) takes a `const unsigned char *data`.

My guess is this was a copy-paste error when adding both read_func and write_func, since read_func _does_ need to take a mutable char pointer.

inklesspen:
This is in line with the [corresponding C types](https://www.cairographics.org/manual/cairo-PNG-Support.html#cairo-write-func-t); they don't need to mutate this data.

Closes #1771.

### DIFF
--- a/cairo/src/stream.rs
+++ b/cairo/src/stream.rs
@@ -200,7 +200,7 @@ struct MutableCallbackEnvironment {
 // so code outside of the `catch_unwind` call must never panic.
 extern "C" fn write_callback<W: io::Write + 'static>(
     env: *mut c_void,
-    data: *mut c_uchar,
+    data: *const c_uchar,
     length: c_uint,
 ) -> ffi::cairo_status_t {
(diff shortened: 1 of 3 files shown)


sdroege (approved):

sdroege:
I'll rebase this here later once it's otherwise passing.

PR #1778, opened 2025-08-08
Bug fixes for `StrV[Ref]`

kawadakk:
Fixes the following bugs:

- `StrV::reserve` can return without actually reserving if `additional` is too large (`>= usize::MAX - self.len()`) and overflow checking is disabled
- `StrV::extend_from_slice` can incorrectly skip `StrV::reserve` before writing elements if it is given a ZST slice containing a large number of elements (`>= usize::MAX - self.len()`) and overflow checking is disabled
- `StrV::extend_from_slice` is not panic safe
- `<StrVRef as PartialEq<[&str]>>::eq` incorrectly evaluates to `false` if two slices have different numbers of elements but their prefixes match

### DIFF
--- a/glib/src/collections/strv.rs
+++ b/glib/src/collections/strv.rs
@@ -800,7 +800,7 @@ impl StrV {
         assert!(index <= self.len);
 
         // Nothing new to reserve as there's still enough space
-        if self.len + 1 + 1 > self.capacity {
+        if 1 >= self.capacity - self.len {
             self.reserve(1);
         }
 
@@ -824,7 +824,7 @@ impl StrV {
     #[inline]
     pub fn push(&mut self, item: GString) {
         // Nothing new to reserve as there's still enough space
-        if self.len + 1 + 1 > self.capacity {
+        if 1 >= self.capacity - self.len {
             self.reserve(1);
         }
 
--- a/glib/src/collections/ptr_slice.rs
+++ b/glib/src/collections/ptr_slice.rs
@@ -713,7 +713,7 @@ impl<T: TransparentPtrType> PtrSlice<T> {
     #[allow(clippy::int_plus_one)]
     pub fn reserve(&mut self, additional: usize) {
         // Nothing new to reserve as there's still enough space
-        if self.len + additional + 1 <= self.capacity {
+        if additional < self.capacity - self.len {
             return;
         }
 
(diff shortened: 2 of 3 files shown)


glib/src/collections/strv.rs
```diff
@@ -1466,6 +1466,10 @@ impl std::hash::Hash for StrVRef {
 
 impl PartialEq<[&'_ str]> for StrVRef {
     fn eq(&self, other: &[&'_ str]) -> bool {
+        if self.len() != other.len() {
```
sdroege, line 1473: Same bug also some lines above in `impl PartialEq<[&'_ str]> for StrV {`

glib/src/collections/strv.rs
```diff
@@ -780,9 +780,11 @@ impl StrV {
             for item in other {
                 *self.ptr.as_ptr().add(self.len) = GString::from(item.as_ref()).into_glib_ptr();
                 self.len += 1;
-            }
 
-            *self.ptr.as_ptr().add(self.len) = ptr::null_mut();
+                // Add null terminator on every iteration because `as_ref`
```
sdroege, line 788: Same issue in `src/collections/ptr_slice.rs`

glib/src/collections/strv.rs
```diff
@@ -772,7 +772,7 @@ impl StrV {
     #[inline]
     pub fn extend_from_slice<S: AsRef<str>>(&mut self, other: &[S]) {
         // Nothing new to reserve as there's still enough space
-        if self.len + other.len() + 1 > self.capacity {
+        if other.len() >= self.capacity - self.len {
```
sdroege, line 779: Same issue in ptr_slice.rs and slice.rs.

And it might make sense to also change this for all the other calls to `reserve()`, e.g. in `insert()` just a few lines below

glib/src/collections/strv.rs
```diff
@@ -712,7 +712,7 @@ impl StrV {
     #[allow(clippy::int_plus_one)]
     pub fn reserve(&mut self, additional: usize) {
         // Nothing new to reserve as there's still enough space
-        if self.len + additional + 1 <= self.capacity {
+        if additional < self.capacity - self.len {
```
sdroege, line 719: Same issue in slice.rs and ptr_slice.rs

sdroege:
The CI failures are unrelated, you can ignore them :)

Once you updated the other parts, I'll merge this and get a bugfix release with that (and the other pending fixes) out. Thanks for fixing these!

sdroege:
Fixing those other occurrences now to get a bugfix release out.

sdroege (approved):

PR #1779, opened 2025-08-08
Use `glib::ExitCode` for `gio::ApplicationCommandLine::{exit_status,set_exit_status}`

pjungkamp:
See https://github.com/gtk-rs/gtk-rs-core/pull/1694#issuecomment-3154057825

Did you think of something like this? @swsnr @sdroege


sdroege (approved):
@swsnr If that also matches what you thought of, let's merge it :)

sdroege:
Oh no, I assumed this was ready apart from getting feedback. Sorry for merging it already.

PR #1782, opened 2025-08-09
gio: Fix ApplicationCommandLineExtManual draft

pjungkamp:
See https://github.com/gtk-rs/gtk-rs-core/pull/1779#issuecomment-3167703612

@sdroege Let's do it properly now. (And wait for most of the CI.)


pjungkamp:
@sdroege I had a typo in the `gio/Gir.toml` but that has already been fixed by the force-push. I did some preliminary tests, and this should be ready now.

PR #1788, opened 2025-08-12
Add new FromGlibPtrBorrow2 and FromGlibPtrBorrowMut2 traits

sdroege:
These borrow more safely the underlying type and are supposed to replace the old FromGlibPtrBorrow trait once all users are gone, as well as the manual implementations for the trait functions which are deprecated now.

### DIFF
--- a/glib/src/translate.rs
+++ b/glib/src/translate.rs
@@ -1587,7 +1587,7 @@ pub trait FromGlibPtrFull<P: Ptr>: Sized {
 }
 
 // rustdoc-stripper-ignore-next
-/// Translate from a pointer type by borrowing, without affecting the refcount.
+/// Translate from a pointer type by borrowing, without affecting the refcount or copying.
 ///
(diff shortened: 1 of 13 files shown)


glib/src/translate.rs
```diff
@@ -1621,6 +1621,64 @@ pub trait FromGlibPtrBorrow<P: Ptr>: Sized {
...
+///
+/// The implementation of this trait as well as the returned type
+/// must satisfy the same constraints together.
+/// They must not take ownership of the underlying value, copy it,
+/// and should not change its reference count.
+/// If it does, it must properly release obtained references.
+///
+/// For more information, refer to module level documentation.
+// FIXME: Rename to FromGlibPtrBorrow once all users are gone
+pub trait FromGlibPtrBorrow2<P>: Sized {
```
sdroege, line 1644: Should also implement this for `&str`, `&GStr`, `&GStrPtr`, `&StrVRef` and various other types. Also boxed derive etc.

PR #1790, opened 2025-08-14
glib: Add logging bridge for structured logs

pjungkamp:
I've found this to be really handy for nice logging across the entire GTK suite.


glib/src/bridged_logging.rs
```diff
@@ -265,6 +265,74 @@ pub fn rust_log_handler(domain: Option<&str>, level: glib_log::LogLevel, message
...
+            continue;
+        };
+
+        match field.key() {
+            "GLIB_DOMAIN" => domain = Some(value),
+            "MESSAGE" => message = Some(value),
+            "CODE_FILE" => file = Some(value),
+            "CODE_LINE" => line = value.parse().ok(),
+            "CODE_FUNC" => func = Some(value),
+            _ => continue,
```
sdroege, line 342: `log` also has key-value support. Could all the other keys here be mapped to that?

glib/src/bridged_logging.rs
```diff
@@ -265,6 +265,74 @@ pub fn rust_log_handler(domain: Option<&str>, level: glib_log::LogLevel, message
...
+        };
+    }
+
+    if let Some(message) = message {
+        let logger = rs_log::logger();
+
+        logger.log(
+            &rs_log::Record::builder()
+                .level(lvl)
+                .target(domain.unwrap_or("<null>"))
```
sdroege: Or just set no target at all if there's no domain?

glib/src/bridged_logging.rs
pjungkamp: Structured logging in `log` is disabled by default so I'd want to gate it behind a `log-kv` or `log-structured` feature.

The actual implementation is pretty trivial. We'd just implement the `log::Source` trait for `glib::LogField` and add the whole array of fields to the `RecordBuilder`.

glib/src/bridged_logging.rs
pjungkamp: That's what I do for my own application, but your `rust_log_handler` above is also using `domain.unwrap_or("<null>")`. I'd want to keep both in sync.

glib/src/bridged_logging.rs
sdroege: Sounds like a plan

glib/src/bridged_logging.rs
sdroege: Maybe best to change that in both places then

pjungkamp: Done!

glib/src/bridged_logging.rs
pjungkamp: Done!

glib/src/bridged_logging.rs
```diff
@@ -262,7 +262,108 @@ pub fn rust_log_handler(domain: Option<&str>, level: glib_log::LogLevel, message
...
+            _ => continue,
+        };
+    }
+
+    if let Some(message) = message {
+        let logger = rs_log::logger();
+        let mut record = rs_log::Record::builder();
+
+        #[cfg(feature = "log_kv")]
+        let fields = LogFields(fields);
```
sdroege, line 351: I wonder if this should duplicate `GLIB_DOMAIN`, `MESSAGE`, etc. I can see arguments for both (personally I think it would make sense to *not* duplicate them into the KVs). What do you think?

glib/src/bridged_logging.rs
sdroege: This is now setting it to an empty string. Does the empty string have the same effect as not setting any target at all?

pjungkamp: The empty string is the default target when it's not set.

https://docs.rs/log/latest/log/struct.MetadataBuilder.html#method.new

glib/src/bridged_logging.rs
pjungkamp: I don't have a preference here, I probably won't use it.

sdroege: @bilelmoussaoui ?

glib/src/bridged_logging.rs
sdroege: Thanks!

glib/src/bridged_logging.rs
bilelmoussaoui: No preferences neither

sdroege:
Let's merge it like this then

PR #1807, opened 2025-09-05
Fix issue 1805

### LINKED ISSUE #1805: [Comprehensive Report] Multiple Soundness and Logical Issues Found in Rust FFI Wrappers
## **Overview**

**Thank you for your contributions to the open-source world\!\!**

I am a security practitioner, and while auditing the project's Rust FFI wrapper code, I discovered a series of issues related to memory safety and logical correctness. Under certain conditions, these issues could lead to Undefined Behavior (UB) or unintended logic, thereby undermining Rust's safety guarantees.

This issue summarizes the multiple problems I have found, covering the following types:

  - **Semantic Mismatch**
  - **Out-of-Bounds Memory Access**
  - **Integer Overflow Risk**
  - **Logical Error**

Some of these issues may be insignificant due to their stringent trigger conditions, but I am providing the information nonetheless. It is recommended that some of these problems be fixed immediately to ensure the soundness and reliability of the wrapper functions.

#### Core Issue

The wrapper function `graphene/src/auto/ray.rs:115, intersect_triangle` unconditionally calls `t_out.assume_init()`. However, the underlying C function only writes to the `t_out` pointer if the ray **intersects** with the triangle. If the ray does not intersect, the memory for `t_out` remains uninitialized. Calling `assume_init` at this point creates a semantic inconsistency with Rust's assumptions, leading to UB.

```rust
#[doc(alias = "graphene_ray_intersect_triangle")]
pub fn intersect_triangle(&self, t: &Triangle) -> (RayIntersectionKind, f32) {
    unsafe {
        let mut t_out = std::mem::MaybeUninit::uninit();
        let ret = from_glib(ffi::graphene_ray_intersect_triangle(
            self.to_glib_none().0,
            t.to_glib_none().0,
            t_out.as_mut_ptr(),
        ));
        (ret, t_out.assume_init())
    }
}
```

#### Recommended Fix:

1.  The `RayIntersectionKind` returned by the C function should be checked first.
2.  Only call `assume_init()` to read `t_out` if the return value indicates an intersection.
3.  Consider changing the function's return type from `(RayIntersectionKind, f32)` to `(RayIntersectionKind, Option<f32>)`, returning `None` when there is no intersection.

#### Core Issue

The wrapper function `cairo/src/region.rs:178, cairo_region_get_rectangle` does not perform bounds checking on the input parameter `nth`. If the value of `nth` is greater than or equal to the total number of rectangles in the region, the C function will attempt to access invalid memory outside the array's bounds, leading to UB.

```c++
void
cairo_region_get_rectangle (const cairo_region_t *region,
          int nth,
          cairo_rectangle_int_t *rectangle)
{
    pixman_box32_t *pbox;

    if (region->status) {
  rectangle->x = rectangle->y = 0;
  rectangle->width = rectangle->height = 0;
  return;
    }

    pbox = pixman_region32_rectangles (CONST_CAST &region->rgn, NULL) + nth; // The C side directly uses addition to get the memory address, without any defensive programming measures

    rectangle->x = pbox->x1;
    rectangle->y = pbox->y1;
    rectangle->width = pbox->x2 - pbox->x1;
    rectangle->height = pbox->y2 - pbox->y1;
}
```

#### Recommended Fix

1.  Before calling `cairo_region_get_rectangle`, first call `cairo_region_num_rectangles` to get the total number of rectangles.
2.  Verify that `nth` is within the valid range `[0, total - 1]`.
3.  If `nth` is out of range, it should return an error early (e.g., `None` or `Result::Err`) instead of calling the C function.

#### Core Issue

In the wrapper function `cairo/src/image_surface.rs:52, create_for_data`, the calculation `height * stride` within the `assert!(len >= (height * stride) as usize);` statement can cause an integer overflow. In Release mode, an overflow will not trigger a panic but will instead wrap around to a small, incorrect result. This allows the assertion to be bypassed, and the subsequent C function call may operate with an incorrect buffer size, leading to a buffer overflow.

#### Recommended Fix

1.  Use `height.checked_mul(stride)` instead of direct multiplication.
2.  If `checked_mul` returns `None` (indicating an overflow), an error should be returned immediately.
3.  Only proceed with the subsequent buffer size check and the C function call if the multiplication is successful and the result is valid.

#### Core Issue

Neither of these wrapper functions verifies that `start_index` is within the byte-length bounds of the input string `text`. If this index is outside the string's boundaries, the underlying C function will read memory beyond the `text` buffer to perform subsequent work, leading to UB.

Reference the following test code:

```rust
#[test]
fn test_itemize() {
    let font_map = pangocairo::FontMap::default();

    let context = font_map.create_context();

    let attrs = AttrList::new();
    let cached_iter = attrs.iterator();

    let pango_context: &Context = unsafe { std::mem::transmute(&context) };
    let text = "a";
    for i in 0..9 {
        let items = itemize(pango_context, text, i, 1, &attrs, Some(&cached_iter));
        eprintln!("items len: {:?}", items.len());
        for item in items {
            eprintln!("item: {:?}",  unsafe{ *item.as_ptr() });  
        }
    }
}
```

In this code, the length of `text` is only 1, but `start_index` exceeds the length of `text`. Running this code, either in Release or Debug mode, has a chance of triggering:

> (signal: 11, SIGSEGV: invalid memory reference)

This indicates that the returned item is illegal.

There is also a chance of producing the following output:

```
items len: 1
item: PangoItem @ 0x7878053f11d0 { offset: 0, length: 1, num_chars: 1, analysis: PangoAnalysis @ 0x7878053f11e0 { shape_engine: 0x0, lang_engine: 0x0, font: 0x78780000c880, level: 0, gravity: 0, flags: 128, script: 25, language: 0x787800008e58, extra_attrs: 0x0 } }
items len: 0
items len: 1
item: PangoItem @ 0x7878053f11d0 { offset: 2, length: 1, num_chars: 1, analysis: PangoAnalysis @ 0x7878053f11e0 { shape_engine: 0x0, lang_engine: 0x0, font: 0x78780000c880, level: 0, gravity: 0, flags: 128, script: 0, language: 0x787800008e58, extra_attrs: 0x0 } }
items len: 1
item: PangoItem @ 0x7878053f11d0 { offset: 3, length: 1, num_chars: 1, analysis: PangoAnalysis @ 0x7878053f11e0 { shape_engine: 0x0, lang_engine: 0x0, font: 0x78780000c880, level: 0, gravity: 0, flags: 128, script: 0, language: 0x787800008e58, extra_attrs: 0x0 } }
items len: 1
item: PangoItem @ 0x7878053f11d0 { offset: 4, length: 1, num_chars: 1, analysis: PangoAnalysis @ 0x7878053f11e0 { shape_engine: 0x0, lang_engine: 0x0, font: 0x78780000c880, level: 0, gravity: 0, flags: 128, script: 0, language: 0x787800008e58, extra_attrs: 0x0 } }
items len: 1
item: PangoItem @ 0x7878053f11d0 { offset: 5, length: 1, num_chars: 1, analysis: PangoAnalysis @ 0x7878053f11e0 { shape_engine: 0x0, lang_engine: 0x0, font: 0x78780000c880, level: 0, gravity: 0, flags: 128, script: 25, language: 0x787800008e58, extra_attrs: 0x0 } }
items len: 0
items len: 0
items len: 0
```

This is not the expected behavior.

#### Recommended Fix

1.  Add a comprehensive bounds check before calling the C function.

#### Core Issue

In the wrapper function `glib/src/functions.rs:138, spawn_async_with_pipes`, the function unconditionally calls `assume_init()` on the output parameters `child_pid`, `standard_input`, etc., regardless of whether the C function `g_spawn_async_with_pipes` succeeded or failed. If the C function fails (for example, because the command was not found or due to insufficient permissions), these output parameters will not be initialized. Calling `assume_init` at this point creates a semantic inconsistency with Rust's assumptions, leading to UB.

#### Recommended Fix

1.  First, check the return value of the C function or the `error` pointer to determine if the operation was successful.
2.  Only call `assume_init()` on the output parameters if the C function executed successfully.
3.  If the execution failed, the `error` should be handled directly and an `Err` should be returned, avoiding any access to uninitialized variables.

#### Core Issue

The validation logic in the `glib/src/date.rs: 275, subtract_days` function is reversed. The code is written as `if julian > n_days`, which incorrectly rejects valid date subtractions. The correct logic should be to check if the subtraction would cause the date to regress to an invalid state (i.e., the Julian day number is less than the number of days to subtract), so it should be `if julian < n_days`.

#### Recommended Fix

1.  Correct the comparison operator by changing `if julian > n_days` to `if julian < n_days`.

zeroy0410:
fix #1805


glib/src/functions.rs
```diff
@@ -179,11 +179,11 @@ pub fn spawn_async_with_pipes<
             standard_error.as_mut_ptr(),
             &mut error,
         );
-        let child_pid = from_glib(child_pid.assume_init());
-        let standard_input = standard_input.assume_init();
-        let standard_output = standard_output.assume_init();
-        let standard_error = standard_error.assume_init();
-        if error.is_null() {
+        if from_glib(success) {
```
sdroege: Why this change?

gdk-pixbuf/src/auto/versions.txt
```diff
@@ -1,2 +1,2 @@
-Generated by gir (https://github.com/gtk-rs/gir @ 7047e1b433e2)
+Generated by gir (https://github.com/gtk-rs/gir @ ab56426172f9)
```
sdroege: I'll get rid of this noise (and the corresponding CI failure) before merging, don't worry about that

glib/src/functions.rs
zeroy0410: I think these two expressions are equivalent, and using a return value might be more intuitive :)

sdroege (changes requested):
Can you split this into 6 separate commits (in the same PR) for the individual, unrelated issues? Each of them with a commit message explaining what is done there (basically what you already wrote in the issue, possibly shortened a bit)

glib/src/functions.rs
sdroege: If we clean that up, usually the pattern we use everywhere is

```
            debug_assert_eq!(is_ok == glib::ffi::GFALSE, !error.is_null());
            if error.is_null() {
```

so let's keep it consistent

zeroy0410: 👌

sdroege (approved):

pango/src/functions.rs
```diff
@@ -93,3 +93,65 @@ pub fn extents_to_pixels(
...
+    attrs: &AttrList,
+    cached_iter: Option<&AttrIterator>,
+) -> Vec<Item> {
+    let total_length = text.len() as i32;
+    assert!(
+        start_index >= 0 && start_index < total_length,
+        "start_index is out of range"
+    );
+    assert!(
+        length >= 0 && start_index.checked_add(length).unwrap() < total_length,
```
sdroege, line 112: This is wrong unfortunately, sorry for missing that. Needs to be `<= total_length`.

sdroege: https://github.com/gtk-rs/gtk-rs-core/pull/1816

PR #1818, opened 2025-09-23
Lower the amount of required dependencies

elmarco:
Allow to build glib crates with lower external dependencies:
- make "proc-macro-crate" optional in glib-macros
- make "futures*" crates optional in glib

### DIFF
--- a/glib-macros/src/utils.rs
+++ b/glib-macros/src/utils.rs
@@ -1,7 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use proc_macro2::{Ident, Span, TokenStream};
-use proc_macro_crate::crate_name;
 use quote::{quote, quote_spanned, ToTokens};
 use syn::{
     meta::ParseNestedMeta, parse::Parse, punctuated::Punctuated, spanned::Spanned, token::Comma,
@@ -185,8 +184,9 @@ pub fn parse_optional_nested_meta_items<'a>(
     }
 }
 
+#[cfg(feature = "proc_macro_crate")]
 pub fn crate_ident_new() -> TokenStream {
-    use proc_macro_crate::FoundCrate;
+    use proc_macro_crate::{crate_name, FoundCrate};
 
     match crate_name("glib") {
         Ok(FoundCrate::Name(name)) => Some(name),
@@ -212,6 +212,12 @@ pub fn crate_ident_new() -> TokenStream {
         let glib = Ident::new("glib", Span::call_site());
         quote!(#glib)
     })
+}
+
+#[cfg(not(feature = "proc_macro_crate"))]
+pub fn crate_ident_new() -> TokenStream {
+    let glib = Ident::new("glib", Span::call_site());
+    quote!(#glib)
 }
 
 // Generate i32 to enum mapping, used to implement
--- a/glib/Cargo.toml
+++ b/glib/Cargo.toml
@@ -19,11 +19,11 @@ name = "glib"
 [dependencies]
 libc.workspace = true
 bitflags.workspace = true
-futures-core = { version = "0.3", default-features = false }
-futures-task = { version = "0.3", default-features = false }
-futures-executor = "0.3"
-futures-channel = "0.3"
-futures-util = "0.3"
+futures-core = { version = "0.3", default-features = false, optional = true }
+futures-task = { version = "0.3", default-features = false, optional = true }
+futures-executor = { version = "0.3", optional = true }
+futures-channel = { version = "0.3", optional = true }
+futures-util = { version = "0.3", optional = true }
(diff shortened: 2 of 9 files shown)


glib-macros/src/utils.rs
```diff
@@ -214,6 +214,12 @@ pub fn crate_ident_new() -> TokenStream {
     })
 }
 
+#[cfg(not(feature = "proc_macro_crate"))]
+pub fn crate_ident_new() -> TokenStream {
+    let glib = Ident::new("glib", Span::call_site());
+    quote!(#glib)
```
sdroege, line 220: This will break the macros when used from `glib` (so `glib` needs to enable the feature which makes this change rather pointless), and also in various other situations (renaming dependencies, ...)

glib/src/functions.rs
```diff
@@ -283,6 +283,7 @@ pub fn file_open_tmp(
 ///
 /// This can be called from any thread and will execute the future from the thread
 /// where main context is running, e.g. via a `MainLoop`.
+#[cfg(feature = "futures")]
 pub fn spawn_future<R: Send + 'static, F: std::future::Future<Output = R> + Send + 'static>(
```
sdroege, line 287: How do these functions depend on the futures crate? They should work with just `std`

glib/src/lib.rs
```diff
@@ -228,12 +228,18 @@ pub use self::bridged_logging::{
 #[macro_use]
 pub mod subclass;
 
+#[cfg(feature = "futures")]
 mod main_context_futures;
+#[cfg(feature = "futures")]
 pub use main_context_futures::{JoinError, JoinHandle, SpawnWithinJoinHandle};
```
sdroege, line 237: The usage of the futures crate should be possible to remove / make optional in here

glib/src/lib.rs
```diff
@@ -228,12 +228,18 @@ pub use self::bridged_logging::{
...
+#[cfg(feature = "futures")]
 mod source_futures;
+#[cfg(feature = "futures")]
 pub use self::source_futures::*;
```
sdroege, line 241: These are a bit harder because of the channels

glib/Cargo.toml
```diff
@@ -39,7 +39,7 @@ trybuild2 = "1"
 criterion = "0.7.0"
 
 [features]
-default = ["gio"]
+default = ["gio", "futures"]
```
sdroege, line 42: `gio` (the crate) needs to enable this feature, and others too probably

glib-macros/src/utils.rs
elmarco: why would it break if the crate name is glib? it doesn't break for glib build itself, what do you mean?

glib/src/functions.rs
elmarco: right

glib/Cargo.toml
elmarco: it's enabled by default, do you think we need to be explicit about it?

glib-macros/src/utils.rs
sdroege: You can't reference `glib` as `glib` from inside itself. The tests should be failing if this feature is disabled (because nothing other than tests currently uses `glib-macros`).

glib/Cargo.toml
sdroege: I think ideally all dependencies on glib would go with `default-features = false` after this PR (so it actually has an effect if you use anything on top of glib), so being explicit about it would be useful

glib/Cargo.toml
```diff
@@ -19,11 +19,11 @@ name = "glib"
...
-futures-channel = "0.3"
-futures-util = "0.3"
+futures-core = { version = "0.3", default-features = false, optional = true }
+futures-task = { version = "0.3", default-features = false, optional = true }
+futures-executor = { version = "0.3", optional = true }
+futures-channel = { version = "0.3", optional = true }
+futures-util = { version = "0.3", optional = true }
 glib-sys.workspace = true
 gobject-sys.workspace = true
 glib-macros.workspace = true
```
sdroege: You probably need `default-features=false` here in some way

glib-macros/src/utils.rs
elmarco: I think glib tests use "glib" name too. otherwise some trick like "extern crate self as glib;" should help, but apparently not needed.

glib/src/lib.rs
elmarco: not so easy, as it needs futures_channel::oneshot. Not sure it's worth it at this point.

sdroege (approved):

sdroege:
@bilelmoussaoui Good for you too?

PR #1827, opened 2025-09-29
glib: fix GStringBuilder::copy_into

elmarco:


elmarco:
btw, is BoxedInline for GString best solution? it does a copy when calling from_glib_full()...

sdroege:
> btw, is BoxedInline for GString best solution? it does a copy when calling from_glib_full()...

It should probably steal the contained string in that case, but also where are you passing `GString`s in public API?

PR #1828, opened 2025-09-30
Streams for D-Bus signals

swsnr:
As briefly discussed in #1280 I'd like to add means to receive signals as streams, which is a lot more convenient to use than callbacks.


swsnr:
@sdroege This PR doesn't actually do what it's said to do yet :grimacing: but I'd like to have some early feedback on the current commit which replaces the `SignalSubscriptionId` with a droppable object to which the lifetime of the signal subscription is attached.

This change is a bit of a prerequisite for actually receiving a signal as stream, because making a signal subscription droppable makes it a lot easier to manage the subscription in the context of streams and futures.

I've left comments on the diff with specific questions.

gio/src/dbus_connection.rs
```diff
@@ -450,7 +462,7 @@ impl DBusConnection {
         arg0: Option<&str>,
         flags: DBusSignalFlags,
         callback: P,
-    ) -> SignalSubscriptionId {
```
swsnr, line 453: @sdroege This is incompatible, obviously.  Is this okay? Or would you rather have me mark `signal_subscribe` and `signal_unsubscribe` as deprecated, and add a new separate method.  If so, can you suggest a new for this new method?

gio/src/dbus_connection.rs
```diff
@@ -117,8 +117,20 @@ pub struct ActionGroupExportId(NonZeroU32);
 pub struct MenuModelExportId(NonZeroU32);
 #[derive(Debug, Eq, PartialEq)]
 pub struct FilterId(NonZeroU32);
+
 #[derive(Debug, Eq, PartialEq)]
-pub struct SignalSubscriptionId(NonZeroU32);
+struct SignalSubscriptionId(NonZeroU32);
+
+#[derive(Debug)]
+pub struct SignalSubscription(DBusConnection, Option<SignalSubscriptionId>);
```
swsnr, line 137: @sdroege In this implementation, the signal subscription keeps a strong reference to the connection so a signal subscription in and by itself is sufficient to keep the connection alive, in my understanding.

I think that's what you'd normally want to have, but I also thought about making the connection a weakref here.  Thoughts?

sdroege:
> PS: I'm not really following the CI errors; are these mine?

No, you can ignore them. Things changed on the github side and now stuff is failing.

gio/src/dbus_connection.rs
sdroege: If it keeps a strong reference then you'd easily create reference cycles, or not? A weak reference might be better

gio/src/dbus_connection.rs
sdroege: Breaking the API is OK but we obviously can't backport this then and it will have to wait until the next breaking release next summer.

bilelmoussaoui: Do it in two steps: one that deprecate the current methods & add new ones & a new commit dropping the deprecated methods & rename the new ones. This way we can backport only the first commit

gio/src/dbus_connection.rs
swsnr: Exactly.

But on the other hand, if the subscription just holds a weak reference then holding on to the subscription alone wouldn't actually maintain the subscription, would it?  I'm not too familiar with GDBus, but presumably all signal subscriptions stop when the connection is destroyed.

So if the subscription doesn't hold a strong reference to the connection, then something like the following:

```rust
let connection = gio::bus_get_future(BusType::Session).await?;
let subscription = connection.signal_subscribe(...)?;
self.subscription_to_my_important_signal.replace(Some(subscription))
```

wouldn't actually work, would it?  

The new `connection` drops when leaving the surrounding block, there's no other strong reference to it, so the connection would be destroyed, taking the signal with it, and the callback given to `.signal_subscribe` would never be invoked.  

I believe that this would be very surprising, and debugging things that _don't_ happen is a bloody nuisance, so this API would arguably not follow the principle of least surprise.

How about adding a `.downgrade()` method on signal subscription to convert to a `WeakSignalSubscription` which only holds a weak reference?  That'd follow the normal API pattern here (APIs normally return strong references and need to be downgraded explicitly), and downstream consumers could decide whether the need a weak or a strong ref.

gio/src/dbus_connection.rs
swsnr: Two steps, but in MR?  That I can do.  Suggestions for a new name?  I'd go for `subscribe_signal`, which is arguably the nicer name anyway.  Slightly confusing if we have `signal_subscribe` and `subscribe_signal` but if the former is clearly deprecated it's perhaps fine?

swsnr: I've added this as a separate `subscribe_to_signal` method, and marked `signal_subscribe` and `signal_unsubscribe` as deprecated.   I'm open for a better name for the new method.

I didn't remove `signal_subscribe` and `signal_unsubscribe`, because the new method is actually implemented on top of these, which sort of nicely separates the FFI binding boilerplate from the Rust-level interface, and `signal_subscribe` at least also offers valuable documentation which I couldn't copy to the new method due to the licensing situation.  

In the next breaking release you could then either just remove the `pub`, or completely inline both, or even indefinitely retain them as deprecated methods, if only for documentation.

gio/src/dbus_connection.rs
swsnr: I've added a `WeakSignalSubscription` counterpart with converse `upgrade`/`downgrade` methods, so that the call to `subscribe_to_signal` now behaves as I believe is least surprising, while still allowing the caller to break reference cycles by downgrading to a weak reference on connection.

swsnr:
@sdroege @bilelmoussaoui I've addressed your answers to my questions, and while I was at it I also pushed a second commit which replaces the lengthy callback signature with four anonymous `&str`s with a struct, so as to properly name things.  I believe that this makes it much easier to understand the signature, and it probably also plays a lot better with editor support, e.g auto-completion.

I don't cling to this tho, so if you dislike that I can back out that commit.

If you agree with the proposed API I'd continue with implementing the `Stream` support for signals, either in this PR with new commits or in a separate one after merging this one, whatever you prefer.

gio/src/dbus_connection.rs
```diff
@@ -515,4 +725,89 @@ impl DBusConnection {
...
+        &self,
+        sender: Option<&str>,
+        interface_name: Option<&str>,
+        member: Option<&str>,
+        object_path: Option<&str>,
+        arg0: Option<&str>,
+        flags: DBusSignalFlags,
+        map_signal: F,
+    ) -> SubscribedSignalStream<SignalSubscription, impl Stream<Item = T> + use<T, F>> {
+        let (tx, rx) = mpsc::unbounded();
```
swsnr, line 757: @sdroege I'm not sure how to do back-pressure here if signals arrive faster than the app consumes them from the channel; currently I just keep filling up the channel until the processes runs out of memory.  That's easy to implement, but probably not a good idea?

sdroege: The only other options you have are dropping once a threshold is reached, or blocking.

Do you know how this is handled in zbus? Probably "blocking" (asynchronously)?

gio/src/dbus_connection.rs
sdroege: Maybe some example for this would be useful to have. I don't know how this would be usually used :)

gio/src/dbus_connection.rs
```diff
@@ -515,4 +725,86 @@ impl DBusConnection {
...
+    /// [`SubscribedSignalStream::downgrade`].
+    pub fn receive_signal<T: 'static, F: Fn(DBusSignalRef) -> T + 'static>(
+        &self,
+        sender: Option<&str>,
+        interface_name: Option<&str>,
+        member: Option<&str>,
+        object_path: Option<&str>,
+        arg0: Option<&str>,
+        flags: DBusSignalFlags,
+        map_signal: F,
```
sdroege, line 755: Why this instead of just having the user call e.g. https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.map ?

sdroege (commented):
Looks good to me otherwise

gio/src/dbus_connection.rs
swsnr: I don't think we can synchronously block here; presumably this callback runs on the main context, so if it blocked you'd have to be careful to put the future consuming the stream on another thread, to avoid deadlocks, wouldn't you?

We could block asynchronously; I did consider this, but then we'd have to schedule another future, and I believe we'd loose order of signal emissions, wouldn't we?

swsnr: PS: I presume zbus just back pressures all the way to the socket buffer, or maintains an intermediate async buffer for back press.  Zbus has the advantage of being all async, so it can do proper backpressure.

But I don't know actually 🤷

With an unbounded channel the downstream consumer can at least make sure to consume all signals right away, and then add their own buffering to constraint memory consumption. 

Alternatively we could use an arbitrary buffer here, and then drop signal emissions when it's full, but I don't really like that 😬

sdroege: Yeah blocking the main context is not going to be very useful indeed, especially as the receiving end of the channel is likely handled on the main context too :)

I don't really have any suggestions here. Maybe just keep it like this and document it carefully, and then we'll see how it works in practice?

swsnr: > I don't really have any suggestions here. Maybe just keep it like this and document it carefully, and then we'll see how it works in practice?

Sure. I'll add some extra documentation and leave the implementation as is.

gio/src/dbus_connection.rs
swsnr: The callback gets references to the sender, objpath, etc. and the parameters variant.  If I wanted to pass these downstream first I'd have to clone them, wouldn't I? I wanted to avoid these extra allocations.

gio/src/dbus_connection.rs
swsnr: Mh, I had hoped that it's sort of obvious 😅

Essentially you just keep the returned subscription somewhere and drop it when you no longer want to receive the signal.  The default subscription keeps a strong reference on the connection, the weak one keeps a weak reference.

I had hoped that this simplifies lifecycle management for signal subscriptions, but if it actually needs an example to be understandable, the API is probably not that good?😬

sdroege: No that part is clear. I mean, how would application code making use of all this actually look like? Keep in mind that I only have knowledge about how DBus works, I never really wrote code for it :)

gio/src/dbus_connection.rs
sdroege: Ah that makes sense, yes

gio/src/dbus_connection.rs
swsnr: Ah I see. I'll make an example then which shows how to use this API in different ways 🙂

gio/src/dbus_connection.rs
swsnr: I'll leave it as is then.  I guess most of the time folks will use `receive_signal_parameters` anyway, as that's the most common case, I guess.

gio/src/dbus_connection.rs
swsnr: I've added another paragraph to the docs to clarify that the stream uses an unbounded channel internally.

gio/src/dbus_connection.rs
swsnr: @sdroege I added a simple example; is that what you had in mind?

PR #1829, opened 2025-10-02
gio: Add DBusObjectManagerClient

bash:
Sadly all constructors are generated manually:
* `new`/`new_future` and `new_for_bus`/`new_for_bus_future` were not generated at all by gir (the `*_future` versions were generated, but I don't know how to set only the callback version to manually generated)
* `new_sync` and `new_for_bus_sync` were generated mostly fine. I only had to fix how the cancellable param is passed to ffi:
  ```patch
  - cancellable.to_glib_none().0,
  + cancellable.map(|p| p.as_ref()).to_glib_none().0,
  ```
  This looks like a bug in gir because for other methods with a cancellable the `.map(...)` part is added just fine.


gio/src/auto/dbus_object_manager_client.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    //#[doc(alias = "get-proxy-type-func")]
+    //fn get_proxy_type_func(&self) -> /*Unimplemented*/Basic: Pointer {
+    //    ObjectExt::property(self.as_ref(), "get-proxy-type-func")
+    //}
+
+    //#[doc(alias = "get-proxy-type-user-data")]
+    //fn get_proxy_type_user_data(&self) -> /*Unimplemented*/Basic: Pointer {
+    //    ObjectExt::property(self.as_ref(), "get-proxy-type-user-data")
+    //}
```
bilelmoussaoui: Those are C APIs, so drop them by ignoring them from the Gir.toml

gio/src/dbus_object_manager_client.rs
```diff
@@ -0,0 +1,449 @@
...
+            );
+        }
+    }
+
+    pub fn new_future(
+        connection: &DBusConnection,
+        flags: DBusObjectManagerClientFlags,
+        name: &str,
+        object_path: &str,
+        get_proxy_type_func: Option<DBusProxyTypeFn>,
```
bilelmoussaoui, line 277: This makes the function signature a bit meh... maybe we should split those into two variants, new_future_with_fn and new_future?

gio/src/dbus_object_manager_client.rs
```diff
@@ -0,0 +1,449 @@
...
+                Some(cancellable),
+                move |res| {
+                    send.resolve(res);
+                },
+            );
+        }))
+    }
+}
+
+impl DBusObjectManagerClient {
```
bilelmoussaoui: Why the separate impl here?

tautropfli: My goal was to group them by "reason" why they're hand-written and not auto-generated.
But I suppose that is already documented in Gir.toml

gio/src/dbus_object_manager_client.rs
tautropfli: Ah, that's a good idea! I suppose the `new_future_with_fn` takes a `get_proxy_type_func` that is non-`Option` then, right?

gio/src/dbus_object_manager_client.rs
tautropfli: Fixed in [9d74f99](https://github.com/gtk-rs/gtk-rs-core/pull/1829/commits/9d74f99f846972fcdb1243de57d1de449788d3ee)

gio/src/auto/dbus_object_manager_client.rs
tautropfli: Fixed in [627529d](https://github.com/gtk-rs/gtk-rs-core/pull/1829/commits/627529de6871ea795e9dacaa907bac85a4b696ae)

bash:
I noticed that the implementation of GDBusObjectManagerClient [uses a mutex](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdbusobjectmanagerclient.c#L132) internally, so I marked it as `send+sync` in Gir.toml. 

Sadly the docs don't mention the MT safety like they [do for other types such as [`DBusProxy`](https://docs.gtk.org/gio/class.DBusProxy.html), so I'm not entirely sure 😅 :
> A `GDBusProxy` instance can be used from multiple threads [...]

gio/src/dbus_object_manager_client.rs
tautropfli: Done in [dc00a1d](https://github.com/gtk-rs/gtk-rs-core/pull/1829/commits/dc00a1dfc8df01d1a74ebd43a3f904cb6a9516c4)

bilelmoussaoui:
You can update the gir submodule & regen, that should get rid of some of the manual code i hope :)

bash:
Wohoo that worked like a charm ✨

gio/src/dbus_object_manager_client.rs
```diff
@@ -0,0 +1,379 @@
...
+    // The checker tries to add a doc alias for `g_dbus_object_manager_client_new_finish`.
+    // checker-ignore-item
+    #[doc(alias = "g_dbus_object_manager_client_new")]
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new<P: FnOnce(Result<DBusObjectManagerClient, glib::Error>) + Send + Sync + 'static>(
+        connection: &DBusConnection,
+        flags: DBusObjectManagerClientFlags,
+        name: &str,
+        object_path: &str,
+        get_proxy_type_func: Option<DBusProxyTypeFn>,
```
sdroege, line 150: This is not useful API. You can't construct a `None` very easily. It would be better to have two variants of this function, one with callback and one without

sdroege: Ah no, `None` is easy because you have it already defined as `Box<dyn>`. But the case where a callback is provided is not very convenient now as it requires wrapping in `Some()` and `Box::new()`.

sdroege (commented):
Looks good to me otherwise. Do you also want to provide a small example making use of this?

gio/src/dbus_object_manager_client.rs
tautropfli: Ah dang it, I changed the async versions already to use two variants but forgot to do the same for the sync and the async+callback ones. Fixed in [23dac07](https://github.com/gtk-rs/gtk-rs-core/pull/1829/commits/23dac07d9bd25904866cf546e7f541cbfa7c2465) and [9728dcf](https://github.com/gtk-rs/gtk-rs-core/pull/1829/commits/9728dcf9c484d4c8a4e676ade4130071a13e129e)

bash:
> Do you also want to provide a small example making use of this?

Yes, my plan was to make `DBusProxy` subclassable and then do an example with UDisks (because that's what I am using it for right now ^^).

PR #1830, opened 2025-10-03
glib: Add Error::with_domain constructor and code getter

thiblahute:
Add a constructor to create errors from a domain quark and code, useful when creating errors with the same domain/code but different messages as we can't get a rust type which implements the `ErrorDomain` in that case.

This also adds a code() getter method to retrieve the error code.

### DIFF
--- a/glib/src/error.rs
+++ b/glib/src/error.rs
@@ -40,6 +40,33 @@ impl Error {
     }
 
     // rustdoc-stripper-ignore-next
+    /// Creates an error with supplied error domain quark, code and message.
+    ///
+    /// This is useful when you need to create an error with the same domain and code
+    /// as an existing error but with a different message.
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// let original = Error::new(FileError::Failed, "Original message");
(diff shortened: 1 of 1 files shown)


glib/src/error.rs
```diff
@@ -39,6 +39,33 @@ impl Error {
...
+    /// ```ignore
+    /// let original = Error::new(FileError::Failed, "Original message");
+    /// let modified = Error::from_quark(
+    ///     original.domain(),
+    ///     original.code(),
+    ///     "Modified message"
+    /// );
+    /// ```
+    #[doc(alias = "g_error_new_literal")]
+    pub fn from_quark(domain: Quark, code: i32, message: &str) -> Error {
```
bilelmoussaoui: Maybe this should be named, with_domain? It has nothing to do with a quark and it is not a traditional From impl to use the from prefix

thiblahute: Much better naming indeed. fixed.

bilelmoussaoui (approved):

sdroege (approved):

PR #1832, opened 2025-10-07
gio: Allow subclassing DBusProxy

tautropfli:


gio/src/subclass/dbus_proxy.rs
```diff
@@ -0,0 +1,112 @@
...
+};
+
+pub trait DBusProxyImpl:
+    ObjectImpl
+    + AsyncInitableImpl
+    + DBusInterfaceImpl
+    + InitableImpl
+    + ObjectSubclass<Type: IsA<DBusProxy>>
+{
+    fn g_properties_changed(
```
bilelmoussaoui, line 20: Remove the g_ prefix

sdroege: That's how it's called though. The signal is `g-properties-changed` for whatever reason :)

gio/src/subclass/dbus_proxy.rs
```diff
@@ -0,0 +1,112 @@
...
+                    invalidated_properties.to_glib_none().0,
+                );
+            }
+        }
+    }
+
+    fn parent_g_signal(
+        &self,
+        sender_name: Option<&str>,
+        signal_name: &str,
```
sdroege:
```suggestion
        sender_name: Option<&glib::GStr>,
        signal_name: &glib::GStr,
```

maybe so we don't have to do heap allocations for all the NUL-termination?

examples/gio_dbus_object_manager/block_proxy.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    #[object_subclass]
+    impl ObjectSubclass for UDisksBlockProxy {
+        const NAME: &'static str = "SampleApplicationUDisksBlockProxy";
+        type Type = super::UDisksBlockProxy;
+        type ParentType = DBusProxy;
+    }
+
+    #[derived_properties]
```
bilelmoussaoui:
```suggestion
    #[glib::derived_properties]
```

gio/src/subclass/dbus_proxy.rs
tautropfli: Yes, all of the [signals](https://docs.gtk.org/gio/signal.DBusProxy.g-properties-changed.html) and [properties](https://docs.gtk.org/gio/property.DBusProxy.g-bus-type.html) on DBusProxy have the `g-` prefix. Probably to avoid name clashes with properties on derived classes :)

PR #1840, opened 2025-10-29
Use BTreeMap instead of HashMap

bmwiedemann:
Use `BTreeMap` instead of `HashMap`
to have deterministic output ordering
to allow for reproducible builds.

See https://reproducible-builds.org/ for why this is good.

Without this patch, ArchLinux' and [openSUSE's `identity` package](https://build.opensuse.org/package/show/openSUSE:Factory/identity) varied between builds.

Special thanks go to @kpcyrd for help in debugging this.

### DIFF
--- a/glib-macros/src/properties.rs
+++ b/glib-macros/src/properties.rs
@@ -5,7 +5,7 @@ use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::format_ident;
 use quote::{quote, quote_spanned};
-use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


bmwiedemann:
I did some more testing and found that this patch helps to make more openSUSE package builds fully deterministic: `mousai identity eww Fragments fractal gstreamer-plugins-rs contrast`

Regarding `IndexMap`:
users had random ordering before, so my guess is that they do not care much if it is alphabetic or insertion order. And performance is not a concern during build (assuming this code is only used during build).

PR #1845, opened 2025-11-14
glib/bytes: relax start offset constraint for `from_bytes`

BiagioFesta:
Maybe I've missed something, but it was faster opening a one-line PR for seeking feedback.

In rust this is totally valid:
```rust
let bytes = b"1234";
let sub = &bytes[4..];
```

But this is not in `glib` at the moment:
```rust
let bytes = glib::Bytes::from_static(b"1234");
let sub = glib::Bytes::from_bytes(&bytes, 4..);  // This crashes without this patch
```

This patch does not contradict [Bytes.new_from_bytes](https://docs.gtk.org/glib/ctor.Bytes.new_from_bytes.html) 'cause `end + 0 <= len` holds.

### DIFF
--- a/glib/src/bytes.rs
+++ b/glib/src/bytes.rs
@@ -109,7 +109,7 @@ impl Bytes {
             Bound::Excluded(v) => v.checked_add(1).expect("Invalid start offset"),
             Bound::Unbounded => 0,
         };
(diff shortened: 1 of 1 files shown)


sdroege (approved):
Thanks, this makes sense

Issue #46, opened 2021-05-14
No way to know when a cancelled Future async operation is actually cancelled


sdroege:
https://github.com/rust-lang-nursery/futures-rs/issues/1278#issuecomment-428071710

Problem here is that the callback-based API would give us a `Cancelled` error at that time, while simply dropping the future allows for no way of signalling that the cancellation has happened. But before cancellation happened, it is impossible to start a new async operation on most objects.

kelnos:
I just got curious about this myself when looking into cancellation for futures running via `MainContext.spawn()`/`.spawn_local()`.

My use case here is a file/folder-view widget.  I want to asynchronously enumerate and load files into the view, but if the user navigates to a new folder while the enumeration is still happening, it should cancel the in-progress enumeration.  I experimented with holding onto the `SourceId` and then calling `.remove()` on it, and that does seem to work?  The stuff in the source does indeed stop running, though I can't be 100% sure that the `GioFuture` that was running inside it actually got dropped (thus causing the `Cancellable` to trigger cancellation on the GIO operation).  (Reading through the `TaskSource` source, I do think this does work how I think it does, though.)

I guess this is also more about Rust async/await behavior: I hope that when there is a Future that wraps an async function, and that function has _another_ Future that is being `.await`'ed on, I expect that the inner future gets dropped as well when the outer future gets dropped; I can't imagine how all this would work otherwise.  But I also expected the GIO operation to return with `gio::IOErrorEnum::Cancelled`, but I guess I can see why that doesn't happen (since the outer future is no longer running at all?).

At any rate, I'm not even sure how this would work in my situation.  Since I'm using `spawn()`, I "lose" the `GioFuture`, so the `TaskSource` that ends up getting built doesn't even know there's a GIO operation running inside it.

One solution I can think of that doesn't involve API breakage might be to add a `.cancellable()` function to `GioFuture`, which returns the underlying `gio::Cancellable`.  Then that could have a new function that returns a Future that waits on the underlying GIO cancellation to finish.  (Or, hmm, `Cancellable` itself could just implement `Future`?)  So any time I call a function that creates a `GioFuture`, I could grab the `Cancellable`, connect to its `cancelled` signal, stash it somewhere, and `.await` if I need to.  The downside here is that the call sites get much more verbose; I would have to do something like:

```rust
let future = gio_file.enumerate_children_future(...);
let cancellable = future.cancellable();
cancellable.connect_cancelled(|cancellable| {
[...]

PR #1846, opened 2025-11-18
CancellableFuture: Some cleanups and few more tests

3v1n0:
Coming from #46 and handled some of the initial changes I was suggesting.

Sadly main improvements would require an API break, so I did just some minimal lifting.

### DIFF
--- a/gio/src/cancellable_future.rs
+++ b/gio/src/cancellable_future.rs
@@ -171,7 +171,7 @@ where
 
 impl From<Cancelled> for glib::Error {
     fn from(_: Cancelled) -> Self {
-        glib::Error::new(IOErrorEnum::Cancelled, "Task cancelled")
+        glib::Error::new(IOErrorEnum::Cancelled, "Operation was cancelled")
     }
 }
 
@@ -179,7 +179,7 @@ impl std::error::Error for Cancelled {}
 
 impl Debug for Cancelled {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "Task cancelled")
+        write!(f, "Operation was cancelled")
     }
 }
 


gio/src/cancellable_future.rs
```diff
@@ -32,14 +37,59 @@ pin_project! {
...
         #[pin]
         future: F,
 
-        #[pin]
-        waker_handler_cb: Option<CancelledHandlerId>,
-
         cancellable: Cancellable,
+        connection_id: Option<CancelledHandlerId>,
+
+        waker: std::sync::Arc<std::sync::Mutex<Option<std::task::Waker>>>,
```
BiagioFesta, line 83: i'd avoid fullspec, `Arc` and `Mutex` can be imported to improve readability (no other `Arc` and `Mutex`, so there is no ambiguity)

Same comment in other part of this diff

gio/src/cancellable_future.rs
```diff
@@ -87,49 +149,37 @@ where
     type Output = Result<<F as Future>::Output, Cancelled>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        if self.is_cancelled() {
-            return Poll::Ready(Err(Cancelled));
+        if self.cancellable.is_cancelled() {
+            // XXX: Whenever we want to break the API, we should return here only
+            // self.cancellable.set_error_if_cancelled() value.
+            return std::task::Poll::Ready(Err(Cancelled));
```
BiagioFesta, line 155: just `Poll::Ready`, useless full-spec

gio/src/cancellable_future.rs
```diff
@@ -87,49 +149,37 @@ where
     type Output = Result<<F as Future>::Output, Cancelled>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        if self.is_cancelled() {
-            return Poll::Ready(Err(Cancelled));
+        if self.cancellable.is_cancelled() {
```
BiagioFesta, line 152: why this change, equivalent to before

gio/src/cancellable_future.rs
```diff
@@ -51,10 +101,22 @@ impl<F> CancellableFuture<F> {
     /// immediately without making any further progress. In such a case, an error
     /// will be returned by this future (i.e., [`Cancelled`]).
     pub fn new(future: F, cancellable: Cancellable) -> Self {
+        let waker = std::sync::Arc::new(std::sync::Mutex::new(None::<std::task::Waker>));
+        let connection_id = cancellable.connect_cancelled(glib::clone!(
+            #[strong]
+            waker,
+            move |_| {
+                if let Some(waker) = waker.lock().unwrap().take() {
```
BiagioFesta, line 109: personal opinion, but I generally never use `unwrap` in PROD code. Rather `.expect("msg")` where `"msg"` is a brief explanation of the logical invariant which holds the condition of why you think that never happens.

Feel free to ignore this comment tho, maybe on `lock` is quite self-explanatory

gio/src/cancellable_future.rs
```diff
@@ -87,49 +149,37 @@ where
...
-            return Poll::Ready(Err(Cancelled));
+        if self.cancellable.is_cancelled() {
+            // XXX: Whenever we want to break the API, we should return here only
+            // self.cancellable.set_error_if_cancelled() value.
+            return std::task::Poll::Ready(Err(Cancelled));
         }
 
-        let mut this = self.as_mut().project();
+        let mut waker = self.waker.lock().unwrap();
+        if waker.is_none() {
```
BiagioFesta, line 159: This changes the logical behavior.

In rust there is no guarantee the a future is polled with the same waker. The waker can change and here we are ignoring the update. The previous code was intentional made because of this.

gio/src/cancellable_future.rs
```diff
@@ -87,49 +149,37 @@ where
...
+            *waker = Some(cx.waker().clone());
```
BiagioFesta, line 160: Nit: `Arc::clone(&t)` is more explicit about why and how the global counter is used

gio/src/cancellable_future.rs
```diff
@@ -87,49 +149,37 @@ where
...
 }
 
 impl From<Cancelled> for glib::Error {
     fn from(_: Cancelled) -> Self {
-        glib::Error::new(IOErrorEnum::Cancelled, "Task cancelled")
+        glib::Error::new(IOErrorEnum::Cancelled, "Operation was cancelled")
     }
 }
 
 impl std::error::Error for Cancelled {}
```
BiagioFesta, line 178: I would move the `XXX:...` comment of yours here, mentioning that maybe we can get rid of a specific error and just wrap the "cancel" error into a glib::Error. With API break of coure

gio/src/cancellable_future.rs
3v1n0: I see. Well we should update the waker then rather than the connection, since the waker is what we can lock on and we've control

BiagioFesta: I have two thoughts on this:

1. The polling pattern could be improved to avoid unnecessary locking. Something like:
```rust
match .poll() {
    Ready() => { .. }
    Pending => {
        self.waker = waker;
    }
}
```
This approach would be cleaner as it only sets the waker when actually needed (on Pending), and eliminates the need for explicit drops, probably. Anyway that should be the typical pattern.

2. As a side consideration: Have you considered the thread-safety implications? Specifically, there could be race conditions if we wake the waker immediately before checking poll. While this may not be an issue given GLib's threading model, it's something I've encountered when working with low-level async Rust code that requires retry loops

PR #1849, opened 2025-11-21
glib: `Fix Regex::match` API

sdroege:


bilelmoussaoui (approved):

glib/src/regex.rs
```diff
@@ -127,11 +128,11 @@ impl Regex {
         string: &'input GStr,
         start_position: i32,
         match_options: RegexMatchFlags,
-    ) -> Result<MatchInfo<'input>, crate::Error> {
+    ) -> Result<(bool, MatchInfo<'input>), crate::Error> {
```
sdroege: This is a breaking API change but the current API makes little sense.

If we want to preserve the API we could omit the `bool` return value and instead require the caller to check `match_info.matches()` which provides the very same information. It's a bit redundant right now but closer to the C API.

@bilelmoussaoui Any opinions?

sdroege: Let's go with this. It's equivalent and has less breakage.

PR #1884, opened 2026-01-11
glib: Use 'static lifetime for nick/name

bilelmoussaoui:


glib/src/enums.rs
```diff
@@ -279,14 +279,14 @@ impl EnumValue {
     // rustdoc-stripper-ignore-next
     /// Get name corresponding to the value.
     #[doc(alias = "get_name")]
-    pub fn name(&self) -> &str {
+    pub fn name(&self) -> &'static str {
```
sdroege:
```suggestion
    pub fn name<'a>(&self) -> &'a str {
```

is IIRC the correct way for reasons I forgot :)

bilelmoussaoui: I thought they were static strings?

sdroege: My version is more generic than yours: it's `'static` or any other lifetime. There was some strange complication with returning `'static` references where that can lead to problems. I'll have to search again.

sdroege (approved):

PR #1885, opened 2026-01-11
Doc: clarify the GObject subclass mechanism

cjao:
The example given in the top-level [subclassing module](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/index.html) is not quite accurate because the wrapper type is not optional.

Update the comments in the example while peeling back some of the magic of how of this framework bridges Rust and GObject. For instance, it wasn't obvious to me just from reading the public documentation how `SimpleObject` "inherits" the public methods of `Object`. Please let me know if I have misunderstood some of the details.

Closes #1870 .

### DIFF
--- a/glib/src/subclass/mod.rs
+++ b/glib/src/subclass/mod.rs
@@ -225,10 +225,10 @@
 //! //                                     ffi::GObject (first member of instance struct)
 //! //                                      |
 //! // |--private data (imp::SimpleObject)--|--instance struct (basic::InstanceStruct)--|
-//! //                    ^
-//! //                    |
-//! //                    |
-//! //               SimpleObject
+//! //                                      ^
+//! //                                      |
+//! //                                      |
+//! //                                 SimpleObject
 //!
 //! pub fn main() {
 //!     let obj = SimpleObject::new();


glib/src/subclass/mod.rs
```diff
@@ -91,7 +100,14 @@
 //!         type Interfaces = ();
 //!     }
 //!
-//!     // Trait that is used to override virtual methods of glib::Object.
+//!     // Trait used to override virtual methods of glib::Object. It requires
+//!     // that the associated `Type` implement the `IsA<Object>` trait declaring that
```
sdroege:
```suggestion
//!     // that the associated `Type` implements the `IsA<Object>` trait declaring that
```

glib/src/subclass/mod.rs
```diff
@@ -166,7 +182,13 @@
...
 //! }
 //!
-//! // Optionally, define a wrapper type to make it more ergonomic to use from Rust
+//!
+//! // Create a wrapper type implementing:
+//! // - the basic traits to support Rust memory management functionality on the
+//! //   raw GType instance pointer defined above
+//! // - the core `IsA<Object>` trait declaring that `SimpleObject` is a subclass of `Object`
+//! // - any public methods on the subclass (what would be declared in a public header file if
+//! //   one were implementing `SimpleObject` in C)
```
sdroege: What do you mean with that?

glib/src/subclass/mod.rs
```diff
@@ -178,10 +200,30 @@
...
+//! //                    | (see `unsafe impl<T: ObjectImpl>IsSubclassable<T> for Object`)
+//! //                   ffi::GObjectClass
+//! //                    |
+//! //                   ffi::GObject (first member of instance struct)
+//! //                    |
+//! // |---private data---|---instance struct (basic::InstanceStruct)---|
+//! //        |           |
+//! //  imp::SimpleObject |
+//! //                    |
+//! //               SimpleObject
```
sdroege: I think it would be good to mark pointer direction here, e.g. by using `^`

cjao: Added. I'm omitting the `PrivateStruct` wrapper around `imp::SimpleObject` unless you think that is an important detail.

glib/src/subclass/mod.rs
cjao: I've tried to clarify this.

glib/src/subclass/mod.rs
sdroege: I don't know how important it is to include that :)

sdroege (approved):

sdroege:
Thanks!

PR #1887, opened 2026-01-24
Add gio::subclass::FileEnumerator async vfuncs

fbrouille:
Add support of async vfuncs for `gio::subclass::FileEnumerator` as it is done in https://github.com/gtk-rs/gtk4-rs/blob/main/gdk4/src/subclass/content_provider.rs

### DIFF
--- a/gio/src/subclass/file_enumerator.rs
+++ b/gio/src/subclass/file_enumerator.rs
@@ -1,8 +1,12 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use glib::{prelude::*, subclass::prelude::*, translate::*};
+use std::pin::Pin;
 
-use crate::{Cancellable, FileEnumerator, FileInfo, IOErrorEnum, ffi, prelude::*};
+use crate::{
+    AsyncResult, Cancellable, FileEnumerator, FileInfo, GioFuture, IOErrorEnum, LocalTask, ffi,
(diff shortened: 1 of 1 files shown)


gio/src/subclass/file_enumerator.rs
```diff
@@ -14,6 +18,16 @@ pub trait FileEnumeratorImpl: ObjectImpl + ObjectSubclass<Type: IsA<FileEnumerat
         self.parent_next_file(cancellable)
     }
 
+    fn next_files_future(
+        &self,
+        num_files: i32,
+        priority: glib::Priority,
+    ) -> std::pin::Pin<
+        Box<dyn std::future::Future<Output = Result<Vec<FileInfo>, glib::Error>> + 'static>,
```
sdroege:
```suggestion
        Box<dyn std::future::Future<Output = Result<glib::PtrSlice<FileInfo>, glib::Error>> + 'static>,
```

would be better here as otherwise you just copy from that to a Vec and back all the time

fbrouille: I opted for `glib::List` since converting the raw `GList` pointer to `glib::PtrSlice` was non-trivial.

sdroege: Ah the C API uses `GList`. Yes then this is more correct, sorry

PR #1901, opened 2026-02-12
binding_group: add transforms with concrete types

A6GibKm:
See https://github.com/gtk-rs/gtk-rs-core/pull/782.

### DIFF
--- a/glib/src/gobject/binding_group.rs
+++ b/glib/src/gobject/binding_group.rs
@@ -477,6 +477,35 @@ mod test {
         assert_eq!(source.name(), "World");
     }
 
+    #[test]
+    fn binding_from_transform_concrete_change_type() {
+        let binding_group = crate::BindingGroup::new();
+
+        let source = TestObject::default();
+        let target = TestObject::default();
+
+        binding_group.set_source(Some(&source));
+        binding_group
+            .bind("name", &target, "enabled")
+            .sync_create()
+            .bidirectional()
+            .transform_to::<&str, _, _>(|_binding, value| Some(value == "Hello"))
+            .transform_from(
+                |_binding, value: bool| if value { Some("Hello") } else { Some("World") },
+            )
+            .build();
+
+        target.set_enabled(true);
+        assert_eq!(source.name(), "Hello");
+        target.set_enabled(false);
+        assert_eq!(source.name(), "World");
+
+        source.set_name("Hello");
+        assert!(target.enabled());
+        source.set_name("World");
+        assert!(!target.enabled());
+    }
+
     mod imp {
         use std::{cell::RefCell, sync::OnceLock};
 


glib/src/gobject/binding_group.rs
```diff
@@ -76,9 +79,33 @@ impl<'a> BindingGroupBuilder<'a> {
...
 
+    // rustdoc-stripper-ignore-next
+    /// Transform changed property values from the target object to the source object with the given closure.
+    ///
+    /// This function operates on concrete argument and return types.
+    /// See [`Self::transform_from_with_values`] for a version which operates on `glib::Value`s.
+    pub fn transform_from<
+        S: FromValue<'f>,
+        T: Into<Value>,
+        F: Fn(&'f crate::Binding, S) -> Option<T> + Send + Sync + 'static,
```
sdroege: I didn't review that in detail yet but this probably needs to be `for<'f> Fn(&'f Binding, ...)` or not?

A6GibKm: Lifetimes are not my forte, i just copied the other MR. Ill double check later today

Edit: well the crate:: part can be removed

sdroege: Which other MR?

A6GibKm: #782 (as per the mr description)

A6GibKm: I removed the unnecessary `crate::`, moved an added import to the commit where it was actually used, and checked that the code here matches the current state of `BindingBuilder`.

glib/src/gobject/binding_group.rs
```diff
@@ -4,37 +4,38 @@ use std::{fmt, ptr};
...
         target_property: &'a str,
-    ) -> BindingGroupBuilder<'a> {
+    ) -> BindingGroupBuilder<'a, 'f, 't> {
         BindingGroupBuilder::new(self, source_property, target, target_property)
     }
 }
 
-type TransformFn = Option<Box<dyn Fn(&Binding, &Value) -> Option<Value> + Send + Sync + 'static>>;
+type TransformFn<'b> =
+    Option<Box<dyn Fn(&'b Binding, &'b Value) -> Option<Value> + Send + Sync + 'static>>;
```
sdroege, line 23:
```suggestion
type TransformFn = for<'b>
    Option<Box<dyn Fn(&'b Binding, &'b Value) -> Option<Value> + Send + Sync + 'static>>;
```

?

sdroege: Basically, if I'm not confused you can get rid of the additional lifetime parameter everywhere.

A6GibKm: It might be possible to make this work, but I am not sure how.

A6GibKm: This https://github.com/A6GibKm/gtk-rs-core/commit/0c2a2b2e2da960e54fb1120f7d2f09579c297b88 compiles, but throws:
```
error: implementation of `FromValue` is not general enough
   --> glib/src/gobject/binding_group.rs:487:9
    |
487 | /         binding_group
488 | |             .bind("name", &target, "enabled")
489 | |             .sync_create()
490 | |             .bidirectional()
491 | |             .transform_to(|_binding, value: &str| Some(value == "Hello"))
    | |_________________________________________________________________________^ implementation of `FromValue` is not general enough
    |
    = note: `FromValue<'0>` would have to be implemented for the type `&str`, for any lifetime `'0`...
    = note: ...but `FromValue<'1>` is actually implemented for the type `&'1 str`, for some specific lifetime `'1`

error: could not compile `glib` (lib test) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
```

when running cargo test

A6GibKm: Another attempt would be

```
        F: for<'t, S: FromValue<'t>> Fn(&'t Binding, S<'t>) -> Option<T> + Send + Sync + 'static
```

but this relies on https://github.com/rust-lang/rust/issues/108185

sdroege: I'll take a look

sdroege: Yeah I see, sucks :)

Issue #659, opened 2022-04-23
Need to fix some pointer handling for strict pointer provenance


sdroege:
Needs this change and recent nightly
```diff
diff --git a/glib/src/lib.rs b/glib/src/lib.rs
index 09327234df6..2f9aa904544 100644
--- a/glib/src/lib.rs
+++ b/glib/src/lib.rs
@@ -1,3 +1,6 @@
+#![feature(strict_provenance)]
+#![deny(fuzzy_provenance_casts)]
+#![deny(lossy_provenance_casts)]
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 #![cfg_attr(feature = "dox", feature(doc_cfg))]
```

Gives

```
error: strict provenance disallows casting integer `usize` to pointer `*const c_void`
   --> glib/src/log.rs:383:24
    |
383 |                 value: data as *const _,
    |                        ^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> glib/src/lib.rs:2:9
    |
2   | #![deny(fuzzy_provenance_casts)]
    |         ^^^^^^^^^^^^^^^^^^^^^^
    = help: if you can't comply with strict provenance and don't have a pointer with the correct provenance you can use `std::ptr::from_exposed_addr()` instead
help: use `.with_addr()` to adjust a valid pointer in the same allocation, to this address
    |
383 |                 value: (...).with_addr(data),
    |                        ++++++++++++++++    ~

error: under strict provenance it is considered bad style to cast pointer `*const c_void` to integer `usize`
   --> glib/src/log.rs:420:38
    |
420 |         (self.0.length == 0).then(|| self.0.value as usize)
    |                                      ^^^^^^^^^^^^---------
    |                                                  |
    |                                                  help: use `.addr()` to obtain the address of a pointer: `.addr()`
    |
note: the lint level is defined here
   --> glib/src/lib.rs:3:9
    |
3   | #![deny(lossy_provenance_casts)]
    |         ^^^^^^^^^^^^^^^^^^^^^^
    = help: if you can't comply with strict provenance and need to expose the pointer provenance you can use `.expose_addr()` instead

error: under strict provenance it is considered bad style to cast pointer `*mut u8` to integer `usize`
   --> glib/src/subclass/types.rs:828:9
    |
828 |         priv_ptr as usize & (mem::align_of::<PrivateStruct<T>>() - 1) == 0,
    |         ^^^^^^^^---------
    |                 |
[...]

PR #1917, opened 2026-03-01
Use the strict provenance API

carlosmn:
I found #659 and then also saw that the strict provenance is available as of 1.84 which is below the minimum version right now so I thought I'd dry.

Using `pointer::addr()` was straightforward and `offset_ptr_by_bytes{,_mut}()` does make sense that we'd have to throw away provenance as we're figuring out a whole other allocation. At least that what I gather from #1115 though I'm not sure why that is. It looks like this is all the same allocation and we just want pointers into different bits.

For `LogField::new_user_data` I was less sure about. That `usize` we receive can just be a number the caller wants us to keep around, but it could also be a pointer that they're planning to use later on. We have no way to know and we don't have a pointer to start from so we have to create one without any provenance. If it was a pointer to start with, the caller code doing the function would then not have any opportunity to make use of any provenance information. I wonder if for the sake of strict provenance that should be two different calls. Then again I'm not sure if there is anything that could make use of that provenance in this codebase.

While here I noticed that we no longer have to write our own `offset_of!` code as of 1.77 so I updated the calculation.

### DIFF
--- a/glib/src/subclass/types.rs
+++ b/glib/src/subclass/types.rs
@@ -1068,16 +1068,7 @@ pub fn register_type<T: ObjectSubclass>() -> Type {
             mem::size_of::<PrivateStruct<T>>(),
         );
         data.as_mut().private_offset = private_offset as isize;
-
-        // Get the offset from PrivateStruct<T> to the imp field in it. This has to go through
-        // some hoops because Rust doesn't have an offsetof operator yet.
-        data.as_mut().private_imp_offset = {
-            // Must not be a dangling pointer so let's create some uninitialized memory
-            let priv_ = mem::MaybeUninit::<PrivateStruct<T>>::uninit();
-            let ptr = priv_.as_ptr();
-            let imp_ptr = ptr::addr_of!((*ptr).imp);
-            (imp_ptr.addr() as isize) - (ptr.addr() as isize)
-        };
+        data.as_mut().private_imp_offset = mem::offset_of!(PrivateStruct<T>, imp) as isize;
 
         let iface_types = T::Interfaces::iface_infos();
         for (iface_type, iface_info) in iface_types {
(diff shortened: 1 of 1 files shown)


glib/src/subclass/types.rs
```diff
@@ -1068,16 +1068,7 @@ pub fn register_type<T: ObjectSubclass>() -> Type {
...
-        // Get the offset from PrivateStruct<T> to the imp field in it. This has to go through
-        // some hoops because Rust doesn't have an offsetof operator yet.
-        data.as_mut().private_imp_offset = {
-            // Must not be a dangling pointer so let's create some uninitialized memory
-            let priv_ = mem::MaybeUninit::<PrivateStruct<T>>::uninit();
-            let ptr = priv_.as_ptr();
-            let imp_ptr = ptr::addr_of!((*ptr).imp);
-            (imp_ptr.addr() as isize) - (ptr.addr() as isize)
-        };
+	data.as_mut().private_imp_offset = mem::offset_of!(PrivateStruct<T>, imp) as isize;
```
sdroege: Not sure what happened here but this looks like it needs a `cargo fmt` run to indent one level more

glib/src/subclass/types.rs
```diff
@@ -1078,7 +1078,7 @@ pub fn register_type<T: ObjectSubclass>() -> Type {
             let priv_ = mem::MaybeUninit::<PrivateStruct<T>>::uninit();
             let ptr = priv_.as_ptr();
             let imp_ptr = ptr::addr_of!((*ptr).imp);
-            (imp_ptr as isize) - (ptr as isize)
+            (imp_ptr.addr() as isize) - (ptr.addr() as isize)
```
sdroege: All this can also use `offset_of!` now or not?

glib/src/subclass/types.rs
carlosmn: Oh odd, my editor decided to use tabs here even though nothing else does.

glib/src/subclass/types.rs
carlosmn: I didn't look into updating it until after fixing up the lints. I can change the order if you prefer so we do this first and avoid these few changes.

sdroege: No that's fine, thanks

sdroege:
Thanks, let's get this in for now then :)

PR #1918, opened 2026-03-05
Enable strict provenance lints on nightly

carlosmn:
Following up from #1917 (and in service to #659) here we add the feature `strict-provenance` that enables these lints. You can only enable them behind a nightly flag so we also have to put it behind a flag.

A bunch of these tests already enable all flags on nightly so we're covered there but those are for specific subprojects, so I've also added it to the general `cargo build` that runs on nightly.


carlosmn:
I did end up wondering if it's worth doing all of this. Once the lints become stable presumably you can detect this via clippy for some time.

PR #1930, opened 2026-03-18
glib: Make sure to acquire the main context and make it thread defaul…

sdroege:
…t in `MainContext::block_on()`

Everything else leads to hard to debug issues and was already documented before as invalid usage.

---

CC @sophie-h @bugaevc any opinions from your side? I'd also like to backport this to avoid wrong usage.

### DIFF
--- a/glib/src/main_context_futures.rs
+++ b/glib/src/main_context_futures.rs
@@ -644,41 +644,47 @@ impl MainContext {
     ///
     /// The given `Future` does not have to be `Send` or `'static`.
     ///
-    /// This must only be called if no `MainLoop` or anything else is running on this specific main
-    /// context.
+    /// # Panics
+    ///
+    /// This panics if the main context can't be acquired on this thread and can't be made the
+    /// thread default, which means that some other thread currently owns it.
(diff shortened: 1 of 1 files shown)


glib/src/main_context_futures.rs
```diff
@@ -644,41 +644,46 @@ impl MainContext {
...
-            // Super-unsafe: We transmute here to get rid of the 'static lifetime
-            let f = LocalFutureObj::new(Box::new(async move {
-                f.await;
-                Box::new(()) as Box<dyn Any + 'static>
-            }));
-            let f: LocalFutureObj<'static, Box<dyn Any + 'static>> = mem::transmute(f);
-            f
-        };
+            let f = unsafe {
+                // Super-unsafe: We transmute here to get rid of the 'static lifetime
```
bugaevc, line 663: Could use a link to https://github.com/rust-lang/unsafe-code-guidelines/issues/282

sdroege: Thanks for finding that :) How did you even find it?

PR #1939, opened 2026-03-30
glib: Create safe bindings for `BookmarkFile`

arun-mani-j:


glib/Gir.toml
```diff
@@ -534,6 +534,14 @@ status = "generate"
...
     manual = true # defined as extension methods on char
 
+[[object]]
+name = "GLib.BookmarkFile"
+status = "generate"
+    [[object.function]]
+    pattern = "*"
+        [[object.function.parameter]]
+        name = "bookmark"
+
```
bilelmoussaoui: You forgot to set const = true. But also, that shouldn't be used for all functions, only for getters/signals. Setters needs a mutable reference

arun-mani-j: Oh I mentioned it in the Matrix chat, removing this is what caused the incorrect parameter error to away - 
https://matrix.to/#/!rxguOwyruYzMcZpIkX:matrix.org/$6fzzmRTPzIZskNNjA_EVDA0MTyK49rZF6IAiYTW34Ok?via=matrix.org&via=gnome.org&via=tchncs.de

Let me add it back selectively.

arun-mani-j: Would not us setting `const` for things where https://docs.gtk.org/glib/struct.BookmarkFile.html does not say `const` be an API break?

A getter may not necessarily mean `const` as it can do lazy lookup and modify the data.

So how can we figure out which should be `const` and which should not be without manually inspecting source code?

bilelmoussaoui: you cannot, you have to inspect the code as far as i know

arun-mani-j: Then would not it be better to stick with non `const` pointers as that means we stick to what API provides? :sweat_smile:

bilelmoussaoui: that means requiring mut when not really needed

arun-mani-j: But how can we determine when it will be mut and when not :)

bilelmoussaoui: By reading the C code :-)

glib/sys/src/manual.rs
```diff
@@ -98,4 +101,14 @@ pub const G_VARIANT_TYPE_VARDICT: &str = "a{sv}";
 
 unsafe extern "C" {
     pub fn g_atomic_int_get(atomic: *const c_int) -> c_int;
+    #[cfg(not(feature = "v2_76"))]
+    pub fn g_boxed_copy(boxed_type: crate::GType, src_boxed: gconstpointer) -> gpointer;
```
sdroege: This will unfortunately break code using only glib-sys and not gobject-sys. I think the workaround for this needs to be inside the glib crate, not glib-sys.

Also maybe we can just implement this type manually? It's not like the API is changing all the time...

arun-mani-j: If it is to be inside `glib` crate, can you also which file it should be? :sweat_smile: 

> Also maybe we can just implement this type manually? It's not like the API is changing all the time...

But that will be 1:1 same as the automatically generated one right? So what's the advantage? :thinking:

sdroege: You can put the workaround to call `g_boxed_copy()` into the `glib::wrapper!` macro invocation then

glib/src/bookmark_file.rs
```diff
@@ -0,0 +1,857 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[cfg(feature = "v2_66")]
+#[cfg_attr(docsrs, doc(cfg(feature = "v2_66")))]
+use crate::DateTime;
+use crate::{ffi, translate::*};
+
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn g_bookmark_file_copy(bookmark: *mut ffi::GBookmarkFile) -> *mut ffi::GBookmarkFile {
```
sdroege: This shouldn't be public

glib/src/bookmark_file.rs
```diff
@@ -0,0 +1,857 @@
...
+
+    #[cfg_attr(feature = "v2_66", deprecated = "Since 2.66")]
+    #[allow(deprecated)]
+    #[doc(alias = "g_bookmark_file_get_app_info")]
+    #[doc(alias = "get_app_info")]
+    pub fn app_info(
+        &self,
+        uri: &str,
+        name: &str,
+    ) -> Result<(crate::GString, u32, libc::time_t), crate::Error> {
```
sdroege, line 113: The `time_t` could probably be a `std::time::SystemTime`?

arun-mani-j: But those methods are deprecated anyway, so I think it is fine to let them be as it is?

sdroege: Ok

glib/src/bookmark_file.rs
arun-mani-j: Done.

arun-mani-j:
Thanks. Will `glib-rs` need a new release for this to be available? In that case, when will be the new release? :sweat_smile:

sdroege:
Maybe next weekend. Ideally more things accumulate by then :)

arun-mani-j:
> Maybe next weekend. Ideally more things accumulate by then :)

Okay thanks :)

sdroege:
Released :)

arun-mani-j:
Thank you so much!!

-------- Original Message --------

PR #1949, opened 2026-04-12
glib: Fix expected output for subclass `Sync` compiler error tests for 1.94

sdroege:


sdroege:
Passes locally after that. I think trybuild2 is just broken.

PR #1960, opened 2026-04-29
glib: collections: Fix theoretical use-after-frees if dropping an ite…

sdroege:
…m panics

This is currently not a problem because all collections require the item type to implement `TransparentType` or `TransparentPtrType`, which are unsafe traits that are only implemented by the `glib::wrapper!` macro and require the `Drop` impl to not panic (among other things).

As such this has no effect right now but is cleaner and might avoid bugs in the future if the requirements of the two traits are ever weakened.

Thanks to JuHyung Son (@tooson9010-spec) for reporting.

### DIFF
--- a/glib/src/collections/slist.rs
+++ b/glib/src/collections/slist.rs
@@ -356,19 +356,19 @@ impl<T: TransparentPtrType> SList<T> {
     /// Only keeps the item in the list for which `f` returns `true`.
     #[inline]
     pub fn retain(&mut self, mut f: impl FnMut(&T) -> bool) {
-        if let Some(head) = self.ptr {
-            unsafe {
-                let mut ptr = head.as_ptr();
-                while !ptr.is_null() {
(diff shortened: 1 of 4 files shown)


glib/src/collections/slist.rs
```diff
@@ -363,8 +363,9 @@ impl<T: TransparentPtrType> SList<T> {
                     let item = &*(&(*ptr).data as *const ffi::gpointer as *const T);
                     let next = (*ptr).next;
                     if !f(item) {
-                        ptr::drop_in_place(&mut (*ptr).data as *mut ffi::gpointer as *mut T);
+                        let mut item_ptr = (*ptr).data;
                         self.ptr = ptr::NonNull::new(ffi::g_slist_delete_link(head.as_ptr(), ptr));
```
sdroege, line 367: There's an actual use-after-free here that was missed. I've fixed it as part of this PR.

PR #1962, opened 2026-05-03
glib: implement HasParamSpec for Type

TitouanReal:
Should allow to have `glib::Type` properties.


sdroege:
@TitouanReal Do you want to fix the CI failures?

bilelmoussaoui:
this is not enough though no? needs some tests to confirm it actually works

TitouanReal:
I might work on this at some point, but I have other priorities right now. Sorry.
I can close this MR and open a work item if you prefer to track this that way.

PR #1964, opened 2026-05-12
 fix: backport VariantStrIter unsoundness fix to 0.18 (RUSTSEC-2024-0429)

shiva-manu:
## Summary
  
  Pass pointer as `&mut p` instead of `&p` to `g_variant_get_child` in
  `VariantStrIter::impl_get` to fix undefined behaviour.
  
  ## Problem
  
  The immutable reference `&p` passed to the variadic C function violates
  Rust's aliasing rules. With recent compiler optimizations, the write
  through `&p` is disregarded entirely, leaving the pointer as NULL and
  causing SIGSEGV crashes when `CStr::from_ptr` is subsequently called.
  
  ## Fix
  
  ```diff
  -            let p: *mut libc::c_char = std::ptr::null_mut();
  +            let mut p: *mut libc::c_char = std::ptr::null_mut();
               ...
  -                &p,
  +                &mut p,
  
  Context
  
  This branch exists because Tauri 2.x depends on the gtk-rs 0.18
  ecosystem and cannot upgrade to glib 0.20+ (which requires gtk4-rs).
  Many Tauri-based projects are affected by this advisory with no
  upgrade path available.
  
  References:
  
  - Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0429 (https://rustsec.org/advisories/RUSTSEC-2024-0429)
  - Upstream fix: https://github.com/gtk-rs/gtk-rs-core/pull/1343 (https://github.com/gtk-rs/gtk-rs-core/pull/1343)


sdroege:
0.18 is EOL since a long time. Please update to the latest stable release.

sdroege:
> ```diff
> This branch exists because Tauri 2.x depends on the gtk-rs 0.18
> ecosystem and cannot upgrade to glib 0.20+ (which requires gtk4-rs).
> ```

tauri will have to maintain GTK3 bindings themselves and keep them up to date, or they have to update to GTK4 which would be a better long-term strategy anyway.

If you care about security it's not an option to depend on old versions of software, ever. Otherwise don't pretend to care about security.

PR #1968, opened 2026-05-21
glib: remove dependency on `g_bookmark_file_get_type` before glib v2.76

### LINKED ISSUE #1966: [BUG] `g_bookmark_file_get_type` is undefined before glib v2.76
**Bug description**

The code path for `#[cfg(not(feature = "v2_76"))]` uses `g_bookmark_file_get_type`:

https://github.com/gtk-rs/gtk-rs-core/blob/7e58f9f00177227296cfb8a5c2892157e7c8767b/glib/src/bookmark_file.rs#L10-L21

However, `g_bookmark_file_get_type` is only available since v2.76:

```c
GOBJECT_AVAILABLE_IN_2_76
GType   g_bookmark_file_get_type   (void) G_GNUC_CONST;
```

`g_bookmark_file_get_type` and `g_bookmark_file_copy` were both added in [GLib commit `552e01e17`](https://gitlab.gnome.org/GNOME/glib/-/commit/552e01e177b309d7f6127816c25d41e12ac2a2d1).

**Backtrace**

N/A

kawadakk:
Fixes #1966 by replacing `glib::wrapper!` with a manual implementation when targeting glib <= v2.76.

This manual implementation is based on the `glib::wrapper!` macro expansion but omits trait implementations and methods that depend on `GBookmarkFile` being a copyable boxed type. Specifically:

- It does not implement `Clone`.

- It does not implement `TransparentPtrType`, a subtrait of Clone`.

- It does not implement `FromGlibPtrNone`.

- `ToGlibPtr::to_glib_full` is unimplemented and panics at runtime.

- `ToGlibContainerFromSlice::to_glib_fill_from_slice` is unimplemented and panics at runtime.

- `FromGlibContainerAsVec::from_glib_{container,none}_as_vec` are unimplemented and panic at runtime.

- `FromGlibPtrArrayContainerAsVec::from_glib_{container,none}_as_vec` are unimplemented and panic at runtime.

[1]: https://gitlab.gnome.org/GNOME/glib/-/commit/552e01e177b309d7f6127816c25d41e12ac2a2d1

### DIFF
--- a/glib/src/bookmark_file.rs
+++ b/glib/src/bookmark_file.rs
@@ -5,31 +5,329 @@
 use crate::DateTime;
 use crate::{ffi, translate::*};
 
-#[allow(clippy::missing_safety_doc)]
-unsafe fn g_bookmark_file_copy(bookmark: *mut ffi::GBookmarkFile) -> *mut ffi::GBookmarkFile {
-    #[cfg(not(feature = "v2_76"))]
-    unsafe {
-        crate::gobject_ffi::g_boxed_copy(
-            ffi::g_bookmark_file_get_type(),
-            bookmark as ffi::gconstpointer,
-        ) as *mut ffi::GBookmarkFile
-    }
-
-    #[cfg(feature = "v2_76")]
-    unsafe {
-        ffi::g_bookmark_file_copy(mut_override(bookmark))
-    }
-}
-
+#[cfg(feature = "v2_76")]
 crate::wrapper! {
     #[doc(alias = "GBookmarkFile")]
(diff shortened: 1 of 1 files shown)


sdroege (commented):
I think instead of that we should simply make it available only in 2.76 and newer by putting the `#[cfg]` in lib.rs.

What do you think?

kawadakk:
@sdroege That's fine by me. I'll open a new PR.

sdroege:
Thanks, seems a lot simpler to me and if someone really needs it pre-2.76 then we can reconsider once that happens.

PR #1970, opened 2026-05-24
Add safe wrapper for `g_log_writer_default_set_debug_domains`

tautropfli:
The function is [explicitly documented](https://docs.gtk.org/glib/func.log_writer_default_set_debug_domains.html) as thread-safe:
> This function is thread-safe.

### DIFF
--- a/glib/src/log.rs
+++ b/glib/src/log.rs
@@ -7,6 +7,8 @@ use std::{
     sync::{Arc, Mutex, OnceLock},
 };
 
+#[cfg(feature = "v2_80")]
+use crate::StrVRef;
 use crate::{GStr, GString, LogWriterOutput, ffi, translate::*};
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


glib/src/log.rs
```diff
@@ -1091,3 +1091,11 @@ pub fn log_writer_default_would_drop(log_level: LogLevel, log_domain: Option<&st
         ))
     }
 }
+
+#[cfg(feature = "v2_80")]
+#[cfg_attr(docsrs, doc(cfg(feature = "v2_80")))]
+#[doc(alias = "g_log_writer_default_set_debug_domains")]
+#[inline]
+pub fn log_writer_default_set_debug_domains(domains: &[&str]) {
```
bilelmoussaoui: Maybe this should be StrV?

tautropfli: Makes sense, fixed 👍🏼

sdroege (approved):

PR #1981, opened 2026-06-08
gio: change write_all/read_all error return type to carry partial count

mvanhorn:
## Summary

Changes the return type of `OutputStream::write_all_async`/`write_all_future` and the symmetric `InputStream::read_all_async`/`read_all_future`/`read_all` from `Result<(B, usize, Option<glib::Error>), (B, glib::Error)>` to `Result<(B, usize), (B, usize, glib::Error)>`, so callers no longer inspect an inner `Option` on the `Ok` path and the error path carries the partial byte count.

## Why this matters

The old `Ok`-with-`Some(error)` shape forced every caller to check the inner `Option` even on success, which is confusing and easy to get wrong (#1942). @sdroege agreed the design is wrong and stated the preferred variant `Result<(B, usize), (B, usize, glib::Error)>`, which also surfaces the partial count on failure. Each async trampoline now builds `Ok((buffer, bytes_written))` when no error occurred and `Err((buffer, bytes_written, ...))` otherwise, dropping the prior three-state branch. The `*_future` wrappers mirror the new callback type and the sync `read_all` surfaces the error directly.

This is a breaking API change, appropriate for the in-progress `0.23.0-alpha` cycle.

## Testing

- Added cancellation tests asserting the error path carries the partial byte count for both `read_all` and `write_all`.

Closes #1942

### DIFF
--- a/gio/src/input_stream.rs
+++ b/gio/src/input_stream.rs
@@ -42,12 +42,13 @@ pub trait InputStreamExtManual: IsA<InputStream> + Sized {
         &self,
         mut buffer: B,
         cancellable: Option<&C>,
-    ) -> Result<(usize, Option<glib::Error>), glib::Error> {
+    ) -> Result<(B, usize), (B, usize, glib::Error)> {
         let cancellable = cancellable.map(|c| c.as_ref());
         let gcancellable = cancellable.to_glib_none();
-        let buffer = buffer.as_mut();
-        let buffer_ptr = buffer.as_mut_ptr();
-        let count = buffer.len();
+        let (count, buffer_ptr) = {
+            let buffer = buffer.as_mut();
+            (buffer.len(), buffer.as_mut_ptr())
+        };
         unsafe {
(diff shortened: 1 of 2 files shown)


sdroege (dismissed):

sdroege:
Test failures need some fixes.

sdroege:
@mvanhorn Do you want to update this PR?

mvanhorn:
Yes - pushed the test fix. read_all_cancelled was failing because it expected a pre-cancelled Cancellable to yield a Cancelled error, but a synchronous MemoryInputStream read completes before the cancel is seen, so read_all returned Ok and unwrap_err panicked. Retargeted it to a closed stream, which reliably hits the (buffer, count, error) path this PR adds; read_all_async_cancelled still covers cancellation. cargo test -p gio input_stream is green (20/0). dd13b01

sdroege:
Rebased on latest main so the CI (hopefully) passes.

mvanhorn:
Thanks @sdroege for rebasing and merging the write_all error type change.

PR #1991, opened 2026-06-24
glib: fix UB in VariantStrIter::impl_get (RUSTSEC-2024-0429)

johncarmack1984:
`g_variant_get_child` writes through the out-pointer — needs `&mut p`, not `&p`. Same fix as 0.20, backported to 0.18 (tauri's gtk stack is still pinned here and trips RUSTSEC-2024-0429).


sdroege:
0.18 is long EOL and there won't be any new releases. tauri will have to either port to GTK4, or they have to maintain their own GTK3 bindings that track up-to-date dependencies.

PR #2003, opened 2026-07-17
sys: bump system-deps 7 -> 8

calvaris:
The `binary` feature of system-deps was released to crates.io as major 8.0.0. Bump all sys crates so the whole graph can unify on system-deps 8, which is a prerequisite for downstream consumers.


calvaris:
It looks like this PR needs https://github.com/gtk-rs/gir/pull/1747 .

thiblahute:
@sdroege could you check that one so we can move forward in gst-rs please?

sdroege:
This needs to be folded into https://github.com/gtk-rs/gtk-rs-core/pull/1998 (and that one needs to be finished)

sdroege:
Updated that one.

Issue #2011, opened 2026-07-29
[BUG] Potential memory leak in block_on + spawn_blocking


sophie-h:
I have the issue that roughly the following setup

```rust
fn main() {
    let main_context = glib::MainContext::new();
    main_context.block_on(async {
        let _ = gio::spawn_blocking(|| std::thread::sleep(std::time::Duration::from_secs(1))).await;
    });
}
```

triggers a memory leak in glycin reported by lsan in gdk-pixbuf tests. It's two different uses of `gio::spawn_blocking` that trigger it.

However, this minimal reproducer doesn't seem to be enough since

```sh
$ LSAN_OPTIONS="verbosity=1:exitcode=1" RUSTFLAGS="-Z sanitizer=leak" cargo +nightly run --target x86_64-unknown-linux-gnu
```

does not report an issue.

I noticed that both debug outputs in the code below return true, They no longer do if the `spawn_blocking` is commented out. I'm not sure if that's relevant. I didn't expect there to be any source that can return `true` for them at this point:

```rust
fn main() {
    let main_context = glib::MainContext::new();
    main_context.block_on(async {
        let _ = gio::spawn_blocking(|| std::thread::sleep(std::time::Duration::from_secs(1))).await;
    });
    dbg!(main_context.pending());
    dbg!(main_context.iteration(true));
}
```

sdroege:
Did you find a way to reproduce it in the meantime? I assume the difference is probably that in the leaking case some source inside the main context keeps a strong reference to the main context so neither the source nor the main context get cleaned up unless the source actually finishes?

bugaevc:
With your example, I do get:

```
#3  0x00007ffff7e39826 in g_task_return_now (task=0x5555555cfb10 [GTask]) at ../gio/gtask.c:1359
#4  0x00007ffff7e39875 in complete_in_idle_cb (task=task@entry=0x5555555cfb10) at ../gio/gtask.c:1377
#5  0x00007ffff7c37524 in g_idle_dispatch (source=0x7fffec000b80, callback=0x7ffff7e39860 <complete_in_idle_cb>, user_data=0x5555555cfb10) at ../glib/gmain.c:6492
#6  0x00007ffff7c35f24 in g_main_dispatch (context=0x5555555ce4b0) at ../glib/gmain.c:3591
#7  g_main_context_dispatch_unlocked (context=0x5555555ce4b0) at ../glib/gmain.c:4451
[...]

PR #2015, opened 2026-08-13
gio: Only return the result from spawn_blocking() from the task compl…

sdroege:
…etion callback

This makes sure (or at least more likely) that all cleanup is actually going to happen before the result is taken out of the channel.

See https://github.com/gtk-rs/gtk-rs-core/issues/2011

----

CC @sophie-h @bugaevc

### DIFF
--- a/gio/src/task.rs
+++ b/gio/src/task.rs
@@ -521,13 +521,35 @@ where
     T: Send + 'static,
     F: FnOnce() -> T + Send + 'static,
 {
-    // use Cancellable::NONE as source obj to fulfill `Send` requirement
-    let task = unsafe { Task::<bool>::new(Cancellable::NONE, Cancellable::NONE, |_, _| {}) };
+    unsafe extern "C" fn free_box<T: Send + 'static>(ptr: glib::ffi::gpointer) {
+        unsafe {
+            let _ = Box::from_raw(ptr as *mut std::thread::Result<T>);
+        }
+    }
+
     let (join, tx) = JoinHandle::new();
+
+    // use Cancellable::NONE as source obj to fulfill `Send` requirement
+    let task = unsafe {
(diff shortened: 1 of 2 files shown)


bugaevc:
Yeah, something like that.

I'm wondering if you still need that channel (`tx.send(res)`); can't you more or less `.wake()` the awaiting coroutine from the task's callback? But maybe that amounts to mostly reimplementing a channel.

Should there be another API that implements this pattern of `new` + `block_on` + `while`-`iteration`? I'm imagining `MainContext::block_on2(some_future)` which does what `block_on` does in other libraries.

sdroege:
> I'm wondering if you still need that channel (`tx.send(res)`); can't you more or less `.wake()` the awaiting coroutine from the task's callback? But maybe that amounts to mostly reimplementing a channel.

The problem with that is that you need to implement a `Future` manually then (or use something simpler than a channel which basically would be equivalent to a oneshot channel passing a `()`).

> Should there be another API that implements this pattern of `new` + `block_on` + `while`-`iteration`? I'm imagining `MainContext::block_on2(some_future)` which does what `block_on` does in other libraries.

I think either that, or adding that iteration to `block_on()`. I'm undecided on that, do you have a preference?

Which other libraries though?

sdroege:
So thinking about it, adding the `iteration(false)` loop would only really cause problems if there was a source that's always ready (e.g. an idle source always returning `G_SOURCE_CONTINUE` which is bad). Always doing it then shouldn't really cause problems that wouldn't have been there during the normal main loop iteration already.

sdroege:
Let's do that then

sophie-h:
I let my test go through 100k iterations and didn't find an issue when using this MR.

sdroege:
Great, thanks for testing. I'll get this into the next bugfix release then :)

PR #2030, opened 2026-08-24
gio: Only return the result from spawn_blocking() from the task completion callback

### LINKED ISSUE #2011: [BUG] Potential memory leak in block_on + spawn_blocking

sophie-h:
This makes sure (or at least more likely) that all cleanup is actually going to happen before the result is taken out of the channel.

Fixes #2011

### DIFF
--- a/gio/src/task.rs
+++ b/gio/src/task.rs
@@ -521,13 +521,35 @@ where
     T: Send + 'static,
     F: FnOnce() -> T + Send + 'static,
 {
-    // use Cancellable::NONE as source obj to fulfill `Send` requirement
-    let task = unsafe { Task::<bool>::new(Cancellable::NONE, Cancellable::NONE, |_, _| {}) };
+    unsafe extern "C" fn free_box<T: Send + 'static>(ptr: glib::ffi::gpointer) {
+        unsafe {
+            let _ = Box::from_raw(ptr as *mut std::thread::Result<T>);
+        }
+    }
+
     let (join, tx) = JoinHandle::new();
+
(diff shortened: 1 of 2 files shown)


sophie-h:
Backport to 0.22

sdroege:
Thanks, this is folded into https://github.com/gtk-rs/gtk-rs-core/pull/2037

PR #2031, opened 2026-08-24
security: Update crossbeam-epoch to 0.9.20

sophie-h:
Security update for <https://rustsec.org/advisories/RUSTSEC-2026-0204>


sophie-h:
Most likely not of practical relevance for gtk-rs, but makes `cargo deny` happy.

PR #2038, opened 2026-09-04
glib: Fix double-free in StrV::clear() by writing NULL terminator

ksj1230:
clear() frees each string via g_free() but did not write a NULL terminator at position 0 afterward. Since Drop calls g_strfreev(), which walks the array until NULL, the stale (already-freed) pointers were freed again, causing a double-free.

### DIFF
--- a/glib/src/collections/strv.rs
+++ b/glib/src/collections/strv.rs
@@ -799,6 +799,10 @@ impl StrV {
                 ffi::g_free(*self.ptr.as_ptr().add(i) as ffi::gpointer);
             }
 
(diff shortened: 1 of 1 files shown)


sdroege:
Do you also want to add a small test for this?

sdroege (approved):

ksj1230:
Added a test in the latest push.
Thanks for the suggestion.

sdroege:
Thanks, also great find :)

ksj1230:
Thank you for the quick review and merge!

